Issue #34, opened 2025-06-03
Page counter property and SCORM/IMS export


pabloamayab:
The ´page counter` property should have no effect on the SCORM/IMS export.

PR #41, opened 2025-06-03
28 the page preference visible on export does not work

pabloamayab:
This PR fixes problems where the page visibility preference in export was not taken into account.
- It adds a function that looks for whether a page is visible or not, both in its preference and in its ancestors.
- Fixed HTML5 export, navigation menu, page counter and previous/next buttons.
- Fixed SCORM/IMS export.
- The content of non-visible pages is not exported.
- Single page (SP) export is fixed.
- Fixed issue #34 where page number export preference was taken into account in SCORM.
- Non-visible content is not exported in the JSON of the search bar.

This PR doesn't delete links reference to non-visible pages. It is neccesary review ePub exportation https://github.com/exelearning/exelearning-web/issues/460

### DIFF
--- a/src/Util/net/exelearning/Util/ExportXmlUtil.php
+++ b/src/Util/net/exelearning/Util/ExportXmlUtil.php
@@ -347,9 +347,7 @@ class ExportXmlUtil
 
         // Pages organization
         foreach ($odeNavStructureSyncs as $odeNavStructureSync) {
-            
-            if(isset($visiblesPages[$odeNavStructureSync->getOdePageId()]))
-                {
+            if (isset($visiblesPages[$odeNavStructureSync->getOdePageId()])) {
                 // Page properties
                 $pageProperties = $odeNavStructureSync->getOdeNavStructureSyncProperties();
                 $pagePropertiesDict = [];


ignaciogros:
Hi @pabloamayab:

It works for me, but I think that all exports except "Single page" should include the page counter. It was like that in eXeLearning 2.9.

The problem is that removing lines 4-7 is not enough:

```
if (
    isset($odeProperties['pp_addPagination'])
    && 'true' == $odeProperties['pp_addPagination']->getValue()
    && in_array(
        $exportType,
        [Constants::EXPORT_TYPE_HTML5, Constants::EXPORT_TYPE_EPUB3, Constants::EXPORT_TYPE_ELP]
    )
) {
    $pageNumber = self::createHTMLPageNumber(
        $odeNavStructureSync,
        $visiblesPages,
        $translator
    );
    self::appendSimpleXml($pageHeader, $pageNumber);
}
```

If you remove those lines, a SCORM package with 10 pages will display "Page X of 1" instead of "Page X of 10".

You also say this:

> The content of non-visible pages is not exported.

But invisible pages are also exported (the HTML files are there). Is that the expected behavior?

Thank you.

ignaciogros (changes requested):
See https://github.com/exelearning/exelearning/pull/41#issuecomment-2935633098

pabloamayab:
Page numbering in SCORM already works well, you needed to have access to the entire array of pages instead of just the one currently being processed.

The ePub3 export needs to be reviewed in another issue.

Now I have detected another problem, the SCORM navigation buttons in the Moodle viewer when there are hidden pages do not work. I will investigate and try to fix it.

The content of the hidden pages is not exported, only the HTML file is exported but without content. What behavior do we expect? It is easier to export the file with its content and not be indexed for searches than not to export the file, because it would involve many changes.

ignaciogros:
> Now I have detected another problem, the SCORM navigation buttons in the Moodle viewer when there are hidden pages do not work. I will investigate and try to fix it.

I saw that the `isvisible` attribute does that:

```
<item identifier="" isvisible="false" identifierref="">
    <title>...</title>
</item>
```

> The content of the hidden pages is not exported, only the HTML file is exported but without content. What behavior do we expect?

I think that it's better to export everything (HTML page + files). Changing the `isvisible` attribute value will be the easiest way to do it.

ignaciogros:
Hi @pabloamayab:

I think it's OK now.

**Two errors:**

- ePub navigation tree. I just opened a different issue for that: #59 
- Procomún only displays the home page link. But that's something to change in Procomún, isn't it?

We can merge this into main if it's OK to you.

pabloamayab:
OK @ignaciogros 
Procomun has to change the algorithm to make the menu. Thanks for opening the ePub issue. 

Merge!

PR #52, opened 2025-06-09
Add api for user editing idevice

jgblasco:
Small API to return the user who is editing an idevice. It has been designed to be as agile as possible since it will be consulted constantly. Therefore, there are no checks. If the idvice does not exist or the request is incorrect, it returns null.


ignaciogros:
Thanks, @jgblasco.

@abertranb:

Just in case you need this to improve the collaborative edition. Use **/api/current-ode-users-management/current-ode-user/get/user/by/current/component/id?current_component_id=IDEVICE_ID** (`idevice-id` HTML attribute) to know who's editing an iDevice. It'll just return null if nobody's editing.

We might need something like that for the pages too. Notice that you can edit the package Properties while other user's doing so. And if you change the title, it becomes a different content and the collaborative edition does not work anymore.

ignaciogros (approved):

PR #66, opened 2025-06-13
Common JavaScript files partial revision

ignaciogros:
- Remove unnecessary code. 
- Move some edition code to a file that won't be exported.

The revision is not finished yet. We'll keep working on this after closing #3


eXeLearningProject:
Checked by @mnarvaezm and @ignaciogros.

PR #67, opened 2025-06-13
Delete all with second Platform

jgblasco:


ignaciogros:
Hi @jgblasco,

**PlatformIntegrationApiControllerTest.php** tries to execute the second type integration. That's why the tests fail.

See `testSecondTypeEditRedirectsWithItemUuid` in that file for details.

If the functionality is no longer in eXe, those tests should be removed too.

Thank you!

pabloamayab:
It's OK!

pabloamayab (approved):
OK!

PR #80, opened 2025-06-16
Remove all hidden classification fields

ignaciogros:
**Only these properties will remain:**

- Content metadata
    - Title
    - Language
    - Authorship
    - License
    - Description
- Export options
    - Editable export
    - Include a copy of the source file
    - Search bar
    - Made with eXe link
    - Page counter
    - Accessibility toolbar
- Custom code
    - HEAD
    - Page footer


ignaciogros:
@pabloamayab, @mnarvaezm, @erseco:

Could you review this, please?

I've deleted many properties, so any metadata that is not in the list above will be removed when opening an old .elp. The information that is not deleted (title, license, description, authorship, etc.) should respect the old .elp values.

erseco (approved):

erseco:
Looks good. As we modify existing properties, some tests will naturally require updates. In the future, once we introduce more stable selectors (like dedicated `data-testid` or semantic roles), tests will be more robust ([see Testing Library guidelines](https://testing-library.com/docs/guide-which-query/)). Still, tests always evolve along with the codebase.

PR #99, opened 2025-06-18
Fix #68 - Tooltips won't hide.

### LINKED ISSUE #68: Tooltips won't hide
Some tooltips don't disappear when they should:

Image

Most users reporting this issue are using Ubuntu + Firefox.

ignaciogros:
Conflict when using Bootstrap's tooltips and jQuery UI.


erseco (approved):
LGTM!

PR #100, opened 2025-06-18
Fixed up-local and added test-local

erseco:
This pull request fixes the `up-local` target and adds a new `test-local` target to the `Makefile`. These allow running eXeLearning outside of Docker, using a local SQLite database and PHP’s built-in web server.

⚠️ This is an **unsupported fallback** mode. It may fail in some cases but can be useful for debugging, quick iterations, or environments where Docker is not available (e.g. OpenAI Codex, Jules, or CI pipelines without Docker-in-Docker support).

### Example usage (Ubuntu CI, like OpenAI Codex):

```bash
export DEBIAN_FRONTEND=noninteractive
sudo apt update
sudo apt install -y make composer php-xml php-sqlite3 php-curl php-mbstring php-intl php-zip

composer install --no-interaction --prefer-dist

make test-local     # run unit tests
# or
make up-local       # start the app locally
```


ignaciogros (commented):
`make up-local` on Windows:

imagen

My PHP version is 8.3.20 (cli).

ignaciogros:
It runs after enabling the `php_pdo_sqlite` extension, but **127.0.0.1:8000** does not work. 

It works in common development environments anyway.

It's already merged.

PR #102, opened 2025-06-19
61-head-and-page-footer-encoding-problems

### LINKED ISSUE #61: HEAD and page footer encoding problems
Add this code in Content properties - Custom code - HEAD and Page footer:

```
<script>
alert("á");
</script>
```

Then click on "Preview".

You'll see this:

Image

As you can see, the "á" is replaced by an HTML entity. There should be no replacement there.

File to review: ExportXmlUtil.php (search `loadHTML`).

pabloamayab:
Improve the processing of custom code in the exported HTML by fixing HTML entities in the head and footer sections, refactoring the code for clarity, and ensuring proper encoding.

I have tested and I think it only affects the head and siteuserfooter code. Can you check if there are any collateral effects?
Now, in the custom code the text should always be inserted between tags (ex: `<p>,<code>`, ... ). IMHO I think that should not be a problem.


pabloamayab:
@erseco  told me that there are now problems with coding on iDevices and navigation, but in my tests it doesn't happen:
- Ubuntu 22 - Firefox
- Kubuntu 24 - Chromium
- Windows 11 - Firefox and Chrome

Can you tell me how to reproduce the problem, can you send me an index.html to check the encoding?

The HTML code looks correct, @ignaciogros can you check what is happening?

pabloamayab:
@erseco @ignaciogros I have already found several ELPs where the coding fails, I am going to fix it.

pabloamayab:
I have already found why some resources are displayed correctly encoded in the export and others are not. Surprisingly, if the search bar is activated, the encoding is altered due to the changes introduced in this PR.

ignaciogros (changes requested):
Open [head-and-footer-tests.zip](https://github.com/user-attachments/files/20852738/head-and-footer-tests.zip) and preview it. You will see that the text ("Atención") included in "Page footer" is missing. I think it's because it's not in any HTML element.

This works:

```
<p>Atención</p>
<script>alert("Atención")</script>
```

But this will not:

```
Atención
<script>alert("Atención")</script>
```

ignaciogros:
Hi @pabloamayab,

I close this PR for the moment. We'll keep #61 open until it's fixed.

v3.0.0-beta3 will have no changes in that part.

PR #124, opened 2025-06-26
Fix: Correct internal link ID migration

franmate:
Bug fixes in the migration from the internal linking system based on file names to the new one that uses IDs.


ignaciogros (changes requested):
Hi @franmate,

They work in eXe now, but not when previewing or exporting a content.

Please download [tema-8.zip](https://github.com/user-attachments/files/20921810/tema-8.zip), click on preview or  export and test the links of the home page. 

Exported links are working on the "main" branch. 

It's probably because "html/" is missing from the href attribute of those links. Only index.html is in the root folder. The rest of the HTML files are in the "html" folder.

I tried different export types: website, IMS, SCORM, ePub...

Single page export links are OK.

PR #139, opened 2025-07-01
82-remove-links-to-a-node-when-deleting-it

Ivan0886:
82-remove-links-to-a-node-when-deleting-it


erseco:
Hi @ignaciogros and @Ivan0886, I think I’ve fixed the issue so that internal links are now properly removed when deleting a node.

In addition, I added a unit test that uses the `tests/Fixtures/tema-10-ejemplo.elp` file.
This test deletes the second node ("Extinción del contrato") and verifies that the links in the "Inicio" page are converted into plain text.

This ensures the problem will not reappear in the future. ✅

erseco:
Hi @ignaciogros if you changed some content from the .elp please update the  unit-test too: `tests/Functional/DeleteInternalLinksOnNodeDeletionTest.php` to pass, because currently is failing, thanks!

ignaciogros:
Sorry about breaking those tests, and thanks for fixing them! Merged!

PR #145, opened 2025-07-03
Api Rest Users (crud), block and unblock (only admin users)

jgblasco:
The url is /apiRest/users
To Block /apiRest/users/{id}/block
To unblock /apiRest/users/{id}/unblock


ignaciogros (changes requested):
Thank you, @jgblasco.

/apiRest/users and /apiRest/users/{id} work, but when I try to block/unblock, I get this message:
 
imagen

@erseco:

Should this be in main without roles and access control?

erseco:
Hi @jgblasco, any news on the API path change to /api/v2/users? Just asking because in PR #149 I added support for generating API tokens for users — these tokens can now be used for Bearer authentication.

It would be great to align this management API with the new versioned path and ensure it's protected by token-based authentication. Let me know if you want help wiring that part up.

erseco:
Great work, @jgblasco — this PR has now been superseded by **PR #201**, which integrates a more complete RESTful API. Many thanks for your contribution and for the ideas you brought in here; they were really valuable for shaping the new implementation.

PR #155, opened 2025-07-17
New iDevice: Magnifier

mnarvaezm:
New iDevice: Magnifier


ignaciogros (changes requested):
Please download [lupa.zip](https://github.com/user-attachments/files/21312712/lupa.zip) and open it with eXe 3. Everything works fine, but the alignment (float) is not recovered. The image is right-aligned in eXe 2.9, but it's on the left in eXe 3.

Thank you!

PR #197, opened 2025-08-21
Second delivery of new interface

JMTomasZafra:
iDevices bottom bar, modals, interface improvements, part of the collaborative mode, fixes to the middle section, drag & drop details, layout system, etc.

There are many pieces still in development, so they may not display correctly, but all functionality should continue to run correctly.


ignaciogros (approved):
Thank you, @JMTomasZafra!

PR #205, opened 2025-08-24
Initial version of teacher-mode (WIP)

erseco:
This PR introduces the **initial version of "Teacher Mode"** in eXeLearning. It adds a toggle button to switch between student and teacher views, a new `visibilityType` property for pages, blocks, and iDevices, and the necessary export logic and translations to support the feature.

At this stage the focus is on **basic functionality** (toggle logic, storage, visibility rules, and export integration).
Future iterations should improve the **visual design and user experience** of the toggle and teacher-only content presentation.

# Screenshots
Captura de pantalla 2025-08-24 a las 13 03 43
Captura de pantalla 2025-08-24 a las 13 03 36


ignaciogros:
Hi @erseco,

I just pushed some changes. That would be the default presentation, but you can change it from any style. I'll create a different button in the default style.

**Do you like that solution?**

Apart from that, there's one detail that will make it better:

Please open [teacher-mode-tests_web.zip](https://github.com/user-attachments/files/22093695/teacher-mode-tests_web.zip) and click on "Preview". You'll see that the third page has no mode toggler. That's because it has no teacher-only boxes.

That behavior is OK, but the user might think it's a bug, because you expect the button to be always in the same place.

To fix that, we'll have to check if the elp (in any of its pages) has that kind of boxes. If so, the `body` tag could have an extra CSS class (like "exe-teacher-contents"), even if the page has no teacher-only boxes.

If you think that that'll work, could you make **ExportXmlUtil.php** add that class?

Thank you.

ignaciogros:
@JMTomasZafra:

exe_export.js creates the mode switcher when teacher-only contents are available in a page. 

If you want to create your own switcher (a button, for example), you can create it in your style's JavaScript file. Just add an element with `id="teacher-mode-toggler"`, and exe_export.js will not create the switcher.

The default behavior is defined in `$exeExport.teacherMode.init`.

PR #218, opened 2025-08-27
207 internal links are not working

ignaciogros:


ignaciogros:
@mnarvaezm will try to find a better temporary solution, updating the content of the `data-idevice-json-data` attribute.

PR #221, opened 2025-08-28
Avoid a implementation error that copied /app/var inside /app/src/var…

erseco:
This pull request updates how cache and log directories are configured and managed in the application, shifting responsibility from custom logic in the `Kernel` class to environment variable resolution in configuration files. It also includes a minor update to service autoloading. The main benefits are simplification of the `Kernel` class and improved flexibility/configurability through environment variables.

**Configuration improvements:**

* Updated `config/packages/framework.yaml` to use the `CACHE_DIR` environment variable for the cache directory, falling back to the default if not set. This removes the need for custom directory management in the `Kernel` class.
* Updated `config/packages/monolog.yaml` to use the `LOG_DIR` environment variable for log file storage, with a fallback to the default path.

**Codebase simplification:**

* Removed the overridden `getCacheDir()` and `getLogDir()` methods from `src/Kernel.php`, as directory management is now handled via configuration and environment variables.

**Service configuration:**

* Added `../src/var/` to the list of resource paths in `config/services.yaml` for autoloading or service discovery.… and set better cache dirs


ignaciogros (changes requested):
Some tests are failing and I get errors running `make fix`:

imagen

And `make up`:

imagen

erseco:
This will be revised

PR #224, opened 2025-08-28
Added paralell testing support

erseco:
This pull request introduces several improvements to the testing infrastructure and configuration, most notably by switching end-to-end (e2e) and parallel test execution to use `paratest`, simplifying environment variables for test users, and making PHPUnit configuration more robust and developer-friendly. Below are the most significant changes:

**Testing and Test Runner Improvements:**

* Switched e2e and parallel test execution to use `paratest` for improved speed and reliability, added `brianium/paratest` as a dev dependency, and introduced new Composer scripts and Makefile targets for parallel test execution. (`composer.json`, `Makefile`)

**Test Environment and User Simplification:**

* Removed the secondary test user (`TEST_USER2_*`) from all environment files, documentation, and Docker Compose configurations, consolidating test setup to use only a single test user. (`.env.dist`, `doc/deploy/README.md`, `doc/deploy/docker-compose.*.yml`, `docker-compose.yml`, `doc/07-testing.md`)

**Test Execution and Environment Updates:**

* Improved the local test execution flow to use an in-memory SQLite database, set appropriate environment variables, and ensure the schema is up-to-date before running unit tests. (`Makefile`)
* Enabled static metadata cache for Doctrine test bundle to speed up test execution. (`config/packages/test/dama_doctrine_test.yaml`)

**Test Code Quality:**

* Refactored a test to always create a unique, ephemeral user for isolation and reliability, instead of relying on a fixed user ID. (`tests/Api/PlatformIntegrationApiControllerTest.php`)

These changes together streamline the test setup, improve reliability and performance, and make the developer experience smoother when running tests locally or in CI.


ignaciogros (changes requested):
I've seen that the server's tests fail too. It can't open basic-example.elp

erseco:
Hi @ignaciogros, 

The `make test-unit-parallel` target currently fails because there is a race condition in the unit tests when writing files. Sometimes this triggers the error:

```
[ERROR] Export failed: Other user is saving changes right now
```

For that reason, parallel execution has only been enabled for `e2e` and `e2e-realtime` tests for now. We still need to investigate how to avoid that race condition, since it would also happen in real usage when two users try to write simultaneously. It’s actually useful that these issues show up, which is why I added the parallel execution option, along with running the tests in random order — so we can be sure they are truly independent.

Issue #64, opened 2025-06-11
Text iDevice. Unnecessary data saved.


ignaciogros:
The content.xml file saves the `<div class="exe-text-template"></div>` wrapper included in the text.html file of the iDevice. It should not be saved. It's not a content created by the user.

PR #225, opened 2025-08-28
Fix internal links

mnarvaezm:
- Fix internal links in edition mode
- Fix internal links in  export mode
- Render HTML again only during export for text-type iDevices in old ELPs
- Add button to open AI chat with textarea content


ignaciogros (approved):
Thank you, @mnarvaezm.

I think that we should merge this changes now to release v3.0.0-beta1 as expected, but maybe we could improve this solution later (see #64) to remove all unnecessary or invalid HTML.

PR #226, opened 2025-08-29
213 add project title to all exports

ignaciogros:


erseco (approved):
It workes and test passing, LGTM!

Issue #232, opened 2025-08-29
Electron. Save instead of download.


ignaciogros:
You can't save, as you'll do in any desktop app. You can only download. The projects are saved in the server, but the user can't define where.

PR #234, opened 2025-08-31
Electron: Desktop-like Save / Open / Export in Offline Mode, plus Offline Tests

erseco:
This pull request introduces full offline (Electron) support for **Open**, **Save**, **Save As**, and **Export**, allowing the app to behave like a true desktop application. Once the file path is chosen, subsequent "Save" actions overwrite the file without prompting again, aligning with common desktop usability patterns.

It also adds offline test coverage and an initial preview of how **Electron E2E tests with Playwright** will look.

### Main Changes

* **Offline (Electron) integration**

  * Implemented persistent Save / Save As functionality for `.elp` projects.
  * Added support for Open and Export dialogs in offline mode.
  * Unified logic so offline behaves as expected in a desktop app.

* **Testing improvements**

  * Added dedicated offline test suite.
  * Provided a preview setup for Playwright + Electron E2E tests (to be expanded later).
  * Updated test scripts and configuration to support offline runs.

### Notes

* ⚠️ **Important:** This PR must be merged **after** #224, since it relies on the same codebase for the new parallelized E2E tests.

### Related Issues

Closes [#232: Electron. Save instead of download](#232)

### DIFF
--- a/.env.dist
+++ b/.env.dist
@@ -26,11 +26,6 @@ TEST_USER_EMAIL=user@exelearning.net
 TEST_USER_USERNAME=user
 TEST_USER_PASSWORD=1234
 
-# Test user data 2 (for realtime)
-TEST_USER2_EMAIL=user2@exelearning.net
-TEST_USER2_USERNAME=user2
-TEST_USER2_PASSWORD=1234
-
 ###> Authentication Configuration ###
 # Available modes (comma-separated):
 # - "none": Offline mode (no authentication)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,6 +80,9 @@ jobs:
       - name: PHPUnit E2E RealTime Tests
         run: make test-e2e-realtime
 
+      - name: PHPUnit E2E Offline Tests
+        run: make test-e2e-offline
+
       - name: Generate HTML report
         if: always()
         run: |
--- a/Makefile
+++ b/Makefile
@@ -26,9 +26,6 @@ PUBLISH_ARG := $(if $(PUBLISH),--publish $(PUBLISH),)
 # Use subst for multiplatform ~ expand
 EXPAND_PATH = $(subst ~,$(HOME),$(1))
 
-# Enable debug if DEBUG or ACTIONS_STEP_DEBUG is set (e.g., in GitHub Actions re-run with debug logging)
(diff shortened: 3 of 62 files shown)


eXeLearningProject:
@erseco:

This PR is great news!

Could you check those conflicts?

Shall we test this one before #221?

erseco:
Hi @ignaciogros I fixed the conflicts, and is not importante, you can mix in the order you prefer

ignaciogros (changes requested):
Thank you for fixing those conflicts.

I just tested in on Windows.

`make run-app` works after upgrading PHP to 8.4 and enabling **openssl**, but there are still some problems:

- You can't open local .elp files (created with eXe 2.9 or eXe 3).
- The "Recent projects" is hidden (just remove the `exe-online` class in menuNavbar.html.twig to show it).
- If I show the "Recent projects" menu and I save a project, that project is there, but if I create a new project, the previous one disappears from the "Recent projects" list.
- I exported an elp that has image that I uploaded via Tools - File manager. The image is missing when I preview/export.

erseco:
Hi @ignaciogros,

I’ve fixed the issue with opening files — as I suspected, it was caused by Windows paths being different from mac/Linux.

Now the file picker in the desktop app also allows selecting both `.elp` and `.zip` files.

Regarding the "Recent projects" menu: handling this properly in the desktop app is not straightforward, so I think it can be postponed to a later iteration.

As for the missing image on export/preview, that problem is not related to this PR (it also happens in `main`), maybe related to #238

ignaciogros:
I works! Thank you, @erseco!

Only one suggestion to improve the solution:

Could the right extension (zip, elp or epub) be included in the file name if it's not typed?

If you don't type the extension when you save or export, the file is saved without it ("my-elp" instead of "my-elp.elp" or "my-website" instead of "my-website.zip" ). If you do that, you won't see the file in the "Open" dialog, because it'll only show zip and elp files.

ignaciogros (approved):
Great improvement! Thank you, @erseco.

I agree with you: the "Recent projects" menu can be added later.

I also think that removing the other version's elements from the HTML will be better than just using the `exe-offline` and `exe-online` classes. I'll open a new issue to check all that later.

There are still some small issues when saving, but maybe we can fix them later too:

* There's no automatic XML extension.
* If you export as SCORM (for example), you open the zip file and you save, it saves a zip file again, instead of an elp, but the content format is not SCORM, but an elp.
* The "elp" extension is hardcoded in some cases. If we use `Constants::FILE_EXTENSION_ELP`, changing the extension will be easier.

PR #289, opened 2025-09-17
⚠️ WIP — Admin panel & System preferences proposal

erseco:
I added a centralized, typed key–value store for runtime configuration: **SystemPreferences**. This replaces scattered per-user settings and many runtime values currently kept in `Constants.php`/`Settings.php`. The goal is a single, extendable place (registry + DB) to read/write config at runtime via **Admin UI**, **CLI**, or **API**.

**What I changed (high level)**

* DB migration: `system_preferences` KV table.
* Registry: `src/Config/SystemPref.php` + attribute `Setting` and `SystemPrefRegistry`.
* Service: `SystemPreferencesService` (cache-aware, typed get/set/delete).
* API: Admin-only endpoints under `/api/v2/system-preferences`.
* CLI: `app:prefs:sync`, `app:prefs:list`, `app:prefs:get`, `app:prefs:set`.
* Admin UI: EasyAdmin screens for Preferences (groups: `additional_html`, `theme`, `maintenance`) including file uploads for theme assets.
* Twig helpers: `sys_pref()` and `sys_pref_bool()` for templates.
* Maintenance handling: `MaintenanceSubscriber` uses prefs to return 503 for non-admins.
* Gradual deprecation note in `Constants.php`: runtime, user-visible options should move to prefs.
* Tests covering API, CLI, maintenance logic and admin pages.

**Why this approach**
A KV registry (key → typed value) keeps the number of entities minimal and lets us add settings without migrations for each new option. It behaves like WordPress options, Chrome/Firefox flags, or `about:config`: typed key/value, editable from UI/CLI/API, effective immediately (no restart).

**How to try locally**

1. Ensure migrations run (the migration creates the table).
2. Promote a user to admin locally:

```bash
make promote-admin EMAIL=user@exelearning.net
```

3. Open the admin: `http://localhost:8080/admin` (login as that admin).
4. Go to **Preferences** → `Additional HTML`, `Theme`, `Maintenance`. You can set login images, favicon, maintenance flag, inject HEAD/BODY/footer HTML, etc.

CLI & API examples:

```bash
# sync registry to DB
bin/console app:prefs:sync

# set maintenance on
bin/console app:prefs:set maintenance.enabled true --type=bool

# read via API (admin JWT)
GET /api/v2/system-preferences
PUT /api/v2/system-preferences/maintenance.enabled  { "value": false, "type": "bool" }
```

**Notes & impact**

* Migration is additive (creates a new table); no legacy data migration is included right now.
* Values are cached; `set()` invalidates cache so changes are immediate.
* File-type prefs are stored under `FILES_DIR/system_preferences_files` and exposed via `public/system_prefs` (symlink created by entry scripts).
* Admin-only API surface by default. Review the security rules if you want different scopes.
* This is an incremental migration path: move runtime-config keys out of `Constants.php` over time.

**What I want feedback on** (short, concrete)

* API shape and security (admin-only is current default).
* Key naming / grouping (`maintenance.*`, `theme.*`, `additional_html.*`) — good granularity?
* File storage strategy and public exposure (symlink + names with hash).
* EasyAdmin UX for types (is editing file/html/bool intuitive?).
* Anything I missed that should be a pref from day one.

I’ll attach three screenshots (admin dashboard, preferences list, maintenance form) to the PR. This is still **WIP** — tests and behaviour look stable but expect tweaks. Please review and tell me which areas you want me to refine first.

Captura de pantalla 2025-09-17 a las 17 15 28
Captura de pantalla 2025-09-17 a las 17 15 31

Captura de pantalla 2025-09-17 a las 17 15 38


ignaciogros:
Amazing work, @erseco!

I suggest testing it properly after v3.0.0 is released.

How about a new setting to write the Terms and Conditions for `$isLopdAccepted`? It'll need a WYSIWYG editor. But I'm not sure about this, because editing a template might be enough...

To review:

I enabled the maintenance mode and now I can't access /admin any more.

**Thank you!**

pabloamayab:
Wow @erseco !!!! 

I think it's great. 

When I edit a preference, upon saving changes, the app returns to a list with all preferences, but I am unable to access that list directly from the menu (my intuition leads me to click on the word PREFERENCES).

jgblasco:
Sorry, I got confused and wrote the comment in the wrong place. 
Very good work @erseco: 
- I agree with @pabloamayab , 
- also when you insert or edit a user, the password field says 'Create user' instead of 'Create pass'."

jgblasco:
I don't know if it's a common error, but I changed preferences and since then it won't let me start Docker and it gives me the following error: 

`failed to solve: failed to checksum file public/system_prefs: archive/tar: unknown file mode ?rwxr-xr-x`

jgblasco:
I've deleted it and it still won't let me do make up. I can boot from Docker Desktop.

PR #291, opened 2025-09-17
Close #290 - Enable the "Save as" option.

### LINKED ISSUE #290: Enable the "Save as" option
The online "Save as" option is hidden.

We should enable it, and check how the collaborative edition works with that.

ignaciogros:
@pabloamayab, @erseco:

The "Save as" option's been hidden for a long time. I've been testing it and I think that it works properly now, even with the collaborative edition.

Could you check it to confirm if it can be enabled?

Thank you.


ignaciogros:
> I supposed is the desired workflow, just to remark. If we are not allowing "Save as..." in collaborative edit, we maybe can disable the option on menu.

Thank you, @erseco.

I think it's the desired behavior. You can't save as while someone else is editing, but once you're alone again you can do it, so you can save a copy of the content. I think it makes sense. I suggest testing this when the collaborative edition enhancements are merged. Then we can see if that's to be hidden in that case.

PR #293, opened 2025-09-18
237 idevice ids mismatch between contentxml and exported html

pabloamayab:
Implement a temporary (and very simple) solution to address the mismatch of IDeviceID between content.xml and the exported HTML. This change modifies the ID generation logic to use existing IDs instead of generating new ones.

@mnarvaezm has already tested this solution and it has worked correctly.

I am going to open an issue to refactor the code because the idevice mapping is no longer necessary and the odeComponentsMapping data structure and the $odeSaveDto->setOdeComponentsMapping function are not needed.

It has not been removed yet in case this change could affect collaborative editing.


mnarvaezm:
Yes, no problem. In this PR the only change is the IDs of the iDevices saved in content.xml when exporting or downloading. You can merge it.
Later, I’ll update the iDevices that use id-resource, since it now matches the iDevice ID and is no longer needed. Then we’ll be able to remove that parameter.

mnarvaezm (approved):

Issue #233, opened 2025-08-29
PHP-generated common_i18n.js


ignaciogros:
common_i18n.js should be generated via PHP so the strings that it includes are in the package language.

PR #296, opened 2025-09-20
PHP-generated common_i18n.js (#233)

Ivan0886:
This pull request updates the way the commoni18n.js file is added to exports to do so dynamically.

To do this, a new helper class called CommonI18nUtil is created within the utilities folder with the strings prepared for translation extraction and resolved on the fly while exporting packages using the OdeExportService service.

Within OdeExportService, the copyBaseFilesToExportDir method is modified for dynamic generation, and a new method called generateCommonI18nJs is added.


ignaciogros (approved):
It works when you export, but /assets/vX/app/common/common_i18n.js (/app/common/common_i18n.js) should change when the package language is changed.

We can merge this PR and do that later. See #303

PR #301, opened 2025-09-21
142 style management errors

franmate:
Environment-based theme installation control.

- Restrict theme installation in online mode.
- Introduce ONLINE_THEMES_INSTALL env variable to override restriction.
- Update theme import logic to check environment configuration
- Implement permission flag validation for themes import.
- Disable import button when installation is not allowed.

.env
ONLINE_THEMES_INSTALL=0
The theme import/install process is completely disabled in online mode.
ONLINE_THEMES_INSTALL=1
Only themes with the necessary installation permissions will be imported/installed.

E2E test error: The export URL format was modified to include a hash directory. The test pattern needs to be updated to match the new structure: /export/[hash]/index.html instead of /export/index.html. @erseco

### DIFF
--- a/src/Controller/net/exelearning/Controller/Api/OdeApiController.php
+++ b/src/Controller/net/exelearning/Controller/Api/OdeApiController.php
@@ -740,9 +740,6 @@ class OdeApiController extends DefaultApiController
         }
         if (!empty($odeValues['theme'])) {
             $responseData['theme'] = $odeValues['theme'];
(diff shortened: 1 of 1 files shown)


ignaciogros:
Sorry about that, @franmate. I saw the new `ONLINE_IDEVICES_INSTALL` option and I hid that button too.

Please confirm if you want me to revert my commit and merge yours.

Thank you.

franmate:
Don't delete it. I'll modify IdeviceApiController and make a new commit before the PR.

PR #302, opened 2025-09-22
Refactor iDevice feedback handling and remove old rendering logic

pabloamayab:
Remove outdated HTML rendering for old iDevices and enhance feedback generation for text-type iDevices by adding feedback buttons and improving the structure of the generated HTML.

The same strategy has been followed in the translations as in eXe 2.9. However, in edit mode, eXe 3 does not take into account the translations that should be made.
Another minor issue is that in the task iDevice; task information in edit and display modes follows different styles: in edit mode it is aligned to the right and in display mode it is aligned to the left.

Consider removing the renderHtmlOldIdevice function in text.js.

### DIFF
--- a/public/files/perm/idevices/base/text/export/text.js
+++ b/public/files/perm/idevices/base/text/export/text.js
@@ -165,13 +165,6 @@ var $text = {
         const $node = $('#' + data.ideviceId);
         const isInExe = eXe.app.isInExe();
 
-        // Only render html old idevice
-        if (!isInExe && $node && $node.length === 1 && $node.find('.exe-text-template').length === 0) {
-            const html = $node.html();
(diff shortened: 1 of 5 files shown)


ignaciogros (commented):
I think it's OK, but @mnarvaezm's revision is required before merging this PR, as he knows better than me how that part should work.

359a0847002399ffb345fbfbd6cefac512d8b55e should fix the Task iDevice presentation issue, but there's still one problem there, because when you export a task created with eXe 2.9 without editing it, you get this HTML:

```
<div class="exe-text-activity">
    [...]
</div>
```

But when you export after saving the iDevice you get this one:
```

<div class="textIdeviceContent">
    <div class="exe-text-activity">
        [...]
    </div>
</div>
```

PR #305, opened 2025-09-22
feat(iDevices): enable/disable custom iDevices import feature.

franmate:
Added enabled/disabled feature for custom iDevices import in online mode.
Use the ONLINE_THEMES_INSTALL environment variable to override this restriction.

# Allow users to import/install iDevices
ONLINE_IDEVICES_INSTALL=0


ignaciogros:
If you prefer, we can revert those changes and use your solution instead. I that case, I suggest hiding the button using the `d-none` CSS class.

Thanks.

PR #323, opened 2025-09-24
Fix #322 - Desktop version can't load styles.

### LINKED ISSUE #322: Beta6 offline version can't load styles
It's probably because the new .env variables are not in main.js.

ignaciogros:


erseco:
Hi @ignaciogros as you said, the line 61 in `src/Helper/net/exelearning/Helper/ThemeHelper.php` was totally wrong, prefixing a DIRECTORY_SEPARATOR before the path, in unix systems you can set a path with double slash // and no problem but on Windows that results on **"\c:\users\..."** that is totally wrong, now the styles should be seen right in windows offline version

ignaciogros:
The PR is merged, and the pre-release is ready, including all its packages. The Windows version works OK, but I couldn't run the .deb one. I get these errors, even after uninstalling a previous version (beta2):

imagen

imagen

It might be my Ubuntu setup.

@franmate, @erseco:

Please confirm if it works for you. It'd be good to test de Mac version too.

It they work for you, I think that we should open that release and start working on the next one.

Thank you.

ignaciogros:
Just as you told me, @erseco:
It works on Ubuntu and Windows after merging #327.
Please confirm if it woks on Mac too. If it does, we can open the pre-release.
Thank you!

PR #324, opened 2025-09-24
Fix wrong iDevice order on pages

mnarvaezm:
Corrected an issue where iDevices were displayed in the wrong order on pages.


pabloamayab (approved):
Tests have been carried out on adding 
- iDevices by dragging and clicking from the side panel
- Clicking from the central button to add a text-type iDevice 
- Moving blocks and iDevices
- Duplicating pages, blocks, iDevices, etc. 

It works correctly. Great work @mnarvaezm !!!!

Issue #151, opened 2025-07-13
Refactor Sharing Mechanism to Use ode_id as Primary Identifier


abertranb:
# Problem Statement

The current sharing mechanism has a critical flaw where it relies on `ode_session_id` as the primary identifier for shared sessions. This creates several issues:

1. **Session Confusion**: Multiple users can have different `ode_session_id` values for the same ODE, leading to data inconsistency
2. **History Desynchronization**: The `ode_session_id` field causes stories to become misaligned when shared
3. **Complex Session Management**: The current system requires complex synchronization between `ode_session_id` and `ode_id` across multiple tables

# Current Architecture Issues

## Database Structure Problems

The current system uses `ode_session_id` across multiple tables:
- `current_ode_users` (primary key for sharing)
- `ode_nav_structure_sync` 
- `ode_pag_structure_sync`
- `ode_components_sync`
- `ode_properties_sync`
- `ode_operations_log`

This creates a complex web of dependencies where session management becomes unreliable.

## Sharing Flow Issues

1. **URL Generation**: Currently generates URLs with `ode_session_id` as `shareCode`
2. **Session Joining**: Users join via `ode_session_id` which may not correspond to the correct `ode_id`
3. **Data Synchronization**: Multiple tables need to be updated when session IDs change

# Proposed Solution

## Use `ode_id` as Primary Sharing Identifier

Change the sharing mechanism to use ode_id instead of ode_session_id:

The key insight is to maintain ode_session_id for initial access but then transition to using ode_id as the primary identifier for all operations.

### Sharing Flow

* Initial Access: User accesses via ode_session_id (for backward compatibility)
* Session Validation: System validates the session and retrieves the corresponding ode_id
* Transition to ode_id: Once validated, all operations use ode_id as the primary identifier
* Consistent Operations: All subsequent operations (editing, saving, syncing) use ode_id

erseco:
This is a fantastic initiative. Switching to `ode_id` as the primary sharing identifier will help us support multiple simultaneous sessions, fixing a major limitation we currently face due to the single shared context.
It might also be a good opportunity to consider using UUIDs for `ode_id` instead of the current predictable format, which would enhance security and reduce collision risks.
In the long run, this change could allow us to reduce redundancy (like `SyncUpdateFlag` and similar constructs), simplifying both the data model and the REST API design #62. Looking forward to seeing this evolve!

erseco:
Fixed in #343

ignaciogros:
[...]

PR #343, opened 2025-10-04
⚠️ WIP — Enhance project and session management to support multiple open documents across tabs and offline mode

erseco:
This pull request introduces a complete update to the project and session management system, aimed at improving the handling of multiple open documents in different browser tabs or windows, both online and offline. The main goal was to modernize session initialization and make document identification explicit and persistent through URL parameters.

**Session and Project Management**

* Added support for multiple independent sessions per user, allowing different documents to be opened simultaneously in separate tabs or windows.
* Introduced explicit `projectId` and `odeSessionId` parameters in URLs to identify each document session.
* Improved session initialization logic in `projectManager.js`, including robust handling for existing, missing, or new sessions, and added support for forced session creation when needed.

**User Experience Enhancements**

* Document titles now include the project name in the browser tab for easier identification when multiple projects are open.
* “New project” actions now open in a new browser tab using the correct base path, ensuring consistent navigation between sessions.
* Share links now include both session and project identifiers for more reliable collaboration links.

These changes resolve a long-standing limitation in the application where multiple documents could not be managed independently across tabs or offline instances, laying the foundation for a more stable and predictable multiwindow experience.

### DIFF
--- a/public/app/rest/apiCallBaseFunctions.js
+++ b/public/app/rest/apiCallBaseFunctions.js
@@ -97,25 +97,48 @@ export default class ApiCallBaseFunctions {
      * @returns
      */
     async doAjax(url, method, data, waiting = true) {
-        if (waiting) this.addWaitingPetition();
-        let response = {};
-        response = await $.ajax({
-            url: url,
-            method: method,
-            data: data,
-            timeout: eXeLearning.config.clientCallWaitingTime,
-            dataType: 'json',
-            success: function (response) {
-                return response;
-            },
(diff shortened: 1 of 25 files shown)


erseco:
This is related to #151

erseco:
@ignaciogros @pabloamayab maybe we can research to only need one parameter like "project" instead of using two

ignaciogros (approved):
It looks good to me, but I suggest testing this in real conditions after rc1 is released.

PR #372, opened 2025-10-08
Implementation of collaborative mode in new interface

JMTomasZafra:
- Modal Preferences Layout
- Properties Accordion, close the bottom two and the button at the top
- Project TITLE layout breaks if the text is too long
- Style/editing inconsistencies
- Collaborative mode implementation
- Collaborative level settings for page or device
- Page lock and avatars on pages


.env.dist
erseco, line 1: This file has no actual changes. You can remove it from the PR with: `git checkout main .env.dist` thanks!

.gitignore
```diff
@@ -1,6 +1,7 @@
 ###> symfony/framework-bundle ###
 .env
 /.env
+/.env.dist
```
erseco: ⚠️ Please **don’t delete** the `.env.dist` file — it’s important for defining default environment variables and should stay in the repository.

docker-compose.yml
```diff
@@ -38,8 +38,32 @@ services:
...
+    restart: unless-stopped
+    environment:
+      MARIADB_ROOT_PASSWORD: "${MARIADB_ROOT_PASSWORD:-root}"
+      MARIADB_DATABASE: "${DB_NAME:-exelearning}"
+      MARIADB_USER: "${DB_USER:-exelearning}"
+      MARIADB_PASSWORD: "${DB_PASSWORD:-exelearning}"
+    ports:
+      - "${DB_PORT:-3306}:3306"
+    volumes:
+      - mariadb-data:/var/lib/mysql
```
erseco: Please remove the `mariadb` service, its `depends_on` entry, and the mariadb-data volume. This `docker-compose is just for local development and should use SQLite, so these local resources are no longer required.

PR #380, opened 2025-10-09
Feature/new markdown idevice

erseco:
## Summary

Adds a new **Markdown Text iDevice** that lets authors compose content in Markdown with a tabbed write/preview editor. The Markdown engine is the project's existing Showdown bundle, hardened a bit so that GitHub-style features and LaTeX both work end-to-end.

## What's in this PR

- New iDevice under `public/files/perm/idevices/base/markdown-text/` with edition (form + live preview) and export (runtime render + feedback toggle).
- `eXe.app.common.markdownToHTML` now enables Showdown's GFM options (tables, task lists, strikethrough, GFM-style sublist indentation) and protects LaTeX delimiters from being mangled by Markdown processing.
- `LIBRARY_PATTERNS.exe_highlighter` extended to also detect Showdown's `language-<lang>` classes, so Prism gets bundled in preview/export wherever fenced code blocks exist (not only TinyMCE-flagged blocks).
- Unit tests (vitest) for edition and export, plus a Playwright E2E spec covering add → edit → save → reload, preview tab, feedback rendering and attribute-escape behaviour.

## Improvements vs the initial version

- **GFM rendering**: tables, task-list checkboxes, strikethrough now render correctly.
- **Code blocks**: monospace fonts and `text-shadow: none` so dark-themed code doesn't look bold; Prism kicks in for fenced blocks in workarea, preview and export.
- **Lists**: ordered, unordered and nested lists work consistently.
- **LaTeX**: `\(...\)`, `\[...\]`, `$$...$$` and `\begin{}...\end{}` are preserved through the markdown pass and rendered by the existing MathJax integration in `renderBehaviour`.
- **Security**: user-supplied values that land inside HTML attributes (feedback button text, duration / participants labels and values) are HTML-escaped.
- **Bug fixes carried in**: LaTeX rendering no longer skipped when there's no feedback button, and the no-op mutation of the `data` argument inside `getHTMLView` (inherited from the text iDevice) is gone.

## How to test

Add the new "Markdown text" iDevice to a page, paste this content into the editor, and confirm the preview tab plus the saved view both render it as expected:

````markdown
# Main Title

## Subtitle

This is a paragraph with **bold text**, *italic text*, ***bold and italic text***, `inline code`, a [link](https://example.com), and an image:

![Random image from Picsum](https://picsum.photos/200/300)

---

## Lists

### Unordered List

- First item
- Second item
  - Nested item
  - Another nested item
- Third item

### Ordered List

1. First step
2. Second step
3. Third step

---

## Checkboxes

- [x] Create the repository
- [x] Add the README file
- [ ] Configure CI/CD
- [ ] Publish the first release

---

## Table

| Name | Role | Status | Priority |
|---|---|---:|:---:|
| Alice | Backend Developer | In progress | High |
| Bob | Frontend Developer | Pending | Medium |
| Carol | DevOps Engineer | Completed | Low |

---

## Quotes

> This is a quote.
>
> It can span multiple lines and highlight important ideas.

---

## Code Blocks

### Bash

```bash
#!/usr/bin/env bash

echo "Hello, world!"
```

### JSON

```json
{
  "name": "example-project",
  "version": "1.0.0",
  "private": true
}
```

### PHP

```php
<?php
/**
 * Return a greeting message.
 *
 * @param string $name User name.
 * @return string
 */
function get_greeting( string $name ): string {
    return sprintf( 'Hello, %s!', $name );
}
```

---

## Collapsible Details

<details>
<summary>Show more information</summary>

This content is hidden until the user expands the block.

- It can contain lists
- Code
- Tables
- Rich text

</details>

---

## Formula

A simple formula:

```text
total = price * quantity
```

LaTeX example for platforms that support it:

\(\cfrac{1}{1 + \cfrac{1}{2 + \cfrac{1}{3}}}\)

---

## Final Line

End of document.
````

## Test plan

- [x] `make fix && make test-unit && make test-integration` are green.
- [ ] `make test-e2e` (or at least `markdown-text.spec.ts` under chromium) passes.
- [ ] Manual smoke with the markdown above: tables render, task-list checkboxes appear, fenced code is colourised in workarea / preview / export, the LaTeX continued-fraction renders via MathJax, and saving + reloading the project preserves everything.

### DIFF
--- a/public/app/common/app_common.js
+++ b/public/app/common/app_common.js
@@ -61,12 +61,40 @@ export default class Common {
   }
 
   /**
-   * Markdown to HTML converter
+   * Markdown to HTML converter.
+   *
+   * LaTeX delimiters (\(...\), \[...\], $$...$$, \begin{...}\end{...}) are
+   * stashed before Showdown runs so that markdown processing does not eat
+   * underscores, asterisks or backslashes inside formulas. They are restored
+   * verbatim afterwards so MathJax can pick them up at render time.
    */
   markdownToHTML(content) {
-    var converter = new showdown.Converter();
-    converter.setOption('noHeaderId', true);
-    return converter.makeHtml(content);
+    var src = String(content == null ? '' : content);
+    var store = [];
+    [
+      /\\\[[\s\S]*?\\\]/g,
+      /\$\$[\s\S]*?\$\$/g,
+      /\\begin\{[^}]+\}[\s\S]*?\\end\{[^}]+\}/g,
+      /\\\([\s\S]*?\\\)/g,
+    ].forEach(function (re) {
+      src = src.replace(re, function (match) {
+        store.push(match);
+        return 'EXELATEXBEGIN' + (store.length - 1) + 'EXELATEXEND';
+      });
+    });
+
+    var converter = new showdown.Converter({
+      noHeaderId: true,
+      tables: true,
+      tasklists: true,
+      strikethrough: true,
+      disableForced4SpacesIndentedSublists: true,
+    });
+    var html = converter.makeHtml(src);
+
+    return html.replace(/EXELATEXBEGIN(\d+)EXELATEXEND/g, function (_, i) {
+      return store[Number(i)] !== undefined ? store[Number(i)] : _;
+    });
   }
 
   /**
--- a/public/app/common/app_common.test.js
+++ b/public/app/common/app_common.test.js
@@ -264,6 +264,28 @@ describe('Common', () => {
       const result = common.markdownToHTML('test');
       expect(typeof result).toBe('string');
     });
+
+    it('preserves inline LaTeX \\(...\\) intact', () => {
+      const result = common.markdownToHTML('Equation \\(x^2 + y_1\\) here');
+      expect(result).toContain('\\(x^2 + y_1\\)');
+    });
+
+    it('preserves display LaTeX $$...$$ intact', () => {
+      const result = common.markdownToHTML('Block: $$\\int_0^1 dx$$ end');
+      expect(result).toContain('$$\\int_0^1 dx$$');
+    });
+
+    it('preserves \\[...\\] block math intact', () => {
+      const result = common.markdownToHTML('Block: \\[x_1 + x_2\\] end');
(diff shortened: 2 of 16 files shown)


pabloamayab:
Great work, Ernesto. I have two questions:
- In order for the iDevice to appear, I had to do a make clean and delete volumes. We should test whether it would appear in the interface when installing a new version in an offline installation.
- I don't understand why task information appears, and I'm not even sure if the feedback area makes sense.
- 
image

erseco:
Exactly.

The new *Markdown Text* iDevice inherits part of the structure and logic from the regular Text iDevice, which is why the task/feedback section appears. It’s not new behavior — it’s just a quick copy of that iDevice. For now, it can remains this way to keep compatibility with how eXe handles text-based activities, but we can review it later if it doesn’t make sense for Markdown.

Regarding the visibility issue — yes, you need to run `make clean` locally because the iDevice configuration files (`config.xml` and the cached iDevice JSON files) are stored in cache. In a clean installation or when volumes are rebuilt, it should appear automatically, even in offline mode.

We could also add a check or cleanup step to the build process to ensure it always loads correctly.

pabloamayab:
Hi @erseco. The only issue is with the feedback button when exporting or previewing, because the button doesn't do anything.

erseco:
Hi @pabloamayab — thanks for catching this!

The feedback button was broken in published exports/previews because the browser-side iDevice config shim (`src/shared/export/browser/idevice-config-browser.ts`) had a hardcoded list of JSON-type iDevices, and `markdown-text` was not in it. Without `componentType: "json"` on the rendered node, `exe_export.js` never reached the branch that wires the click handler, so the button looked right but did nothing.

Fix: added `markdown-text` to the `jsonIdevices` list in that shim, plus a unit test asserting `getIdeviceConfig('markdown-text').componentType === 'json'` and an E2E test that adds a markdown iDevice with feedback content, opens the preview iframe and verifies the toggle actually shows and hides the panel.

I also squashed the branch to a single commit on top of main and force-pushed.

pabloamayab (approved):
LGTM!

PR #390, opened 2025-10-10
Feature/Implement session expiration detection and automatic login redirect

erseco:
This PR adds a backend endpoint and frontend logic to detect expired user sessions.
When the session is no longer valid, the app automatically disconnects the active Mercure connection (if any) and redirects the user to the login page to prevent console errors and invalid states.


ignaciogros (approved):

ignaciogros:
Hi @erseco,

This is already merged, and It should be working on "exemain", but after a long time without activity I'm not being redirected.

It might just be the session expiration time.

Could you please check that?

Thank you.

ignaciogros:
:-D Thanks!

PR #415, opened 2025-10-15
Use twig for dynamic background images on login page

pabloamayab:
Implement a system to randomly generate background images for the login page using Twig, enhancing visual variety. 

Some questions:
- These images are examples; you could select a theme every X times or version (birds, animals, cities, landscapes, etc.).
- The process of generating random images could have been done using PHP or JS, so you need to decide whether this method with the Twig template is right for you.


ignaciogros (changes requested):
Amazing pictures, @pabloamayab! I guess they are yours... Thank you for sharing them!

I only see a problem in this solution:

You need 9 pictures in that folder, and that number might change. 

I don't think that Twig can do that without PHP.

Sometimes I've used a solution like `background-image: url('rotator.php');`

And rotator.php returns the path of a random image included in a folder (/public/images/ sin this case).

That'd be good, because you can upload as many images as you want in that folder, without even caring about their names.

There are many open issues, so if you don't have time to improve that now, maybe we can use any of those images to replace /public/style/workarea/images/exe-login-image.jpg. As you prefer.

Thank you!

PR #421, opened 2025-10-15
Added css generation in Makefile and pipeline

erseco:
This pull request introduces improvements for managing SCSS/CSS assets in the project, focusing on better build processes and developer experience. The main changes include new Makefile commands for building and watching CSS, updates to the CI workflow to ensure CSS is built during production builds, and enhancements to Docker Compose for development workflows. Additionally, `.gitattributes` is updated to properly handle SCSS files.

**SCSS/CSS build and workflow improvements:**

* Added new Makefile targets: `css`, `css-dev`, and `css-watch` for building production and development CSS, and for watching SCSS changes during development. These use a Dockerized Node/Sass setup for consistency.
* Updated the `Makefile` help output to document the new CSS-related commands for easier discoverability.
* Updated the CI workflow (`.github/workflows/ci.yml`) to run `make css` and ensure production CSS is built as part of the build process.

**Docker Compose development enhancements:**

* Improved the `docker-compose.yml` Sass service for development: added environment variables for reliable file watching and switched to expanded output with source maps for easier debugging. The service is now only included in the "dev" profile.

**File handling and configuration:**

* Updated `.gitattributes` to treat `.scss` files as text with LF line endings, ensuring proper handling in version control.

### @ignaciogros @JMTomasZafra maybe we should always generate the main.css and avoid pushing it to the repository, what do you think about?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -528,7 +528,7 @@ pull-vendor: check-docker check-env upd
 	@echo "✅ Done. Local ./vendor directory updated from container."
 
 
-.PHONY: css css-prod css-watch
+.PHONY: css css-prod
 
 # Prevent MSYS path conversion breaking Docker paths on Windows Git Bash
 # (same approach used elsewhere in this Makefile for docker compose commands)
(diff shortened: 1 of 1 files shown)


ignaciogros (changes requested):
Very good idea, @erseco.

I've tested the new commands, but they don't work on Windows:

```
$ make css
docker run --rm -v C:/.../exelearning:/app -w /app node:24-alpine sh -lc "npm i -s --no-fund sass && npx sass assets/styles/main.scss public/style/workarea/main.css --style=compressed --no-source-map"
docker: Error response from daemon: the working directory 'C:/Program Files/Git/app' is invalid, it needs to be an absolute path
```

**About not saving main.css:**

See https://github.com/exelearning/exelearning/pull/421#issuecomment-3409800723

~I think that some developers might find useful to have it in the original source so they can fix any CSS problem without using SAAS. It's saved right now, so I'll keep it for the moment.~

erseco:
I added a patch for work on windows

ignaciogros (changes requested):
`make css` and `make css-dev` work on Windows now, but main.css does not change when editing the SCSS files after running `make css-watch`.

If that's difficult to fix now, we can merge it as it is, or even removing that command, because editing the CSS file and pasting the right code in the SCSS one when you've finished would be enough.

Thank you!

ignaciogros:
I was wrong...

main.css and its map cause many conflicts. 

I agree with you, @erseco (and @JMTomasZafra too): 

We should avoid pushing main.css to the repo.

PR #459, opened 2025-10-19
Hotfix/reorder text idevice elements

erseco:
This pull request makes a small adjustment to the order of elements in the HTML generated by the `text.js` file. The textarea is now placed before the first `.exe-parent` container, instead of after it.

@mnarvaezm I don't know if this can break anything, if you consider you can discard it


mnarvaezm:
@erseco 

No, that doesn’t break anything, but I think it’s a bit inconsistent with the behavior of the other iDevices. In the others, during export, the elements are shown in the same order as in the editor: instructions (minimized), options or relevant game information (visible), and then the content or feedback (minimized).

With the change you made, in the editing area the user might get the impression that the task information will appear between the text and the feedback in the export. I think it would be more consistent to keep the current order: Task information (optional, minimized) → text (maximized) → feedback (minimized)
But that’s just my opinion, it’s not really a big deal.

PR #460, opened 2025-10-19
Feature/tabs

erseco:
This pull request introduces significant improvements to multi-tab and window management in the Electron app, with a focus on providing a native tabbed experience on macOS and consistent behavior across platforms. It adds macOS-specific handling for native tabs, ensures safe window/tab closing, and introduces a small preferences helper for persistent settings. The changes also refactor how new windows and tabs are created and managed.

**Platform-specific window/tab management:**

* On macOS, new tabs are opened as native tabs using `tabbingIdentifier`, and the native tab bar is forced to appear at least once. The `createMacTab` function is introduced to create tabs, and `bindMacCloseSafety` ensures tabs/windows can always be closed, even if the page tries to block unload. (`main.js`)
* On Windows and Linux, new windows are created as independent `BrowserWindow` instances with consistent sizing and behavior. The force-close logic is retained for these platforms to ensure windows close even if a page tries to prevent it. (`main.js`) 

Captura de pantalla 2025-10-19 a las 19 13 08

@ignaciogros  I can investigate to add tab support also on Windows and Linux, if you think can be interesting

### DIFF
--- a/src/Service/net/exelearning/Service/Export/ExportSCORM12Service.php
+++ b/src/Service/net/exelearning/Service/Export/ExportSCORM12Service.php
@@ -131,7 +131,7 @@ class ExportSCORM12Service implements ExportServiceInterface
         $odeId = $this->currentOdeUsersService->getOdeIdByOdeSessionId($user, $odeSessionId);
 
         // imslrm.xml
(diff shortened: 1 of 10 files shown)


ignaciogros:
We close this until the multi-doc compatibility is fixed. See #343

PR #465, opened 2025-10-20
Use switches instead of checkboxes in Project, Page, and iDevice properties

mnarvaezm:
Replace checkbox inputs with switch-style toggles across: 
-  Project properties
-  Page properties 
- iDevice properties


ignaciogros (changes requested):
Some of them are not working in these dialogs (I'm testing on Firefox):

* Page properties.
* Box properties.

Could you please check them?

Thank you!

PR #471, opened 2025-10-20
fixes for revision iii

JMTomasZafra:
1. We discussed in the last meeting that this is not possible with the current collaborative architecture, since there is nothing to persist the events that Mercure executes to consult them in each user interaction.
2. Ook, seen, and we also apply the solution within the styles
3. Solved
4. We have uploaded the styles that solve this and other errors, in any case Javi will send you an email with the 8 styles (4 without logo and 4 with logo), which you can replace in the base folder.

### DIFF
--- a/public/app/workarea/menus/navbar/items/navbarUtilities.js
+++ b/public/app/workarea/menus/navbar/items/navbarUtilities.js
@@ -270,43 +270,82 @@ export default class NavbarFile {
      * @returns {String}
      */
     json2Csv(objArray, headerTitles) {
-        var array =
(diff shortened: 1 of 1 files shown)


ignaciogros:
I think It's just a JS error in menuIdevicesBottom.js.

JMTomasZafra:
> I think It's just a JS error in menuIdevicesBottom.js.

Screenshot_20251020-165814

I think it's because of this line that the string for the button to change favorite devices needs to be added. I started to do it and got stuck (I had a day working on several projects at once, and it slipped my mind). Changing that to the appropriate string should work. (Sorry, but I don't have a computer nearby to change it myself)

Issue #457, opened 2025-10-19
Fix string extraction


ignaciogros:
Some string are not found when running make translations.

We should revert See 8ad125a6656ee2e2ee1246073a5b4bcef8b05eb8 after this is fixed.

**Here's how to test it:**

- Improve `extractMessagesFromJs` in **JsTranslationExtractor.php**.
- Leave only `'en'` in `LOCALES` (**Settings.php**), because the extraction process is quite slow and that'd be better for testing.
- Run `make translations` and check if any new string included in **map/edition/map.js** is there.

ignaciogros:
We should also fix some original strings, like " Reformat".

PR #483, opened 2025-10-21
457 fix string extraction

mnarvaezm:
- Add missing translations for Map and AZ-Quiz-Game iDevices.
- Remove tmpStringList function.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -3706,16 +3706,16 @@
         },
         {
             "name": "symfony/form",
-            "version": "v7.3.3",
+            "version": "v7.3.4",
             "source": {
                 "type": "git",
                 "url": "https://github.com/symfony/form.git",
-                "reference": "f151b4a027fa67769268b80111f7fdb63edb5e79"
+                "reference": "7b3eee0f4d4dfd1ff1be70a27474197330c61736"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/symfony/form/zipball/f151b4a027fa67769268b80111f7fdb63edb5e79",
-                "reference": "f151b4a027fa67769268b80111f7fdb63edb5e79",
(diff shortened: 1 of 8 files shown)


ignaciogros (changes requested):
Thank you for changing that, @mnarvaezm.

If you delete **messages.en.xlf** and run `make translations` you get an XLF file in which some required strings are missing:

- Create image maps: Images with interactive hotspots to reveal images, videos, sounds, texts...
- Show questionnaire
- Show active areas
- Click here to do this activity
- Select the correct options and click on the "Reply" button.
- Mark all the options in the correct order and click on the "Reply" button.
- Write the correct word o phrase and click on the "Reply" button.
- Click on
- Everything is perfect! Do you want to repeat this activity?
- Great! You have passed the test, but you can improve it surely. Do you want to repeat this activity?
- Almost perfect! You can still do it better. Do you want to repeat this activity?
- It is not correct! You have clicked on
- and the correct answer is
- Great! You have visited the required dots.
- You can do the test.
- Select a subtitle file. Supported formats:
- Map
- Return
- Questionnarie
- Arrow
- Map marker
- Do you want to save the changes of this presentation?
- Do you want to save the changes of this quiz?
- Do you want to save the changes of this map?
- Provide a slide title.
- Hide score bar
- Play the sound when scrolling the mouse over the points.
- Show when the mouse is over the icon or active area.
- Hide areas
- Provide a word and its definition. May toggle between: "Word starts" or "Word contains", by clicking on %s

Most of those strings (it not all) are in **edition/map/map.js**.

So **the string extraction process is not working properly** (See  `extractMessagesFromJs` in **JsTranslationExtractor.php**).

I think that we shouldn't merge this before v3.0.0 is released. Apart from those problems, some collaborators are already working on the translation of the current XLF files.

I'll close this PR for the moment and keep #457 open.

PR #507, opened 2025-10-22
 Exclude data-idevice-json-data and data-idevice-template attributes for text type idevices in all export formats

mnarvaezm:
Exclude data-idevice-json-data and data-idevice-template attributes for text type idevices in all export formats


ignaciogros (approved):
I think it's OK. Thank you, @mnarvaezm.

Just one thing:

I removed the `id-resource` for the Text iDevice. 

**Please review my changes.**

In fact, we should consider removing or renaming it to `data-id-resource` in all the iDevices to avoid invalid HTML.

PR #528, opened 2025-10-24
Fix: Correct file naming on export

franmate:
This PR fixes a bug where the exported file names were generated using incorrect characters/symbols


ignaciogros (approved):
Thank you, @franmate!

You can still use special characters that generate HTML files with names like ⚠-aviso-⚠.html

I think that we can merge this PR now and try to fix that later.

Replacing them will avoid some problems. Maybe we could define a whitelist and remove or replace anything else. It's not important now.

PR #544, opened 2025-10-25
Initialize FX effects on Text iDevice save. Fix 533

mnarvaezm:
Initialize FX effects on Text iDevice save.


ignaciogros (approved):
Thank you!

PR #545, opened 2025-10-25
Mermaid diagrams rendering support via TinyMCE plugin

### LINKED ISSUE #474: New Diagrams iDevice
Create an iDevice to generate diagrams from markdown-like text. See https://github.com/mermaid-js/mermaid

lrlopez:
This PR adds a new plugin into TinyMCE that allows pasting [Mermaid's](https://mermaid.js.org/intro/) [text definitions](https://mermaid.js.org/intro/syntax-reference.html) into any HTML element. Those definitions will render into SVG images.

Mermaid is a BSD-licensed diagramming and charting tool that creates different kind of images from a text definition, i.e. flowcharts, class diagrams, gantt diagrams, timelines, mindmaps, E/R diagrams, kanban boards, user journeys, network packets diagrams, etc.

When exporting/previewing/printing HTML that contains Mermaid content, library will be also exported so that rendering will still work.

Limitations:
- Diagrams are always stored as text definitions within the HTML code, not as images. Loading mermaid.js is needed for rendering.
- No image fallback when JS is not available.

image

image

image

image

Closes #474

### DIFF
--- a/src/Controller/net/exelearning/Controller/Api/ParameterApiController.php
+++ b/src/Controller/net/exelearning/Controller/Api/ParameterApiController.php
@@ -317,8 +317,6 @@ class ParameterApiController extends DefaultApiController
         $this->translator->trans('The exported content will be editable with eXeLearning.');
         $this->translator->trans('"Made with eXeLearning" link');
         $this->translator->trans('Help us spreading eXeLearning. Checking this option, a "Made with eXeLearning" link will be displayed in your pages.');
-        $this->translator->trans('Include a copy of the source file');
-        $this->translator->trans('A copy of the eXeLearning file will be included when exporting the content.');
         $this->translator->trans('Page counter');
         $this->translator->trans('A text with the page number will be added on each page.');
         $this->translator->trans('Search bar (Website export only)');
--- a/src/Properties.php
+++ b/src/Properties.php
@@ -495,14 +495,6 @@ class Properties
             'category' => 'properties',
             'groups' => ['export'],
         ],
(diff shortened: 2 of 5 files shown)


lrlopez:
> I think it's ready to be merged, but we still don't know if a new iDevice instead of a TinyMCE plugin will be better for that.

I've been working on both possibilities. In fact, I've also coded an specific iDevice for Mermaid diagrams. It still needs some minor adjustments, but I could send a PR for you to review and compare.

Maybe it's because of my personal needs, but I find more useful interleaving diagrams among paragraphs of text and source code blocks inside the same text iDevice. Also, embedding diagrams into TinyMCE allows inserting them on many places, like DUA content, games, etc.

But I also do understand that an independent iDevice makes sense. It'd allow better customization and is a direct approach for those who need just to insert a diagram in a page.

Hmmmm... what if we could have them both ways? I mean, we could keep the plugin while implementing the feature as an independent iDevice.

I will get the iDevice PR ready just in case. Thanks!

ignaciogros:
OK!

Thanks a lot, @lrlopez.

mnunezcedec:
Hello @ignaciogros 

Something seems to happen somewhat randomly with diagrams using Mermaid code.

In the online version of eXeLearning, everything works fine: you enter the code, save it, and the corresponding diagram is displayed immediately.

However, in the desktop version (4.0.3), when you save, the diagram does not appear immediately. After some time, it eventually appears and everything works correctly.

In short, it seems that in the desktop versions there is a delay while the diagram is being generated or rendered.

PR #547, opened 2025-10-26
Internal links are not working in exe

franmate:
This PR resolves the internal link malfunction.


ignaciogros (approved):
Thank you, @franmate!

PR #577, opened 2025-10-28
Add support for multimedia source extraction in OdeXmlUtil

pabloamayab:
Enhance OdeXmlUtil to extract sources for video, audio, images, and pictures, improving multimedia handling capabilities.


pabloamayab:
@ignaciogros Do you think there could be another tag with src?

ignaciogros:
Hi @pabloamayab,

`source` is OK for audio, video and picture. 

They accept more than one `source`, but eXe is not supporting editing that for the moment. In fact, a `picture` tag can't be editted.

With `src`, apart from the ones that are already included:

- `iframe`
- `track`
- `embed`
- `input type="image"` (not supported)

Thank you!

ignaciogros (approved):
Fixed! Thank you, @pabloamayab.

PR #580, opened 2025-10-30
iDevice and box title simultaneous edition.

ignaciogros:
@pabloamayab, @erseco:

Do you think that checking if someone else is editing the box title or the icon is necessary?


erseco (approved):
This functionality is exceptionally effective and addresses a common and frustrating issue encountered during the editing process. IMHO right now is not necessary to check if another using is editing the title. Maybe in the future as many users use the collaborative feature.

PR #582, opened 2025-10-30
Style review and documentation

ignaciogros:


ignaciogros:
Hi @pabloamayab, @erseco, @jesusbautistaq:

This PR includes a Style review (cleanup, organization, and formatting) and **documentation for creating new styles**.

**Please review the documentation and let me know if you miss anything.** It's just an explanation for designers. No information about where are the styles stored is included.

**Notice that all styles are downloadable now**, because you might want to use any of them as a base to create your own style. Creating a style from scratch will be more difficult.

**Important:**

I didn't include in the code a tool to make style designing easier, but It could be interesting for style developers. I attach an example: [style-designer.zip](https://github.com/user-attachments/files/23283739/style-designer.zip). Just open index.html on your browser to see it.

You can switch between different views while you edit your style (just change something in /theme/content.css to see how it works).

If you put that folder in eXe's public directory you can even see how the style looks inside eXe.

The problem is that maintaining that tool will be quite difficult, because the HTML of the exports and eXe might change.

If you think that including that in eXe is interesting, we should automate the creation of those files. Another option will be creating a new repo for this kind of tools, or just sharing them with the developers.

ignaciogros:
After merging this changes into main we should check if the previous CSS and JS files are removed from the server in all versions.

See 8d3a64e2ca4052e0de95cfcbca3eef796aac15c7 and 3dc3829dd66440eb04a6a51865302a25a25656bd.

ignaciogros:
I broke some tests after renaming the styles CSS files. 

I'd like to rename those files, but if it can cause problems we can keep the previous names.

I can't see right now why the tests are failing.

erseco (approved):
I downloaded many styles and It worked! The guide is also ok! GJ

ignaciogros:
> The tests were not actually failing. After each test, the system checks that there are no errors in the browser console, and in this case TinyMCE was trying to load the `content.css` file from the selected theme (not sure why). I renamed it to `style.css`, and now all tests pass.

My fault. I should've seen that. I'll have to update the documentation to avoid those errors.

TinyMCE loads that file so you can see what you edit with the final design. You can also select CSS clases included in the style.

Thanks, @erseco!

ignaciogros:
eXeLearning Style Designer initial version: https://github.com/ignaciogros/exelearning-style-designer

ignaciogros:
@erseco, @pabloamayab:

This PR's been merged, but we should check if It's OK.

After merging it, the "exemain" online version should:

- Allow downloading all predefined styles (you can't download them).
- Use style.css instead of content.css in all styles (content.css is still used).
- Use style.js instead of default.js in all styles (default.js is still used).

We should test the desktop versions too.

It changing those files name can cause any problems, maybe we should rever those changes.

PR #589, opened 2025-11-01
Mermaid diagrams rendering support via iDevice

### LINKED ISSUE #474: New Diagrams iDevice

lrlopez:
### [This is an alternate version of PR #545, which is implements a TinyMCE plugin instead of a new iDevice.]

This PR adds a new iDevice that allows pasting [Mermaid's](https://mermaid.js.org/intro/) [text definitions](https://mermaid.js.org/intro/syntax-reference.html) into any HTML element. Those definitions will render into SVG images.

Mermaid is a BSD-licensed diagramming and charting tool that creates different kind of images from a text definition, i.e. flowcharts, class diagrams, gantt diagrams, timelines, mindmaps, E/R diagrams, kanban boards, user journeys, network packets diagrams, etc.

When exporting/previewing/printing HTML that contains Mermaid content, library will be also exported so that rendering will still work.

Limitations:

- Diagrams are always stored as text definitions within the HTML code, not as images. Loading mermaid.js is needed for rendering.
- No image fallback when JS is not available.

image
image
image

Closes #474

### DIFF
--- a/.env.dist
+++ b/.env.dist
@@ -86,15 +86,17 @@ API_JWT_SECRET=dev_secret_change_me
 # API_JWT_AUDIENCE=exelearning_clients
 ###< API JWT (local Bearer tokens) ###
 
+###> symfony/framework-bundle ###
+# See: https://symfony.com/doc/7.3/deployment/proxies.html
+# ATTENTION: if the app is behind an SSL reverse proxy, set the proxies below.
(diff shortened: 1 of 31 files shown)


erseco:
Hi @lrlopez, it looks like the latest main updates introduced some linting errors. You can fix them easily by running `make fix` in your branch and pushing again. Thanks for your work on this!

PR #618, opened 2025-11-05
608 unexpected error importing file

erseco:
This pull request introduces optimizations for handling large files (over 100MB) in the ODE import workflow, focusing on ZIP extraction, file reading, and XML processing. The main goal is to improve memory management and performance when dealing with very large ELP (eXe Learning Package) or XML files, reducing the risk of memory exhaustion and long processing times.

Key changes include:

**Optimized Large File Handling**

* In `OdeService.php`, added logic to detect large files and use optimized methods for ZIP extraction, file reading, and XML parsing when files exceed 100MB. This includes logging for when these optimized paths are used. 

**File Utility Enhancements**

* Added `FileUtil::extractZipToOptimized`, which extracts large ZIP files in batches, uses chunked streaming for big files inside the ZIP, and manages memory and execution time limits for better performance and reliability.
* Added `FileUtil::getFileContentOptimized`, which reads large files in 8MB chunks to avoid loading the entire file into memory, with periodic garbage collection to further reduce memory usage.

**XML Processing Optimization**

* Added `OdeXmlUtil::readOdeXmlOptimized`, which parses large XML files with increased memory and execution limits, uses `LIBXML_PARSEHUGE`, and processes XML in chunks with explicit garbage collection to prevent memory leaks.

These changes make the import process significantly more robust and scalable for large learning packages and resources.

Thanks @jgblasco for providing us a really huge .elpx file, it was really helpful! Maybe you should create bigger ones!


erseco:
Yes, there's still an error when opening very large files. With these changes, at least we can prevent them from being opened until we identify which core function is failing. Sometimes the file loads, sometimes it doesn’t — really frustrating…

erseco:
I think I encounter one of the main problems! https://github.com/exelearning/exelearning/blob/a5f694655567edd9085a4ca4f72747c63781f80c/src/Constants.php#L321 this is limiting the import to 20 seconds, I'm going first to increase this, and later a better progress response when parsing the .xml

PR #628, opened 2025-11-06
Add custom 404 error page and update error handling mechanism

pabloamayab:
Introduce a user-friendly 404 error page and modify the error handling to serve this page when a resource is not found. The system now attempts to render specific error templates before falling back to a generic one if necessary.


pabloamayab:
The work had already been done at some point.

PR #619, opened 2025-11-05
Add progress dialog for uploads

erseco:
This pull request introduces significant improvements to the user experience when uploading large files in the application, particularly by adding a dedicated upload progress modal and enhancing session management for file uploads. The changes ensure that users receive clear feedback during lengthy uploads, handle unsaved changes more robustly, and streamline error handling and session confirmation flows.

Captura de pantalla 2025-11-05 a las 1 17 44

**Large File Upload Experience Improvements**

* Added a new `ModalUploadProgress` class to display real-time upload and processing progress, including percentage complete, current phase (uploading, extracting, etc.), and error handling. 
* Updated the `largeFilesUpload` method in `modalOpenUserOdeFiles` to show the progress modal during uploads, update progress as file chunks are uploaded, and display extraction/processing phases. Error handling now uses the modal for user feedback. 
**Session Management Enhancements**

* Before uploading large files, the app now checks for unsaved changes in the current session and prompts the user with a confirmation modal if needed, preventing accidental data loss.
* Improved the flow for resuming uploads after a session logout: the session logout modal now correctly resumes large file uploads or regular file opening based on the user's choice and file type.

**Code Organization and Robustness**

* Refactored methods in `modalOpenUserOdeFiles` and `modalSessionLogout` to support new parameters (such as `skipSessionCheck` and `forceCloseSession`) and to ensure that modals are shown/hidden at appropriate times during the upload and error handling process. 

These changes collectively provide a smoother, safer, and more transparent workflow for users uploading large files, with clear progress updates and robust handling of session-related issues.

PR #636, opened 2025-11-07
Hotfix / Fix open file from user projects

erseco:
This pull request introduces a small but important update to the `openElp` method in `OdeService.php`, adding a new parameter to support additional functionality.

- **API Enhancement:**
  * The `openElp` method now accepts an optional boolean parameter `$isImportIdevices`, defaulting to `false`. This allows callers to specify whether iDevices should be imported when opening a project from the user projects.


erseco:
This issue was introduced by #619

erseco:
Integrated in PR #637

Issue #643, opened 2025-11-09
Portuguese, Romanian and Esperanto automatic translations


ignaciogros:
We have added two new languages (Portuguese and Romanian), and the translation into Esperanto is incomplete.

@erseco:

Could you please generate automatic translations (files **messages.pt.xlf** and **messages.ro.xlf**)?

Translations should start with "~", and the Spanish translation (**messages.es.xlf**) should be used as a reference. No string that is not translated in messages.es.xlf should be translated.

Untranslated strings should remain as `<target></target>`, not `<target/>`.

**messages.eo.xlf** should be completed following the same instructions.

As soon as we have those translations, we will notify the contributors so they can complete the work.

Thank you very much.

PR #644, opened 2025-11-09
Added automatic Portuguese, Romanian and Esperanto translations

erseco:
See #643


ignaciogros (changes requested):
Thank you, @erseco.

I've seen that you've provided the EU translation instead of EO. 

We'll merge the EU translation too, because there were many untranslated strings, but could you please include messages.eo.xlf too?

ignaciogros (approved):
Thanks!

PR #664, opened 2025-11-13
Allow importing EPUB files in file upload selectors

pabloamayab:
Enable support for importing EPUB files by updating the file upload selectors to accept the .epub extension. This change enhances the file upload functionality for users.


ignaciogros (changes requested):
Hi @pabloamayab,

You can open an epub, but you cant import it. 

When I try to do that, I get this messages:

imagen

imagen

Thank you.

PR #673, opened 2025-11-16
Automatically save the package properties.

ignaciogros:


ignaciogros:
You'll see that the "Save metadata and properties" is no longer needed.

Maybe we should create a new function for that, but that's the fastest solution, because `checkOpenIdevice` is called in many parts of the application.

We could also rename the current function.

Please test the solution and let me know if you want me to change anything.

Thank you!

ignaciogros:
I think that the problem's caused by some changes in the navbarFile.js file.

**Before:**

```
    setSaveProjectEvent() {
        this.saveButton.addEventListener('click', () => {
            if (eXeLearning.app.project.checkOpenIdevice()) return;
            
            // ...
        });
    }
```

**Now:**

```
    setSaveProjectEvent() {
        this.saveButton.addEventListener('click', async () => {
            // Check if an iDevice is open, but without saving the package properties
            if (eXeLearning.app.project.checkOpenIdevice(true)) return;

            // ...
        });
    }
```

PR #710, opened 2025-11-25
Form iDevice improvements and responsive iDevices updates

mnarvaezm:
- Form iDevice improvements: edition/form.js refactor and new feedback/hint options.
- Fix for floated images overflowing in Form iDevice questions.
- Responsive improvements for Puzzle and Word Search.
- SCORM fixes: better score recovery and improved suspend_data handling

### DIFF
--- a/.github/workflows/build-electron-installers.yml
+++ b/.github/workflows/build-electron-installers.yml
@@ -21,7 +21,7 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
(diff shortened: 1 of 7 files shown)


ignaciogros (commented):
Thank you, @mnarvaezm.

I only have doubts about the XLF files. This will make future exports more complex, and those files should work along with eXe's current i18n mechanism.

Could you please confirm if they should be there and how i18n would work?

Maybe we could include that later. In the JavaScript version?

ignaciogros (approved):

PR #715, opened 2025-11-26
 Improvements to TinyMCE media plugins (exeimage & exemedia)

mnarvaezm:
- exeimage: replaces ID-based image selection with a class-based approach and improves horizontal/vertical alignment and editing of figures with or without attribution.

- exemedia: improves class-based selection, editing and updating of media, and enhances horizontal alignment logic for attributed and non-attributed media.

### DIFF
--- a/public/files/perm/idevices/base/text/export/text.js
+++ b/public/files/perm/idevices/base/text/export/text.js
@@ -187,8 +187,12 @@ var $text = {
         );
         if ($btn.length === 1) {
             const [textA, textB = textA] = $btn.val().split('|');
-            $btn.val(textA).attr('data-text-a', textA).attr('data-text-b', textB);
-            $btn.off('click').closest('.feedback-button').removeClass('clearfix');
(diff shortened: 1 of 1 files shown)


ignaciogros (changes requested):
Thank you very much, @mnarvaezm.

It’s a very complex task. After some quick tests of "exeimage" in Firefox, I’ve noticed there are significant functionality issues. I suggest postponing merging these changes into “main.” Maybe the best approach is to rebuild these plugins from scratch.

**Some issues:**

- I add Image + Text ( + Text). I edit the image to add a `figcaption`. The text gets placed inside `figure`.
- I edit an image with `figcaption` and change the title. The new value is placed after the previous one instead of replacing it.
- The same happens when changing the license: 2 `figcaption` elements are loaded.
- “Custom license” doesn’t allow entering data.
- If you remove the image caption (title), `figure` is not deleted.

**Please close the PR if you believe the task should be approached differently.**

PR #716, opened 2025-11-26
Fix image paths on Windows - normalize separators (#711)

### LINKED ISSUE #711: Images lost in preview
Some images are disappearing from the preview. Some of them had a "double extension," but even after correcting them, they still don't appear.

These errors are detected in this file. The images with "double extensions" have been corrected in this version.

[programa-tu-propio-juego.zip](https://github.com/user-attachments/files/23743361/programa-tu-propio-juego.zip)

Looking at the source code of the web page generated in the preview, the images that are not visible are referenced to "files/tmp/..." while those that are visible are referenced to "../content/resources/ "

erseco:
### Problem
On Windows, images inserted via file manager retain backslashes in HTML src attributes, causing them to disappear in preview and export.

### Root Cause
`FilemanagerMethodController::getPath()` uses `DIRECTORY_SEPARATOR` to build paths. On Windows this inserts `\` characters that later fail to match in `replaceFilemanagerResourcesLinks()`.

### Solution
Normalize all path separators to forward slashes before returning to TinyMCE:
```php
$relativePath = str_replace('\\', '/', $relativePath);
```

Forward slashes work in URLs on all platforms and are correctly handled by PHP file operations on Windows.

Closes #711

### DIFF
--- a/src/Controller/net/exelearning/Controller/Filemanager/FilemanagerMethodController.php
+++ b/src/Controller/net/exelearning/Controller/Filemanager/FilemanagerMethodController.php
@@ -660,6 +660,9 @@ class FilemanagerMethodController extends DefaultApiController
         // Relative path
         $relativePath = $pathIni.DIRECTORY_SEPARATOR.$path;
 
+        // Hotfix #711: Normalize path separators for URLs (Windows compatibility)
+        $relativePath = str_replace('\\', '/', $relativePath);
+
         $data = [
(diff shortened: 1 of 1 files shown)


ignaciogros (approved):
Thank you, @erseco.

I think that this should be merged into main, because, as you said, there's no need to work with different separators, but this won't fix #711. Those missing images are just not part of the attached zip file.

PR #789, opened 2025-12-19
fix: escape quotes in JSON generation

### LINKED ISSUE #557: Error opening a file exported to the new extension
With this resource.

https://descargas.intef.es/recursos_educativos/ODES_SGOA/Codigo_Escuela/Explorando_fisica_con_Scratch/Explorando_fisica_con_Scratch.elp
It opens and saves with the new extension (elpx), and some content is lost upon opening.

Image

Image

However, the preview shows all the content.

franmate:
This PR fixes a bug where unescaped double quotes within string values were causing JSON parsing errors.

Important: Please perform thorough testing on both export and import processes to ensure data integrity remains intact after these escaping fixes.


ignaciogros (commented):
Hi @franmate,

Thank you for fixing that.

v3.0.2 will be released on December 22. If this PR might introduce any new problems, maybe we should wait until we release that version. Any problems in that part of the code might cause malfunctions in many iDevices.

PR #888, opened 2026-01-04
New 3d viewer idevice

### LINKED ISSUE #1810: 3D Viewer iDevice does not render STL after publishing/exporting
## Description

The **3D Viewer** iDevice does not display the 3D object once the content is published/exported.

The 3D Viewer plugin/iDevice is available here:

https://c33106eb.exelearning.pages.dev/

The object renders correctly inside the eXeLearning editor and also in the integrated preview, but it fails in the published/exported version and also in the preview opened in a new browser tab.

## Expected behavior

The 3D object should render correctly in all contexts:

- Editor
- Integrated preview
- Preview opened in a new tab
- Published/exported content

## Actual behavior

In the published/exported content, the 3D viewer does not load the model. The browser console shows errors related to JavaScript module loading, Three.js, `model-viewer`, and the STL resource.

## Console errors

```text
Uncaught SyntaxError: export declarations may only appear at top level of a module three.module.min.js:6:661468
Uncaught SyntaxError: import declarations may only appear at top level of a module STLLoader.js:1:1
Uncaught SyntaxError: import declarations may only appear at top level of a module OrbitControls.js:1:1
SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data
    parse http://127.0.0.1:8000/idevices/three-d-viewer/model-viewer.min.js:456
    load http://127.0.0.1:8000/idevices/three-d-viewer/model-viewer.min.js:456
    load http://127.0.0.1:8000/idevices/three-d-viewer/model-viewer.min.js:18
    promise callback*load http://127.0.0.1:8000/idevices/three-d-viewer/model-viewer.min.js:18
    load http://127.0.0.1:8000/idevices/three-d-viewer/model-viewer.min.js:456
    e http://127.0.0.1:8000/idevices/three-d-viewer/model-viewer.min.js:456
    e http://127.0.0.1:8000/idevices/three-d-viewer/model-viewer.min.js:456
    preload http://127.0.0.1:8000/idevices/three-d-viewer/model-viewer.min.js:456
    load http://127.0.0.1:8000/idevices/three-d-viewer/model-viewer.min.js:456
    i http://127.0.0.1:8000/idevices/three-d-viewer/model-viewer.min.js:456
    setSource http://127.0.0.1:8000/idevices/three-d-viewer/model-viewer.min.js:456
    updateSource http://127.0.0.1:8000/idevices/three-d-viewer/model-viewer.min.js:456
    e/Uy/this[cy]< http://127.0.0.1:8000/idevices/three-d-viewer/model-viewer.min.js:456
    Uy http://127.0.0.1:8000/idevices/three-d-viewer/model-viewer.min.js:456
    n http://127.0.0.1:8000/idevices/three-d-viewer/model-viewer.min.js:463
    h http://127.0.0.1:8000/idevices/three-d-viewer/model-viewer.min.js:463
    d http://127.0.0.1:8000/idevices/three-d-viewer/model-viewer.min.js:463
    y http://127.0.0.1:8000/idevices/three-d-viewer/model-viewer.min.js:463
    s http://127.0.0.1:8000/idevices/three-d-viewer/model-viewer.min.js:463
    s http://127.0.0.1:8000/idevices/three-d-viewer/model-viewer.min.js:463
    s http://127.0.0.1:8000/idevices/three-d-viewer/model-viewer.min.js:463
    r http://127.0.0.1:8000/idevices/three-d-viewer/model-viewer.min.js:463
    e http://127.0.0.1:8000/idevices/three-d-viewer/model-viewer.min.js:463
    <anonymous> http://127.0.0.1:8000/idevices/three-d-viewer/model-viewer.min.js:463
    <anonymous> http://127.0.0.1:8000/idevices/three-d-viewer/model-viewer.min.js:1
    <anonymous> http://127.0.0.1:8000/idevices/three-d-viewer/model-viewer.min.js:1
model-viewer.min.js:456:110553
WARNING: Multiple instances of Three.js being imported. three.module.min.js:6:661384
Error in source map: Error: request failed with status 404
Stack in the worker:networkRequest@resource://devtools/client/shared/source-map-loader/utils/network-request.js:43:9

Resource URL: http://127.0.0.1:8000/idevices/three-d-viewer/model-viewer.min.js
Source map URL: model-viewer-umd.min.js.map
XHRGET
http://127.0.0.1:8000/content/resources/dcf55a78-717e-c463-3ce5-ef8a908c4478.stl
[HTTP/1 404 File not found 0ms]

[3D Viewer] Failed to render STL: RangeError: invalid array length
    parseBinary http://127.0.0.1:8000/idevices/three-d-viewer/STLLoader.js:196
    parse http://127.0.0.1:8000/idevices/three-d-viewer/STLLoader.js:404
    renderSTL http://127.0.0.1:8000/idevices/three-d-viewer/three-d-viewer.js:828
    init http://127.0.0.1:8000/idevices/three-d-viewer/three-d-viewer.js:668
    ThreeDViewerRuntime http://127.0.0.1:8000/idevices/three-d-viewer/three-d-viewer.js:624
    boot http://127.0.0.1:8000/idevices/three-d-viewer/three-d-viewer.js:1161
    boot http://127.0.0.1:8000/idevices/three-d-viewer/three-d-viewer.js:1156
    promise callback*renderBehaviour http://127.0.0.1:8000/idevices/three-d-viewer/three-d-viewer.js:1165
    initJsonIdevice http://127.0.0.1:8000/libs/exe_export.js:297
    initJsonIdevice http://127.0.0.1:8000/libs/exe_export.js:244
    initJsonIdeviceInterval/window[intervalName]< http://127.0.0.1:8000/libs/exe_export.js:227
    setInterval handler*initJsonIdeviceInterval http://127.0.0.1:8000/libs/exe_export.js:226
    initJsonIdevices http://127.0.0.1:8000/libs/exe_export.js:215
    initJsonIdevices http://127.0.0.1:8000/libs/exe_export.js:214
    init http://127.0.0.1:8000/libs/exe_export.js:47
    <anonymous> http://127.0.0.1:8000/libs/exe_export.js:452
    jQuery 13
three-d-viewer.js:896:25
```

## Notes

There may be more than one related issue here:

1. `three.module.min.js`, `STLLoader.js`, and `OrbitControls.js` contain `import`/`export` declarations, but they seem to be loaded in a non-module context.
2. Multiple instances of Three.js are being imported.
3. The STL file is requested from `content/resources/` but returns `404 File not found`.
4. After the STL request fails, `STLLoader.js` appears to try to parse the response and ends with `RangeError: invalid array length`.

## Environment

- eXeLearning version: not specified
- OS: MX Linux 23.6, based on Debian 12
- Browser: not specified
- The issue is visible in published/exported content and in the preview opened in a new tab.

erseco:
## 3D Viewer iDevice

Adds a **3D Viewer iDevice** so authors can embed interactive 3D models inside eXeLearning content, plus the file-manager, asset-resolution and exporter work needed to make 3D assets round-trip end to end across editor, workarea preview, preview panel and static HTML / SCORM / EPUB exports.

Closes #1810.

Captura de pantalla 2026-01-04 a las 17 11 57

### What the iDevice does

- Embeds 3D models that learners can rotate, zoom and inspect inline.
- Native support for `.glb`, `.gltf` and `.stl`. GLB/GLTF render through Google's `<model-viewer>` web component (PBR, IBL, animations); STL renders with Three.js (`STLLoader` + `OrbitControls`) directly — no upload-time conversion, no duplicate asset.
- Editor, workarea preview panel and exported HTML/SCORM/EPUB use the same renderer — author preview matches learner view.
- Configurable behaviour: **auto-rotate** (default, adjustable speed) or **navigation controls** (fullscreen toggle + 4-direction arrow pad). Mutually exclusive — enabling one disables the other.
- Animation playback for GLB models with named animations.
- Per-instance background colour, **STL model colour** (visible always, disabled in the form when the selected file isn't STL), alt text, and accessibility (`aria-label`, screen-reader announcements).

### File-manager and picker

- `ModalFilemanager` recognises, filters and labels 3D models alongside images/audio/video/PDF. Type detection works from both MIME and extension (`.glb`, `.gltf`, `.stl`, `.obj`, `.fbx`).
- New `accept='3d'` filter for the iDevice file picker; auto-selected by `legacyExeIdevicesFilePicker` when the input id contains `3d` or `model`.
- Default upload `<input>` accepts the new 3D extensions.

### Fixing the export crash (#1810)

The exported HTML was failing to render the STL with a cascade of errors:

1. `SyntaxError: import declarations may only appear at top level of a module` — the exporter was emitting every iDevice JS file as a classic `<script src=…>`, so the ES modules shipped by the plugin (`three.module.min.js`, `STLLoader.js`, `OrbitControls.js`) failed to evaluate.
2. With Three.js / STLLoader unavailable, `<model-viewer>` tried to load the STL itself and 404'd because the rewriter pointed at `content/resources/<uuid>.stl` while the file shipped under its friendly filename.
3. `RangeError: invalid array length` — downstream symptom of #2; STLLoader was fed the HTML 404 response body.

Fixed in two layers:

- `IdeviceRenderer` now detects ES modules (by `.mjs` extension, `*.module.js` filename convention, or content-sniffed top-level `import`/`export`) and emits `<script type="module">`. Same detection covers HTML5, SCORM 1.2 / 2004, EPUB3 and IMS exports.
- The STL URL handling was restructured (see Refactor below) so the runtime can recover the canonical AssetManager handle and serve the asset correctly across editor, preview iframe and static export contexts.

### Refactor — shared runtime, strict `asset://`, STL color

The editor, workarea view, preview iframe and static export all share a single rendering runtime at `public/files/perm/idevices/base/three-d-viewer/export/three-d-viewer-runtime.js`. Both edition and export load it via `<script>` injection and call into `window.eXe3DViewer`:

- `init(wrapper, options)` — synchronous registration + async boot, idempotent per wrapper
- `destroy(wrapper)` / `destroyAll()`
- `detectModelType` / `normalizeColor` / `normalizeModelSource`
- `resolveModelSource(src, ctx)` — `asset://` → blob via AssetManager
- `configureRendererColorManagement(renderer)` — sRGB + `THREE.ColorManagement.enabled`
- `disposeObject3D` / `disposeMaterial`

The previous ~130 lines of STL code duplicated between edition and export are gone.

**Flat `data-*` attributes, no base64 envelope.** `renderView` emits the canonical model reference as flat attributes on the wrapper (`data-model-src`, `data-model-type`, `data-model-color`, `data-background-color`, `data-camera-controls`, `data-auto-rotate`, `data-auto-rotate-speed`, `data-show-nav-controls`, `data-animation-*`, `data-alt`). `<model-viewer>` is emitted without `src` — the runtime sets it at boot from `data-model-src`. This eliminates the STL parse crash path and removes any place where a `blob:` URL could leak into persisted HTML. A `migrateLegacyConfig` one-shot decoder upgrades any pre-existing wrapper still carrying the old base64 `data-config` blob in place at first render, so existing projects keep working without a save round-trip.

**`data-model-asset-ref` for engine resolution.** The workarea engine and `IdeviceRenderer.fixAssetUrls` rewrite `asset://` URLs on the way to the DOM (workarea → blob URL, static export → `content/resources/...`). The wrapper carries a parallel `data-model-asset-ref="<uuid>.<ext>"` (no `asset://` prefix, so the rewriter ignores it). The runtime uses it to recover the canonical handle and ask AssetManager for the live blob URL in workarea and preview iframe contexts; in static export contexts it falls back to the rewritten `data-model-src`.

**Strict `asset://` discipline.** `handleModelSelection` refuses to store a `blob:` URL as the canonical source. `get3DViewerJSON` strips any `blob:` from `state.src` before returning the JSON. The runtime blob URL lives in an instance field (`this.previewBlobUrl`), never on `state._previewBlobUrl` or `formElements.src.dataset.blobUrl`. `set3DViewerJSON` recovers `asset://<id>.<ext>` from a `blob:` URL via `AssetManager.reverseBlobCache` + `getAssetMetadata` so the form picker doesn't show empty after reload.

**STL `modelColor`.** New colour-input form field inside the existing Display Options fieldset. Visible always, `disabled` when the selected file isn't STL. Default `#888888`. Applied as `MeshStandardMaterial({ color, metalness: 0.0, roughness: 0.55 })` — `metalness: 0` is important: any non-zero metallic component reflects the (empty) environment as black and silently kills the chosen albedo. GLB/GLTF embedded materials are not touched.

**Lifecycle / multi-instance.** Per-wrapper instance object tracking renderer/scene/geometry/material/textures/RAF/listeners/objectURLs. `destroy()` cancels RAF, removes listeners, disposes every GPU resource, revokes only the object URLs the instance itself created. `destroyAll()` bound once to `beforeunload`. Edition's `init()` destroys the previous preview container before rebuilding the form, preventing accumulating WebGL contexts across repeated open/close cycles.

**Single source of truth for Three.js libs.** `STLLoader`, `OrbitControls`, `three.module.min.js` and `model-viewer.min.js` live under `export/` only and are shared by editor and exports — saves ~700 KB of duplicated assets and removes drift risk between the two contexts.

**Asset URL replacement bug fix.** `AssetManager.resolveHTMLAssets` now sorts asset URL patterns longest-first before substitution. The previous shortest-first order produced partial matches that broke 3D asset references in exported HTML.

**Test-pollution fix.** `resetIdeviceConfigCache()` now also clears `idevicesBasePath`. Without this, a spec that called `setIdevicesBasePath(tempDir)` in `beforeEach` left the global pointing at a deleted directory and the next spec's `getIdeviceExportFiles(image-gallery, '.css')` fell back to `['image-gallery.css']` only, dropping `simple-lightbox.min.css` and breaking the `IdeviceRenderer > getCssLinks` test under CI.

### Tests

- **6949** unit tests green (Bun) including the new 46-spec runtime suite (`three-d-viewer-runtime.test.js`) and the extended edition/export specs.
- **185** frontend Vitest tests across edition/export/runtime files green.
- **E2E** Playwright suite extended with specs for STL colour enable/disable, colour persistence through save+reload, `data-model-asset-ref` recovery on reload, flat `data-*` (no base64) discipline.
- `make fix` clean across `src/`, `public/app/`, `test/`.
- Patch coverage 98.5% on changed lines (1 untested line in `idevice-config.ts`).


erseco:
Pushed 0c52433a on this branch to address issue #1810 (3D Viewer iDevice fails to render after publish/export).

**Root cause.** The exporter emits every iDevice JS file as a classic `<script src="…">`. The plugin ships ES modules (`three.module.min.js`, `STLLoader.js`, `OrbitControls.js`), so the browser threw `SyntaxError: import declarations may only appear at top level of a module`. Once the modules failed to evaluate, `model-viewer` could no longer fetch STL assets — the `content/resources/<uuid>.stl` 404 in the report was a downstream symptom.

**Fix.** `IdeviceRenderer.getJsScripts()` / `getJsScriptInfo()` now mark a file as `type="module"` when any of these holds:
1. `.mjs` extension,
2. `*.module.js` filename convention (catches `three.module.min.js`),
3. content sniff for a top-level `import` / `export` statement (catches `STLLoader.js`, `OrbitControls.js`).

UMD bundles such as `model-viewer.min.js` that only mention `"exports"` inside a string literal are explicitly excluded. The detection lives in `src/services/idevice-config.ts` (`isIdeviceJsModule`) with a per-file cache. Every Html5Exporter subclass (HTML5, SCORM 1.2 / 2004, EPUB3, IMS) benefits transparently — the renderer is shared.

**How to verify.**
- `bun test src/services/idevice-config.spec.ts src/shared/export/renderers/IdeviceRenderer.spec.ts`
- `bun test src/shared/export/` (whole export pipeline: 1850 tests pass on this branch)
- Spot-check the generated HTML for any subpage and confirm e.g. `<script type="module" src="../idevices/three-d-viewer/three.module.min.js"></script>`.
- BASE_PATH coverage: tests assert the attribute is present with `basePath=""`, `"../"`, and `"/myapp/"`.

Thanks @jjdeharo for the very detailed console trace — it made the cascade obvious.

erseco:
Follow-up 7d0513bb addressing the residual symptoms from the post-export trace.

After 0c52433a the static export still showed two problems when served from a plain `http-server`:

1. **`STLLoader.js` / `OrbitControls.js` → `Cannot use import statement outside a module`.** The browser bundle of the exporter only does filename-based module detection (no FS access for content-sniffing), and those two files don't follow the `.module.js` / `.mjs` convention.
2. **`content/resources/<uuid>.stl` 404.** The exporter correctly rewrote `asset://<uuid>.stl` to `content/resources/model.stl` in the outer `data-idevice-json-data`, but the runtime kept reading the inner wrapper's base64 `data-config` blob — opaque to the URL rewriter — so the stale `asset://` URL survived and `STLLoader` then choked on the HTML 404 body.

**Fix 1 (browser exporter).** New static registry `IDEVICE_JS_MODULES` in `src/shared/export/browser/idevice-config-browser.ts`. `isIdeviceJsModule()` now consults it alongside `.mjs` / `*.module.js`. The Node-side helper still content-sniffs, so this only matters for browser-driven exports.

**Fix 2 (plugin runtime).** New `resolveBootConfig(data, wrapper)` helper on `$threedviewer`. It prefers the engine-supplied `data` (URLs already resolved by the exporter) and falls back to the base64 `data-config` only when no `src` is present (editor / preview path, where AssetManager resolves blobs). `renderBehaviour()` delegates to it.

**TDD:** 8 new browser-config specs (filename + registry + UMD/classic negatives), 3 plugin specs covering the exact UUID URL from the bug report.

**Verification:** `bun test src/shared/export/`, `npx vitest run public/files/perm/idevices/base/three-d-viewer/`, and `make lint` all green. Bundle regenerated with `bun run bundle:exporters` (artifact is gitignored — anyone testing the PR needs to run that command after pulling).

Thanks for the http-server trace, @jjdeharo — the two-layer config storage wasn't visible from issue #1810's original output alone.

erseco:
Follow-up 0dc2c7b8: the previous commit fixed the static export but broke the editor / in-app preview, because `resolveBootConfig` always preferred the engine's resolved URL.

In the workarea and the preview iframe the project assets live in IndexedDB and are served as blob URLs via the AssetManager. The exporter-resolved path \`content/resources/model.stl\` does not exist on the live server — only the original \`asset://<uuid>.stl\` URL does (it gets translated to a blob URL by \`AssetManager.resolveAssetURLSync\`).

\`resolveBootConfig\` now picks based on AssetManager presence:

| Context | AssetManager | Source of truth |
|---|---|---|
| Editor / preview | available | wrapper's \`data-config\` (asset:// URL → blob) |
| Static export (HTML/SCORM/EPUB/IMS) | absent | engine \`data.src\` (rewritten by exporter to friendly filename) |

Tests reshaped to exercise both contexts plus three edge cases (missing data-config, absolute https URL, empty engine payload). \`npx vitest run public/files/perm/idevices/base/three-d-viewer/\` → 104 / 104 green. \`bun test src/shared/export/\` still green. \`make lint\` clean. Bundle rebuilt with \`bun run bundle:exporters\` (artifact gitignored).

Sorry about the regression — the editor path wasn't visible from the first trace.

erseco:
Follow-up 30824cf4 — the \`COLOR= "... is not valid JSON\` error.

**Why it appeared now.** \`buildModelMarkup\` was always passing the model URL straight into \`<model-viewer src="…">\`, including for STL files. model-viewer dispatches its src through the GLB / GLTF / USDZ loaders and chokes on the ASCII STL header (\`COLOR=\` ⇒ "not valid JSON"). Previously the file 404'd before model-viewer's loader saw any bytes; once the blob-URL path started returning 200 in the live preview, model-viewer finally received the data and crashed.

**Fix.** Drop the \`src\` attribute from \`<model-viewer>\` for \`.stl\` files. \`ThreeDViewerRuntime.renderSTL()\` already hides the model-viewer host (\`style.display = 'none'\`) and renders the geometry on its own \`<canvas>\`. GLB / GLTF / GLTF passthrough is unchanged.

**TDD.** Three new specs in \`three-d-viewer.test.js\` cover the STL suppression, GLB / GLTF passthrough, and the \`asset://\` variant — \`npx vitest run public/files/perm/idevices/base/three-d-viewer/\` → 107 / 107 green. \`make lint\` clean. No bundle changes needed (this is plugin runtime, not the exporters bundle).

Reload the workarea and the STL should render via Three.js without any model-viewer JSON error.

erseco:
Follow-up 31b7ea14 — the "COLOR=" error was still firing because of persisted HTML.

**Why the previous fix wasn't enough.** 30824cf4 only changed \`buildModelMarkup\` (which runs at save time). Projects saved BEFORE the fix landed still carry \`<model-viewer src="asset://<uuid>.stl">\` (or \`src="blob:…"\` for older sessions) inside the persisted iDevice HTML — and as soon as the model-viewer custom element activates, it fetches that URL and dispatches the bytes through GLB / GLTF / USDZ loaders, which throws the JSON error on the ASCII STL header.

**Fix.** \`renderBehaviour\` now runs a synchronous pre-pass BEFORE \`ensureModelViewerModule()\` loads the custom element definition. For every wrapper it strips the \`src\` attribute from the embedded \`<model-viewer>\` whenever EITHER the attribute itself OR the resolved \`cfg.src\` looks like an STL — the blob-URL case (no extension) is covered via \`cfg.src\`. GLB / GLTF are left untouched.

**Tests.** 4 new specs cover asset://*.stl, content/resources/*.stl, blob:*.stl (decided via cfg.src), and the GLB negative. \`npx vitest run public/files/perm/idevices/base/three-d-viewer/\` → 111 / 111 green. Lint clean.

**Separate UX point (input shows blob:URL).** Pinging this on a separate note: the file picker (\`legacyExeIdevicesFilePicker\` in ideviceNode.js) already sets \`e.value = result.assetUrl\` and \`e.dataset.blobUrl = result.blobUrl\`, and the edition's \`preResolveAssetUrl\` / \`renderSTLWithThreeJS\` only touch \`dataset.blobUrl\` — none of the runtime paths overwrite \`.value\` with the blob URL. If the input is still showing the blob URL after reloading a saved project, the most likely culprit is that \`result.assetUrl\` from the file-manager modal is returning the blob URL (not \`asset://<uuid>\`). Could you confirm what the modal returns in \`onSelect\`? That's outside the issue #1810 scope (export & rendering) but easy to fix in the picker plumbing.

erseco:
## 3D Viewer iDevice refactor — final summary

This comment used to be the long refactor write-up plus the CI triage follow-up; consolidating both into a single chronological log now that the work is squashed to one commit on the branch.

### Architecture (refactor of PR #888's original wiring)

**Shared runtime module** — `public/files/perm/idevices/base/three-d-viewer/export/three-d-viewer-runtime.js` (new). Owns the STL Three.js scene/camera/renderer setup, `STLLoader` parsing, `OrbitControls`, the animate loop, dispose helpers, per-wrapper instance registry, and `beforeunload` cleanup. Both edition and export load it via `<script>` injection and call into `window.eXe3DViewer`:

- `init(wrapper, options)` — synchronous registration + async boot, idempotent per wrapper
- `destroy(wrapper)` / `destroyAll()`
- `detectModelType` / `normalizeColor` / `normalizeModelSource`
- `resolveModelSource(src, ctx)` — `asset://` → blob via AssetManager
- `configureRendererColorManagement(renderer)` — sRGB + `THREE.ColorManagement.enabled`, mirrors PR #1689's three-sixty-viewer pattern
- `disposeObject3D` / `disposeMaterial`

The previous ~130 lines of STL code duplicated between edition and export are gone.

**Flat `data-*` attributes, no base64 envelope.** `renderView` emits flat `data-model-src`, `data-model-type`, `data-model-color`, etc. directly on the wrapper. `<model-viewer>` is emitted without `src` — the runtime sets it at boot. Eliminates the STL parse crash root cause and the blob:-in-persisted-HTML leak vector. `migrateLegacyConfig(wrapper)` one-shot decoder upgrades pre-existing wrappers still carrying the base64 `data-config` blob in place.

**`data-model-asset-ref` for engine resolution.** The workarea engine and `IdeviceRenderer.fixAssetUrls` rewrite `asset://` URLs on the way to the DOM. The wrapper carries a parallel `data-model-asset-ref="<uuid>.<ext>"` (no `asset://` prefix, invisible to the rewriter) so the runtime can recover the canonical handle and ask AssetManager for the live blob URL in workarea + preview iframe contexts. Falls back to the rewritten `data-model-src` in static export.

**`resolveBootConfig` simplified.** The PR #1810 dual-context kludge (engine data vs base64 wrapper, gated on AssetManager presence) collapses into a single read from flat attributes.

**`modelColor` for STL.** Visible-always / disabled-when-not-STL form field. Default `#888888`. Applied as `MeshStandardMaterial({ color, metalness: 0.0, roughness: 0.55 })` — `metalness: 0` is mandatory because any metallic component reflects an empty scene as black and silently kills the chosen albedo.

**Strict `asset://` discipline.** `handleModelSelection` refuses `blob:`. `get3DViewerJSON` strips `blob:` from `state.src`. The runtime blob URL lives in `this.previewBlobUrl` (instance field), never on state or DOM dataset. `set3DViewerJSON` recovers `asset://<id>.<ext>` from a `blob:` URL via `AssetManager.reverseBlobCache` + `getAssetMetadata` so the form picker doesn't show empty after reload.

**Lifecycle / multi-instance.** Per-wrapper instance object tracking renderer/scene/geometry/material/textures/RAF/listeners/objectURLs. `destroy()` cancels RAF, removes listeners, disposes every GPU resource, revokes only the object URLs the instance itself created. `destroyAll()` bound once to `beforeunload`. Edition's `init()` destroys the previous preview container before rebuilding the form.

### Issue #1810 close-out

Every numbered cause from the issue's "Notes" section is addressed:

1. **ES module loading** — exporter emits `<script type="module">` for `.mjs`, `*.module.js`, and content-sniffed top-level `import`/`export`.
2. **Multiple Three.js instances** — informational; the workarea (parent) and preview iframe are different JavaScript realms, expected behaviour, gone once only one viewer renders in the document context.
3. **STL 404 at `content/resources/...`** — fixed via `data-model-asset-ref` recovery path in workarea / preview-with-AssetManager contexts.
4. **`RangeError: invalid array length`** — downstream symptom of #3; goes away once the fetch succeeds.

### Bug-fix iterations (from live testing in this session)

Seven targeted fixes that landed alongside the refactor:

1. Preview STL color picker didn't re-render after change → `renderSTLWithThreeJS` now uses an options-aware skip key (was comparing only `blobUrl`).
2. `/content/resources/...` 404 in preview iframe → runtime was prepending a leading `/`, missing the preview SW's `/viewer/*` interception. Now returned relative.
3. Workarea view didn't render the model after save → `IdeviceRenderer.fixAssetUrls` was rewriting `data-model-src` away from `asset://`. Added `data-model-asset-ref`.
4. STL rendered gray regardless of `modelColor` → material had `metalness: 0.2` (copied from old code); spec required `0.0`.
5. `modelColor` not respected in workarea view-mode → `ThreeDViewerRuntime.normalizeConfig()` enumerated fields and silently dropped `modelColor`. Added to the whitelist.
6. Reload / page navigation showed empty state and empty file picker → the workarea engine resolves `asset://` → `blob:` when reading JSON. Both `renderView` (via `buildWrapperAttrs`) and edition `set3DViewerJSON` now use `AssetManager.reverseBlobCache` + `getAssetMetadata` to reconstruct `asset://<id>.<ext>`, and `resolveBootConfig` accepts `blob:` directly (browser can fetch a live blob URL).
7. Re-init didn't destroy the previous WebGL context → added `eXe3DViewer.destroy(previousWrapper)` to edition's `init()`.

### CI

**Build & Test Docker Images** — initially failed with `IdeviceRenderer > getCssLinks > should include CSS dependencies for image-gallery`. Verified pre-existing on `31b7ea14` (the last commit on this branch before the refactor session), but visible to CI because the polluting spec (`isIdeviceJsModule` in `src/services/idevice-config.spec.ts`) had been added by `0c52433a` and isn't on `main`. Root cause: the spec called `setIdevicesBasePath(tempDir)` in `beforeEach` but the outer `afterEach` only called `resetIdeviceConfigCache()` — which left `idevicesBasePath` pointing at a now-deleted directory. Fix: `resetIdeviceConfigCache()` also clears `idevicesBasePath`. CI now green: `make test-unit` 6949/6949.

**E2E (static)** — three of my new specs were failing on the post-reload wrapper inspection. Two causes, both spec-side:

- Wrong selector: the iDevice template (`export/three-d-viewer.html`) wraps the engine content in its own outer `<div class="three-d-viewer-wrapper" data-three-d-viewer>`, so there are two nested divs sharing the class. `querySelector('article .three-d-viewer-wrapper')` matched the outer one (no `data-*` payload). Qualified every selector with the `[data-three-d]` attribute filter to land on the inner wrapper.
- Wrong assertion target: two specs expected `data-model-src` to start with `asset://` and the wrapper attrs not to contain `blob:`. The workarea engine resolves `asset://` → `blob:` by design, so `data-model-src` is a blob URL after reload — that's expected. The canonical handle that survives is `data-model-asset-ref`. Rewrote the specs to assert on that instead.

### Tests

- Bun: 6949/6949
- Vitest (3D viewer subtree): 185/185
- Playwright suite extended with five new specs covering STL color, asset-ref recovery, no-base64 discipline
- `make fix` clean
- Patch coverage 98.5% (1 untested line in `src/services/idevice-config.ts`)

### Files

- `public/files/perm/idevices/base/three-d-viewer/edition/three-d-viewer.{js,test.js}`
- `public/files/perm/idevices/base/three-d-viewer/export/three-d-viewer.{js,test.js}`
- `public/files/perm/idevices/base/three-d-viewer/export/three-d-viewer-runtime.{js,test.js}` (new)
- `src/services/idevice-config.{ts,spec.ts}` (test-pollution fix)
- `test/e2e/playwright/specs/idevices/three-d-viewer.spec.ts`
- `IdeviceRenderer.{ts,spec.ts}` and `idevice-config-browser.{ts,spec.ts}` (PR #1810 baseline)
- assorted file-manager + asset-manager touches from the original PR

erseco:
CI failure analysis folded into the parent comment above. The remaining failure (`IdeviceRenderer > getCssLinks > image-gallery`) was a pre-existing test-pollution bug in `src/services/idevice-config.spec.ts` where `setIdevicesBasePath(tempDir)` was leaking past the spec's teardown; fixed in the squashed commit by making `resetIdeviceConfigCache()` also clear `idevicesBasePath`. The E2E spec fixes (wrong selector and post-reload assertions that didn't account for the workarea's `asset://` → `blob:` resolution) are also in the squashed commit.

mnunezcedec (approved):
Thanks

PR #917, opened 2026-01-07
Embed file manager extracted assets (nested webpages)

erseco:
This pull request improves how HTML are resolved and displayed inside iframes, both in TinyMCE and in the preview panel.
As a result, **embedded HTML pages can now be rendered correctly in the preview**, including their internal resources and relative links.

From a purely technical perspective, this works as intended and is robust.
That said, I want to be explicit about the broader context and the trade-offs involved.

## What this PR achieves

* HTML assets embedded via iframes are now resolved as self-contained documents, with their internal resources correctly loaded.
* Relative links inside embedded HTML pages can be followed, thanks to cross-frame communication.
* Asset URLs are preserved consistently for all media types (HTML, PDF, etc.), improving reliability in both editor and preview modes.
* TinyMCE integration has been adjusted so HTML assets are embedded as iframes rather than being misclassified as media elements.

In short: **embedded web pages now display correctly and predictably in the preview**.

## Important considerations

Although this behavior is now supported, embedding full HTML pages inside other documents is **not a pattern I personally like nor one I would encourage**.

This approach is conceptually similar to classic OLE (Object Linking and Embedding):
[https://en.wikipedia.org/wiki/Object_Linking_and_Embedding](https://en.wikipedia.org/wiki/Object_Linking_and_Embedding)

From experience, this kind of embedding tends to create more problems than it solves:

* It increases the attack surface and raises security concerns, especially when arbitrary HTML and JavaScript are involved.
* For this reason, most LMS and CMS platforms block or heavily restrict embedded HTML in corporate or institutional environments.
* The real-world use case is rare. Wanting to embed a full website inside another document is usually a sign that the content should be integrated differently, not nested.
* Supporting this adds complexity to the codebase, both in asset resolution and in preview logic.

An analogy would be embedding photos inside a Word document by linking to an external website instead of inserting the images directly: it works, but it is not a good way to use the technology.

## Position

This PR makes the behavior **work correctly and safely within the current constraints**, and that is its goal.
It does **not** mean that embedding full web pages inside learning content should be promoted as a standard workflow.

In my view, this should remain a niche, opt-in capability, handled carefully, and ideally restricted or disabled by default in environments where security policies apply.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -459,7 +459,7 @@ endif
 
 .PHONY: test-frontend-legacy
 test-frontend-legacy: check-node check-env ## Run frontend tests with Node.js (Vitest) + coverage (CI) - for systems without Bun
-	npm run test:frontend:node
+	npm run test:frontend:legacy
 
 .PHONY: lint-legacy
 lint-legacy: check-node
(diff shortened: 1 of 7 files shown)


erseco:
I’ve reviewed this case and agree that fully supporting HTML websites opened in a new window during **preview** is non-trivial. While embedded HTML in iframes works correctly (link interception + postMessage resolution), the same approach breaks in new windows because there is no parent context to handle internal navigation.

Making this fully functional would require dynamically generating a self-contained version of the entire website (resolving all linked HTML pages and assets recursively), which is essentially equivalent to running an export on demand. This adds significant complexity and performance cost for a use case with limited impact.

Given that:

* this issue only affects **preview**, not **export** (which already works correctly),
* and this is an edge case rather than a core workflow,

My proposed solution is to intercept these links in preview and show a clear message informing the user that HTML websites must be exported to be viewed correctly. This keeps preview behavior predictable, avoids executing complex logic client-side, and aligns with the idea that this capability should remain restricted and opt-in.

Issue #932, opened 2026-01-09
Problems when moving idevices


cristinavaldera:
When moving a idevice inside or outside a block, sometimes it doesn´t work, like in the video. 
It is also possible to move ir over the title (see video), that shouldn´t be possible.

How to reproduce the error:

1. Create idevice 1 (first block)
2. Create idevice 2 (second block)
3. Move idevice 2 inside block 1
4. Move it out again
5. Tests:
- Try to move it again into block 1 - it doesn´t work
- Try to move to over the title of block 2 - it works and it shouldn´t
- Move idevice 2 outside any block - it creaters a new block (3) to contain it but doesn´t delete the block 2 (it doesn´t  warn either)
- You can move idevice from block 2 to 3 and otherwise but not to block 1.
- If you delete block 1, all idevices disappear, only block2 and 3 (empty) stand

https://github.com/user-attachments/assets/41f83f35-fc9a-453b-8b82-6407058d145b

PR #934, opened 2026-01-09
Fix iDevice Drag & Drop: Block Synchronization, Invalid Drop Zones, and State Corruption

erseco:
This PR addresses several critical issues related to iDevice movement between blocks (boxes) and the main page content, as reported in issue #932. The fixes ensure that the internal state (Yjs and local components list) remains synchronized with the UI after drag-and-drop operations.

### Problems Solved

1.  **Block "Lock-out"**: Fixed an issue where a block (e.g., Block 1) would stop accepting iDevices after they were moved out.
    *   *Cause*: The `idevices` array within the block node was becoming stale or not properly reset, leading to failed validation during subsequent drag-over events.
2.  **Invalid Drop Zone (Block Titles)**: Previously, iDevices could be dropped on block headers/titles.
    *   *Fix*: Updated `isDragableInside` and related positioning logic to explicitly exclude `.box-head` and its children as valid drop targets. iDevices can now only be dropped into the block body.
3.  **Yjs Desync on Auto-created Blocks**: When an iDevice was dragged outside any block, a new block was created in the DOM but not always correctly registered in Yjs.
    *   *Fix*: Modified the creation flow to ensure the block is created in the Yjs shared state **before** the iDevice is attached to it, preventing "target block not found" errors.
4.  **Cascading Deletion Bug**: Deleting one block sometimes caused all iDevices on the page to disappear.
    *   *Cause*: Stale references in the `idevices` array of the block being deleted. The deletion logic was inadvertently triggering `remove()` on iDevices that had already moved to other blocks.
    *   *Fix*: Added a strict `blockId` check during deletion to ensure only iDevices belonging to the target block are removed.
5.  **Empty Block Dialog Spam**: The "Delete empty box?" dialog was appearing for blocks that were already empty, not just the one currently affected by a move.
    *   *Fix*: Refactored `setParentsAndChildrenIdevicesBlocks` to accept a specific `blockIdToCheck`. The dialog now only triggers for the block that actually lost its last iDevice.
6.  **Drag-and-Drop Order Sync**: Reordering iDevices within the same block or moving blocks themselves via drag-and-drop was not always syncing the new `order` to Yjs.
    *   *Fix*: Implemented DOM-based order calculation during drop events to explicitly update the `order` property in the shared state.

### Technical Changes

*   **`idevicesEngine.js`**: 
    *   Improved `isDragableInside` validation.
    *   Fixed `dropIdeviceContentInContent` to handle state sync and specific empty-block checks.
    *   Synchronized auto-created block IDs with Yjs bridge.
*   **`blockNode.js`**:
    *   Protected `removeIdevices()` against stale object references.
    *   Improved `removeIdeviceOfListById` to support both internal and sync IDs.
*   **`ideviceNode.js`**:
    *   Updated callers of block/page movement to pass previous block IDs for proper cleanup.

### Testing & Quality Assurance

*   **Unit Tests**: Added **11 new tests** to `idevicesEngine.test.js` and `ideviceNode.test.js` covering:
    *   Same-block vs. cross-block reordering.
    *   Specific block-emptiness validation.
    *   Invalid drop target rejection.
*   **E2E Tests**: Created a new comprehensive Playwright spec: `test/e2e/playwright/specs/idevices-draganddrop.spec.ts`. This test reproduces the exact steps from the issue video to ensure no regressions.
*   **Manual Verification**: 
    *   Verified that moving an iDevice out and back into a block works repeatedly.
    *   Confirmed that deleting a block preserves iDevices in other blocks.
    *   Verified real-time sync across two browser windows (Yjs order sync).

---

### How to reproduce (to verify fix):
1. Create Block 1 (Text 1) and Block 2 (Text 2).
2. Move Text 2 into Block 1.
3. Move Text 2 out (Block 3 is created).
4. Move Text 2 back into Block 1 -> **Should now work**.
5. Try to drop Text 2 on the title "Block 1" -> **Should be rejected**.
6. Delete Block 1 -> **Only Text 1 should disappear; Text 2 in Block 3 must remain**.

### DIFF
--- a/public/app/yjs/LegacyXmlParser.js
+++ b/public/app/yjs/LegacyXmlParser.js
@@ -591,11 +591,63 @@ class LegacyXmlParser {
   }
 
   /**
+   * Check if a page is a descendant of another page
+   *
+   * @param {Object} page - The page to check
+   * @param {string} ancestorId - The potential ancestor page ID
+   * @param {Map} pageIdMap - Map of page ID to page info
+   * @returns {boolean} True if page is a descendant of ancestor
+   */
+  _isDescendantOf(page, ancestorId, pageIdMap) {
+    let currentParentId = page.parent_id;
+    while (currentParentId) {
+      if (currentParentId === ancestorId) {
+        return true;
+      }
+      const parent = pageIdMap.get(currentParentId);
+      if (!parent) break;
+      currentParentId = parent.parent_id;
+    }
+    return false;
+  }
+
+  /**
+   * Build path from a promoted parent to a descendant page
+   *
+   * @param {Object} page - The descendant page
+   * @param {string} promotedParentId - The ID of the promoted parent
+   * @param {Map} pageIdMap - Map of page ID to page info
+   * @returns {string} Path from promoted parent to page (inclusive of both)
+   */
+  _buildPathFromPromotedParent(page, promotedParentId, pageIdMap) {
+    const pathParts = [page.name];
+    let currentParentId = page.parent_id;
+
+    // Walk up until we reach the promoted parent (inclusive)
+    while (currentParentId) {
+      const parent = pageIdMap.get(currentParentId);
+      if (!parent) break;
+      pathParts.unshift(parent.name);
+      if (currentParentId === promotedParentId) {
+        break;
+      }
+      currentParentId = parent.parent_id;
+    }
+
+    return pathParts.join(':');
+  }
+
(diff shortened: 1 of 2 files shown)


ignaciogros (approved):
In one of my tests, It didn't ask if I wanted to delete an empty box, but I can't reproduce it. I'll open a new issue if I find the way to do it.

Thank you.

Issue #930, opened 2026-01-09
iDevices: CSS url() images and icons are not displayed inside eXe


mnarvaezm:
Inside eXe, images and icons referenced from iDevice/theme CSS files using url() are not shown. As a result, the UI loses toolbar icons/backgrounds and looks incomplete.

Steps to reproduce:
- Open eXe and create a new project.
- Add a Game iDevice: Rosco.
- Save the changes.
- Check the iDevice: toolbar images/icons loaded via CSS are missing.

mnarvaezm:
Yes, I’ve tested it and it’s working correctly now. ✅
#935 fixes the issue on my side, so I’m closing the incident. Thanks!

Issue #918, opened 2026-01-08
Link to .elpx is not working


ignaciogros:
In addition to the iDevice for downloading the `.elpx` file, users can create a link to the file itself:

```
<a href="exe-package:elp" id="link_0" download="exe-package:elp-name">Download file</a>
```

A link to `exe-package:elp` is a link to the `.elpx` file, and those links are not working now.

`fflate.umd.js` and `exe_elpx_download.js` should be exported in those cases too.

[Example](https://github.com/user-attachments/files/24492912/download-elpx.zip)

PR #935, opened 2026-01-09
Fix .elpx links, user themes and iDevice css

erseco:
This update resolves critical issues with manual download links and CSS resource rendering, while implementing a more secure architecture for custom styles.

**Key Changes:**

*   **Fix for #918 (Manual .elpx Links):** Added regex detection for the `exe-package:elp` protocol in content. This ensures that the required libraries (`fflate.umd.js` and `exe_elpx_download.js`) are included in HTML5 and SCORM exports even when the download iDevice is not explicitly used.
*   **Fix for #930 (CSS Resources in iDevices):** Resolved the issue where images and icons referenced via `url()` in iDevice/theme CSS were not displaying inside eXe. The CSS parser now correctly rewrites relative paths to absolute URLs (preserving quotes and formatting), restoring missing toolbar icons and backgrounds.
*   **Secure Theme Architecture:** To prevent the execution of insecure code on the server, **User Themes** are no longer stored or served by the backend. @ignaciogros maybe now we can set `ONLINE_THEMES_INSTALL=1` by default.
    *   User themes are now stored on the client-side (**IndexedDB**) and synchronized via **Yjs**.
    *   Imported and user themes are now distributed within the `.elpx` file (similar to assets), making them portable and allowing "Online Themes Install" to be enabled by default.
*   **Preview & UI Improvements:**
    *   Improved **HTML asset resolution in iframes**: Relative URLs, CSS, and JS within embedded HTML files (from extracted ZIPs) now resolve correctly in preview mode.
    *   Added click interception for HTML assets in the workarea with a warning to prevent broken navigation before export.


ignaciogros:
Maybe it would be possible to allow local installation in the online version when `ONLINE_IDEVICES_INSTALL == -1` or "browser".

PR #973, opened 2026-01-13
Fix image compressor

erseco:
This pull request enhances the image optimizer functionality in TinyMCE by ensuring that optimizing an image updates the existing asset in the AssetManager rather than creating a new one. It also improves robustness in asset ID extraction and image size handling, and adds a comprehensive end-to-end test to verify the new behavior.

**Image Asset Management Improvements:**

* Refactored asset ID extraction logic to reliably determine the correct asset ID from various sources, including asset URLs, blob URLs, and DOM attributes, to ensure the correct asset is updated during optimization. [[1]](diffhunk://#diff-2cd047826d4f6627045ba19d8592a4edf1726576bf3ffd3b17a7d2da995ae6baR125) [[2]](diffhunk://#diff-2cd047826d4f6627045ba19d8592a4edf1726576bf3ffd3b17a7d2da995ae6baR135-R161)
* Modified the asset update flow so that, when optimizing an image, the existing asset is replaced in the AssetManager (including updating metadata and blob caches) instead of inserting a new asset, preventing asset duplication. [[1]](diffhunk://#diff-2cd047826d4f6627045ba19d8592a4edf1726576bf3ffd3b17a7d2da995ae6baL170-R200) [[2]](diffhunk://#diff-2cd047826d4f6627045ba19d8592a4edf1726576bf3ffd3b17a7d2da995ae6baR464-R525) [[3]](diffhunk://#diff-2cd047826d4f6627045ba19d8592a4edf1726576bf3ffd3b17a7d2da995ae6baR534-R548)

**Image Optimizer UI Robustness:**

* Improved logic for determining the maximum image size in the optimizer UI, handling edge cases where dimension values may be missing or invalid, and providing sensible fallbacks.

**Testing:**

* Added an end-to-end Playwright test to verify that optimizing an image replaces the existing asset rather than creating a new one, ensuring asset count remains unchanged and metadata (such as hash and size) is updated as expected.

### DIFF
--- a/public/app/workarea/menus/structure/menuStructureBehaviour.js
+++ b/public/app/workarea/menus/structure/menuStructureBehaviour.js
@@ -7,7 +7,8 @@
 import ImportProgress from '../../interface/importProgress.js';
 
 // Use global AppLogger for debug-controlled logging
-const Logger = window.AppLogger || console;
+// Use global AppLogger for debug-controlled logging
+const Logger = (typeof window !== 'undefined' && window.AppLogger) || console;
 
 export default class MenuStructureBehaviour {
     constructor(structureEngine) {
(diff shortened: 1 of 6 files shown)


ignaciogros (approved):
Thanks, @erseco.

I’ve pushed a few changes to complete the functionality. Now, when you click “Finish”, the “Source” field in TinyMCE is updated, as well as “Width” and “Height”, and the editor dialog is closed.

Each time you do that, a new base64 image is inserted. That means that the File Manager will contain two images (the original one and the new one). This can be fixed, but it would involve deleting user content without confirmation. Considering the decisions already made regarding the File Manager, I think we can:

a) Release the new version with the optimizer as it is now.
b) Hide the optimizer and leave only the new tool in the Utilities menu.

If the changes I’ve just pushed look good to you, I think that this branch can be merged into main, and the optimizer can be hidden later if we decide to go with option b.

ignaciogros (changes requested):
We should use @erseco's original solution to replace the original image after optimizing it using `assetManager`. I'll fix that, @erseco. Thanks for the explanations.

PR #976, opened 2026-01-14
fix: toolbar accessibility text visible in preview (all themes)

### LINKED ISSUE #929: Game iDevices toolbar accessibility text: visible in Preview with default theme, and .sr-av { display:none } in other themes breaks accessibility
In Preview mode, for Game iDevices, the toolbar shows the icon buttons and also the descriptive accessibility text.

Default theme: the accessibility text becomes visible because the .sr-av styling is not being loaded/applied correctly (or it’s being overridden), so content that should be “screen-reader only” is rendered on screen.

Other themes/styles: the .sr-av class exists but is defined only as display:none, which hides the text and also removes it from screen readers, breaking accessibility

mnarvaezm:
- Replace display:none with visually-hidden technique
- Add clip-path: inset(50%) as modern replacement for deprecated clip
- Keep clip: rect(0,0,0,0) for legacy browser compatibility
- Ensures content is hidden visually but accessible to screen readers

Fixes #929

### DIFF
--- a/public/files/perm/themes/base/base/style.css
+++ b/public/files/perm/themes/base/base/style.css
@@ -700,19 +700,6 @@ label[for="teacher-mode-toggler"] {
     }
 }
 
-.sr-av {
-    position: absolute;
-    width: 1px;
-    height: 1px;
-    margin: -1px;
-    padding: 0;
-    overflow: hidden;
-    clip: rect(0, 0, 0, 0); 
-    clip-path: inset(50%); 
-    white-space: nowrap;
-    border: 0;
-}
-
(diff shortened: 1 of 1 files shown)


mnarvaezm:
@ignaciogros @erseco 

Ok. I’ve seen that, after Ernesto’s PR this afternoon, the base.css file is used both in preview and in export.

I’ve updated this PR to remove the .sr-av class from the zen, flux, nova, and neo themes, since it added display: none, which made the text of elements marked with it not accessible.

ignaciogros (approved):

PR #978, opened 2026-01-14
977 icons not loading in version 40

erseco:
This pull request updates the backend and tests to ensure that theme icons are returned in a structured format required by the frontend, preventing issues where icons appeared as 'undefined'. It introduces a new `ThemeIcon` interface, updates the theme API to use this structure, and adds comprehensive tests to verify correct icon handling both at the API and UI level.

**Backend: Theme icon structure improvements**
- Added a new `ThemeIcon` interface defining the required fields (`id`, `title`, `type`, `value`) for icons returned by the API, ensuring consistency for frontend consumption.
- Updated the `ThemeConfig` and `scanThemeIcons` implementations to use `Record<string, ThemeIcon>` instead of `Record<string, string>`, so each icon now includes all required metadata. 

**Testing: API and UI validation**
- Added a new Playwright end-to-end test (`block-icons.spec.ts`) that verifies the block icon selection modal displays icons correctly and that the API returns icons with the expected structure, preventing regressions where icons could be undefined.
- Enhanced the themes route unit tests to check that the API returns icons with the proper `ThemeIcon` structure, including all required fields and correct types/values.

### DIFF
--- a/public/app/yjs/ElpxExporter.js
+++ b/public/app/yjs/ElpxExporter.js
@@ -123,6 +123,7 @@ if (typeof window !== 'undefined' && window.ElpxExporter) {
       let xml = '<odeProperties>\n';
       const props = {
         pp_title: metadata.get('title') || 'Untitled',
+        pp_subtitle: metadata.get('subtitle') || '',
         pp_author: metadata.get('author') || '',
         pp_lang: metadata.get('language') || 'en',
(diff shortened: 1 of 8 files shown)


erseco:
I fixed the issue with the icons and the toggle (was no exporthing the content if no text in box title) also solved the issue of generating the common_i18n.js

PR #982, opened 2026-01-14
fix(e2e): Cross-platform compatibility for Playwright webServer config

pabloamayab:
## Description
This Pull Request addresses an issue where the End-to-End (E2E) test suite's web server configuration failed on Windows environments. The previous configuration used Unix-style inline environment variable assignment in the shell command and hardcoded `/tmp/` paths, which are not supported or valid on Windows.

## Changes Made
- **Refactored `webServer` configuration in `playwright.config.ts`**:
  - Moved environment variable definitions from the shell `command` string to the dedicated `env` property. Windows `cmd.exe` and PowerShell do not support the `VAR=value command` syntax used by default in Unix-like shells.
  - Replaced the hardcoded `FILES_DIR='/tmp/exelearning-e2e/'` with a dynamic, cross-platform path using `path.join(os.tmpdir(), 'exelearning-e2e')`.
- **Added Imports**: Imported `path` and `os` modules to support the dynamic path generation.

## Technical Details
- **Cross-Platform Compatibility**: By using `os.tmpdir()`, the temporary directory is resolved correctly based on the operating system (e.g., `%TEMP%` on Windows, `/tmp` on Linux/macOS).
- **Environment Variable Handling**: Playwright's `webServer.env` property ensures that environment variables are correctly set for the spawned process regardless of the OS shell being used.

## Motivation
To allow developers and contributors working on Windows machines to run the full E2E test suite locally (`npx playwright test`) without needing manual workarounds or external tools (like WSL or Git Bash) just for the environment setup.

## Verification
- Confirmed that `npx playwright test` correctly starts the web server on Windows.
- Verified that the temporary directory is created in the correct OS-specific location.


erseco (approved):
It works! nice contribution!

PR #1012, opened 2026-01-15
Fix the properties setting export to fix visibility and teacher mode to fix #1009

### LINKED ISSUE #1009: idevice properties not working
idevice properties ("Visible in exportation" and "teacher mode") are not working. However, the same functionalities in Block properties do work.

erseco:
This pull request refactors how display-related properties (`visibility`, `teacherOnly`, and `cssClass`) are handled in `IdeviceRenderer` and its related tests.

These properties are now consistently read from `structureProperties`, which is the correct and intended source for structural metadata shared across all iDevices. A fallback to the legacy `properties` field is preserved to ensure backward compatibility.

In parallel, test coverage has been updated and expanded to validate the new behavior and cover relevant edge cases.

This change fixes the root cause of #1009.

---

### Why `structureProperties` instead of `properties`?

`ExportComponent` intentionally separates two different kinds of data:

```ts
export interface ExportComponent {
    id: string;
    type: string;
    order: number;
    content: string;
    properties: Record<string, unknown>;          // iDevice-specific JSON config
    structureProperties?: ExportComponentProperties; // shared structural metadata
}
```

#### `properties` (formerly `jsonProperties`)

This field contains **iDevice-specific configuration**, which varies by type:

* Rubric: rows, columns, scoring criteria
* Quiz: questions, answers, feedback
* Crossword: words, clues, grid
* Text: usually empty or minimal

These values are used to render and configure the internal behavior of each iDevice.

#### `structureProperties`

This field contains **structural metadata shared by all iDevices**:

* `visibility` – whether the component is exported
* `teacherOnly` – visibility restricted to teachers
* `cssClass` – custom CSS classes
* `identifier` – custom HTML id

These properties affect layout and rendering, not the internal logic of the iDevice.

---

### Root cause

During export, `YjsDocumentAdapter.convertComponent()` already separates these concerns:

* `jsonProperties` → mapped to `component.properties`
* structural properties → mapped to `component.structureProperties`

However, `IdeviceRenderer` was incorrectly reading `visibility`, `teacherOnly`, and `cssClass` from `component.properties`, which contains iDevice-specific JSON data instead of structural metadata.

As a result, these display-related properties did not work correctly for iDevices.

Blocks were unaffected because `ExportBlock.properties` directly stores structural properties and does not have this separation.

---

#### Renderer refactor

* `IdeviceRenderer` now reads `visibility`, `teacherOnly`, and `cssClass` from `structureProperties`
* A fallback to legacy `properties` is kept for backward compatibility
* iDevice-specific configuration continues to be read from `properties` (now referred to as `jsonProps`) for JSON data attributes

#### Tests

* Updated all relevant unit and integration tests to use `structureProperties`
* Added new tests covering:

  * missing or undefined `structureProperties`
  * legacy property formats
  * multiple properties combined
  * non-string `cssClass` values
  * correct CSS class assignment and removal

#### Types

* Added explicit import of `ExportComponentProperties` to improve type safety and readability

---

### Outcome

This refactor aligns the renderer with the intended data model, fixes incorrect property resolution for iDevices, and strengthens test coverage. The result is clearer separation of concerns, safer rendering logic, and a more maintainable foundation going forward.


ignaciogros (changes requested):
The box properties are not saved, so you can’t have a box that is hidden in exports or available only to teachers. The iDevice properties work, but the ID field always shows "undefined".

ignaciogros (approved):
Thanks, @erseco!

PR #1015, opened 2026-01-16
Image optimization (losless)

erseco:
Reduced image size by 10.8 MB (from 54 MB, 29.6% average reduction per file, up to 99% in some cases).


pabloamayab (approved):
It works correctly. Compression is good and loss is minimal with standard values. The image shows that some files increase in size because their quality percentage was very low, and setting it to 85% actually increases it.

image

ignaciogros (approved):
This change affects many files. We might not want to modify images of components developed by third parties, but in the checks I’ve done I haven’t seen any issues. That’s why I think we can merge this into `main` and try to fix any errors if we find them.

erseco:
Hi @ignaciogros and @pabloamayab 

Just to clarify: this pull request **does not add or modify the image optimizer feature**.

What this PR does is **optimize the images that are bundled with eXeLearning itself** (application and iDevice assets) using a **lossless optimization process**. These are internal assets shipped with the program, not user content.

User images and the runtime image optimizer are **not affected** by this change. This is a one-time optimization of the assets we distribute to reduce the application size.

I’m adding this comment to avoid mixing this work with the image optimizer functionality, as they are two different concerns.

PR #1030, opened 2026-01-16
Fix the url when base path is configured

erseco:
This pull request improves how the viewer base path is resolved in the `PreviewPanelManager` class.

Instead of accessing the `basePath` property directly, the code now relies on the `getBasePath()` method when building the viewer URL. This ensures the configured base path is always respected.

For example, with the following configuration:

```
BASE_PATH=/web/exelearning
```

**Previous behavior** **(WRONG)**
The extract preview opened in a new tab ignored the base path:

```
http://localhost:8080/viewer/index.html
```

**Current behavior (with this fix)**
The viewer URL correctly includes the configured base path:

```
http://localhost:8080/web/exelearning/viewer/index.html
```

This change avoids broken preview links when the application is deployed under a subpath instead of the web root.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,10 @@ on:
       - tikismaker
       - 'release/*'
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 permissions:
   contents: read
   packages: write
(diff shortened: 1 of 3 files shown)


ignaciogros (changes requested):
After running `make clean-local` and `make up-local`, I see a 404 page with no styling:



If I try to go to `:8080/iamthebasepath/`, it redirects to `:8080/iamthebasepath/iamthebasepath/workarea`, and I see this (also note the “Return to login” link):

ignaciogros (approved):
My fault, sorry… I forgot to add the `/` to the `BASE_PATH`, like this: `/mybasepath`.

It works.

PR #1033, opened 2026-01-17
Integrate v3.0.2 TinyMCE fixes and remove the image optimization tool

ignaciogros:
This PR includes all the changes made to the TinyMCE editor in v3.0.2, while preserving the necessary modifications to work with the new asset system. 

The image optimization tool from the `exeimage` plugin has also been removed.

Please review it and check if anything is missing or any tests need to be modified or removed.

**Important:**

The audio recording plugin does not work properly in export or preview, but I think it’s better to open a new issue for that, as it seems to be more related to the asset system than to the plugin itself.

### DIFF
--- a/test/e2e/playwright/specs/undo-redo-idevice-title.spec.ts
+++ b/test/e2e/playwright/specs/undo-redo-idevice-title.spec.ts
@@ -544,6 +544,22 @@ test.describe('Undo/Redo iDevice Icon - Issue #956', () => {
         // Press Ctrl+Z to undo
         await pressUndo(page);
 
+        // Wait for icon to revert (poll for the change - should no longer have an img, should have svg)
+        await page.waitForFunction(
+            () => {
+                const block = document.querySelector('#node-content article.box');
+                if (!block) return false;
+                const iconBtn = block.querySelector('header.box-head button.box-icon');
+                if (!iconBtn) return false;
+                // After undo, the icon should be empty (no img, has exe-no-icon class or svg)
+                const hasImg = iconBtn.querySelector('img') !== null;
+                const hasEmptyClass = iconBtn.classList.contains('exe-no-icon');
+                const hasSvg = iconBtn.querySelector('svg') !== null;
+                return !hasImg && (hasEmptyClass || hasSvg);
+            },
+            { timeout: 10000 },
+        );
+
         // Verify icon reverted visually (should be empty again)
         const hasEmptyIconAfterUndo = await blockHasEmptyIcon(page, 0);
         expect(hasEmptyIconAfterUndo).toBe(true);


erseco (changes requested):
Good work overall, the attribute handling looks correct and behaves as expected 👍

There are two points I’d like to highlight based on the attached screenshots:

1. **Missing accessibility warning message**
   It looks like we have lost the confirmation message:

Image

   > “Are you sure you want to continue without including an Image Description? Without it the image may not be accessible to some users with disabilities, or to those using a text browser, or browsing the Web with images turned off.”

   This warning is important from an accessibility perspective and was useful to guide authors. It would be good to confirm whether its removal was intentional or if it should be restored.

2. **Image replacement and sizing behavior**
   When replacing an image inside a template, should the editor adapt to the *template-defined size* or to the *new image’s intrinsic size*?
   From the current behavior, selecting a larger image causes the layout to break or become misaligned. My expectation would be that replacements respect the existing template dimensions to avoid visual inconsistencies.

Image

Additionally, to strengthen coverage around these flows, it might be worth adding (or re-enabling) the following E2E tests to this PR:

* [https://github.com/exelearning/exelearning/blob/7766968242d696ee01f2141a0f2256642f314f91/test/e2e/playwright/specs/idevices/text-idevices-templates.spec.ts](https://github.com/exelearning/exelearning/blob/7766968242d696ee01f2141a0f2256642f314f91/test/e2e/playwright/specs/idevices/text-idevices-templates.spec.ts)
* [https://github.com/exelearning/exelearning/blob/7766968242d696ee01f2141a0f2256642f314f91/test/e2e/playwright/specs/idevices/template-image-replace-full-flow.spec.ts](https://github.com/exelearning/exelearning/blob/7766968242d696ee01f2141a0f2256642f314f91/test/e2e/playwright/specs/idevices/template-image-replace-full-flow.spec.ts)

These scenarios are directly related to the changes introduced here and would help catch regressions early.

Apart from these points, the work is clear and well structured. Thanks for the solid update to the TinyMCE integration.

ignaciogros:
Thanks for checking everything, @erseco!

The commit `d96ded6a46118682ff42517570d6e85efa18f3df` should fix the alt text issue. The check was not missing; the problem was that when inserting an image, the file name was used as the alt text. Since the field was not empty, the accessibility warning didn’t appear.

The commit `0b6dbc9d2407d72f69c037d656d3d8e383662264` sets a `max-width` for `FIGURE` elements so the layout is not broken inside TinyMCE.

The commit `79558085211481022394af0de9720fccf5857dec` fixes an unnoticed bug: when a custom license was added in the caption, if the attribution tab wasn’t opened, the value was lost. That commit also fixes the loss of classes on `FIGURE` elements (also in "exemedia").

I don’t think it’s really necessary to add tests for these issues: if the ALT is empty, the warning will appear, and the broken layout is a presentation problem solved with a simple `max-width`. I also think none of the plugin CSS files should be updated without manual review. They are simple, and **their content should be manually controlled**. If you think they are necessary, let me know and I’ll include them.

erseco:
> Hi, @erseco.
> 
> Thank you for your complete review. Since this PR is quite blocking, I suggest merging the current changes into `main` as soon as possible and addressing those problems in a new issue. What do you think?

I agree

PR #1035, opened 2026-01-17
1034 cant reproduce recorded audios

ignaciogros:
**Do not test this before merging #1033 into "main".**

When recording audio inside eXe, its path was: `/files/tmp/uploads/content/resources/idevice-x/y.webm`

And when exporting: `/content/resources/y.webm`

I removed the initial slash, and it works, but maybe we should use the Asset Manager API instead.

@erseco: 

Could you please review this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@
 
 ## About the Project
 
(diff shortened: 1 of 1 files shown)


erseco (approved):
It works, thanks @ignaciogros !

ignaciogros:
Perfect solution for the recorded audios. Thank you, @erseco.

PR #1043, opened 2026-01-18
1042 tinymce issues when using a different base path

ignaciogros:
Mermaid and the CodeMagic editor now work with any correct `BASE_PATH`.


public/libs/tinymce_5/js/tinymce/plugins/codemagic/plugin.min.js
```diff
@@ -73,8 +73,8 @@ jQuery(document).ready(function ($) {
 				title: _('Edit source code'),
 				width: 900,
 				height: 650,
-                // maximizable: true,
-				url: basePath + '/api/codemagic-editor/codemagic.html'
+				// maximizable: true,
+				url: basePath + '/' + eXeLearning.version + '/libs/tinymce_5/js/tinymce/plugins/codemagic/codemagic.html'
```
erseco: Concatenate the version can be a little bit "tricky" let's explore a more centralised option

PR #1050, opened 2026-01-18
Fix legacy pbl task import hide text jsondata

mnarvaezm:
- Fix: fix legacy PBL task import.
- Export: hide/remove data-idevice-json-data for text iDevices during export (avoids duplicatingfix(legacy-import): default iDevice
- Titles in legacy projects are generated using the project language.
- Support: add support for FPD Solved Exercise by mapping it as Case Study.


ignaciogros (commented):
Thank you, @mnarvaezm.

I think this could be merged as it is, but there are still some elements that could be removed.

**Before:**

```
data-idevice-json-data='{"ideviceId":"...","textInfoDurationInput":"","...",...}'
data-idevice-template="text.html"
```

**After:**

`data-idevice-json-data='{"ideviceId":"..."}'`

It’s valid HTML and it hardly affects the export size, but I’m not sure whether it was left intentionally or not. Could `data-idevice-json-data` be removed?

mnarvaezm:
No, the ideviceId can’t be removed; it’s needed in the export renderBehaviour function to identify the feedback button for each text iDevice when there are multiple on the same page.
@ignaciogros  @erseco

PR #1055, opened 2026-01-19
Fix broken asset export in CLI and browser exports

erseco:
This pull request fixes a **high-priority bug** affecting project exports that caused **missing images/media and broken content** in both CLI and browser-based exports.
The issue impacts real user projects and makes exported HTML content incomplete or unusable.

The problem was detected while exporting ELP projects via CLI and confirmed across multiple ELP versions.

---

### 1. Broken asset resolution after folder-based storage migration

After the migration to folder-based asset storage (`content/resources/{folderPath}/{filename}`), asset references were no longer resolved consistently during export.

Internally, the editor, Yjs, and the backend use logical references such as:

```
asset://<uuid>
asset://<uuid>.ext
```

What was going wrong:

* These logical `asset://` URLs were leaking into exported HTML and `content.xml`.
* Exporters partially assumed a flat structure.
* Assets existed on disk but were not rendered in the exported pages.
* Images and media appeared missing despite being present in the project.

**Fix**

* Asset resolution is now centralized in `BaseExporter`.
* All `asset://` URLs are **fully resolved during preprocessing** to:

  ```
  {{context_path}}/content/resources/{folderPath}/{filename}
  ```
* This conversion is applied consistently to:

  * Component HTML content
  * Component JSON properties
* XML generation (`content.xml`) now receives already-resolved paths, avoiding duplicated or conflicting logic.

This guarantees that exported projects never contain unresolved `asset://` URLs.

---

### 2. CDATA parsing failure (ELP 4.0)

Projects generated with ELP 4.0 wrap HTML and JSON content inside CDATA blocks.

When parsed with `fast-xml-parser`, CDATA content is exposed as:

```ts
{ __cdata: '...' }
```

The existing code assumed plain strings, which caused:

* HTML not being rendered.
* JSON properties failing to parse.
* Silent export corruption.

**Fix**

* Added normalization logic to safely extract content from both:

  * Plain strings
  * `__cdata` wrapper objects
* This restores correct rendering and JSON parsing for CDATA-based projects.

---

### 3. Removal of legacy ElpxExporter fallback (browser)

A legacy JavaScript fallback implementation of `ElpxExporter` was still present and could silently override the shared TypeScript exporter depending on load order.

**Fix**

* The fallback `public/app/yjs/ElpxExporter.js` and its tests have been fully removed.
* The shared TypeScript `ElpxExporter` is now explicitly exposed on `window`.
* This guarantees a **single exporter implementation** across:

  * Browser
  * CLI
  * Tests

---

[...]

PR #1056, opened 2026-01-19
Fix image gallery when importing form eXe 2.9

erseco:
This PR fixes a regression affecting **Image Gallery preview** when importing legacy eXe 2.9 projects.

#### Root cause

There are two underlying issues:

1. **Unresolved `asset://` URLs (dependency on #1055)**
   After legacy import, Image Gallery properties was setting logical `asset://` references.
   Without the asset resolution fixes introduced in **#1055**, images cannot render correctly in preview or export.

2. **Incorrect Image Gallery rendering logic**
   A previous change attempted to prevent Image Gallery from re-rendering in certain contexts to avoid repainting issues.
   This logic was incorrect added by me, causing the gallery HTML not to be regenerated in preview, resulting in empty galleries. My apologies for that.

#### Fix

* Reverted Image Gallery rendering behavior to match **v3.0.2**, with small adjustments.
* Ensured `ideviceId` is **always included in `jsonProperties`** during legacy imports, which is required by `image-gallery.js` and many other json iDevices.
* Restored proper re-rendering in preview without relying on export-only assumptions.

This PR restores correct Image Gallery preview behavior for imported eXe 2.9 projects like this
[galeria_2.9.zip](https://github.com/user-attachments/files/24722685/galeria_2.9.zip)

### DIFF
--- a/public/app/yjs/AssetManager.test.js
+++ b/public/app/yjs/AssetManager.test.js
@@ -2225,6 +2225,62 @@ describe('prepareJsonForSync', () => {
   });
 });
 
+describe('getAssetUrlFromBlobUrl', () => {
+  let assetManager;
+
+  beforeEach(() => {
+    global.Logger = { log: mock(() => {}), warn: mock(() => {}) };
+    assetManager = new AssetManager('project-123');
+    // Set up some cached blob URLs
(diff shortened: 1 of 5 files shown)


ignaciogros (approved):
These changes work well, but the gallery HTML needs a complete review. I’ll open a new issue and try to fix that later.

PR #1059, opened 2026-01-20
Fix the export of blocks and iDevices

erseco:
This PR fixes component (block and iDevice) export/import to correctly handle assets, aligning it with the XML/ELPX export logic and avoiding `asset://` URLs in exported files.
It also adds dedicated **E2E tests** to ensure images are preserved and correctly resolved after export and import.

### What was fixed

* Asset URLs are converted to `{{context_path}}/content/resources/...` during export, matching the ELPX format.
* Only referenced assets are included in the exported ZIP.
* Exported components are now interoperable across different servers.
* No mutation of the original document during export.

### Tests added

* E2E tests for exporting **blocks** and **iDevices** containing images.
* Verification that images:

  * Are included in the exported ZIP.
  * Are correctly imported into a new project.
  * Load correctly (blob URLs, non-zero dimensions) after import.
* Coverage for single and multiple image scenarios.

### How to test manually

1. Export a **block** with images.
2. Export an **iDevice** with images.
3. Import both into:

   * A new document, or
   * A clean instance (fresh server).
4. Verify that all images are present and load correctly after import.

This confirms interoperability between different installations and servers.

#### ⚠️ Important merge order

**This PR must NOT be merged before #1055.**


ignaciogros (approved):
It works well, except for one detail:

When importing a block, the icon selected in the original `.elpx` is not preserved. It’s not important al tall, and if it’s complex to fix, I think it’s OK to merge it as it is.

When importing an iDevice, the block is imported (including its title) but without an icon (same result). In that case, it's even less important. If the icon and title are imported, it's OK; if not, fine too. The important thing is that if the block contains more than one iDevice, only the selected one is imported, and that works correctly.

I get the same results on a clean installation.

I attach a ZIP with an `.elpx`, a block, and an iDevice: [testing-1059.zip](https://github.com/user-attachments/files/24741286/testing-1059.zip)

PR #1063, opened 2026-01-20
Refactor Page Export to Client-Side logic to fix #1058

### LINKED ISSUE #1058: Images missing when exporting pages from older resources and opening them in 4.0
When exporting a page from a resource created with version 2.9 as well as from one created with version 3.0, and then opening it in version 4.0, the images are not displayed.
After renaming the .elpx file to .zip to inspect the exported content, it can be seen that the images are not included in the exported file.

erseco:
This PR refactors the "Export Page" functionality to perform the operation entirely on the client-side, replacing the previous server-side implementation.

**Why this change is necessary:**
1.  **Unsaved Assets:** The previous server-side approach failed to include assets (images, media) that had not yet been saved to the file system (residing only in IndexedDB). This ensures that recently added content is exported correctly even if the project hasn't been saved.
2.  **Serverless/Desktop Compatibility:** Moving this logic to the client removes the dependency on the Python backend for page exports. This is a requirement for the future static HTML/Desktop version where no backend server will be available.
3.  **Efficiency:** It leverages the existing JavaScript exporters, reducing unnecessary server round-trips.

**Key Changes:**
*   Updated `menuStructureBehaviour.js` to trigger the export via `pageExportHelper.js` instead of calling the backend API.
*   Implemented asset filtering in `PageElpxExporter` to ensure only assets referenced by the specific page subtree are included (fixing an issue where the entire project's assets were being exported).
*   Added comprehensive E2E tests covering these scenarios.

**How to test:**
1.  Open eXeLearning and create a new project.
2.  Add a page and insert an image (do **not** save the project yet).
3.  Right-click the page in the outline and select **Export Page**.
4.  Verify that the `.elpx` file is downloaded.
5.  Import this file into a new project and confirm the image is visible (proving IndexedDB assets were included).

**Testing:**
*   Added `test/e2e/playwright/specs/page-export-import.spec.ts` verifying export/import flows with both saved and unsaved assets.

#### ⚠️ Important merge order

**This PR must NOT be merged before #1055.**

### DIFF
--- a/test/e2e/playwright/helpers/workarea-helpers.ts
+++ b/test/e2e/playwright/helpers/workarea-helpers.ts
@@ -1098,6 +1098,10 @@ export async function exportPage(page: Page, nodeId: string): Promise<Download> 
     const navElement = page.locator(`.nav-element[nav-id="${nodeId}"]`);
     await navElement.waitFor({ state: 'visible', timeout: 10000 });
 
+    // Hover over the nav element to reveal the settings trigger
+    await navElement.hover();
+    await page.waitForTimeout(300);
+
     // Find and click the dropdown trigger (three dots menu)
     const dropdownTrigger = navElement.locator('.page-settings-trigger');
     await dropdownTrigger.waitFor({ state: 'visible', timeout: 5000 });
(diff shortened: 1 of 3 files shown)


ignaciogros (changes requested):
The page was not imported. I followed these steps:

- New content.
- New Text iDevice with two images.
- Save the iDevice, but do not save the content.
- Export the page.
- Open a previous content.
- Select the first page (index).
- Click on the three dots and then on **“Import page”**.

I get this error: **“Please wait for the project to fully load before importing.”**

[.zip with the exported page](https://github.com/user-attachments/files/24741486/new-page.zip)

Issue #1060, opened 2026-01-20
Empty license when opening projects created with exe 3


cristinavaldera:
When opening a elpx created with eXe 3, the license field seems to be empty.

Issue #1049, opened 2026-01-18
Handle empty licenses


ignaciogros:
eXe 2.9 allowed content with an empty license, but when opening such content in the new eXe, the license is changed to the default one, which is not correct.

I’m attaching [two contents without license](https://github.com/user-attachments/files/24695538/20260117.zip): one created with eXe 2.9 and another from eXe 3 that does not have `pp_license`.

**How should eXe behave?**

- Open them with the proprietary license.
- Add support for an empty license (which in practice is equivalent to a proprietary license, since the actual license is unknown).

I think it would be good to get @cristinavaldera’s and @mnunezcedec opinion on this.

cristinavaldera:
In my opniion, if there is no license selected, it should remain without license.

mnunezcedec:
I agree with @cristinavaldera; if there's no license selected in the original, the new one should remain unlicensed

PR #1070, opened 2026-01-20
Align license handling with eXe 3 and restore missing legacy licenses

erseco:
Fixes issues #1060 and #1049.

This PR restores full license compatibility with eXe 3 by:
- Adding the 19 missing legacy licenses used in eXe 3.
- Introducing a centralized license registry as a single source of truth.
- Reusing and adapting the same license handling logic used in eXe 3 for import, export, and rendering.

Empty or unknown licenses from legacy projects are now preserved instead of being
silently replaced with a default license, ensuring correct behaviour when opening
projects created with older eXe versions.

### DIFF
--- a/src/shared/export/constants.spec.ts
+++ b/src/shared/export/constants.spec.ts
@@ -24,6 +24,7 @@ import {
     getLicenseClass,
     getLicenseUrl,
     LICENSE_REGISTRY,
+    shouldShowLicenseFooter,
 } from './constants';
 import { resetIdeviceConfigCache, loadIdeviceConfigs } from '../../services/idevice-config';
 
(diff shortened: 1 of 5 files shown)


ignaciogros (changes requested):
I’ve discussed this with @cristinavaldera, and for now we suggest the following behavior:

- We add an empty license to eXe 4, so if a project has that license, the preference is preserved.
- eXe 4 will keep the reduced list of licenses it has had so far.
- Any Creative Commons license (regardless of its version) is mapped to the CC license available in eXe 4. Public Domain and Proprietary licenses are preserved too.
- Any other license is left empty for now.

That would be the first step.

**Step 2 (possible solutions):**

a) If a license is not in the list, warn the user. Something like: The content license is not available in this version of eXeLearning. Please review the "License" field.
b) If a license is not in the list, store it as text. For now, it’s a regular SELECT and doesn’t allow saving free text.

**Question:**  

Should we warn the user when converting, for example, from CC-x 3 to CC-x 4?

erseco:
We shouldn’t “upgrade” or auto-convert licenses (e.g. CC 3.x → CC 4.0). That would be a relicensing action and we can’t assume users have the rights to do that.

**Proposed behavior (simple, minimal code):**

* Keep the **complete 3.0 license registry** (including legacy ones) as the single source of truth.
* Some entries in that registry will be marked as **obsolete/legacy** (e.g. `obsolete: true` or `legacy: true`), so we can deprecate licenses centrally over time.
* When opening a project:

  * If the stored license exists in the registry (even if obsolete), **preserve it exactly as stored** and show it as the current value.
  * Obsolete/legacy licenses must be **read-only in practice**: they should not be selectable from the dropdown.

* UI approach:

  * Render the `<select>` with **only non-obsolete** licenses.
  * If the current project license is obsolete/legacy, inject a synthetic option at the top:

    * `selected + disabled + hidden`, label like `EUPL (obsolete)` / `GPL-2.0 (legacy)`.
    * This keeps the displayed current value, but it won’t appear in the dropdown list.
  * Optionally show a small warning/badge near the field: “Obsolete/legacy license – please review”.
* Saving:

  * Do **not** replace obsolete/legacy licenses by default.
  * Only overwrite the stored value if the user explicitly selects another (non-obsolete) license from the dropdown.

This keeps full backward compatibility, avoids implicit relicensing, and lets us manage deprecations from a single place (the registry).

PR #1077, opened 2026-01-22
Remove redundant Mermaid library detection code

erseco:
Mermaid diagrams are always pre-rendered to static SVG in the workarea before export. The `skipMermaid` option and related detection code were unnecessary and caused a bug where `mermaid.min.js` was trying to be loadod when opening preview panel.

### Changes
- Remove `skipMermaid` option from `LibraryDetectionOptions`
- Remove Mermaid pattern from `LIBRARY_PATTERNS`
- Clean up conditional logic in exporters

### Testing
1. Create a page with a Mermaid diagram
2. Click Preview/Export
3. Check the generated `index.html`

**Before:** The file references `mermaid/mermaid.min.js` (incorrectly included)
**After:** No Mermaid library reference (correct, since diagrams are already SVG)

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,7 +113,7 @@
         "@vitest/coverage-v8": "^4.0.16",
         "@vitest/ui": "^4.0.16",
         "cross-env": "^10.1.0",
-        "electron": "^39.2.7",
+        "electron": "^40.0.0",
         "electron-builder": "^26.0.12",
         "esbuild": "^0.27.2",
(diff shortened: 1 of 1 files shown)


ignaciogros (approved):
Thanks. 

I didn't notice before that the JS was referenced, but the file was missing:

`<script src="libs/mermaid/mermaid.min.js"> </script>`

PR #1103, opened 2026-01-26
Feature: Static Runtime Unification and Redesigned Desktop App

erseco:
This PR introduces a **Static Runtime** mode for eXeLearning. This version allows the application to run entirely from static assets without requiring a backend server. This architecture is now the foundation for our desktop installers (Electron) and static web deployments.

### Motivation: Why this change?
1.  **Security & Antivirus Compatibility:** Previously, the desktop app ran a local HTTP server. Many antivirus programs flag internal servers as malicious activity. By serving files directly through a custom protocol (`app://`), we eliminate these false positives and improve security.
2.  **Performance & Efficiency:** Removing the local server layer results in faster boot times and lower memory consumption on desktop versions.
3.  **Unified Codebase:** We’ve implemented a "Hexagonal Architecture" pattern using Data Providers and Adapters. Whether the app is "Online" (Elysia API) or "Static" (Bundled JSON), the frontend code uses the same API calls.
4.  **Reliability:** By testing the exact static version that users install, we ensure that "offline" bugs are caught before release.

### Key Changes
*   **Redesigned Installer Core:** The Electron app now uses a custom protocol handler instead of an internal HTTP server.
*   **New Build Commands:**
    *   `make build-static`: Bundles all iDevices, themes, and translations into a single `dist/static` folder.
    *   `make up-static`: Serves the static version locally for quick testing.
*   **E2E Testing Expansion:** Our Playwright suite now runs three parallel projects:
    *   **Chromium** (Online mode)
    *   **Firefox** (Online mode)
    *   **Static** (The bundled offline mode)
*   **Test Refactoring:** We unifed many test helpers to handle the differences between URL structures in static vs. online modes, fixing several "flaky" tests in the process.
*   **Automated PR Previews:** Every Pull Request now automatically deploys its static version to **GitHub Pages**. This allows reviewers to test changes instantly in a browser without downloading anything.

### Technical Note
We have moved `main.js`, `preload.js`, and `update-manager.js` into an `/app` directory to clearly separate the desktop-specific logic from the core web application. 

*Apologies for the large diff; this unification was necessary to ensure that the API-based and Static-based versions of eXeLearning don't drift apart in functionality.*

--- 

**Testing Instructions:**
1. Run `make build-static` to generate the bundle.
2. Run `make up-static` to verify the application works without a backend.
3. Run `make test-e2e-static` to trigger the new specialized test suite.

### DIFF
--- a/public/app/workarea/modals/modals/pages/modalOpenUserOdeFiles.test.js
+++ b/public/app/workarea/modals/modals/pages/modalOpenUserOdeFiles.test.js
@@ -1824,48 +1824,83 @@ describe('modalOpenUserOdeFiles', () => {
       expect(window.eXeLearning.app.project.refreshAfterDirectImport).toHaveBeenCalled();
     });
 
-    it('should throw error when yjsBridge is not available in static mode', async () => {
+    it('should fallback to legacy flow when yjsBridge is not available in static mode', async () => {
       window.eXeLearning.app.capabilities = { storage: { remote: false } };
       window.eXeLearning.app.project._yjsBridge = null;
 
+      // Mock legacy upload path
+      window.eXeLearning.app.api.postLocalLargeOdeFile = vi.fn().mockResolvedValue({
+        responseMessage: 'OK',
+        odeFileName: 'test.elpx',
+        odeFilePath: '/tmp/test.elpx',
+      });
+
+      const consoleErrorSpy = vi.spyOn(console, 'error').mockImplementation(() => {});
+
       const mockFile = new File(['test'], 'test.elpx', { type: 'application/octet-stream' });
 
-      await expect(modal.largeFilesUpload(mockFile, false, false, false, false)).rejects.toThrow('Yjs bridge not initialized.');
+      await modal.largeFilesUpload(mockFile, false, false, false, false);
+
+      // Should log error and fallback to legacy flow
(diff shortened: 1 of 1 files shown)


public/app/common/edicuatex/js/edicuatex-tools.js
```diff
@@ -97,6 +97,35 @@ document.addEventListener("DOMContentLoaded", function() {
     var url = "https://cdnjs.cloudflare.com/ajax/libs/mathjax/3.2.2/es5/tex-svg.min.js";
     if (isInExe) {
         url = parent.tinymce.activeEditor.settings.edicuatex_mathjax_url;
+
+        // Detect app base path from edicuatex iframe URL for subdirectory deployments
```
ignaciogros, line 101: “edicuatex”'s been developed by third parties. Maybe we can contribute the changes to the project so it can be used as-is from the original repository. We’ll need to decide what to do with the `.test` file.

erseco: The educuatex test is not testing edicuatex, just is testing that exe load on static/online with differente base paths, but we can remove from there if you prefer.

erseco: Ok to send a PR to edicuatex project to improve the base_path support, this will ease our codebase too!

PR #1121, opened 2026-01-27
UI and exports revision (IV)

ignaciogros:
**Changes:**

- Improve the progress bar presentation.
- New option to display toast messages inside modals.
- System JavaScript alerts replaced with toasts in the File Manager.
- Image Gallery iDevice presentation revision.
- Revision of third-party library licenses and attributions.
- Global fonts licenses and attributions.
- ~The Base style now uses SVG icons instead of PNG.~
- Review the Structure menu presentation and visually emphasize highlighted pages within the application.

---

**Important information about the last point:**

When saving the page properties (e.g., marking a page as "highlighted"), `compose()` was being called twice:

1. First by `resetStructureData()` with correct properties from Yjs (highlight: true)
2. Then by `syncStructureToLegacy()` via `setDataFromYjs()`, which passed incomplete data without `odeNavStructureSyncProperties`.

The second call overwrote the correctly-built menu, causing the new "nav-element-highlighted" class to disappear.

**Solution:**

Modified `setDataFromYjs()` to call `getStructureFromYjs()` instead of using the passed data parameter. 

This ensures the method always retrieves complete page data including properties.

### DIFF
--- a/assets/styles/components/_menu_nav.scss
+++ b/assets/styles/components/_menu_nav.scss
@@ -457,7 +457,7 @@
                 }
                 &.nav-element-highlighted:not(.selected) {
                     > .nav-element-text {
-                        color: var(--brand-primary-dark);
+                        color: var(--brand-primary);
                         font-weight: 600;
                     }
 


erseco (changes requested):
Thanks for this PR, the changes look good overall! Just a couple of observations:

1. **SVG icons quality**: Some of the new SVG icons appear slightly blurry or pixelated. Could you check if they need optimization or if the viewBox/dimensions need adjustment?

2. **Highlighted page styling**: The green color for highlighted pages actually draws less attention than the default black text. Consider adding `font-weight: bold` or a different visual indicator. Also, relying solely on color can be problematic for users with color vision deficiency (particularly red-green colorblindness). A secondary visual cue like bold text or an icon would improve accessibility.

erseco (approved):
Hi agree with both!

PR #1125, opened 2026-01-27
Fix v3.0 ELP import: detect custom/ folder assets and handle filename spaces

erseco:
ELP files created with eXeLearning v3.0 (PHP/Symfony) store File Manager assets in a `custom/` folder. The current import code does not detect these assets, causing broken image references.

There is also a filename normalization bug in v3.0: files in `custom/` keep original names with **spaces** (e.g., `11 A1.png`), but XML references use **underscores** (e.g., `11_A1.png`).

**Example from a real v3.0 ELP:**
- File on disk: `/custom/11 A1.png` (with space)
- XML reference: `{{context_path}}/20251211173343CEGCIN/11_A1.png` (with underscore)
- Result: Broken image

### Changes

1. **Detection**: Add `custom/` folder to the list of recognized asset locations during ELP extraction

2. **Normalized mapping**: When extracting files from `custom/` that contain spaces, create an additional mapping with underscores so both lookup paths work

3. **Enhanced lookup**: Add fallback strategies in `findAssetUrl()`:
   - Try `custom/` + filename
   - Try denormalized filename (underscores → spaces) in `custom/`
   - Search `custom/` folder for matching filename

4. **Unit tests**: Added test coverage for all new functionality

### How to test

**Automated tests:**
```bash
make test-unit
```

**Manual testing with a v3.0 ELP:**
1. Open a v3.0 ELP file that uses File Manager (has `custom/` folder with assets)
2. Verify images load correctly in the editor
3. Check browser console for `[AssetManager] Detected custom/ folder asset` log messages

**Backwards compatibility:**
- Test with modern ELPX files (should work as before)
- Test with legacy ELP using `resources/` folder (should work as before)

### DIFF
--- a/public/app/yjs/AssetManager.test.js
+++ b/public/app/yjs/AssetManager.test.js
@@ -3289,6 +3289,62 @@ describe('convertContextPathToAssetRefs', () => {
       expect(result).toContain('asset://');
       expect(result).toContain('uuid-search');
     });
+
+    it('finds asset in custom/ subfolder via loop when denormalized filename matches', () => {
+      // This tests Strategy 3: files in SUBFOLDERS of custom/
+      // Strategy 1 (custom/ + filename) won't match because actual path has subfolder
+      // Strategy 2 (custom/ + denormalized) won't match for same reason
+      // Strategy 3 loops through all custom/ entries and finds the match
+      const assetMap = new Map([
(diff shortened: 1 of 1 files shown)


pabloamayab (approved):
Great job!!! Thanks!!!

Issue #1151, opened 2026-02-02
None of the iDevices work correctly.


mnarvaezm:
The main issue is that the scripts for each iDevice type are loaded multiple times on the same page.
**How to reproduce the issue:**
- Open a project and add a rosco iDevice.
- Complete a word in the game and save.
- Click Start game.

**Expected result**
The iDevice should work normally.

**Actual result**
The iDevice does not respond, shows errors in the console, and, in JSON-based iDevices, exhibits erratic behavior (duplicated buttons, corrupted data, etc.).

ignaciogros:
Hi @mnarvaezm:

#1161 has been merged, but there are still errors. Please open [idevices-tests.zip](https://github.com/user-attachments/files/25068858/idevices-tests.zip). You'll se duplicated buttons, and some JavaScript events are not working.

ignaciogros:
Hi @mnarvaezm,

Please confirm if this issue can be closed. I think it's fixed.

Thank you.

mnarvaezm:
Fixes #1161 in this PR

Issue #1107, opened 2026-01-26
When editing a map iDevice, the text of the first point disappears


3DCube-KNX:
When editing a map iDevice, the text of the first point disappears, and it won't allow me to proceed until I type something in the text editor.
This happens every time I open it for editing.
The map is part of an exlearning file created with previous versions.

Regards

Issue #979, opened 2026-01-14
Interactive video iDevice from version 2.9 displays error and prevents saving in version 4.0


mnunezcedec:
When opening an .elp file created in version 2.9 that contains an "Interactive Video" iDevice in version 4.0, the video title and the "Start" button do not appear. Instead, the message "This video has no interactive elements" is displayed.

<img width="561" height="214" alt="Image" src="https://github.com/user-attachments/assets/4eb3e753-9de7-4469-9ece-620075be1d13" />

Additionally, if you enter edit mode for this iDevice, everything appears to be configured correctly, but it is impossible to save changes. The page must be refreshed to continue working with eXe.

Note: If the "Interactive Video" iDevice is created from scratch in version 4.0, it seems to function correctly and can be edited.

PR #1161, opened 2026-02-02
This PR fixes multiple iDevice-related issues that were caused by duplicate loading of iDevice scripts, incorrect mappings during import, and a few additional bugs.

mnarvaezm:
**Key fixes:**
- Prevents duplicate script loading in single-page application mode
- quick-questions-multiple-choice: Adds null guards to prevent crashes when the question index is invalid (branching mode with corrupted data)
- interactive-video: Fixes local asset playback by allowing blob: URLs in the editor and improves MIME type handling
- quick-questions-video: Fixes video playback using asset:// URLs and media library validation
- ELP import: Fixes trivial iDevice mapping (it was incorrectly mapped to quick-questions)
- Asset Manager: Fixes duplicate /api prefix in asset API URLs and improves server response handling
- Legacy iDevices: Fixes duplicate feedback buttons in reflection/activity iDevices and optimizes JSON sanitization
**Fixes** #1151, #1107, #979

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -146,7 +146,7 @@ bundle: deps
 # APP_ENV=dev (default): Elysia with hot-reload + SCSS watcher
 # APP_ENV=prod: Pre-compiled Elysia, no watchers
 .PHONY: up-local
-up-local: check-bun deps css bundle
+up-local: check-bun check-env deps css bundle
 ifeq ($(APP_ENV),prod)
 	@echo "Starting local (prod mode)..."
 	@$(MAKE) bundle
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -517,7 +517,11 @@ check-coverage: check-bun ## Check that all files have at least 90% coverage
 	@bun run scripts/check-coverage.ts < /tmp/exe-coverage.txt
 
 # Test environment: in-memory database for isolation, no BASE_PATH
+ifeq ($(SYSTEM_OS),windows)
```
erseco, line 520: Perfect!

Issue #1117, opened 2026-01-27
Improve document change detection to avoid unnecessary save prompts


cristinavaldera:
The current change detection logic could be improved to provide a better user experience.

When the application starts with an empty workarea (no actual changes made), the static version currently:
- Shows the save button red dot
- Prompts "Do you want to save?" when opening an existing document or creating a new one

Since no real changes have been made to the empty document, these prompts are unnecessary and interrupt the workflow.

**Expected behavior:**
- Save icon red dot should only turn on when actual changes are made
- No save confirmation should appear when switching from an untouched empty workarea to opening/creating a document

This would streamline the user experience, especially for users who frequently open existing documents right after launching the application.

PR #1195, opened 2026-02-04
fix: prevent data loss by warning user that language changes require a reload

franmate:
Simple solution to prevent window.location.reload() in static mode.
Adds a warning: "Language changes will take effect after restarting or creating a new project."

Note: This does not resolve the default license change issue.

### DIFF
--- a/public/app/workarea/user/preferences/userPreferences.js
+++ b/public/app/workarea/user/preferences/userPreferences.js
@@ -11,7 +11,8 @@ export default class UserPreferences {
         type: 'text',
         hide: true,
     };
-
+    userLang='';
+    
     /**
      * Load user preferences
      *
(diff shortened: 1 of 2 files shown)


erseco:
Maybe this will be superseded by #1117

erseco (changes requested):
Thanks for the fix, @franmate. Preventing data loss here is important.

It would be great to add an E2E test that validates this behavior: changing the language in static mode should show the warning instead of reloading the page. The static E2E suite is already failing, so it'd be a good opportunity to cover this case explicitly.

Also, it might be worth extending the test coverage to other user preference changes beyond language. If any other preference triggers a reload or has similar side effects in static mode, we should make sure those are handled and tested too.

PR #1212, opened 2026-02-05
Fix SCORM/IMS Export: Library Inclusion, Pre-rendering, and Interface Cleanup

pabloamayab:
## Summary
Refactored SCORM 1.2, SCORM 2004, and IMS exporters to align with `Html5Exporter` logic, ensuring all required libraries and assets are included and content is correctly pre-rendered.

## Critical Changes
*   **Library Detection**: Implemented `LibraryDetector` integration in SCORM/IMS exporters. This ensures that when a user uses FX effects (Accordion, Tabs), Music notation (abcjs), or other interactive elements, the necessary JavaScript and CSS libraries are automatically detected and included in the exported ZIP. Previously, these were missing, causing 404 errors in LMSs.
*   **Pre-rendering**: Added support for pre-rendering complex content:
    *   **LaTeX**: Both standard LaTeX and DataGame LaTeX are now pre-rendered to SVG/MathML/HTML during export. This removes the strict dependency on runtime MathJax loading, which can be problematic in some restrictive LMS environments.
    *   **Mermaid Diagrams**: Diagrams are pre-rendered to static SVGs, ensuring they display correctly even if the LMS blocks external scripts or has restrictive CSPs.
*   **Global Fonts**: Added logic to bundle global fonts (e.g., OpenDyslexic) and inject the corresponding CSS, ensuring accessibility settings are preserved in the export.

## Technical Details
*   **Code Porting**: Logic was ported from `Html5Exporter.ts` to `Scorm12Exporter.ts`, `Scorm2004Exporter.ts`, and `ImsExporter.ts` to establish feature parity across all export formats.
*   **Interface Improvements**:
    *   Updated `ZipProvider` interface in `src/shared/export/interfaces.ts` to use a strict `ZipGenerateOptions` type instead of `any`, resolving lint warnings.
    *   Added missing properties (`author`, `license`, etc.) to `ScormManifestOptions` to ensure valid manifest generation.
*   **Cleanup**: Removed obsolete comments and cleaned up debug logging.

## Testing
*   **Unit Tests**: ran `bun test src/shared/export/exporters`
    *   Result: **500 passed**, 0 failed.
    *   Verified structure of ZIP files, presence of `imsmanifest.xml`, inclusion of library files, and correct HTML generation.

## Reason for Diff Size
The diff is large because the export logic for SCORM/IMS was significantly rewritten to support modern eXeLearning features (libraries, pre-rendering) that were previously only available in the HTML5 export. This is a necessary modernization to support current content types in LMS exports.

### DIFF
--- a/src/shared/export/exporters/ElpxExporter.ts
+++ b/src/shared/export/exporters/ElpxExporter.ts
@@ -91,7 +91,8 @@ export class ElpxExporter extends Html5Exporter {
             // 1.0 Pre-fetch theme to get the list of CSS/JS files for HTML includes
             const { themeFilesMap, themeRootFiles, faviconInfo } = await this.prepareThemeData(themeName);
 
-            // 1.1 Generate HTML pages (with optional Mermaid pre-rendering)
+            // 1.1 Generate HTML pages with optional Mermaid pre-rendering, store for later — manifest script tag injection happens after manifest is created)
+            const pageHtmlMap = new Map<string, string>();
             let mermaidWasRendered = false;
 
             for (let i = 0; i < pages.length; i++) {
@@ -150,7 +151,7 @@ export class ElpxExporter extends Html5Exporter {
                 const encoder = new TextEncoder();
                 baseCss = encoder.encode(baseCssText);
             }
-            this.zip.addFile('content/css/base.css', baseCss);
+            addFile('content/css/base.css', baseCss);
 
             // 1.4 Add eXeLearning logo for "Made with eXeLearning" footer
             try {


ignaciogros (changes requested):
Thank you, @pabloamayab.

It's still happening in the Single page and the ePub exports.

Could you please check that?

pabloamayab:
Hi @ignaciogros !! Could you review it?

cristinavaldera:
Hi, @pabloamayab. I´ve idetified several issues (with the elpx attached):

**1. SCORM progress tracking does not work in Moodle**
SCORM packages do not register progress in Moodle: Page navigation does not mark the completion check. The issue occurs both when uploading the project as a SCORM activity and using the plugin.

**2. Style icons not visible in “Single Page”**
In Single Page mode, the style icons are not displayed for the following styles:
- Nova
- Flux
- Universal
- Zen

**3. Space used by “Classify” iDevice in Single Page export**
When exporting to Single Page, the space required by the Classify iDevice seems to be proportional to the size of the .elp file, resulting in an unpredictable layout. A video is attached showing the described issue.

https://github.com/user-attachments/assets/d4f692b8-fd8f-4171-bbf3-ab29011aab27

[pruebas-scorm-y-single-page---nova.elpx.zip](https://github.com/user-attachments/files/25148349/pruebas-scorm-y-single-page---nova.elpx.zip)

ignaciogros:
After the latest changes, the original issue (Library Inclusion and Pre-rendering) is resolved in all exports except the first one: Download as + .elpx.

In that case, Mermaid diagrams are not pre-rendered.

I confirm that when styles with SVG icons are used, they are not displayed because the path points to the .png. Example: ``. I open a separate issue for this: #1248 

I have not been able to reproduce the SCORM traceability issues mentioned by Cristina.

I am also opening another issue for the "Classify" presentation problem: #1249

pabloamayab:
Hi @ignaciogros @cristinavaldera I have been testing Mermaid, Tooltip, Mathjax, Music Notation, etc., in EPX (Save as static and download online), Website, Single Page, SCORM, IMS, and EPUB in both versions of eXe. I have also checked that hidden pages are not displayed and that SCORM works correctly in Moodle and SCORMCloud.

I think everything you mentioned is now working. Please run your tests again.

Issue #1220, opened 2026-02-05
Non-editable exports


ignaciogros:
Export any project with the default value for "Editable export" (enabled).

**These exports are editable (they have the content.xml file):**

- .elpx
- IMS
- SCORM
- Web site

**These are not (no content.xml file):**

- ePub
- Single page

erseco:
Fixed in #1225.

The problem had two parts. SCORM 1.2, SCORM 2004 and IMS were always including `content.xml` without checking the `exportSource` flag. ePub and Single page were never including it at all.

Now all formats respect the `exportSource` metadata property, matching the behavior from eXe 3:

| Format | exportSource=true (default) | exportSource=false |
|---|---|---|
| .elpx | content.xml | content.xml (always) |
| IMS | content.xml | NO content.xml |
| SCORM 1.2 | content.xml | NO content.xml |
| SCORM 2004 | content.xml | NO content.xml |
| Web site | content.xml | NO content.xml |
| ePub | content.xml | NO content.xml |
| Single page | content.xml | NO content.xml |

PR #1225, opened 2026-02-06
Fix: Respect exportSource flag in all export formats (#1220)

erseco:
All export formats now respect the `exportSource` metadata property ("Editable export" in the UI). Previously only the Web site exporter checked this flag. SCORM 1.2, SCORM 2004 and IMS always included `content.xml` regardless of the setting, while ePub and Single page never included it.

With this change, when `exportSource` is `true` (the default), all formats include `content.xml` so the exported file can be re-imported into eXeLearning. When `exportSource` is `false`, `content.xml` is excluded. The `.elpx` format always includes it regardless of the flag.

This matches the behavior from eXe 3:

| Format | exportSource=true (default) | exportSource=false |
|---|---|---|
| .elpx | content.xml | content.xml (always) |
| IMS | content.xml | NO content.xml |
| SCORM 1.2 | content.xml | NO content.xml |
| SCORM 2004 | content.xml | NO content.xml |
| Web site | content.xml | NO content.xml |
| ePub | content.xml | NO content.xml |
| Single page | content.xml | NO content.xml |

**Changes:**

- `Scorm12Exporter.ts`, `Scorm2004Exporter.ts`, `ImsExporter.ts`: wrapped `content.xml`/DTD addition with `if (meta.exportSource !== false)`
- `Epub3Exporter.ts`, `PageExporter.ts`: added `content.xml` generation when `exportSource !== false` (previously these formats never included it)
- Updated and added tests for all five exporters (6376 tests pass, no lint issues)

### DIFF
--- a/src/db/queries/admin.spec.ts
+++ b/src/db/queries/admin.spec.ts
@@ -732,7 +732,7 @@ describe('Admin Queries', () => {
             const userId = await seedTestUser(db, { email: 'owner@test.com', user_id: 'owner' });
             await seedTestProject(db, userId, { title: 'Active Project' });
 
-            // Create archived project
+            // Create archived but saved project
             const now = Date.now();
             await db
                 .insertInto('projects')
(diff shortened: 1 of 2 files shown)


ignaciogros:
Hi @erseco,

This is fixed in #1212. Since that ticket also includes other fixes, I suggest closing this one and continuing the discussion in #1212.

erseco:
No problem, you can close it. I just did it because you assigned it to me on the #1220 issue.

PR #1229, opened 2026-02-06
Fix collaborative editing: user list sync (#1164) and content visibility (#1169)

### LINKED ISSUE #1164: Collaborative editing. Synchronize user list.
The list of users should be synchronized for the user who has shared the content. Right now you can't see other users.

### LINKED ISSUE #1169: Collaborative editing. Can't see changes made by other user.
A user is on a page, another user enters an iDevice on that page and edits its content, and then saves. User A does not see the changes unless reloading the page or navigating to another one and then come back.

erseco:
When the initiator (first user) opened a project alone, the Yjs sync protocol handshake never completed bidirectionally. The `_checkForOtherUsers()` optimization (300ms timeout) made the initiator skip the full sync wait. When a joiner arrived later, only the joiner received the initiator's state (via SyncStep2), but the initiator never received the joiner's state back. This caused two problems: the initiator couldn't see other users in the concurrent users list, and couldn't see content changes made by joiners without reloading.

**Fix:** When the initiator was alone during init, an awareness listener is registered. When the first remote user is detected, the WebSocket provider disconnects and reconnects to re-run the full Yjs sync handshake bidirectionally. This only triggers once per session (`_hasResynced` flag). As a fallback, the server also sends a `trigger-resync` message to existing clients when a new client joins.

Changes:
- Set user info in awareness early (before WebSocket connects) instead of after UI init
- Added `_setupCollaborationResync()` and `_forceResync()` to `YjsDocumentManager`
- Server sends `trigger-resync` to existing clients when a new client joins
- `AssetWebSocketHandler` handles the `trigger-resync` message
- Un-skipped 2 previously flaky E2E tests
- Added new E2E tests for user presence sync and bidirectional content sync
- Added unit tests for all new functionality (8 new tests in YjsDocumentManager, 2 in yjs-websocket, 1 in message-parser)

Closes #1164, closes #1169

### DIFF
--- a/public/app/workarea/interface/elements/concurrentUsers.js
+++ b/public/app/workarea/interface/elements/concurrentUsers.js
@@ -17,6 +17,8 @@ export default class ConcurrentUsers {
         this.currentUsers = [];
         this.app = app;
         this.unsubscribe = null;
+        this._initRetryTimer = null;
+        this._subscribedDocumentManager = null;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


ignaciogros (changes requested):
Hi @erseco,

- user (Firefox) creates a content and shares it with user2
- user2 (Google Chrome) uses the link provided by user to see/edit the content.
- user2 sees the online users button.
- user does not see that button, not even after some seconds or reloading the page. When user reloads the page, the "Online users" button is display for a second, but then it's hidden.

The "Online users" dialog works well now: It includes only user and user2, with no duplicated names.

PR #1231, opened 2026-02-06
Fix collaborative editing: disable delete/move on locked iDevices (#1165)

### LINKED ISSUE #1165: Collaborative editing. Do not delete or move iDevice in edition mode.
An iDevice that someone is editing cannot be deleted or moved.

erseco:
When a user is editing an iDevice during collaborative editing, other users could still delete, move, or drag that iDevice, which could cause data loss or conflicts.

The lock detection infrastructure already existed (awareness state, `lockedByRemote`, `updateLockIndicator()`), but only the Edit button was being disabled. Delete, Move Up, Move Down, the Actions dropdown, and drag-and-drop remained active.

**Fix:**
- Disable Delete, Move Up, Move Down, and Actions dropdown buttons when an iDevice is locked by another user
- Disable drag-and-drop on locked iDevices (`draggable=false`)
- Add runtime guards in click handlers for delete, move up, and move down to prevent action even if the button state is stale between renders
- Fix `toogleIdeviceButtonsState()` selector (was using `.button-action-idevice` which didn't match any element, changed to `.btn-action-menu`)
- Added unit tests covering disabled button rendering, handler guards, and lock state transitions

Closes #1165

### DIFF
--- a/public/app/common/common.js
+++ b/public/app/common/common.js
@@ -59,9 +59,26 @@ window.MathJax = window.MathJax || (function() {
             }
         }
     }
+    // Generic logic to detect if we are in the index page (root) or a subpage
+    // We check the src of the common.js script itself.
+    var scriptPath = '';
+    var scripts = document.getElementsByTagName('script');
+    for (var i = 0; i < scripts.length; i++) {
+        var src = scripts[i].getAttribute('src');
+        if (src && src.indexOf("common.js") !== -1 && src.indexOf("common_i18n") === -1) {
+            scriptPath = src;
+            break;
+        }
+    }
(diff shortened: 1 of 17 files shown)


ignaciogros (changes requested):
Thank you, @erseco.

Some issues have been fixed, but there are still some problems.

- user creates a content and shares it with user2.
- user2 uses the link provided by user to see/edit the content.
- user edits iDevice X.
- user2 can't edit or delete it. That's correct.
- But user2 can delete the box in which iDevice X is, without even confirmation.

ignaciogros (approved):

PR #1234, opened 2026-02-06
 Speed up ELPX download generation (#1065) + fix missing manifest in ELPX export

### LINKED ISSUE #1065: Speed up package generation during download
elpx generation process during download can be optimized to run faster and more efficiently, helping to reduce user waiting time.

erseco:
The client-side ELPX generation (`exe_elpx_download.js`) was slow with large projects due to three bottlenecks. This PR addresses all three and also fixes a bug where the ELPX manifest was missing from ELPX exports.

**Performance optimizations in `exe_elpx_download.js`:**

1. **Sliding-window fetch pool (6 → 10 concurrent):** Replaced the batch-of-6 pattern (which waited for the slowest request in each batch before starting the next) with a worker-pool approach where 10 workers consume from a shared index. New fetches start immediately as previous ones complete, eliminating idle time between batches.

2. **Per-file compression levels:** Added a `STORE_EXTENSIONS` regex for already-compressed formats (jpg, png, mp4, woff2, pdf, etc.) and set `level: 0` (STORE) for those files. Text files still get `level: 6`. This avoids wasting CPU on files that won't get any smaller.

3. **Parallel folder picker file reads:** Replaced the sequential `for` loop with `filter().map()` + `Promise.all()` to read all folder files concurrently instead of one at a time.

**Bug fix in `ElpxExporter.ts`:**

The `ElpxExporter` overrides `Html5Exporter.export()` with its own implementation that never called `needsElpxDownloadSupport()` and never generated the ELPX manifest. The web export worked correctly, but when the user downloaded an `.elpx` file and opened the HTML inside it, `libs/elpx-manifest.js` was missing. Fixed by adding manifest detection, file tracking, and script tag injection to the ELPX export path.

Closes #1065

### DIFF
--- a/public/files/perm/idevices/base/interactive-video/edition/editor/index.html
+++ b/public/files/perm/idevices/base/interactive-video/edition/editor/index.html
@@ -83,12 +83,12 @@
             adminStyle.rel = 'stylesheet';
             adminStyle.href = window.__LOCAL_RESOURCE_PREFIX__ + 'css/admin.css';
             document.head.appendChild(adminStyle);
-
-            // Load langs script (local resource)
-            var langsScript = document.createElement('script');
-            langsScript.src = window.__LOCAL_RESOURCE_PREFIX__ + 'langs/all.js';
-            document.head.appendChild(langsScript);
         })();
+    </script>
+
+    <!-- Load langs script synchronously (required before admin.js) -->
+    <script>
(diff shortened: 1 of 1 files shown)


ignaciogros (approved):
It works well in all cases now.

Thank you, @erseco.

I suggest keeping this option for the moment and see what users say. If they see it as a problem/bug, be can disable it.

PR #1236, opened 2026-02-06
 Fix collaborative editing: user-uploaded styles show wrong title and can't be selected (#1167)

### LINKED ISSUE #1167: Collaborative editing. Styles while editing collaboratively.
A user uploads a style while editing collaboratively. Other users can see the style, but cannot select it. The first user sees the style’s `title` in their interface, while the other users see `name`. That might be the problem.

erseco:
When User A uploads a custom style during collaborative editing, other users could see the style in the list but could not select it. User A saw the correct `<title>` from `config.xml`, while other users saw the raw `<name>` tag instead.

**Root cause:** `_parseThemeConfigFromFiles()` in `YjsProjectBridge.js` had two bugs in how it builds the theme config from `config.xml`:

1. `config.name` was set to `getValue('name')` (the raw `<name>` tag from `config.xml`, e.g. "Collab Test Theme"), but it should be set to `themeName` (the Yjs map key, which is the sanitized `dirName`, e.g. "collab_test_theme"). This caused the theme to be registered under `installed["Collab Test Theme"]` but looked up by `getThemeInstalled("collab_test_theme")`, so the theme was never found by collaborators.
2. `config.title` was set to `getValue('name')` instead of `getValue('title')`. For a theme with `<name>base</name>` and `<title>Default</title>`, collaborators saw "base" instead of "Default".

The upload path in `navbarStyles.js` already did this correctly (`name: dirName`, `title: getValue('title') || themeName`), so User A always worked fine. Only the Yjs sync path (`_parseThemeConfigFromFiles`) was broken.

**Fix in `YjsProjectBridge.js`:**

```
// Before (broken)
config.name = getValue('name') || themeName;
config.title = getValue('name') || themeName;

// After (fixed)
config.name = themeName;
config.title = getValue('title') || getValue('name') || themeName;
```

**Tests added:**
- 3 new unit tests in `YjsProjectBridge.test.js` covering `<title>` tag parsing, `<name>` fallback, and `themeName` vs raw `<name>` tag distinction
- 1 new E2E test in `theme-import-collaborative.spec.ts`: Client A uploads a theme with distinct `<name>` and `<title>`, selects it, verifies Client B receives the theme with the correct title and sanitized name
- New test fixture `test-theme-collab.zip` with distinct `<name>` and `<title>` tags

Closes #1167

### DIFF
--- a/public/files/perm/idevices/base/interactive-video/edition/editor/index.html
+++ b/public/files/perm/idevices/base/interactive-video/edition/editor/index.html
@@ -83,12 +83,12 @@
             adminStyle.rel = 'stylesheet';
             adminStyle.href = window.__LOCAL_RESOURCE_PREFIX__ + 'css/admin.css';
             document.head.appendChild(adminStyle);
-
-            // Load langs script (local resource)
-            var langsScript = document.createElement('script');
-            langsScript.src = window.__LOCAL_RESOURCE_PREFIX__ + 'langs/all.js';
-            document.head.appendChild(langsScript);
         })();
+    </script>
+
+    <!-- Load langs script synchronously (required before admin.js) -->
+    <script>
+        document.write('<script src="' + window.__LOCAL_RESOURCE_PREFIX__ + 'langs/all.js"><\/script>');
     </script>
 
     <!-- Load external scripts after path computation -->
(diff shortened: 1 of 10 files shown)


ignaciogros (dismissed):
- user shares a content and user2 enters the URL.
- user installs a new style.
- user2 cannot see it (there is nothing in the "Imported" section of the Styles panel).
- user selects the new installed style they.
- Then, user2 can see it and select it without problems.

I think the only thing left to fix is the first part. 

Thanks, @erseco.

ignaciogros (approved):

PR #1258, opened 2026-02-09
Fix iDevice SVG icon resolution in all export formats

### LINKED ISSUE #1248: iDevice icons are not displayed (SVG)
When styles with SVG icons are used, these are not displayed because the path has the .png extension. Example: `` instead or ``.

erseco:
Fixes #1248

This PR ensures that `IdeviceRenderer.setThemeIconFiles()` is called across all export formats, preventing the renderer from falling back to the default `.png` extension when a theme uses SVG icons.

**Changes:**
- **Html5Exporter:** Moved `setThemeIconFiles()` into `prepareThemeData()`. This automatically fixes **SCORM 1.2** and **SCORM 2004** exports, as they inherit this method.
- **Epub3Exporter:** Added the `setThemeIconFiles()` call to its `prepareThemeData()` implementation.
- **PrintPreviewExporter:** Updated to fetch the theme files and configure the renderer before generating the preview.
- **Tests:** Added unit tests for every exporter (Epub3, PrintPreview, SCORM 1.2, SCORM 2004) to verify that SVG icons are correctly resolved in the output.

### DIFF
--- a/public/files/perm/idevices/base/interactive-video/edition/editor/js/admin.js
+++ b/public/files/perm/idevices/base/interactive-video/edition/editor/js/admin.js
@@ -20,7 +20,7 @@ var iAdmin = {
         currentSlide: '', // The order of the slide that's being edited (a number): InteractiveVideo.slides[X],
         warnWhenSave: true, // Tell the user to click on Save before quitting
     },
-    changeMode: function (mode) {
+    changeMode: mode => {
         top.interactiveVideoEditor.hasChanged = true;
         if (mode == 'add') {
             $('BODY').removeClass('edition-mode');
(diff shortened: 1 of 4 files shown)


ignaciogros (dismissed):
With the base style, I’m still seeing the old green icons in preview and export, but that style has orange icons now. I think that's a different problem, but I mention it just in case it can be a cache issue, because most styles used to have .png icons.

When I switch to another one using `.svg`, I can see the icons while editing, but not in preview or when exporting, because the path still points to the `.png` files instead of the `.svg`.

ignaciogros (approved):
That was it... I should've seen it.

Thank you, @erseco.

PR #1260, opened 2026-02-10
1241 Race conditions in the initialization of multiple iDevices

mnarvaezm:
This PR fixes **race conditions** when loading pages with **multiple iDevices of the same type** in the workarea preview.

The root cause was that export scripts with auto-initialization (`$(function(){ init() })`) could run **before all iDevice HTML was present in the DOM**, causing the last iDevices on the page to not activate.

**Main fix: Phased loading architecture**

A **4-phase** loading architecture was implemented in `loadComponentsPage()`:

| Phase | Action                                                             |
| ----- | ------------------------------------------------------------------ |
| 1–2   | Insert **ALL** iDevice HTML into the DOM (without loading scripts) |
| 3     | Load styles and scripts (auto-exec finds all HTML already present) |
| 4     | Initialize JSON-type iDevices (`ideviceInitExport()`)              |

✅ With this approach, when `init()` runs, **all instances** of that iDevice type are already in the DOM, so a **single** `init()` call initializes them all correctly.

---

### Bug #1: `forEach` with `async/await` doesn’t wait for promises

**Location:** `resetCurrentIdevicesExportView()`
**Problem:** `forEach(async () => ...)` doesn’t await each iteration.
**Fix:** replaced with sequential `for...of` + `await`.

### Bug #2: Legacy init functions executed before HTML was ready

**Location:** `idevicesEngine.js` → `loadIdevicesExportScripts()`
**Problem:** legacy initializers (`$exeAparic.init()`, `$exeMemorypair.init()`, etc.) ran immediately after script load, before the DOM was complete.
**Fix:** integrated legacy initialization into the phased loading flow, ensuring HTML is present before calling `init`.

### Bug #3: “Cancel edit” didn’t await content regeneration

**Location:** `idevicesEngine.js` → `cancelEdition()`
**Problem:** `generateContentExportView()` was called without `await`, causing race conditions.
**Fix:** added the missing `await`.

### Bug #4: Arbitrary `setTimeout` used to “fix” timing issues

**Location:** various initialization paths
**Problem:** `setTimeout(fn, 100)` is unreliable and hides the real issue.
**Fix:** removed arbitrary timeouts and relied on the phased loading architecture.

### Bug #5: ABC Music accumulated duplicate click handlers

**Location:** `exe_abc_music.js` → `setButtonsBehaviour()`
**Problem:** each re-init added new handlers without removing previous ones.
**Fix:** remove previous handlers before registering new ones.

### Added a guard to **skip activities already processed**

---

### DIFF
--- a/src/shared/export/exporters/Epub3Exporter.spec.ts
+++ b/src/shared/export/exporters/Epub3Exporter.spec.ts
@@ -903,3 +903,79 @@ $("." + k + "-accordion").each(function (i) {
         expect(contentStr).toContain('href="javascript:void(0)"');
     });
 });
+
+describe('Icon Resolution via setThemeIconFiles', () => {
+    let document: MockDocument;
+    let resources: MockResourceProvider;
+    let assets: MockAssetProvider;
+    let zip: MockZipProvider;
+    let exporter: Epub3Exporter;
+
+    beforeEach(() => {
+        resources = new MockResourceProvider();
+        assets = new MockAssetProvider();
+        zip = new MockZipProvider();
+    });
+
+    it('should resolve SVG icons when theme has SVG icon files', async () => {
+        const pagesWithIcon: ExportPage[] = [
+            {
+                id: 'page-1',
+                title: 'Test Page',
+                parentId: null,
+                order: 0,
+                blocks: [
+                    {
+                        id: 'block-1',
+                        name: 'Block with Icon',
+                        order: 0,
+                        components: [],
+                        iconName: 'activity',
+                    },
+                ],
+            },
+        ];
+
+        document = new MockDocument({}, pagesWithIcon);
+
+        resources.fetchTheme = async (_name: string) => {
+            const files = new Map<string, Buffer>();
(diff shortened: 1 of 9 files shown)


test/e2e/playwright/specs/component-export-import.spec.ts
```diff
@@ -981,20 +988,30 @@ test.describe('Block Icon Preservation during Export/Import', () => {
...
 
         // Verify icon src is displayed (may have different blob URL but should be present)
-        const importedIconSrc = await getBlockIconSrc(page, 0);
+        const importedIconSrc = await getBlockIconSrc(page, importedBlockIndex);
         console.log('Imported icon src:', importedIconSrc?.substring(0, 80));
         expect(importedIconSrc).toBeTruthy();
 
+        // Verify the icon src contains 'activity' or is a theme icon blob URL
+        // The src may be a direct URL or a blob URL, but should exist
+        expect(importedIconSrc).toBeTruthy();
```
erseco, line 1013: This assert is checked twice, is not important, but we can remove one of them

PR #1272, opened 2026-02-11
fix(padlock): fix invisible texts in Lock iDevice edition

### LINKED ISSUE #1269: Lock idevice
It seems there are some texts missing in the Gerenal options (in edition):

Image

In eXe 2.9 it shows like this:

Image

mnarvaezm:
Fixes an issue where the “Instructions” and “Feedback” switch texts were not visible in the Lock iDevice editor. Fix #1269


cristinavaldera (approved):
Fixed!

PR #1281, opened 2026-02-12
Fix opening files in app

erseco:
This pull request improves the reliability and user experience when opening `.elpx` files via file association in the Electron app. The changes ensure that files are only sent to the renderer process after it is fully ready to handle them, preventing race conditions and lost events. Additionally, the application now queues files if the UI is not yet ready, and processes them as soon as possible.

**Electron main process improvements:**

* Introduced a `pendingOpenFileByWebContentsId` map to track which file should be sent to each renderer, and only send the file path after the renderer signals readiness via the new `app:renderer-ready-for-open-file` IPC event. 
* Updated the project window creation logic to hide the window until it is ready to show, improving perceived performance and preventing visual glitches.

**Renderer process and preload improvements:**

* Added `notifyRendererReadyForOpenFile` to the preload and mock Electron APIs, allowing the renderer to notify the main process when it is ready to receive file open events. 
* Modified the renderer app (`App` class) to register the file-open handler as early as possible, track pending files if the UI is not ready, and flush the queue once initialization is complete.

### DIFF
--- a/public/app/workarea/interface/elements/previewPanel.js
+++ b/public/app/workarea/interface/elements/previewPanel.js
@@ -47,9 +47,23 @@ export default class PreviewPanelManager {
     init() {
         this.bindEvents();
         this.subscribeToChanges();
-        this.restorePinnedState();
+        this.resetToDefaultState();
         this._setupVisibilityHandler();
         Logger.log('[PreviewPanel] Initialized');
+    }
+
+    /**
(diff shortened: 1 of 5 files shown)


cristinavaldera:
In Mac, It works

eXeLearningProject (approved):
tested in mac, windows and linux

PR #1309, opened 2026-02-17
Fix: Remove custom/.gitkeep from ELPX exports and filter placeholder files on import

### LINKED ISSUE #1307: Unexpected “custom” folder with “.gitkeep” file appears in File Manager after exporting page as ELPX in eXeLearning
When a resource is open in eXeLearning and a single page is exported as an ELPX resource, opening the exported ELPX file results in an unexpected folder appearing in the File Manager.

Specifically, a folder named “custom” is displayed, containing a file called “.gitkeep”. This folder and file do not appear to be relevant to the exported page and should likely not be visible in the File Manager interface.

This behavior may indicate that internal repository or packaging files are being included in the export process and exposed to end users, which could cause confusion and should be reviewed.

I'm including a link that discusses these .gitkeep files: https://www.campusmvp.es/recursos/post/que-son-los-archivos-gitkeep-en-git.aspx?srsltid=AfmBOool6jDUgq_w1PtD1-IsLUjbp-FSH0-LE2pL8sLlDv765xsfAUFj

erseco:
The ELPX exporter was adding an empty `custom/.gitkeep` marker file to force the `custom/` directory into the ZIP. This is a leftover from v3.x, where the `custom/` folder stored user assets uploaded through the File Manager. The directory is no longer needed as an empty placeholder, and when users reopened the exported ELPX, the `.gitkeep` file appeared as a visible entry in the File Manager, causing confusion.

**Changes:**

- Removed the `custom/.gitkeep` creation line from `ElpxExporter.ts`. Real `custom/` assets (if any) are still included normally.
- Added a filter in `AssetManager.js` to skip hidden placeholder files (names starting with `.`) inside `custom/` during import, as a safety net for older exports.
- Fixed a regex in `convertContextPathToAssetRefs` that was breaking resolution of `custom/` asset paths containing multiple spaces uploaded via legacy File Manager.

Fixes #1307

### DIFF
--- a/public/app/workarea/menus/menuEngine.js
+++ b/public/app/workarea/menus/menuEngine.js
@@ -150,13 +150,13 @@ export default class MenuEngine {
                     const menus = document.querySelectorAll(
                         '[data-bs-toggle="dropdown"].show'
                     );
+                    const clickedDropdownToggle =
+                        event.target?.closest?.('.dropdown-toggle');
                     if (menus.length > 0) {
                         menus.forEach((menu) => {
                             if (
                                 !menu.contains(event.target) &&
-                                !event.target.classList.contains(
-                                    'dropdown-toggle'
-                                )
+                                !clickedDropdownToggle
                             ) {
                                 menu.click();
                             }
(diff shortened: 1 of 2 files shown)


mnunezcedec (approved):
Tested online and on desktop. Everything works perfectly. Thank you @erseco

PR #1313, opened 2026-02-18
Fix: render LaTeX inside `<span>` elements and clarify MathJax checkbox

### LINKED ISSUE #1292: MathJax option not preserved when opening eXe 2.9 project in eXe 4
When opening an .elp file created with eXe 2.9 that has the “MathJax” option enabled, the MathJax option appears disabled in eXe 4 project properties. Although the formulas are displayed correctly during editing, they are not rendered in Preview mode or in exported formats.

Everything works fine if you create a elpx from scratch.

For tests: [tipografia (1).elpx.zip](https://github.com/user-attachments/files/25290309/tipografia.1.elpx.zip)

Image

erseco:
Closes #1292

---

### Root causes

After reviewing the test file from the report, two separate root causes were found.

**1. LaTeX inside `<span>` elements was silently skipped**

The pre-renderer was skipping any `<span>` with a `style` attribute containing `color:`, under the assumption that colored spans always contained example code rather than real formulas. This was too broad. In the reported file, the actual formulas in the second table column were wrapped in `<span style="color: #000000;">` and were silently dropped by both `ServerLatexPreRenderer.ts` and `LatexPreRenderer.js`.

**2. Mixed content (pre-rendered + pending raw LaTeX) was also skipped**

Both pre-renderers had an early-exit guard that skipped an entire content block if it already contained any `exe-math-rendered` span. If a page had even one already-rendered formula, all remaining raw LaTeX in the same block was ignored.

**3. The MathJax checkbox label was misleading**

The label "MathJax (formulas)" suggested that formulas would not display unless the option was enabled. That is not accurate: formulas are always pre-rendered to SVG/MathML. The MathJax library is only needed for optional advanced features (accessibility toolbar, context menus), at the cost of ~8 MB added to the export. This confusion made users believe rendering was broken when the setting was not preserved on import.

### DIFF
--- a/src/routes/assets.spec.ts
+++ b/src/routes/assets.spec.ts
@@ -399,12 +399,56 @@ describe('Assets Routes', () => {
             expect(res.status).toBe(404);
         });
 
-        it('should return 400 for invalid asset ID', async () => {
+        it('should return 404 for non-existent project UUID', async () => {
+            const res = await app.handle(new Request(`http://localhost/api/projects/non-existent-uuid/assets/1`));
+            expect(res.status).toBe(404);
+            const body = await res.json();
+            expect(body.error).toContain('Project not found');
+        });
+
+        it('should return 404 when numeric asset belongs to another project', async () => {
+            const filePath = path.join(testDir, 'other-project-asset.txt');
+            await fs.writeFile(filePath, 'Other project content');
+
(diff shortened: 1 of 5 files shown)


cristinavaldera (commented):
It works!

PR #1315, opened 2026-02-18
Improve image optimizer avoiding optimize already optimized images

erseco:
This pull request enhances the image optimizer modal by improving how images that are already optimized are handled. It introduces logic to automatically detect images that cannot be further optimized, updates their status display, and prevents users from selecting them for optimization. The changes also include comprehensive tests to ensure this new behavior works as expected.

**Image Optimization Logic Improvements:**

* Added logic to detect when an image is already optimized (savings below 1% or no reduction possible), and introduced a new status (`already_optimized`) for such cases. These images now display an "Already optimized" badge and are automatically made unselectable in the UI. 
* Updated selection logic throughout the modal to prevent already optimized assets from being (de)selected, including in bulk actions like "Select All", and ensured the UI reflects their unselectable state. 

**Testing Enhancements:**

* Added and updated unit tests to verify detection of already optimized images, correct badge rendering, and that non-selectable assets are properly excluded from selection actions and UI state.

### DIFF
--- a/public/app/workarea/modals/modals/pages/modalImageOptimizer.js
+++ b/public/app/workarea/modals/modals/pages/modalImageOptimizer.js
@@ -50,6 +50,9 @@ const STATUS_BADGE_CONFIG = {
     [STATUS.FAILED]: (error) => `<span class="badge bg-danger" title="${error || _('Error')}">${_('Failed')}</span>`,
 };
 
+const ALREADY_OPTIMIZED_MIN_SAVINGS_PERCENT = 1;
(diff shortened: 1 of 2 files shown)


cristinavaldera (commented):
It works. Just one concern, maybe for improving it later on. Does it have any sense to say how much the image size would grow after optimization, when it is already optimized? Maybe is would be better to show a "-" or "N/A" in "Stimated" and "Savings" columns when it is already optimized.

cristinavaldera (approved):

PR #1352, opened 2026-02-22
refactor(flux): submenu collapse, sidebar state, CSS cleanup

### LINKED ISSUE #1717: Flux theme: collapse arrow on expanded tree does not collapse it
Reported by @greynoa.

## Summary

With the **Flux** theme, when the navigation tree is expanded in the viewer, clicking the collapse button (up arrow) does not collapse the tree. The arrow is visible but has no effect.

## Steps to reproduce

1. Open a project with multiple pages/sections in eXeLearning.
2. Apply the **Flux** theme.
3. Preview / view the exported site.
4. Expand the navigation tree.
5. Click the up arrow to collapse it.

## Expected behavior

Either:
- The collapse button actually collapses the tree, **or**
- The up arrow should not be shown at all (other themes do not display it), since its presence without function is confusing to users.

## Actual behavior

The up arrow is displayed but clicking it does nothing — the tree stays expanded.

## Environment

- Theme: Flux
- Affected area: viewer / exported site navigation

## Video

A screen recording reproducing the issue will be attached as a comment below.

https://github.com/user-attachments/assets/cd210e79-e116-44da-b103-778a6268ad9b

erseco:
## Summary

Originally aimed at reducing JS in the flux theme, the work expanded to fix several related issues uncovered along the way.

## Submenu collapse / expand

- Ports the universal `open-ul`/`closed-ul` button pattern. Each `.daddy` link now gets a real toggle button with rotating chevron; the submenu animates with `slideUp`/`slideDown`.
- The chevron uses `background-color` + CSS `mask`, so a single SVG drives both the white-on-dark default and the dark-on-light highlighted / hover / focus states without extra assets.

## Page title layout

- Reintroduces a minimal `movePageTitle()` that moves `.page-title` into `.page-content` on init.
- Long titles push the content card down naturally; empty pages no longer leave the title orphan over the dark backdrop.

## Sidebar state persistence (no flicker)

- Synchronous IIFE at the top of `flux/style.js` (loaded in `<head>`) reads `sessionStorage` / `?nav=false` **before paint** and toggles `.siteNav-off` on `<html>`.
- CSS selectors migrated from `body.siteNav-off` to `.siteNav-off` so the pre-paint guard takes effect immediately, eliminating the flicker on navigation.
- `sessionStorage` is canonical; `?nav=false` is honored only on the first visit.
- New `setNavOff()` helper keeps `<html>`, `<body>` and `sessionStorage` in sync atomically.

## Search UX

- Clicking the search button no longer collapses the sidebar (at any viewport width).

## CSS simplifications

- New CSS custom properties `--flux-header-h` and `--flux-title-gap` replace magic numbers (100px / 20px) used across header, sidebar and content.
- Four duplicated sr-only patterns (`clip: rect(1px ...)`) consolidated into a single utility block matching the canonical `.sr-av` definition.
- Removed redundant `text-shadow` rule on `.page-title` (already in base).
- Consolidated `.page-counter` (had duplicate `margin`, `text-align` and `color` declarations).
- Removed dead/empty rules (`.exe-export main {}`, `.exe-web-site main {}`).
- Removed `display: inline` on a flex item.
- Fixed `#packageLicense.propietary` sr-only (was leaving a 1px visible dot).
- Removed the per-state chevron backgrounds on `.daddy` (3 image variants); only `background-color` remains — chevron now lives on the toggle button.

## Test plan

- [x] `make fix` clean
- [x] PageRenderer unit tests pass (131/131)
- [ ] Visual QA in preview:
  - [ ] Collapse / expand submenus on desktop and mobile widths
  - [ ] Long page title pushes the content card down
  - [ ] Empty page does not leave the title orphan
  - [ ] Sidebar state survives Anterior / Siguiente without flicker
  - [ ] Search button preserves sidebar state on any width


ignaciogros (dismissed):
Hi @erseco,

Could you please review this details?

**background-image instead of mask:**

Please replace the current CSS `mask`/`-webkit-mask` usage with `background-image` to avoid local `file://` CORS issues when opening the HTML without a web server.

Please ensure the icon remains visually correct across different background colors, since replacing `mask` may affect color inheritance and transparency behavior.

**ePub, SCORM and IMS:**

Current header presentation:

Image

Expected:

Image

**Single page:**

Header contrast problems and missing node titles:

Image

Expected:

Image

**Variables:**

The style uses CSS variables, which are not supported by older browsers such as IE11. In this case, replacing the variables with their actual values would be fast.

Could you please do that? The other styles do not use variables in order to preserve compatibility with those browsers.

Thank you!

erseco:
Hi @ignaciogros, thanks for the review. All four points addressed in f0cc8ddd:

- **`mask` → `background-image`**: The submenu chevron now uses `background-image` so it works under `file://`. Added a white companion SVG (`img/arrow_down_white.svg`) alongside the existing dark `arrow_down.svg`; the default/highlighted states swap images instead of recoloring a mask, which keeps the icon visually correct on both the dark sidebar and the highlighted (yellow) states.
- **ePub / SCORM / IMS header**: `.main-header` now uses `flex-direction: column` for these exports and `.package-title` is no longer absolutely positioned. Result matches the expected mock — small grey project title on top, purple page title below.
- **Single-page**: only `.package-header` is fixed at the top; each section's `.main-header` is static and transparent, so the white `.page-title` (e.g. "Inicio", "1. El contrato de trabajo") renders on the dark backdrop between content boxes instead of being hidden by stacked fixed headers.
- **CSS variables**: the two custom properties (`--flux-header-h`, `--flux-title-gap`) are gone — values are inlined throughout, matching the IE11-friendly approach of the other themes.

Tested under `file://` and via local web server with the website / SCORM / IMS / ePub / single-page layouts at desktop and mobile widths. `make fix` clean, PageRenderer specs green.

ignaciogros (approved):

Issue #1356, opened 2026-02-23
Link to anchor in other page


ignaciogros:
You can add a link to an anchor in the same page, but not in a different one.

PR #1366, opened 2026-02-23
Fix cross-page and same-page anchor links (#1356)

erseco:
The link dialog only showed anchors from the active TinyMCE instance, so anchors in other components on the same page were invisible. Cross-page anchor links (`exe-node:pageId#anchor`) were also impossible to create from the UI, and even if inserted manually, the exporter would strip the fragment when resolving the final URL.

**What changed:**

- `BaseExporter.ts`: `replaceInternalLinks()` now preserves the `#anchor` fragment when resolving `exe-node:pageId#anchor` links to real URLs.
- `YjsProjectBridge.js`: Added `_extractAnchorsFromPageMap()`, `getPageAnchors(pageId)`, and `getAllPageAnchors(excludePageId)` to scan anchor bookmarks (`<a id>` / `<a name>`) across all Yjs page content.
- `exelink/plugin.min.js`: The anchor dropdown now includes anchors from other components on the same page (deduplicated against the active editor scan), plus anchors on other pages as `exe-node:pageId#anchor` entries.
- `BaseExporter.spec.ts`: 4 new tests covering anchor fragment preservation from index page, subpage, and unknown page fallback.


ignaciogros (changes requested):
The anchor links work in preview and export, and are displayed in TinyMCE, whether they are internal links (to the same page) or external links (to another page). But:

- Links to other nodes (with or without anchors) don't work within eXe.
- Links don't work in the Single page export. This could be quite tricky, because if you use the same ID for anchors on different pages (which is common), that ID would need to be replaced. An easier solution would be to remove anchor links in this export.

Issue #1298, opened 2026-02-13
Mixed languages on first launch


mnarvaezm:
When opening the application for the first time, some texts appear in Spanish while others appear in English.

Cause: The server and the client determine the language independently. On the first access, there is no saved  user preference, and the different fallbacks do not match — some use 'en' while others detect the browser language. Additionally, the detected language is not automatically persisted, so it may vary between requests.

Expected Solution: The app should start in a single, consistent language (either the one configured in APP_LOCALE or the browser’s language) and automatically persist that selection on the first access.

PR #1382, opened 2026-02-25
Fix mixed languages on first launch (#1298)

mnarvaezm:
## Problem

On first launch, the application showed a mix of languages: the interface appeared in the user's browser language, but the new project title and page names were created in English, causing an inconsistent experience.

The root causes were:

- The default locale was hardcoded to `'en'` across the backend (`DEFAULT_LOCALE`, `DEFAULT_PREFERENCES`), ignoring `APP_LOCALE`.
- The locale resolution order in the pages route ignored `APP_LOCALE`, jumping directly from the user preference to the browser's `Accept-Language`.
- Static mode did not persist the detected locale on first launch, so preferences could silently change if the browser language changed.
- When `_ensureNewProjectLanguage` fixed the project language in a new document, it updated the `language` field but left the default titles (`Untitled document`, `New page`) in English.

### Backend

- `src/services/translation.ts` — `DEFAULT_LOCALE` now reads the `APP_LOCALE` environment variable, with fallback to `'es'`.
- `src/routes/user.ts` — `DEFAULT_PREFERENCES.locale` now follows the same logic: `APP_LOCALE || 'en'`.
- `src/routes/pages.ts` — locale resolution order updated to: user preference → `APP_LOCALE` → `Accept-Language` → default; on the first visit, the resolved locale is automatically saved as the user preference to ensure consistency in later requests.

### Frontend — static mode

- `userPreferences.js` — the fallback locale in static mode now reads `window.eXeLearning.config.locale` (set by `initStaticMode`) instead of hardcoding `'en'`.
- `userPreferences.js` — on first launch (with nothing in `localStorage`), all detected preferences are immediately persisted under `exe_user_preferences` so they remain fixed in future sessions.

### Frontend — new project language

- `YjsProjectBridge.js` — `_ensureNewProjectLanguage` now re-translates the document title (`Untitled document`) and the root page `title`/`pageName` (`New page`) after fixing the project language, ensuring all default content is consistent with the user's locale from the start.

## Tests

All changes are covered by new or updated tests:

- `userPreferences.test.js` — 3 new tests for first-launch auto-save behavior in static mode.
- `YjsProjectBridge.test.js` — existing test expanded to verify re-translation of `title` and `pageName`.
- `pages.spec.ts` / `user.spec.ts` / `config.spec.ts` — assertions updated to reflect the new default value `'en'`.

### DIFF
--- a/assets/styles/layout/_idevices.scss
+++ b/assets/styles/layout/_idevices.scss
@@ -914,4 +914,14 @@ article.box {
     .idevice-content-block.out {
         opacity: 0.3;
     }
+
+    /* Teacher mode visual indicator in workarea */
+    article.box.exe-teacher-highlight {
+        border-color: var(--brand-primary-500) !important;
+        border-style: dashed !important;
+    }
+
+    .idevice_node.exe-teacher-highlight {
+        border: 1px dashed var(--brand-primary-500);
+    }
 }
--- a/public/app/workarea/project/idevices/content/blockNode.js
+++ b/public/app/workarea/project/idevices/content/blockNode.js
@@ -268,6 +268,10 @@ export default class IdeviceBlockNode {
                 this.blockContent.classList.add(cls);
             });
         }
+        // teacher only - workarea visual indicator (separate class to avoid export hide rule)
+        if (this.properties.teacherOnly?.value == 'true') {
+            this.blockContent.classList.add('exe-teacher-highlight');
+        }
         // allow toggle
         if (this.properties.allowToggle.value != 'true') {
             // This should always be available while editing:
--- a/public/app/workarea/project/idevices/content/blockNode.test.js
+++ b/public/app/workarea/project/idevices/content/blockNode.test.js
@@ -432,6 +432,20 @@ describe('IdeviceBlockNode', () => {
             expect(block.blockContent.classList.contains('class3')).toBe(true);
         });
 
+        it('adds exe-teacher-highlight class when teacherOnly is true', () => {
+            block.properties.teacherOnly = { value: 'true' };
+            block.setPropertiesClassesToElement();
+
+            expect(block.blockContent.classList.contains('exe-teacher-highlight')).toBe(true);
+        });
+
+        it('does not add exe-teacher-highlight class when teacherOnly is false', () => {
+            block.properties.teacherOnly = { value: 'false' };
+            block.setPropertiesClassesToElement();
+
+            expect(block.blockContent.classList.contains('exe-teacher-highlight')).toBe(false);
+        });
+
         it('calls toggleOff when minimized is true', () => {
             const spy = vi.spyOn(block, 'toggleOff');
             block.properties.minimized.value = 'true';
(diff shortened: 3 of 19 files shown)


public/app/workarea/user/preferences/userPreferences.js
```diff
@@ -34,9 +34,11 @@ export default class UserPreferences {
 
         // Final fallback to minimal defaults
         // Note: advancedMode defaults to 'true' in static mode so all features are visible
+        // locale defaults to the app's configured locale (from server config)
         if (!preferencesConfig) {
+            const appLocale = app.eXeLearning?.config?.locale || 'es';
```
erseco: Maybe the fallback locale should be "en", no?

mnarvaezm: No, I believe the default language should be Spanish.

One thing is that, for technical reasons or by tradition, code comments, issues, and PRs are in English, and quite another is the default language in which the application is presented to the user.

If 99.9% of the real audience for our application has Spanish as one of their official languages, I don’t think it makes much sense for the application’s default language to be English.

cristinavaldera: Ideally, the online version should take the default language from the browser, and the desktop version from the operating system.

If that’s not possible, I’d suggest setting English as the default language.

Even though most users are Spanish speakers, in my opinion English fits better with the international nature of the project. That’s how it worked in eXe 2.9, and it felt more consistent with the project’s global approach.

mnarvaezm: Hi,

Right now it already works like this: it takes the language from the browser and, if it can’t be applied because we don’t have that translation, it falls back to English.

I reverted the default back to English in my previous commit, even though it goes against my own criteria and what I’d prefer, because the E2E tests are full of assertions against English strings (aria-labels, modal text, button titles, etc.), and when the app started in Spanish they were failing all over the place.
@cristinavaldera @erseco @ignaciogros

public/app/workarea/user/preferences/userPreferences.test.js
```diff
@@ -104,7 +104,7 @@ describe('UserPreferences', () => {
       await userPreferences.load();
 
       expect(userPreferences.preferences).toBeDefined();
-      expect(userPreferences.preferences.locale).toEqual({ title: 'Language', value: 'en', type: 'select' });
+      expect(userPreferences.preferences.locale).toEqual({ title: 'Language', value: 'es', type: 'select' });
```
erseco: This is still setting "es" as default

public/app/workarea/user/preferences/userPreferences.js
erseco: The code is still in "es"

src/services/translation.ts
```diff
@@ -111,7 +114,7 @@ interface XlfParseResult {
 const catalogues: Map<string, Map<string, string>> = new Map();
 
 /**
- * Current locale for the session (default: from env or 'en')
+ * Current locale for the session (default: from APP_LOCALE env or 'es')
```
erseco: The comment still reflect "es"

PR #1412, opened 2026-02-28
SandboxJS Integration in eXeLearning

### LINKED ISSUE #1297: Evaluate SandboxJS implementation to safely support JavaScript inside .elpx projects
Study the possible implementation of SandboxJS ([https://github.com/nyariv/SandboxJS](https://github.com/nyariv/SandboxJS?utm_source=chatgpt.com) in order to enable the use of JavaScript inside .elpx projects with reduced security risks.

erseco:
## 1. The Security Problem (Why this change is being made)
Until now, eXeLearning allowed authors to freely input HTML and JavaScript code into iDevices, headers (`extraHeadContent`), and footers (`userFooterContent`). 

This opened the door to severe vulnerabilities known as **XSS (Cross-Site Scripting)**. If a teacher downloads a eXe package (scorm, web...), file from an untrusted source, or if an attacker manages to inject code into a collaborative project, that JavaScript code would execute with **full privileges** in the end-user's browser (students, or the teachers themselves when previewing/exporting).

**The potential risks that existed were:**
1. **Session Hijacking:** The script could read `document.cookie` or `localStorage` and send the user's access tokens (e.g., their Moodle or eXeLearning session) to an external server.
2. **Phishing Attacks (Defacement):** A malicious script could modify the DOM (the page structure) to erase educational content and display a fake form asking for passwords instead.
3. **Malicious Redirects:** Forcing the browser to leave the educational resource and redirect the student to a page that downloads malware or viruses (`window.location.href = ...`).
4. **Silent Execution:** Injected scripts hidden within inline events (like an `` or a `<button onclick="...">`) that act without the user noticing.

## 2. The Solution (What this modification mitigates)
We have integrated the **SandboxJS** library through the `SandboxHtmlProcessor` service. 

From now on, before the HTML content of an iDevice (or project properties) is rendered or exported, the processor parses it. All code inside `<script>` tags or inline events (`onclick`, `onload`, `onerror`, etc.) is captured and **encapsulated in an isolated environment (Sandbox)**.

**What does this mean?** The code still executes, but its access to dangerous browser objects is "cut off". There is no longer real access to `window`, `document`, `localStorage`, or outbound `fetch` requests. Therefore, the malicious script is neutralized, protecting the student, the teacher, and the LMS platform where the content is hosted.

---

## 3. QA Testing Guide (4 Validation Examples)

To verify that the `SandboxHtmlProcessor` is doing its job correctly, the tester must perform the following tests. 

**General Instructions:**
1. Open a project in eXeLearning.
2. Add a **Free Text iDevice** (iDevice de Texto Libre).
3. Enable the **HTML Source Code** view in the editor (the `<>` button).
4. Paste the following codes, one by one.
5. Save the iDevice, click on **Preview**, and open the **Browser Console (F12)**.

### Example 1: Session / Cookie Hijacking Attempt
*This script attempts to steal the session cookie and display it (or send it).*
* **Paste this code into the iDevice:**
  ```html
  <div class="security-test">
    <p>Test 1: Cookie Hijacking</p>
    <script>
      try {
        const myCookies = document.cookie;
        console.log("MALICIOUS SUCCESS! Cookies stolen: ", myCookies);
        alert("I have stolen your cookies");
      } catch (e) {
        console.error("SANDBOX BLOCKED ATTACK 1: " + e.message);
      }
    </script>
  </div>
  ```
* **Expected Result:** **No** `alert` showing the cookies should pop up. In the console (F12), you should see the message from the `catch` block or an error indicating that `document` is not defined or is restricted by the sandbox.

### Example 2: Forced Malicious Redirect
*This script attempts to force the student out of the educational resource and take them to another website immediately.*
* **Paste this code into the iDevice:**
  ```html
  <div class="security-test">
    <p>Test 2: Malicious Redirect</p>
    <script>
      try {
        window.location.href = "https://www.google.com";
        console.log("MALICIOUS SUCCESS! Redirecting...");
      } catch(e) {
        console.error("SANDBOX BLOCKED ATTACK 2: Redirect intercepted.");
      }
    </script>
  </div>
  ```
* **Expected Result:** The eXeLearning page must stay exactly where it is. It should not reload or redirect to Google.

### Example 3: Page Destruction (Defacement / Phishing)
*This script attempts to delete the entire eXeLearning interface (menus, headers) and replace it with fake content.*
* **Paste this code into the iDevice:**
  ```html
  <div class="security-test">
    <p>Test 3: DOM Destruction</p>
    <script>
      try {
        document.body.innerHTML = "<h1 style='color:red; font-size:50px;'>YOUR SESSION HAS EXPIRED! Please enter your password again.</h1>";
      } catch(e) {
        console.error("SANDBOX BLOCKED ATTACK 3: Could not modify the main DOM.");
      }
    </script>
  </div>
  ```
* **Expected Result:** The menu, header, and layout of eXeLearning must look **exactly the same as always**. The attempt to overwrite `document.body` must fail.

### Example 4: Evasion via Inline Events (Inline Handlers)
*Attackers sometimes avoid `<script>` tags and instead put malicious code inside attributes of hidden buttons or images (`onclick`, `onerror`).*
* **Paste this code into the iDevice:**
  ```html
  <div class="security-test">
    <p>Test 4: Inline Events</p>
    <button onclick="try { alert('Malicious button clicked! Destroying website...'); document.body.style.display='none'; } catch(e) { console.error('SANDBOX BLOCKED ATTACK 4'); }">
      Click here to claim your prize
    </button>
    <img src="fake-image.jpg" onerror="console.log('Attempted attack via broken image blocked by Sandbox');" style="display:none;">
  </div>
  ```
* **Expected Result:** 
  1. When clicking the button, the page must **not** disappear, nor should it launch dangerous alerts that affect the main flow (depending on sandbox configuration, it will neutralize the event or the access to `document` will fail).
  2. The image with the loading error must not be able to execute privileged logic behind the user's back.

> **Acceptance Criteria:** If after performing these 4 tests the eXeLearning application remains unaffected, does not redirect, does not expose data in the console, and the injected scripts throw reference errors (*undefined* or *access denied*), **the SandboxJS integration is considered validated and successful.**


ignaciogros (commented):
Thank you, @erseco.

I have tested all the proposed code changes and compared their behavior against `main`.

The validation works in all cases, but the current solution prevents the execution of simple JavaScript code. For example:

```html
<script>alert('Test')</script>
```
That will no longer work. When exporting, you'll have this code:

```
<script>
    try {
        const __code = `alert("Test")`;
        const __sandbox = new Sandbox({
            globals: {
                ...Sandbox.SAFE_GLOBALS,
                document,
                window,
                console,
                alert,
                Math,
                RegExp,
                Date,
                setTimeout,
                setInterval,
                clearTimeout,
                clearInterval
            }
        });
        const __exec = __sandbox.compile(__code);
        __exec({
            this: this,
            event: typeof event !== 'undefined' ? event : null
        }).run();
    } catch (e) {
        console.error("Sandbox execution error:", e);
    }
</script>
```

That limits the JavaScript functionality that many users commonly use.

If the goal is to restrict the use of JavaScript, it may be better to remove all inline `script` tags. I think it is a decision that should be discussed with @eXeLearningProject and the development team.

Maybe the current solution could still be valid if the JavaScript were included normally in exports. In that case, we would need to check that opening a document containing such code never results in execution in the application.

I suggest waiting before merging this PR and discussing it with the rest of the team.

ignaciogros (dismissed):
Hi @erseco,

Right now, I can add a simple `<script>alert('Message')</script>` in the `HEAD` section and the footer, but when I tested your examples, I got this JS error, both when previewing and exporting:

`Sandbox execution error: ReferenceError: Sandbox is not defined`

I attach [the file I've been testing with](https://github.com/user-attachments/files/25813224/sandbox-tests_web.zip).

ignaciogros (commented):
I should have seen that it was my commit that broke the file loading. Sorry about that.  

I’ve tested again, and everything is working fine. It allows me to run simple scripts, but not the malicious ones. 

**I think we should discuss this in the group**, because the Sandbox could block common uses, such as DOM manipulation for generating links or forms, scripts for resizing iframes, etc.

Example:

```
<script>
    document.addEventListener("DOMContentLoaded", (event) => {
        const footer = document.getElementById("siteFooter");
        if (footer) footer.innerHTML = 'Example';
    });
</script>
```

erseco:
After reviewing this again, I think we should close this PR without merging it.

The security boundary should not be implemented by rewriting user-authored JavaScript in the editor/export pipeline. If we need to protect the eXeLearning application while previewing or rendering untrusted content, the right place for that isolation is the viewer (plugins) layer, by loading the content in a properly configured sandboxed iframe as @franmate said many times.

That gives us a browser-enforced boundary and keeps the authoring/exported content semantics intact. In contrast, this PR changes how `<script>` tags and inline handlers are executed, which may break legitimate JavaScript use cases and makes the editor/exporter responsible for a security policy that belongs to the rendering context.

There is also a maintenance/security concern with relying on SandboxJS as an additional security layer. Recent SandboxJS advisories include multiple sandbox escapes and prototype-pollution-related issues, so even if it is useful as a defense-in-depth mechanism, we should not treat it as the primary protection boundary.

A better approach would be:

- keep the editor/exported content as authored, unless we explicitly decide to strip JavaScript as a product policy;
- ensure previews/viewers load potentially unsafe content inside a sandboxed iframe;
- carefully define the sandbox flags, avoiding unsafe combinations such as `allow-scripts` + `allow-same-origin` for same-origin content;
- implement any extra restrictions in the viewers, not in the editor/export renderer.

For these reasons, I think this PR should be closed and the issue should be reframed around hardening the viewer (plugins) isolation instead of integrating SandboxJS into the export/editor pipeline.

ignaciogros:
Thanks, @erseco. We will need to review it in the Moodle plugins. It's an important security issue to take into account.

PR #1414, opened 2026-02-28
Fix: Desktop no longer closes silently with unsaved changes

erseco:
**Problem**
On Desktop (Electron), closing the app after opening **Preview** could exit without prompting to save, losing unsaved changes. This happened because the main window close flow previously forced `destroy()` and bypassed any renderer-side unsaved-change protection (and `beforeunload` isn’t reliable/active in this desktop path).

**Fix**

* Intercept `BrowserWindow` `close` and query the renderer dirty state (`documentManager.isDirty` / `assetManager.hasUnsavedAssets()`).
* If dirty, show a native dialog: **Stay** / **Close without saving**.
* Prevent double-triggering with small in-memory guards.
* Register `app://` protocol handler only once per process to avoid macOS re-activate crashes.
* Quit the app when the last window closes (macOS included).

**How to test**

1. Edit without saving → open Preview → close main window.
2. Expected: dialog appears; **Stay** keeps the app open; **Close without saving** exits.

Captura de pantalla 2026-02-28 a las 19 46 48


erseco:
Good point about using `_`.

These strings live in `app/main.js` (Electron main process). The `_` helper we use in the UI is available in the renderer bundle, but it’s not automatically available in the main process (and native dialogs are created from main), so we can’t just call `_()` here unless we introduce a shared i18n layer for Electron.

In 3.x we had `en.json` / `es.json` files for the desktop UI, but they were never properly wired into the Electron main process (resources packaging + initialization + locale selection), so they never worked in practice.

What we should do for 4.x is:
- Add a small i18n module that can be required from the main process (e.g. load `locales/en.json`, `locales/es.json`), pick the language from user settings or `app.getLocale()`, and expose a `_(key)` function.
- Then replace the current hardcoded fallbacks with `_('desktop.unsavedChanges.*')` keys.

For this beta I kept safe fallbacks (`tOrDefault(...)` + en/es literals) so we don’t block testing. Happy to follow up with the shared Electron i18n module and add the translation keys once the close-guard behavior is validated.

PR #1415, opened 2026-03-01
Refactor block icons to use Bootstrap Icons and custom file-manager assets

erseco:
This is an experimental PR to rethink how block/iDevice icons are handled in eXeLearning. 

**The Problem**
Currently, our icon model is too tightly coupled to the selected theme. In my opinion, this approach has several drawbacks:
* The default collection of theme icons is quite limited and lacks standardization.
* Themes can define their own icon names, which means that switching a project's theme often breaks the existing icons.
* It forces theme designers to anticipate every possible pedagogical need. For example, in the Canary Islands, we needed very specific icons for the *"Código Escuela 4.0"* resources. Creating or modifying a whole visual theme just to add specific icons overly complicates the design of OERs (REAs).

**The Solution**
This PR decouples the icon logic from the themes by providing a unified and standard system:
1. **Bootstrap Icons:** We integrated the full Bootstrap Icons library, providing over 2,000 high-quality SVG icons out of the box. This covers almost any pedagogical need.
2. **File Manager Integration:** Users can now select any custom image directly from the file manager to act as a block icon.

**The Advantages**
* **Empowers the Author:** The decision of which icon to use is delegated to the resource creator, rather than the theme developer. It is a much more elegant, sustainable, and simple approach.
* **Safe Theme Switching:** Because icons are now standardized (Bootstrap) or embedded as assets (custom), changing the theme will no longer break the icons.
* **Maintains Visual Harmony:** Themes still control the integration. Each theme defines the default CSS color for the SVG icons to match its palette. If users need a specific color for a specific box, they can still easily override it via custom CSS classes.

*(Please see the attached screenshots to check the new UI selector and the rendered results).*

Let me know your thoughts on this new approach!

Captura de pantalla 2026-03-01 a las 7 44 44
Captura de pantalla 2026-03-01 a las 8 48 34


cristinavaldera:
I think that Material icons would fit better than the others (I like tabler icons also, but it is too large!!).

Regardsless the icons bank used finally, would it be possible (in the future, as enhancement) to locate in the first positions the most recently used icons? Or favourites or similar. Looking for one specific icon among 2500 can be difficult if you don´t remember which concept you searched by!

As this topic is related with styles, I would like to have @ignaciogros opinion too!

erseco:
Hi @cristinavaldera, yes, they are related — both PRs address the same underlying change (replacing theme-bundled icons with a standard library). The difference is that #1497 was opened specifically to evaluate Material Icons as an alternative to the Bootstrap Icons used here.

The plan would be: once we agree on the icon set, we merge whichever branch reflects that decision and close the other. If the team goes with Material Icons, we continue on #1497 and close this one. If we stick with Bootstrap Icons, we merge this PR and close #1497.

erseco:
Closed this because in favor of https://github.com/exelearning/exelearning/pull/1497

PR #1425, opened 2026-03-02
Public viewer URL for projects (read-only, isolated opaque-origin sandbox)

### LINKED ISSUE #348: Public URL
@jesusbautistaq suggested creating a public URL, so users (teachers) can share their projects without using any other tool than eXe.

Adding a new configuration option to allow/disallow that would be interesting too.

New "Publish" UI options will be required.

erseco:
## Summary

Adds a **public read-only viewer** for projects at an opaque, shareable URL (`/view/:publicViewId`), with no authentication required.

Because the viewer serves **author-provided HTML/JS** (which must be treated as untrusted — a teacher may paste AI-generated/copied code, and the link can be opened by anyone, including an admin), the content is **isolated so it can never reach the authenticated session**. This implements the mitigation from the "untrusted content in educational resources" security study (opaque-origin isolation, requirements R1–R4).

## How it works (server-served + opaque-origin sandbox)

> Note: this supersedes the original client-side approach. The viewer no longer downloads a ZIP and renders it via a same-origin Service Worker — that would run untrusted content in the app origin.

- The trusted loader page `/view/:publicViewId` renders a single `<iframe sandbox="allow-scripts …">` **without `allow-same-origin`** → the content runs in an **opaque origin**.
- The content files are served from the server at `GET /view/:publicViewId/_/*`, built once from the project's HTML5 export and cached in memory (keyed by `publicViewId:updated_at`).
- Each content response carries `Content-Security-Policy: sandbox …` (R3), so isolation holds **even fullscreen, in a new tab, or via the raw content URL** — not only when embedded. Plus `Permissions-Policy`, `X-Content-Type-Options: nosniff`, `Cache-Control: no-store`, and `referrerpolicy="no-referrer"` on the iframe.

A service worker cannot control opaque-origin clients, which is why public content is served from the server rather than via the workarea preview SW. The workarea editor preview stays same-origin (the author previews their own content).

With the opaque origin, untrusted content cannot reach `window.parent`/`top`, read/send the `auth` cookie, call the authenticated API, or open the app's IndexedDB/Cache.

## Identifiers & access

- `public_view_id` (opaque `randomUUID`, distinct from the internal UUID) + `public_view_enabled` flag, independent of edit `visibility`; regeneration invalidates old links. Routes return 404 for missing/disabled and never accept the internal UUID.

## Hardening (review follow-up)

- **No unauthenticated bulk ZIP endpoint** — the orphaned `GET /api/v1/export/public/:publicViewId/preview` was removed.
- **Concurrency coalescing** — parallel first-time export builds share one in-flight promise.
- **Size/file-count guard** on the cached export (memory/DoS protection).
- **Selectable CSP profile** via `PUBLIC_VIEW_CSP_PROFILE`: `compatible` (default; external `https:` assets allowed) or `strict` (`connect-src 'none'`, no external resources).

## CSP profile trade-off

Default `compatible` keeps external `https:` assets working (CDN, MathJax, external images, YouTube/Vimeo) since the opaque origin already protects the session. `strict` additionally cuts exfiltration for sensitive deployments. Per-IP rate limiting is left to the reverse proxy.

## Tests

- Unit: sandbox/CSP helper (incl. strict profile + env resolution), content service (path-traversal safety, caching, concurrency coalescing, oversize rejection), `/view/:id/_/*` route.
- E2E (`test/e2e/playwright/specs/public-view-sandbox.spec.ts`): from inside the content frame asserts `window.origin === 'null'`, parent/cookie access blocked, and the response `Content-Security-Policy: sandbox` header without `allow-same-origin`.

## Docs

- `doc/architecture.md` §8.6 (Public View & Untrusted Content Isolation) and `AGENTS.md` §7.11 + the `PUBLIC_VIEW_CSP_PROFILE` env var.

## Architecture records

Design and durable decisions for this change are recorded using the ADR/SDD workflow proposed in #2149:

- **SDD:** `doc/architecture/sdd/SDD-0004-public-read-only-viewer-opaque-origin.md`
- **ADRs:**
  - `doc/architecture/adr/ADR-0017-public-view-opaque-origin-isolation.md`
  - `doc/architecture/adr/ADR-0018-public-view-identifier-and-enablement.md`
  - `doc/architecture/adr/ADR-0019-public-view-export-caching-and-cost-guards.md`

The existing operational docs (`doc/architecture.md` §8.6, `AGENTS.md` §7.11) remain the reference/how-it-works material; the SDD/ADRs capture the design and durable decisions and link to them as evidence. This opaque-origin isolation is a sibling of the editor-preview work in #1968 (ADR-0006–0012) and references it as related. Records follow the paths/templates from #2149 (added minimally here; indexed once #2149 merges); IDs avoid collisions with #2147 (0001–0005) and #1968 (0006–0012).


pabloamayab:
Great job, @erseco . I've been testing it, and it works perfectly. I can't find a single flaw. The only thing—and it's a minor detail—is that the public view doesn't update until you save it, but I don't think it's a big deal. 
I've tested the collaborative features with a public link, with guests, and with the public... it works great.
I think we should take down the “work in progress” label now!
Thanks a million!!!

pabloamayab (approved):
LGTM!

erseco:

The opaque-origin isolation is sound at head: `PUBLIC_VIEW_SANDBOX` deliberately omits `allow-same-origin`, the `sandbox` directive is emitted in the response CSP (not only the iframe attribute), and `normalizePublicViewPath` rejects `..`/encoded-traversal/NUL — covered by both unit and the e2e opaque-origin probe. One minor issue is worth fixing:

- **Minor (error-handling):** `public/app/rest/apiCallManager.js:2357-2361` (`_jsonRequest`) reads `errorData.message` on a non-ok response, but every public-view/sharing handler returns its human text in `detail` (e.g. `src/routes/project.ts:1062` → `{ responseMessage: 'FORBIDDEN', detail: 'Only the project owner can change this' }`). So `_jsonRequest` yields `detail: 'HTTP 403'`, and because that string is truthy, `modalShare.js:773`'s `response.detail || _('Failed…')` surfaces a raw `HTTP 403`/`HTTP 500` toast instead of the server message or the generic fallback. Fix by reading `errorData.detail || errorData.message`.

Verified: static inspection at head only, no local build/test run. CI: all checks green on 204766e ✅.

Residual risk: (1) an oversize export (>100 MB / >5000 files) throws in `getFiles` and is never cached (`src/services/public-view-content.ts:185-243`), so each hit re-runs reconstruct+export+`unzipSync` and returns 500 — owner-gated edge case, low practical risk. (2) `buildHtml5PreviewExport` (`public-view-content.ts:63`) is exercised only by the Playwright e2e spec (unit tests stub it via DI), so its real wiring is not covered by the unit-flag gate; acceptable while the e2e stays enabled. Both proposed findings are accurate but do not rise above minor/hardening.

cristinavaldera (changes requested):
I can only test in the on render deployment, so I cannot see how it looks when you have other saved projects. I like this first page, but I would improve the UI (for example, the button in the right up)

PR #1459, opened 2026-03-06
Desktop: make `Save` always prompt in Electron and remove remembered-path overwrites

erseco:
## Why

The Desktop/Electron save flow was behaving like a partial native desktop app, but the current architecture does not safely support that model.

At the moment, eXeLearning Desktop exports the whole `.elpx` file from the in-memory/Yjs project state each time. It does not maintain a true document/file binding like a traditional desktop editor. The previous implementation tried to simulate that behavior by remembering a save path per project and reusing it on later saves.

That approach introduced several risks:

* silent overwrites on subsequent saves
* fragile path association based on `projectKey` with fallback to `"default"`
* possible collisions across windows/projects
* stale remembered paths after imports or temp-file flows
* extra complexity in Electron main/preload and frontend save branches

Given the current architecture, always prompting for the destination is the safer and simpler behavior. It also matches the product decision to prioritize stability and avoid hidden overwrite risks.

## What this PR changes

This PR makes **Save** in Desktop/Electron always behave like a save dialog flow.

### Main behavior changes

* **Save always prompts for a destination**
* **Save As uses the same flow as Save**
* silent overwrite through remembered full file paths is removed
* only the **last used directory** is remembered, to improve the dialog default path
* suggested filenames still use the project/export name and enforce the right extension

### Electron main process

* removed the remembered full-path save behavior
* replaced `getSavedPath()` / `setSavedPath()` with `getLastSaveDir()` / `setLastSaveDir()`
* simplified `promptSave()` so it always opens the OS save dialog
* updated `will-download` handling to always ask for the target path
* unified `app:save` and `app:saveAs` into the same prompt-based flow
* unified `app:saveBuffer` and `app:saveBufferAs` into the same prompt-based flow
* kept `app:setSavedPath` and `app:clearSavedPath` as no-ops for compatibility

### Electron preload API

* removed the distinction between `save` and `saveAs` in practice
* removed `saveBufferAs`
* removed `setSavedPath` / `clearSavedPath` from the exposed API

### Frontend save/export flow

* removed the `saveAs` branching from Yjs `.elpx` export
* `exportToElpxViaYjs()` no longer receives save-mode options
* Desktop save now always calls the same Electron save flow
* offline “Save As” now delegates to the same save behavior as “Save”

### Cleanup

* removed code that stored original local file paths for later overwrite reuse
* removed template/import cleanup related to remembered save paths
* updated mocks and tests to match the always-prompt model

## Result

This change makes the Desktop save flow:

* safer for user data
* easier to reason about
* easier to test
* closer to the real export-based architecture of the app
* more consistent with Static/PWA behavior

The trade-off is that the user confirms the destination on each save, but the dialog is prefilled with a sensible filename and the last used directory, which keeps the flow fast while avoiding silent overwrite risks.

### DIFF
--- a/doc/development/customization.md
+++ b/doc/development/customization.md
@@ -1,13 +1,26 @@
 # Customization
 
-Educators and admins can safely apply CSS/JS customizations without modifying core code.
+There are two complementary approaches depending on your deployment type:
 
-## Custom CSS (Look & Feel)
+| Approach | How | Works in static mode | Requires server |
+|---|---|---|---|
+| **Custom files** | Edit `custom.css` / `custom.js` on the filesystem | Yes | No |
+| **App name** | Admin → Customization → App Identity | No | Yes |
+| **Favicon** | Admin → Customization → App Identity | No | Yes |
+| **Custom HEAD HTML** | Admin → Customization → Custom HEAD HTML | No | Yes |
+| **Custom Assets** | Admin → Customization → Custom Assets | No | Yes |
 
-- Path: `/public/style/workarea/custom.css`
-- Add your overrides here to change fonts, colors, spacing, etc.
+---
 
-Example
+## Option A: Custom Files (filesystem / static mode)
+
+For sysadmins with filesystem access, Docker volume mounts, or static deployments where there is no server or database.
+
+### Custom CSS
+
+- Path: `public/style/workarea/custom.css`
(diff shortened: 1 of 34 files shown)


pabloamayab:
@erseco Perfect!!! Thanks!!!

Issue #1522, opened 2026-03-10
Enhancement: Apply custom code (CSS/JS) in Edit Mode, not just in Preview


jcayetano-spain:
Currently, custom code such as CSS classes and inline styles are only applied and rendered when the user switches to Preview mode. They are not visible while in Edit mode.

This behavior significantly reduces usability and makes the authoring process uncomfortable, as creators cannot see the real-time visual impact of their structural changes. In previous versions of eXeLearning, the visual results of custom styles were visible directly in the editor (except, of course, when actively editing the specific iDevice).

The exact same issue occurs when inserting custom scripts (JS) into the HTML of an iDevice—they only execute/render in Preview.

**Proposed Solution:**
Please restore or enable the rendering of custom CSS and JavaScript within the Edit mode workspace. Providing a design more aligned with WYSIWYG (What You See Is What You Get) experience is essential for an intuitive design workflow.

ignaciogros:
After reviewing it with the Development Team, it has been decided not to apply custom CSS within eXe to avoid breaking eXe’s UI.

It was also agreed to revisit it in the future, once an option that allows better scoping of that code’s impact is available.

Thank you for the suggestion, @jcayetano-spain.

PR #1527, opened 2026-03-10
Fix #1430 reduce peak memory usage in save, preview and export flows

### LINKED ISSUE #1430: Exelearning freezes when saving this 150 MB file
Exelearning behaves very strangely with this file
https://fromsmash.com/4iFjX0t0PY-ct?e=amF2aWVyYXJpYW5AZ21haWwuY29t

If you open it and click save, the page remains blank and the save dialog box doesn't even appear.
We've tested this on two different computers, with windows 11. 
With de offline versión

But if you keep trying, there comes a point where eXeLearning stops working with any file and gets stuck on the metadata screen, even if you select another page from the sidebar menu.
You need to restart your computer.

erseco:
This PR reduces peak memory usage when working with very large projects, especially the file reported in #1430.

The main goal is to avoid V8 out-of-memory crashes and freezes during:

- save
- preview
- export as website

This affects the Electron desktop app most clearly, but some of the same memory-heavy paths are also shared by other export flows.

## What changed

- replaced O(n²) base64 conversion with chunked conversion in Electron-related flows
- added sequential asset processing to avoid loading all assets into memory at once
- added lightweight asset metadata listing so exporters can build path maps without loading binary data
- reduced memory pressure during ZIP generation
- reduced intermediate HTML buffering in HTML export paths

## Main areas touched

- Electron save/export related base64 conversion
- asset providers used by export flows
- HTML export
- ZIP generation
- preview/export asset handling

## How to test

Please test with the large ELPX file from #1430.

#### In Electron desktop app

Run:

```bash
make run-app
```

Then verify:

* opening the large project works
* page navigation remains responsive
* **Save** works and shows the dialog correctly
* **Export to website** completes without freezing or crashing
* **Preview** opens correctly without running into OOM
* exported website opens correctly and assets are present

#### In other relevant flows

Please also verify the same project in the non-Electron paths that share export logic:

* website export outside the desktop app flow, if applicable in your environment
* preview-related flows outside Electron, if applicable
* generated export contains pages, images and media correctly

## What to pay special attention to

Please confirm that these specific operations work with the large project:

* **Save from Electron**
* **Export to website from Electron**
* **Preview from Electron**
* exported website works correctly after generation

## Related issue

* Closes #1430

### DIFF
--- a/scripts/build-static-bundle.ts
+++ b/scripts/build-static-bundle.ts
@@ -29,6 +29,8 @@ import { XMLParser } from 'fast-xml-parser';
 // Import centralized configuration
 import { LOCALES, LOCALE_NAMES, PACKAGE_LOCALES, LICENSES } from './static-bundle/static-config';
 import { buildConfigParams } from '../src/routes/config-params';
+import { STATIC_ROUTES } from '../src/routes/api-routes';
+import { buildParameterResponse } from '../src/routes/parameter-response';
 import { VOID_ELEMENTS } from '../src/shared/utils/html-constants';
 
 // Re-export config for external use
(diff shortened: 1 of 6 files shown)


erseco:
Hi @pabloamayab I did a update that improves the save/export flow by reducing complexity and making memory handling more predictable.

### Key improvements

- Removed upfront loading of all asset blobs in `save()` and switched to metadata-only flow
- Implemented true per-batch blob loading in upload paths:
  - large assets → loaded per asset
  - small assets → loaded per batch
  - session uploads → loaded per chunk
- Simplified asset size handling by relying on metadata (`size`) instead of blob presence
- Ensured consistent batching logic across all upload methods
- Preserved Cache API as a fallback source while safely releasing in-memory blobs

### Impact

- Eliminates the initial memory spike during save
- Keeps peak memory bounded to the current batch instead of all assets
- Makes the upload flow easier to follow and reason about
- Maintains correct post-save behavior (preview, export, continued editing)

Overall, this makes the save path both more efficient and easier to maintain without changing expected functionality.

Plese, review it again to know if now it solves the problems #1522

pabloamayab:
The save function still uses the same amount of resources

I've compared it to beta2 and can't really tell difference in performance or efficiency

PR #1551, opened 2026-03-15
Fix TinyMCE link plugin issues

ignaciogros:
- Use the selected element instead of adding an unnecessary `id` attribute to all links.
- ~Fix the "Include File Information" option.~

We should carefully test the functionality of the links. This is a widely used feature and must work correctly.


cristinavaldera (changes requested):
Some aspects need to be fixed (above)

Issue #1557, opened 2026-03-16
Untranslated strings in the static version


ignaciogros:
Please do not start to work on this issue before https://github.com/exelearning/exelearning/pull/1553 is merged into "main".

I think that the i18n mechanism of the static version is not working properly.

**Server version, in Spanish:**

<img width="1002" height="702" alt="Image" src="https://github.com/user-attachments/assets/184df78a-1703-43ae-bc3b-548563581c81" />

**Static versión, in Spanish:**

<img width="1002" height="702" alt="Image" src="https://github.com/user-attachments/assets/ec8a681a-d406-4bbf-8052-1ef583aef551" />

Related issue: #1556

PR #1562, opened 2026-03-17
Fix untranslated strings in static bundle UI

erseco:
This PR fixes several untranslated strings that were still present in the static bundle UI.

The changes focus on static HTML elements and inline JavaScript messages that were not being passed through the translation system, even after the translation work merged in #1553.

## Changes

- Added `data-i18n` to the **Styles** sidebar title
- Added `data-i18n` to the **System** tab button
- Added `data-i18n` to the **Imported** tab button
- Updated the recent projects empty state to use the existing translation key:
  - from `No recent projects`
  - to `No recent projects...`
- Replaced the hardcoded delete confirmation message with translated `_()` calls:
  - `Delete this project?`
  - `This action cannot be undone.`

## Tests

Added coverage in `scripts/build-static-bundle.spec.ts` for:

- static HTML i18n coverage
- required translation key existence in `messages.es.xlf`
- config parameter parity between `buildApiParameters()` and `buildConfigParams()`

These tests help prevent regressions in the static bundle translation flow and verify that the static and server config structures stay aligned.

## Context

This addresses untranslated strings reported in #1557 and builds on the translation-related work already merged in #1553.


ignaciogros (changes requested):
I ran `make run-app` from this branch, and I tested the application in Spanish, but the File manager is still en English.

Could you please check that, @erseco?

Thanks!

cristinavaldera (dismissed):
I find the following strings missing:

- Project properties > Export options > Global font - "**Style default**"
- User preferences: the 3 info globes.
- Page properties: "**Advanced (SEO)**"
- Structure panel and page options: "**Import content**"
- Tooltips in Preview: "**Open in a new tab**", "**Fix preview**", "**Refersh preview**"

PR #1564, opened 2026-03-17
Refactor static parameter response to reuse shared config builder

erseco:
This PR fixes the response-shape duplication between the online config endpoint and the static bundle builder.

Until now, both online and static code paths assembled the parameter response separately. Although the config groups were already generated from a shared `buildConfigParams()` factory, the final response structure and static routes were still duplicated, which made static and server outputs drift over time.

## Changes

- reused a shared `buildParameterResponse()` builder in both server and static flows
- replaced the custom static parameter assembly in `build-static-bundle.ts`
- extracted shared API routes into `api-routes`
- reused `STATIC_ROUTES` instead of keeping a separate hardcoded route map
- kept the static-specific behavior explicit with `disableThemeEdition: true`

## Result

The static bundle and the live server now share the same source of truth for the parameter response shape, reducing the risk of future i18n regressions and config mismatches.

## Notes

Static mode still keeps its intentional differences:
- limited route set
- no server app settings
- theme edition disabled


ignaciogros (approved):
Good job! That's important for maintenance.

Tested in different versions adding new strings and translations.

Thanks, @erseco!

PR #1565, opened 2026-03-17
Preserve original filenames in preview downloads

erseco:
This PR fixes document downloads from the preview panel so that browsers keep the original filename instead of using an internal generated name.

Previously, non-HTML resources were fetched, converted to a `blob:` URL, and opened in a new tab. Since blob URLs do not preserve filenames, downloaded files could end up with UUID-based names.

## Changes

- updated the external link handler in `public/preview-sw.js`
- replaced the blob URL tab-opening flow with an anchor-based download using the `download` attribute
- extracted the filename from the resource URL path
- added safe filename decoding with fallback to the raw path segment when decoding fails
- appended the temporary anchor to the DOM before triggering the download for better browser compatibility
- kept the fallback behavior by opening the original URL in a new tab if the fetch-based download fails

## Tests

Updated `public/preview-sw.test.js` to verify that:

- downloads use the `a.download` pattern
- the old `window.open('about:blank')` flow is no longer used for non-HTML downloads
- filenames are extracted correctly from URL paths
- filenames with spaces, non-ASCII characters, multiple dots, and malformed encodings are handled correctly

## Result

Preview downloads now retain the original filename from the exported resource path, improving usability and avoiding confusing UUID-based download names.


erseco:
@mnunezcedec try again!

mnunezcedec (approved):
Hi @erseco, it works and behaves as expected, both in edit mode and in preview. Thank you.

PR #1582, opened 2026-03-20
TinyMCE link plugin. Get file size and extension.

ignaciogros:
"Include File Information" retrieves the file size and extension. The data is now properly updated when changing the file or the unit (B, KB, MB).

Minor HTML error fixed: an extra space before a closing parenthesis.


cristinavaldera (changes requested):
- With an image, it takes size and extension ok. But if you change unit and go to another tab, it is still deleting your changes (it comes back to Bytes)
- With a pdf, it does´nt take size or extension.

https://github.com/user-attachments/assets/f9d92dd7-bba2-4ca5-a17d-b179332d0ffc

cristinavaldera (approved):
YEs!! THan you so much!!!

PR #1593, opened 2026-03-23
Feature: Removed bundled libraries and generate from package.json dependencies

erseco:
This PR stops committing a large set of generated frontend library artifacts to the repository and makes them reproducible from Bun dependencies during the build.

The main change is a new `bundle:vendor` step that runs before the rest of `build:all`. It rebuilds the Yjs browser shims and copies the required distributable files from `node_modules` into the existing runtime paths under `public/`.

This keeps the current runtime model intact:

- the app still loads libraries through static files under `public/`
- exported/offline resources still use the same paths
- `scripts/build-resource-bundles.js` can keep packaging assets from `public/`
- no runtime loading strategy was changed

## Why

The repository was mixing two sources of truth for third-party frontend assets:

1. declared dependencies in `package.json` / `bun.lock`
2. committed built files under `public/libs/`, `public/app/common/`, and some iDevice export directories

That caused a few problems:

- dependency versions could drift from the files actually shipped
- updating a library required manual file replacement
- Dependabot could not manage many of these assets properly
- the repository was storing generated artifacts that can be reproduced during the build

This PR makes Bun the source of truth for the migrated libraries while preserving the public file layout expected by the application and exported content.

### New build step

Added a new script:

- `bundle:vendor`

It is now executed before the parallel bundle steps in `build:all`:

- `bun run build`
- `bun run bundle:vendor`
- then the rest of the existing bundle tasks

This ordering matters because other build steps package files from `public/`, so vendored assets must exist first.

#### `scripts/build-yjs-shims.js`

Builds the Yjs browser files into `public/libs/yjs/`:

- `yjs.min.js`
- `y-websocket.min.js`
- `y-indexeddb.min.js`

This preserves the current Yjs browser integration, including the global `window.Y` pattern used to avoid duplicate Yjs instances.

It also removes the need for the nested `public/libs/yjs/build/package.json`, since the build now uses the root Bun dependencies.

#### `scripts/copy-vendor-libs.js`

Copies required distributable files from `node_modules` into the existing public paths used at runtime and in exported content.

The script fails loudly if an expected source file is missing.

### Already managed in the project and now generated during build

- `yjs`
- `y-websocket`
- `pdfjs-dist`
- `mermaid`

### Added as Bun devDependencies and now generated during build

- `jquery`
- `bootstrap`
- `showdown`
- `interactjs`
- `jquery-ui`
- `abcjs`
- `html2canvas`
- `simplelightbox`

### Existing dependency now copied from package-managed source

- `DOMPurify`

## Files no longer tracked in git

This PR removes the generated vendor artifacts from git and ignores them in `.gitignore`.

That includes generated files under:

- `public/libs/yjs/`
- `public/libs/pdfjs/`
- `public/app/common/mermaid/`
- `public/libs/jquery/`
- `public/libs/bootstrap/`
- `public/libs/showdown/`
- `public/libs/fflate/`
- `public/libs/abcjs/`
- `public/libs/interact/`
- `public/libs/jquery-ui/`
- `public/libs/simplelightbox/dist/`
- `public/app/common/edicuatex/js/DOMPurify/`
- duplicated export assets such as `html2canvas` and `simplelightbox` files in iDevice export directories

For `simplelightbox`, the old committed directory contained several unused variants. This PR keeps only the minified JS/CSS artifacts that are actually needed and reproduces them from Bun dependencies.

## Architecture records

- **SDD:** `doc/architecture/sdd/SDD-0007-vendored-frontend-libs-build-pipeline.md`
- **ADRs:**
  - `doc/architecture/adr/ADR-0029-vendored-frontend-libs-from-npm.md`
  - `doc/architecture/adr/ADR-0030-yjs-browser-shims-esbuild-window-global.md`
  - `doc/architecture/adr/ADR-0031-dependabot-grouped-frontend-dependency-updates.md`

## Fix — undeclared dep, lost CSS override, and dead script specs

1) `fflate` shipped to the browser but resolved only transitively via `@vitest/ui`; now declared in `dependencies` (`^0.8.3`) and grouped under a new `frontend-libs` Dependabot group. 2) The regenerated `simple-lightbox.min.css` had dropped the local `SDWEB` override (aqua links + caption/license padding); it is re-appended in `copy-vendor-libs.js` so exported galleries keep those styles. 3) `scripts/*.spec.ts` were never run in CI; a new `test:scripts` script is wired into `make test-unit`/`test-unit-ci` (not `ci.yml`, which self-guards). Commit `5fcc26d7`; 137 scripts tests pass.


ignaciogros (commented):
Thanks for the PR, @erseco . 

I think that what you're proposing would the best thing to do, but it requires manually reviewing the files that will no longer be part of the application code and will be loaded as dependencies, because some of these files:

- Have changes (even they shouldn't). Example: `simple-lightbox.min.css`. These differences would need to be reviewed to include those changes in the right files.
- Contain different versions, which could generate problems. Example: the version of `abcjs-basic-min.js` which is currently included in the application is older than the imported one. We would need to ensure everything works OK.
- Some automatic updates could introduce issues. Example: upgrading jQuery or jQuery UI. That'll require a review to ensure nothing breaks.

I really like the idea, but I think it might cause problems. Perhaps we should change one dependency at a time.

erseco:
Thanks for the review, @ignaciogros.

I agree with your concerns. This PR is meant as an initial step to start moving in that direction, not as the final state of the migration.

That is also why I did not include the libraries that I think are more sensitive or more likely to cause issues, such as MathJax. The idea is to use this PR as the base to start reviewing the migrated libraries one by one, validate the differences carefully, and leave the path ready so this becomes automatic and maintainable.

Right now we have a bigger long-term problem: some key frontend dependencies are not being updated at all, including Bootstrap, jQuery, and especially Yjs. In several cases, known upstream fixes are already available, but we are not getting them because these assets are effectively detached from package management. That also increases the risk of having a _dependency hell_ over time.

My goal here is to:
- establish the build path for generating/copying these assets from `package.json` dependencies
- review the migrated libraries incrementally on top of this work, directly in the PR
- add tests in the areas where we know regressions are more likely
- make future updates happen through the usual Dependabot flow, like the rest of the dependencies

The initial effort is the hard part. Once the process is in place and covered by tests, future updates should be much easier to review and maintain, since they will arrive as regular dependency update PRs instead of manual vendor file replacements.

So I fully agree that the review should be careful. My intention with this PR is to open that path and start working through it in a controlled way.

erseco:

- **Important:** `fflate` ships to the browser (`window.fflate`, used in `public/app/core/ProjectFolderStorage.js` and `public/app/yjs/ResourceCache.js`) and is copied by `scripts/copy-vendor-libs.js` (fflate COPIES entry → `node_modules/fflate/umd/index.js`), but it is declared in neither `dependencies` nor `devDependencies`. `bun.lock` resolves it only transitively via `@vitest/ui` (test tooling). So the shipped runtime asset floats ungoverned (currently 0.8.3), no Dependabot group/package covers it — contradicting ADR-0031/SDD-0007 which list fflate as npm-declared — and if `@vitest/ui` ever stops depending on fflate the copy step throws and the production build breaks on an unrelated test-tooling change. Declare fflate directly and add it to a Dependabot group.
- **Important (back-compat):** the regenerated `public/files/perm/idevices/base/image-gallery/export/simple-lightbox.min.css` (COPIES → upstream simplelightbox@2.14.3) drops a local customization the deleted committed copy carried: the `/* +++++ SDWEB +++++ */` block (`.sl-wrapper a{color:aqua}` + license/caption `padding-right` rules). That file is listed in the image-gallery `config.xml` and ships in every exported HTML5/SCORM/ePub gallery, so exports silently lose those styles. The rules exist nowhere else at head, and SDD-0007:48's "duplicated byte-for-byte" premise is false here (the `libs/` copy never had the block). The maintainer already flagged `simple-lightbox.min.css` in the PR discussion — re-apply the override as a tracked layer or record the intentional removal in ADR-0029.
- **Test gap:** `scripts/copy-vendor-libs.spec.ts` and `scripts/build-resource-bundles.spec.ts` never execute in CI — `test:unit`/`:ci` run `bun test ./src ./test/helpers`, integration runs `./test/integration`, frontend is Vitest over `public/`, and no step runs `bun test ./scripts`. So assertions like "3 html2canvas destinations" and "buildLibsBundle throws on a missing required lib" are inert: deleting the rubric `html2canvas` COPIES entry would still pass. Add a `scripts/` test step so these guards actually run.

Verified: static inspection at head e3a2582 (diff, deleted blobs vs upstream, `package.json`/`bun.lock`, `.gitignore`, CI/Makefile test globs, PR discussion). CI: all checks green on e3a2582 ✅ — but green CI does not exercise the `scripts/` specs, nor validate the vendored version bumps.

Residual risk (not run locally): the intentional one-time upgrades of abcjs (6.0.4→6.6.3, with the still-tracked `public/libs/abcjs/abcjs-audio.css` left un-bumped), jQuery UI, Bootstrap, interact.js and SimpleLightbox have no rendering/E2E coverage (no music-iDevice E2E); the maintainer's "ensure everything works OK" concern for abcjs stands and needs a runtime check. `buildIdevicesBundle` also lacks the fail-loud `required` guard that `buildLibsBundle` has — minor/defensive, since `build:all` always runs `bundle:vendor` first.

---
**Update — 3 findings resolved (`5fcc26d7`):** `fflate` declared + Dependabot group; `SDWEB` image-gallery CSS override re-applied on every vendor rebuild; `scripts/` specs now run in CI via `make test-unit`/`test-unit-ci` (137 pass).

PR #1614, opened 2026-03-27
Fix legacy .elp internal links broken only in workarea (#1609)

### LINKED ISSUE #1609: Internal links are not working
Open the attached .elp file (saved with eXe 2.9) and click on any of the internal links within the application. Those links don't work, and the JavaScript console displays a message in Spanish: `Navega a exe-node:page-X`.

erseco:
## Summary

Fixes #1609 — Internal links from legacy `.elp` files (eXe 2.9) were broken in `/workarea` but worked fine in preview/export.

## Root cause

During legacy `.elp` import, `remapInternalPageLinks()` remapped `exe-node:page-X` links to new page IDs **only in `htmlView`**, but missed `properties` (e.g., `textTextarea` for text idevices).

The workarea renders text idevice content from `jsonProperties.textTextarea` (not from `htmlView`), so the links kept the legacy `page-9` format. Then `enableInternalLinks()` could not find `.nav-element[nav-id="page-9"]` in the nav tree (which uses the new IDs), so no click handler was attached and the browser tried to follow the raw `exe-node:` URI scheme, producing the error:

```
Failed to launch 'exe-node:page-9' because the scheme does not have a registered handler.
```

Preview was unaffected because the export path reads `htmlView` from Yjs (which was correctly remapped).

## Fix

Extended `remapInternalPageLinks()` to also recursively process `comp.properties`, matching the same approach used by `LegacyXmlParser.convertLinksInObject()`.

## Test coverage

- **New test:** `properties.textTextarea` is remapped for text idevices
- **New test:** nested properties objects with `exe-node:` links are remapped
- All 6721 existing unit tests pass, 0 regressions

## How to test manually

1. Import the attached `enlaces-rotos.zip` (contains a legacy `.elp` file with internal links)
2. In `/workarea`, click an internal link → should navigate to the correct page
3. Open preview → internal links should still work
4. Export as HTML5 → verify internal links resolve correctly


ignaciogros (approved):
Thanks, @erseco.

PR #1620, opened 2026-03-30
Avoid pasted `blob:`-type paths and prepare v4.0.0-beta4 release

### LINKED ISSUE #1621: Missing classes of the selected style in TinyMCE
The CSS classes of the selected style should be available in TinyMCE.

ignaciogros:
- Add `onpaste` warnings to TinyMCE and other fields so the user knows that `blob` type pasted URLs will not work.
- Fix alert warnings `button:focus` presentation.
- Update changelog.
- Updated Spanish (ES) translation. 
- Closes #1621 
- Added automated placeholder translations for incomplete translations.

@erseco,

Please test the solution that informs users when they paste HTML containing `blob:`-type paths. It detects pasting both inside and outside TinyMCE. It does not block the action; it only shows a message. It is important that users are aware of the situation. That's why the messages is not automatically closed.

@mnunezcedec,

Please test the TinyMCE Insert / Edit attributes plugin. In the CSS class selector, there are now two groups: "Predefined classes" and "Current style classes". They are separate so that advanced users are aware of the implications of selecting a style-specific class. I think that's better than keeping all classes together, because style-specific classes might break the presentation (or even functionality) when using a different style.

Thanks.

### DIFF
--- a/public/CHANGELOG.md
+++ b/public/CHANGELOG.md
@@ -14,6 +14,7 @@
 - TinyMCE: usability and accessibility improvements across the editor
 - TinyMCE media plugin: YouTube Live and Shorts URLs are now recognized correctly
 - TinyMCE: toolbar visibility is preserved between editing sessions in the desktop app
+- TinyMCE: missing CSS style classes in the "Insert/Edit Attributes" selector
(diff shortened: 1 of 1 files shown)


erseco (approved):
Works like a charm!
Image

PR #1655, opened 2026-04-09
1618 fix platform related issues

### LINKED ISSUE #1618: Fix platform-related issues
- Publish to %s should work.
- The "File" menu should be hidden, or at least those options that allow you to open a project and loose the context.
- The collaborative options should be hidden.

I think that hiding them with `d-none` would be enough for now.

pabloamayab:
# feat: improve resource initialization and streamline UI for LMS integrations

### 1. Background / Motivation
When a new eXeLearning session is launched from a platform like Moodle via LTI (`edit_ode` request with a `jwt_token`), Moodle sends the initial base64-encoded ELP resource. Previously, eXeLearning correctly created the project session but failed to retrieve and load this base64 string because the callback mechanism was detached from the Yjs collaborative startup flow. Furthermore, when users are within an LMS context, certain standalone controls (like creating new projects, opening local files, sharing, or manual saving) become irrelevant and can cause workflow confusion.

This PR fixes the Base64 loading bug and gracefully hides irrelevant UI elements when the editor detects an active platform integration.

### 2. Technical Decisions & Architecture
* **Stateful LMS Import Trigger:** In `src/routes/pages.ts`, we now append the specific flag `fetchPlatformElp=1` when redirecting to the workarea after creating a new session from a Moodle `odeId`. This explicitly instructs the frontend to seek the remote resource without muddying the existing embedding initialization configs.
* **Base64 Fetch & Client-Side Hydration:** During the `checkAndImportElp()` execution hook in `projectManager.js` (before the workspace renders), the client makes an asynchronous fetch to `POST /api/platform/integration/openPlatformElp`. The `elpFile` (Base64) payload is converted to an `application/octet-stream` File Blob on the fly, and then piped directly into `importFromElpxViaYjs()`. This guarantees the resource leverages the Yjs Document Manager seamlessly and is saved accurately in the collaborative database via `this._yjsBridge.documentManager.saveToServer()`.
* **Robust Nunjucks UI Rendering:** To hide unsupported commands ("New", "New from Template", "Open", "Recent Projects", "Share", "Save", and their corresponding dividers), we avoided brittle DOM queries. Instead, we used the backend variables directly within Nunjucks templates `{% if config.platformIntegration %}d-none{% endif %}`. 
* **Preventing JS Overwrites:** An early return was added to `checkAndShowNewFromTemplateButton()` in `navbarFile.js` because its native implementation removes the `d-none` class dynamically when it successfully queries the API for templates. 

### 3. Commits Overview
1. `fix(lti): append fetchPlatformElp param for new LMS sessions in backend routers`
2. `feat(yjs): intercept LTI creation flag and hydrate base64 resources via openPlatformElp`
3. `style(ui): conditionally hide irrelevant File menu interactions and headers during LMS sessions`

### 4. Testing & Verification
* **Existing Tests:** Since no core Yjs structures or standard vanilla importing APIs were altered, the integration acts as an invisible proxy. `make test-unit` and HTML tests (`make test-frontend`) were executed and finished with exit code `0`.
* **Linting:** Validated against Biome rules (`make fix` / `make lint`), confirming stylistic integrity across `pages.ts`, `projectManager.js`, and `navbarFile.js`.
* **Manual Verification:** 
  - Simulated LTI LMS URL containing `odeId` and `jwt_token`.
  - Verified UI loader displays `"Downloading from platform..."`.
  - Verified `File` menu hides unsupported dropdown elements without disrupting standard desktop standalone mode workflows.

### DIFF
--- a/public/app/workarea/menus/navbar/items/navbarFile.js
+++ b/public/app/workarea/menus/navbar/items/navbarFile.js
@@ -351,6 +351,11 @@ export default class NavbarFile {
         if (capabilities && !capabilities.storage.remote) {
             return;
         }
+        
+        // Hide templates if this is an LMS session
+        if (eXeLearning?.config?.platformIntegration) {
+            return;
+        }
 
         try {
             // Get current locale from eXeLearning config or default to 'en'
--- a/public/app/workarea/menus/navbar/items/navbarFile.test.js
+++ b/public/app/workarea/menus/navbar/items/navbarFile.test.js
@@ -650,6 +650,25 @@ describe('NavbarFile', () => {
             expect(li.classList.contains('d-none')).toBe(true);
         });
 
+        it('checkAndShowNewFromTemplateButton should keep button hidden and return early when platform integration is enabled', async () => {
+            const li = document.createElement('li');
+            li.classList.add('d-none');
+            li.appendChild(mockButtons.newFromTemplateButton);
+            navbarElement.appendChild(li);
+
+            global.eXeLearning.config.platformIntegration = true;
+            global.fetch = vi.fn();
+
+            await navbarFile.checkAndShowNewFromTemplateButton();
+
+            // Should not fetch and stay hidden
+            expect(global.fetch).not.toHaveBeenCalled();
+            expect(li.classList.contains('d-none')).toBe(true);
+            
+            // Clean up config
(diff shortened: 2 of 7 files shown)


erseco (approved):
LGTM!

PR #1660, opened 2026-04-13
fix(export): Export Page/Box/iDevice now writes files in Electron (#1659)

### LINKED ISSUE #1659: Export options are not working
Desktop versión (tested on Windows).

The following export options are not working:

**Page:**

Image

**Box:**

Image

**iDevice:**

Image

I select the folder in which I want to download the contents, but no files are saved.

erseco:
## Summary

Fixes #1659. On the Electron desktop build (reported on Windows), **Export Page**, **Export Box**, and **Export iDevice** opened the save dialog, let the user pick a folder — and then silently wrote nothing. The fix is a single renderer-side contract bug: all three partial-export flows funnel through `componentDownloadHelper.runElectronDownload`, which converted the blob to a **base64 string** and passed it as the first positional arg to `electronAPI.saveBufferAs`.

The main process helper `normalizeBinaryPayload` (`app/main.js:1419`) only treats the first arg as binary when it is a `Uint8Array` / `ArrayBuffer` / `Array`; a string falls through to check a `base64Data` arg that the preload never forwards. So the helper returned `null`, `saveBufferWithDialog` short-circuited with `"Failed to normalize binary payload"`, and `fs.writeFileSync` never ran. To make things worse, the old renderer treated any resolved value as success, so the `{ saved: false, error: ... }` response was swallowed and the browser fallback did not fire either.

The working save-project path (`YjsProjectBridge.js:3185`) and the working per-iDevice exports (`checklist.js`, `rubric.js`, `progress-report.js`) all already pass a `Uint8Array`. This helper was the outlier.

### Changes

- `public/app/workarea/project/idevices/content/componentDownloadHelper.js`
  - Blob branch now reads `response.arrayBuffer()` and hands a `Uint8Array` to `saveBuffer` / `saveBufferAs`, aligning with the working pattern.
  - Removed the now-unused `blobToBase64` helper (no workaround, single source of truth).
  - Added `isElectronSaveSuccessful` so `{ saved: false, ... }` is treated as a failure and the browser-download fallback can still fire.
- `public/app/workarea/project/idevices/content/componentDownloadHelper.test.js`
  - Added 2 TDD regression tests for #1659:
    - `saveBufferAs` receives a `Uint8Array` (not a base64 string) for `blob:` URLs.
    - Browser download fallback fires when Electron reports `{ saved: false }`.
  - Updated 3 pre-existing `blob:` URL tests that were pinning the old (buggy) base64-string contract to the new `Uint8Array` contract.

No changes to `app/main.js` or `app/preload.js`.

## Test plan

- [x] New regression tests confirmed **red** on the unfixed code (`expected 'string' not to be 'string'`, fallback not fired), **green** after the fix.
- [x] `npx vitest run public/app/workarea/project/idevices/content/componentDownloadHelper.test.js` — 56/56 passed.
- [x] `make test-frontend` — 11,637/11,637 passed.
- [x] `make fix` — no lint issues.
- [ ] Manual Electron smoke on `make run-app` (macOS + Windows):
  - [ ] Export Page writes an `.elpx` file at the selected folder.
  - [ ] Export Box writes an `.elpx` file at the selected folder.
  - [ ] Export iDevice writes an `.elpx` file at the selected folder.
  - [ ] Main project Save / Save As still works (no regression on the `YjsProjectBridge` Uint8Array path).

### DIFF
--- a/public/app/workarea/project/idevices/content/componentDownloadHelper.js
+++ b/public/app/workarea/project/idevices/content/componentDownloadHelper.js
@@ -2,32 +2,22 @@ const rememberedComponentKeys = new Set();
 const STORAGE_PREFIX = '__exe_component_download:';
 
 /**
- * Convert a Blob to base64 string.
- * Used for Electron downloads where blob: URLs cannot be streamed via Node.js HTTP.
- * @param {Blob} blob - The blob to convert
- * @returns {Promise<string>} Base64 encoded string
- */
-async function blobToBase64(blob) {
-    return new Promise((resolve, reject) => {
-        const reader = new FileReader();
-        reader.onload = () => {
-            const dataUrl = reader.result;
-            // Extract base64 part after "data:...;base64,"
-            const base64 = dataUrl.split(',')[1];
-            resolve(base64);
-        };
-        reader.onerror = reject;
-        reader.readAsDataURL(blob);
-    });
-}
-
-/**
  * Check if a URL is a blob: URL that needs special handling in Electron.
  * @param {string} url - The URL to check
  * @returns {boolean} True if it's a blob: URL
  */
 function isBlobUrl(url) {
     return typeof url === 'string' && url.startsWith('blob:');
+}
+
+// The Electron main process (normalizeBinaryPayload in app/main.js) accepts
+// Uint8Array / ArrayBuffer / Array as the first positional arg. When the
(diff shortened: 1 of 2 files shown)


erseco:
blocked by #1661

ignaciogros (approved):
Export and import are working.

Thanks!

PR #1667, opened 2026-04-13
fix(yjs): block arrow reorder uses Y.Doc as source of truth (closes #1665)

### LINKED ISSUE #1665: Block reorder via arrows: order is lost / random and breaks exported HTML links
## Summary

A user reports that **moving blocks (\"cajas\") within the same page using the up/down arrow buttons** does not produce a stable order:

- The arrows seem to work in the UI, but the resulting order is not preserved.
- The order behaves randomly after subsequent moves.
- When the project is then exported as a website, the generated site shows \"page not found\" / broken-link errors.

The user clarified that this is about reordering **blocks inside a single page**, not about moving pages in the navigation tree.

## Reproduction (to be confirmed by an E2E test in the same PR)

1. Open / create a project with at least one page containing several blocks (each with one or more iDevices inside).
2. Use the up/down arrow buttons on a block several times in quick succession.
3. Save and navigate to a different page.
4. Navigate back to the original page.
5. Observe that the order of the blocks does not match the order resulting from the moves.
6. Export the project as a website. Open the generated HTML — links between pages and inside the page may resolve to non-existent files.

A new Playwright spec (`block-reorder-stability.spec.ts`) and a Vitest stress test for `YjsStructureBinding.updateBlockOrder` will be added in the fix branch to lock the expected behavior.

## Suspected root causes

These are based on a code-level investigation, not yet confirmed by a failing test (TDD step is in progress).

### 1. \`Y.Array.delete\` + \`Y.Array.insert\` anti-pattern in \`updateBlockOrder\`

`public/app/yjs/YjsStructureBinding.js:1603-1633` implements block reordering as:

```js
this.manager.getDoc().transact(() => {
    const clonedBlock = this.cloneBlockMapForMove(blockMap);
    blocks.delete(currentIndex, 1);
    const insertIndex = newOrder > currentIndex ? newOrder : newOrder; // ← always == newOrder
    const finalIndex = Math.min(Math.max(0, insertIndex), blocks.length);
    blocks.insert(finalIndex, [clonedBlock]);
    // ...then a second pass forces order = arrayIndex on every block
}, this.manager.getDoc().clientID);
```

Two issues:

- **No native move**: `Y.Array` has no `move` operation. Using `delete + insert` is the standard anti-pattern: the inserted item is a *new* element from Yjs's point of view, not the same one that was deleted. Combined with `cloneBlockMapForMove` (which re-creates the `Y.Map`), the moved block loses its Yjs identity — undo/redo history breaks and concurrent edits from other clients can produce duplicates / drops because the deleted-then-reinserted block appears twice in the merged state.
- **Insert-index math is dead code**: `newOrder > currentIndex ? newOrder : newOrder` always returns `newOrder`. Whatever the intent was, the conditional has no effect, so when moving a block downwards the insertion index isn't compensated for the preceding `delete`, and the block ends up one position off.

Same anti-pattern is used for moving pages in `movePage` (`YjsStructureBinding.js:283-288`), which suggests the bug is general, not block-specific.

### 2. DOM-first move with a `setTimeout`-cleared lock

`public/app/workarea/project/idevices/content/blockNode.js:783-829` (`addBehaviourButtonMoveUpBlock`) and lines 835-881 (`addBehaviourButtonMoveDownBlock`):

- The handler reads the previous/next sibling **from the DOM**, mutates `this.order` locally, calls `apiUpdateOrder()`, and then physically reorders the DOM with `insertBefore()`.
- A class `moving` is added to gate against double clicks, but it's removed inside a `setTimeout` (`reorderViaYjs`, lines 1757-1761), which means two arrow clicks fired close together both pass the guard, both compute their new `this.order` from a DOM that's already in flight, and both call `updateBlockOrder` with stale indices.
- `apiUpdateOrder` returns a synthetic `{responseMessage: 'OK'}` regardless of whether the Yjs mutation actually succeeded, so the UI never recovers when the Yjs side disagrees.

The combination of \"DOM is the source of truth for the next move\" + \"Yjs ack is fake\" + \"lock cleared on a timer\" is a textbook race for fast successive arrow clicks.

### 3. Two-source-of-truth ordering: array position vs. `order` property

Each block carries both an array index in the page's `blocks: Y.Array` **and** a numeric `order` field on the `Y.Map`. The reorder code touches both, in two passes inside the same transaction. The export adapter (`src/shared/export/adapters/YjsDocumentAdapter.ts:250-259`) iterates the `Y.Array`, then re-sorts by the `order` field:

```ts
blocksArray.forEach((blockMap, index) => {
    blocks.push(this.convertBlock(blockMap as YMap, index));
});
// Sort blocks by order property to ensure correct rendering order
blocks.sort((a, b) => (a.order ?? 0) - (b.order ?? 0));
```

If the `order` field gets out of sync with the array position (which the delete-insert + double-pass `set('order', k)` makes possible under any kind of concurrent edit), the export silently picks the stale field instead of the real array order. There's no validation that the two agree.

### 4. Page filename map is positional, breaking links

`src/shared/export/exporters/BaseExporter.ts:746-794` (`buildPageFilenameMap`) treats whichever page is at `pages[0]` as the root and writes it to `index.html`. `src/shared/export/renderers/PageRenderer.ts:520-535` (`getPageLink`) checks `page.id === allPages[0]?.id` to decide whether to emit `index.html` or a slugified file path.

If the upstream array is reordered incorrectly (which the same `delete + insert` anti-pattern does to the navigation array), then:

- A page that should be `index.html` ends up with a slug filename.
- Internal links that were generated assuming a different root point to a file that doesn't exist.
- → The exported site shows \"page not found\" exactly as the user reports.

This explains the second half of the report: the reorder bug doesn't only affect blocks visually; the same broken \`delete+insert\` machinery is used for pages, and the export pipeline trusts whatever the post-reorder array looks like.

## Affected files

| File | Lines | Role |
|---|---|---|
| `public/app/workarea/project/idevices/content/blockNode.js` | 695-696, 783-881, 1697-1770 | Arrow click handlers, DOM mutation, calls into `apiUpdateOrder` / `reorderViaYjs` |
| `public/app/yjs/YjsStructureBinding.js` | 283-288, 1193-1240, 1573-1647 | `movePage`, `cloneBlockMapForMove`, `updateBlockOrder` |
| `src/shared/export/adapters/YjsDocumentAdapter.ts` | 250-259 | `convertPage` blocks sort by `order` field |
| `src/shared/export/exporters/BaseExporter.ts` | 746-794 | `buildPageFilenameMap` (positional root assumption) |
| `src/shared/export/renderers/PageRenderer.ts` | 520-535 | `getPageLink` (positional root assumption) |

## Plan

1. **TDD branch** (`tdd/block-reorder-yjs-bug`):
   - Add a Playwright E2E spec `block-reorder-stability.spec.ts` that creates several blocks with iDevices, performs many arrow moves (including back-and-forth and across page-switches), saves and re-opens the page, and asserts the resulting order. Then exports the project and asserts that `index.html` and the inter-page links are valid.
   - Add a Vitest stress test for `YjsStructureBinding.updateBlockOrder` against a real `Y.Doc`: N blocks, many random moves, assert no duplicates / no losses / `order` field === array index.
   - Both must be **red on `main`** to lock in the bug.

2. **Fix branch** (after the test reproduces): replace `delete+insert+clone` with a stable approach (single source of truth — the array position — and remove the redundant `order` field, or fix the insert-index math and drop `cloneBlockMapForMove`). The export adapter should stop re-sorting and trust the array. The page filename map should not depend on positional root identification.

The fix is **not** included in this issue; this issue exists to track the reproducer and the eventual fix.

## Reporter

User report (translated from Spanish):

> Moving the blocks (\"cajas\") with the arrows is failing for me. It looks like it works, but when you view the result the order is not respected. It works at random. And when you export the project as a website it even shows \"page not found\" errors.
>
> *(After clarification: this is about reordering blocks inside a single page, not pages in the tree.)*

erseco:
Closes #1665.

Two commits on this branch:

1. **`14389e63` — failing reproduction tests** (TDD red step). Adds a `describe('updateBlockOrder regression #1665')` block in `public/app/yjs/YjsStructureBinding.test.js` and an end-to-end Playwright spec in `test/e2e/playwright/specs/block-reorder-stability.spec.ts`. Three of the new unit tests fail on `main` and reproduce the bug deterministically.

2. **`eceeb56a` — the fix.** Switches the arrow-driven path in `blockNode.js` to a new `moveBlockRelative` helper on `YjsStructureBinding` that reads the block's current index from the Y.Doc instead of trusting a stale per-instance counter. All 234 unit tests in the binding spec and all 157 in `blockNode.test.js` now pass, plus the rest of the frontend suite (11652 / 0 / 0).

## Root cause

The legacy arrow handlers in `public/app/workarea/project/idevices/content/blockNode.js` kept a per-instance `this.order` field and bumped it directly:

```js
// addBehaviourButtonMoveDownBlock
this.order++;
this.apiUpdateOrder().then(...)

// reorderViaYjs
structureBinding.updateBlockOrder(this.blockId, this.order);
```

`updateBlockOrder` *itself* is correct — the seven baseline tests in this PR (`moving a block from index 0 to index 2`, etc.) prove it. The bug is one level up: after any reorder, the JS instances of every other `IdeviceBlockNode` on the same page still hold their pre-move `order` values, because nothing reconciles them with the Y.Doc. The next click on a neighbour bumps a stale counter and feeds `updateBlockOrder` a target index computed from that snapshot. The block then \"jumps\" two slots instead of the one slot the user wanted, which is exactly the \"funciona random\" the report describes.

The same bug class explains the \"page not found\" half of the report when exporting: `movePage` in `YjsStructureBinding.js` uses the same machinery for navigation and `BaseExporter.buildPageFilenameMap` / `PageRenderer.getPageLink` identify the root page by `allPages[0]`, so any positional drift in the page array rewrites which page becomes `index.html` and breaks the cross-page links. The page side is **not** included here — this PR is scoped to the in-page block reorder. Tracking the page+export half as a follow-up so we can land them independently.

### `public/app/yjs/YjsStructureBinding.js`

Two new methods, no signatures or exports removed.

```js
findBlockLocation(blockId) {
  // walks the navigation Y.Array once, returns
  // { pageMap, blocks, index } | null
}

moveBlockRelative(blockId, delta) {
  if (!Number.isFinite(delta) || delta === 0) return false;
  const location = this.findBlockLocation(blockId);
  if (!location) return false;
  const { blocks, index } = location;
  const targetIndex = Math.min(Math.max(0, index + delta), blocks.length - 1);
  if (targetIndex === index) return false;
  return this.updateBlockOrder(blockId, targetIndex);
}
```

`updateBlockOrder` is left intact — it is still the right entrypoint for callers that genuinely need an absolute index (drag-and-drop, REST imports, ...). `moveBlockRelative` is the right entrypoint for the arrow buttons because the user is asking for a *neighbour* swap, not for a position.

### `public/app/workarea/project/idevices/content/blockNode.js`

The arrow handlers now branch on `isYjsEnabled()`:

- **Yjs path**: call a new `reorderViaYjsRelative(delta)` that delegates to `moveBlockRelative`, then reconciles `this.order` with whatever index the block ended up at (so any leftover code path that still reads `this.order` — drag-and-drop fallback, REST writes — stays sane). `this.order++` / `this.order--` is no longer the source of truth.
- **Legacy path**: unchanged. Still mutates `this.order` and forwards to `apiUpdateOrder`. Builds without Yjs are not affected by this PR.

`reorderViaYjs` (the absolute version) is left as a fallback in case a future bridge revision exposes `updateBlockOrder` but not `moveBlockRelative`.

### `public/app/yjs/YjsStructureBinding.test.js`

The three reproduction tests are rewritten to drive the binding via `moveBlockRelative` instead of the stale-counter simulation. They now pass with the fix.

Five new tests cover the helper itself:

- `moveBlockRelative is a no-op at the array edges` — top/bottom guard.
- `moveBlockRelative clamps larger deltas into the valid range` — `+10` lands at the last index, `-10` at the first.
- `moveBlockRelative returns false when the block does not exist` — fallback contract.
- `findBlockLocation returns the page and the index of an existing block` — positive case.
- `findBlockLocation returns null when the block does not exist` — negative case.

### `test/e2e/playwright/specs/block-reorder-stability.spec.ts`

End-to-end repro left in place. It seeds a page with five blocks, performs seven real arrow clicks alternating between blocks and directions, navigates away and back to verify the order survives a page rebuild from Yjs, and asserts both the Y.Doc and the rendered DOM match the in-memory reference. After the fix it should be green; not run in CI here (would require server-backed E2E) — to be exercised by the existing `test-e2e-chromium` job.

## Test plan

- [x] `npx vitest run public/app/yjs/YjsStructureBinding.test.js` — 234 / 0 / 0 (was 7 pass + 3 fail + 219 skipped on the previous commit of this branch).
- [x] `npx vitest run public/app/workarea/project/idevices/content/blockNode.test.js` — 157 / 0 / 0.
- [x] `make test-frontend` — 11652 / 0 / 0 across 207 files.
- [x] `make fix` — Biome clean.
- [ ] `bun x playwright test --project=chromium test/e2e/playwright/specs/block-reorder-stability.spec.ts` — to be run by CI.
- [ ] Manual smoke in `make run-app`: open a project, add several blocks to a page, click up/down arrows on different blocks rapidly, navigate to another page and back, confirm the order matches the user's intent.

## Out of scope (follow-up)

- Apply the same \"read current position from Y.Doc\" pattern to `movePage`, which has the same anti-pattern at `YjsStructureBinding.js:283-288`.
- Make the export pipeline robust to navigation reordering: stop relying on `allPages[0]` for root identification in `BaseExporter.buildPageFilenameMap` and `PageRenderer.getPageLink`. That is the second half of \"export shows page not found\" in #1665 and lives in the server-side export, so it should be a separate PR with its own backend tests.


erseco:
Coverage bump pushed in \`e94088d2\`.

The previous commit landed the fix but the new code in \`blockNode.js\` (\`reorderViaYjsRelative\` plus the modified up/down arrow handlers) had no direct unit tests, so Codecov flagged the patch coverage at **43.18%**. This commit adds 19 unit tests so the touched lines are exercised:

- \`describe('reorderViaYjsRelative')\` — 9 tests covering happy path, the \`findBlockLocation\` reconciliation that updates \`this.order\`, the \`this.id\` retry fallback, the legacy-bridge fallback to \`reorderViaYjs\`, the throw / ERROR error paths, and the \`moving\` CSS class timer.
- \`describe('addBehaviourButtonMoveUpBlock / addBehaviourButtonMoveDownBlock')\` — 9 tests driving real DOM \`button.click()\` for both arrows: Yjs branch routes through \`reorderViaYjsRelative\` with the right delta and reinserts the DOM block next to its neighbour on success; \`checkOpenIdevice\` short-circuit; \`isAvalaibleOdeComponent\` failure path with the alert; ERROR response leaves the DOM untouched; legacy non-Yjs branch still mutates \`this.order\` and calls \`apiUpdateOrder\`.
- One test for \`YjsStructureBinding.moveBlockRelative\` covering the \`delta=0 / NaN / ±Infinity\` early-return guard.

Pollution avoided: the new hooks **save and restore** \`global.$\` (\`vitest.setup.js\` loads real jQuery as \`global.$\` — deleting it broke 7 unrelated tests in \`generateBlockContentNode\` / \`makeBlockHeadElement\` on a first attempt), \`eXeLearning.app.project.checkOpenIdevice\`, and \`isAvalaibleOdeComponent\`.

### Verification

- \`npx vitest run blockNode.test.js YjsStructureBinding.test.js\` — **410 / 0 / 0** (was 391 / 0 / 0 before this commit).
- \`make test-frontend\` — **11671 / 0 / 0** across 207 files (was 11652).
- \`make fix\` — Biome clean.

Codecov should report a much higher patch % on the next CI run.

erseco:
E2E spec fix pushed in \`acf704a2\`.

The previous version of \`test/e2e/playwright/specs/block-reorder-stability.spec.ts\` timed out in CI at \`waitForDomBlockCount\` right after the \`addBlockViaYjs\` seeding loop. Root cause: \`bridge.addBlock\` writes to the Y.Doc but the node content renderer needs an explicit reload (\`project.idevices.loadApiIdevicesInPage(false, pageElement)\`) to repaint the new DOM nodes. The existing \`test/e2e/playwright/specs/undo-redo-block-structure.spec.ts\` already handles that via a local \`reloadCurrentPageFromYjs\` helper and a combined DOM+Yjs wait — I just hadn't replicated it.

### What changed

- Copied \`reloadCurrentPageFromYjs\`, \`createTargetPageViaYjs\`, \`clearUndoHistory\` and \`waitForDomAndYjsBlockCount\` verbatim from the undo-redo spec.
- Seeding now does: add N blocks via \`addBlockViaYjs\` → \`reloadCurrentPageFromYjs\` → \`waitForDomAndYjsBlockCount(N)\`. Same pattern after the post-moves page bounce.
- Use \`selectNavNode\` from \`test/e2e/playwright/helpers/workarea-helpers.ts\` (same helper the reference spec uses) instead of a hand-rolled selector.
- Move the \"create other page\" step to **after** the arrow moves, so the initial seeding is unambiguously on the source page and nothing can auto-switch during setup.
- Lifted the combined DOM+Yjs wait timeout to 15s (the reference uses 10s; the extra headroom is for the post-bounce reload).

### No production code touched

The unit-level reproduction and fix verification stay in:

- \`public/app/yjs/YjsStructureBinding.test.js\` → \`describe('updateBlockOrder regression #1665')\` — 15 tests.
- \`public/app/workarea/project/idevices/content/blockNode.test.js\` → \`describe('reorderViaYjsRelative')\` (9 tests) and \`describe('addBehaviourButtonMoveUpBlock / addBehaviourButtonMoveDownBlock')\` (9 tests).

All of those still pass:

- \`npx vitest run YjsStructureBinding.test.js blockNode.test.js\` — 410 / 0 / 0.
- \`make test-frontend\` — 11671 / 0 / 0.
- \`make fix\` — Biome clean.
- \`bun x tsc --noEmit\` — clean except the unrelated pre-existing \`baseUrl\` deprecation warning.

Waiting on the next CI run to confirm the E2E is green.

ignaciogros (dismissed):
Please open [arrows.zip](https://github.com/user-attachments/files/26702652/arrows.zip) and add an iDevice B at the bottom of the page (after iDevice A). You'll have 2 boxes. Then click on the down arrow of box A. 

Box A should be placed after box B, but nothing happens.

erseco:
Thanks for the repro @ignaciogros — reproduced live against the branch preview and fix pushed in `7ba40d03`.

### Root cause

`public/app/workarea/project/idevices/content/blockNode.js` had two sibling helpers that drifted apart:

```js
getContentPrevBlock() {
    let previousBlock = this.blockContent.previousElementSibling; // ← element-only, correct
    ...
}

getContentNextBlock() {
    let nextBlock = this.blockContent.nextSibling;                // ← includes text nodes, wrong
    ...
}
```

When a page is rendered from an imported `.elpx`, the `#node-content` container ends up with whitespace text nodes between its `<article.box>` children. `this.blockContent.nextSibling` then returned a `#text` node, the `classList.contains('box')` check failed, `getContentNextBlock()` returned `false`, and `addBehaviourButtonMoveDownBlock` exited silently with `if (nextBlock)` falsy — so the click appeared to do nothing even though the Y.Doc binding and the `#1665` reorder path were fine.

### How I verified it

Loaded `arrows.elpx` (your attachment) into the branch preview (`tdd-block-reorder-yjs-bug.exelearning.pages.dev`), added a text iDevice, and inspected the live DOM:

```json
{
  "nextSiblingType": "#text",
  "nextSiblingIsBlock": false,
  "nextElementSiblingId": "block-1776166857619-qjzmlmakf",
  "nextElementSiblingIsBlock": true
}
```

`nextSibling` returned a text node, `nextElementSibling` correctly pointed at block B. Clicking A's down arrow was a visible no-op in both DOM and Y.Doc, exactly as you described.

### Fix

One line: use `nextElementSibling` in `getContentNextBlock()` so the two helpers stay symmetric. The `moveBlockRelative` / Y.Doc-source-of-truth work from the earlier `#1665` commits is preserved unchanged.

### Tests added

- `public/app/workarea/project/idevices/content/blockNode.test.js` — 4 new cases that pin the behaviour:
  - `getContentNextBlock` skips whitespace text nodes between blocks (regression `#1667`)
  - `getContentNextBlock` skips comment nodes
  - `getContentPrevBlock` symmetric regression lock
  - existing cases kept green
- `test/e2e/playwright/specs/block-reorder-stability.spec.ts` — new E2E `regression #1667`: loads the real `arrows.elpx` fixture, adds a second text iDevice, asserts A's down arrow moves A past B in both Y.Doc and DOM, then round-trips B down and A up-at-top as a no-op.
- `test/fixtures/arrows.elpx` — your attachment checked in so the E2E is reproducible from the repo.

### Verification

- `make test-frontend` → **11675 / 0 / 0** across 207 files (was 11671 before this commit, +4 new cases)
- `make fix` → Biome clean
- `bun x tsc --noEmit` → clean (only the pre-existing `baseUrl` deprecation warning)

Could you re-run your exact scenario on the next branch preview once Cloudflare publishes it? Happy to dig further if there's any other edge case you can think of.

ignaciogros (approved):

Issue #1519, opened 2026-03-10
Enhancement: Comprehensive revamp of the "Rubric" iDevice (CSV import, UX, and interactive assessment)


jcayetano-spain:
The current Rubric iDevice needs a major update to improve both the authoring experience for teachers and the final utility for students. The following essential features are proposed:

**-     Complementary text fields**: Add rich text editor fields to include content before and after the rubric table.
- 
**-     Editing UX improvements**: Currently, editing the content of each cell/matrix box is quite cumbersome. The data entry interface needs to be redesigned to make filling in the descriptors much faster, smoother, and more user-friendly.
- 
**-     Import/Export functionality**: Implement the ability to import and export rubrics using a standard spreadsheet format (e.g., CSV). This would save an enormous amount of time for teachers who already have rubrics created on other platforms.
- 
**-     Interactivity and Self-assessment**: Enhance the generated rubric so that students can interact directly with it. Students should be able to click to mark different performance levels for each criterion, and the iDevice should visually indicate and automatically calculate the corresponding final grade/score.

PR #1670, opened 2026-04-13
fix(desktop): remember last saved file name on subsequent saves (#1666)

### LINKED ISSUE #1666: File name not remembered when saving in desktop version (RC1)
In previous versions of the desktop application, when clicking Save, the application remembered both:

The file path, and
The file name previously used

However, in the current RC1 version, only the path is remembered, while the file name is not.

This means that each time the user saves, they need to re-enter or reselect the file name.

erseco:
## Summary

Fixes #1666 — in `v4.0.0-rc1` the desktop Save dialog stopped remembering the file name the user had previously chosen, only the directory was kept. The behaviour worked in `v4.0.0-beta4`.

**Root cause:** commit 355ae9d9 (PR #1519, the rubric revamp) flipped the precedence in `promptSave` from `storedName || suggestedName` to `suggestedName || storedName`. Because the renderer (`YjsProjectBridge`, `navbarFile`) always passes a freshly computed `suggestedName` derived from the current project title, the stored name became dead code and the native Save dialog never pre-filled the file name the user had typed earlier.

## Fix

- Extracted the precedence rule (and the other inline save helpers `getExt`, `ensureExt`, `getDialogFilterForExt`, `proposeSavePath`) into a new pure module `app/save-utils.js` so it can be unit-tested without booting Electron.
- Added `resolveEffectiveSaveName(suggestedName, storedName)` with the rule:
  - If a `storedName` exists and its extension matches the current export target, prefer `storedName` (restores beta4 behaviour, fixes the bug).
  - Otherwise fall back to `suggestedName` so switching export type (for example `.elpx` project → `.zip` download) still produces a sensible default.
  - A bare project title (no extension) is treated as compatible with any stored extension.
  - Comparison is case-insensitive.
- Wired `app/main.js` `promptSave` to use the helper and removed the duplicated inline implementations.

## TDD evidence

- `app/save-utils.spec.ts` adds 22 unit tests, including the explicit regression case from #1666:
  - `resolveEffectiveSaveName('fresh-project.elpx', 'user_chose.elpx')` must return `'user_chose.elpx'`.
- `bun test app/save-utils.spec.ts --coverage`:

```
-------------------|---------|---------|-------------------
File               | % Funcs | % Lines | Uncovered Line #s
-------------------|---------|---------|-------------------
All files          |  100.00 |   98.41 |
 app/save-utils.js |  100.00 |   98.41 |
-------------------|---------|---------|-------------------
 22 pass
 0 fail
 37 expect() calls
```

100% function coverage and 98.41% line coverage on the new module — well above the 90% project target for both total and patch coverage.

## Test plan

- [x] `bun test app/save-utils.spec.ts` — 22/22 pass.
- [x] `make fix` (Biome) — clean, no fixes applied.
- [x] `node -c app/main.js` — syntactically valid after refactor.
- [x] Verified failing tests in the broader `bun test src/` run are pre-existing and reproduce on `main` without this change (test pollution in API v1, websocket and access-notifier suites — unrelated to this fix).
- [ ] Manual smoke test in Electron desktop build: open a project, Save As `MyCourse.elpx`, edit, Save again — dialog must pre-fill `MyCourse.elpx` (not the project title).
- [ ] Manual: with the same project, trigger an HTML5 export — dialog must propose a fresh `.zip` name, not `MyCourse.elpx`.

Closes #1666


ignaciogros (dismissed):
Thank you, @erseco.

**I think that there are still some issues:**

- If you open a file and save it, it doesn’t remember that file’s name.
- That is problematic in this case: I save a file and then open another one, and it remembers the name of the first saved file, making it easy to overwrite it unintentionally.

**Ideally:**

- Every time you create a new file, it should start with no remembered value (default name: the project title, “Untitled document,” or even empty).
- Every time you open a project, it should remember the file name associated with it.
- Every time you save a project, it should remember the last name you saved it with (I think it already does this).

erseco:
@ignaciogros thanks for the review — you're right, my first commit missed two of the scenarios you described. I just pushed `7f3e26fa` which addresses them following your three rules.

**Root cause**

The `lastSaveInfo` slot was keyed by `projectKey` (the project UUID). In desktop/static mode that UUID **is regenerated on every `location.reload()`** — which is exactly what Open and New trigger — so the per-project slot was useless outside the "save the same project multiple times" flow. That explained both (a) opening a file never pre-filled its name and (b) the risk of overwriting the previous file with the name left over from the last session.

**Fix**

Added a single global `currentFileSave` slot in `settings.json` that survives the full reload, plus two new IPCs:

- `app:setSavedPath({ filePath })` → persists `{ dir, name }` via the new pure helper `splitSavePath()`.
- `app:clearSavedPath()` → forgets the associated file.

`promptSave` still checks the per-project slot first (so repeated saves of the same project keep working) and falls back to the global slot. Both save handlers now write to both slots, so every save also refreshes the global one.

**Wiring for the three rules**

| Rule | Where it's called |
|---|---|
| Open project → remember its name | `public/app/app.js` (file association) and `navbarFile.openUserOdeFilesEvent` (File > Open) → `setSavedPath(filePath)` |
| New project → no remembered value | `projectManager.transitionToProject({action:'new'})` and `modalTemplateSelection.loadTemplate` → `clearSavedPath()` |
| Save → remember last saved name | already worked, but it now also mirrors into the global slot so it survives the reload |

I also covered the `import` case in `transitionToProject` in case the flow enters there.

**TDD**

`app/save-utils.spec.ts` went from 22 to 29 tests. I added:

- 4 tests for the new `splitSavePath()` (POSIX, Windows, nullish, bare basename).
- 3 scenario tests that literally encode your three rules, including the exact case you reported: "save A.elpx, open B.elpx → the dialog shows B.elpx, never A.elpx".

```
 29 pass
 0 fail
 52 expect() calls
```

I also ran the affected vitest suites (`app.test.js`, `navbarFile.test.js`, `modalTemplateSelection.test.js`, `projectManager.test.js`) → 361 + 224 pass, 0 fail. `make fix` clean.

**Manual smoke-test plan** (I can't run this myself — it needs the Electron build):

1. Start the desktop app clean → Save → dialog proposes the project title. ✅
2. Save as `one.elpx` → Save again → dialog pre-fills with `one.elpx`. ✅
3. File > Open → pick `two.elpx` → Save → dialog pre-fills with `two.elpx` (not `one.elpx`). ✅
4. File > New → Save → dialog proposes the project title, **not** `two.elpx`. ✅
5. Export to HTML5 after saving a `.elpx` → suggests `.zip`, not the `.elpx` (the extension filter in `resolveEffectiveSaveName` still works). ✅

If you notice any flow I missed while testing (e.g. opening from "Recent" or drag-and-drop), let me know and I'll wire it to the same `setSavedPath`.

ignaciogros (dismissed):
Hi @erseco,

Please check this behavior:

- Open eXe, type a title and save the file. Example: "Documento sin título" is saved as "documento-sin-titulo-1.elpx".
- Click on File - New. Then click on save (with or without changing the project title).
- The predefined name is still "documento-sin-titulo-1.elpx".

The risk is that a user might click on Save without reading the predefined title and loose the original "documento-sin-titulo-1.elpx".

I think that the predefined value should be cleaned each time you open a file or create a new one.

Thanks.

erseco:
@ignaciogros good catch — your repro was spot-on. Just pushed `5c583f64` with the actual fix and a regression test for the exact flow you described.

**Root cause (round 1 was incomplete)**

PR #1670 round 1 added a global `currentFileSave` slot to remember the file across the full page reload that follows File → Open / File → New, but `promptSave` still read `perKey.name || globalInfo.name` — i.e. the per-project cache (`lastSaveName[<projectKey>]`) took priority over the global slot. That broke two of your three rules:

1. **Save A → Open B → Save.** `setSavedPath(B)` only updated the global slot. `lastSaveName[uuid-A]` still pointed at A, and perKey won, so the dialog proposed A. Exactly what you reported.
2. **Save `documento-sin-titulo-1.elpx` → File → New → Save.** `clearSavedPath()` only removed the global slot. On the fresh-Electron / fallback path the key collapses to `'default'` (the renderer hasn't attached an `odeId` yet), so `lastSaveName['default']` survived `window.location.reload()` and the dialog still proposed the old name.

**Fix**

Two pure helpers in `app/save-utils.js`, both unit-tested without Electron:

- `pickStoredSaveInfo(perKey, globalInfo)` — the **global slot now wins** for the name. The per-project cache is only a fallback when the global slot has none (so repeated saves of the same project still prefill, which was the whole point of round 1).
- `clearSavedNameCache(settings)` — nukes `lastSaveName` (keeping `lastSaveDir`, because "last used folder" is still useful).

Wiring in `app/main.js`:

- `setCurrentFileSaveInfo` and `clearCurrentFileSaveInfo` now both call `clearSavedNameCache` before touching the global slot. That way `setSavedPath` (Open) and `clearSavedPath` (New) really forget every stale per-project name.
- Save handlers (`saveUrlWithDialog`, `saveBufferWithDialog`) call `setCurrentFileSaveInfo` **before** `setLastSaveInfo`, so the just-saved project's slot is repopulated *after* the cache wipe instead of being wiped along with the stale ones.
- Both handlers now read the stored info through `pickStoredSaveInfo` instead of the old `perKey.name || globalInfo.name` precedence.

**TDD**

`app/save-utils.spec.ts` grows from 29 to 40 tests. The new block literally encodes your repro against the pure state helpers (`applySave` / `applySetCurrentFile` / `applyClear` mirror the three IPC entry points):

- `save A → open B → Save dialog pre-fills B.elpx, never A.elpx`
- `save documento-sin-titulo-1.elpx → File > New → Save dialog does NOT pre-fill that name`
- `save A → File > New → save B → File > New → dialog is empty again` (belt-and-braces, no cross-session leak)
- `save A → save A again uses A.elpx` (happy path; makes sure the round-2 fix doesn't regress round 1)

```
 40 pass
 0 fail
 71 expect() calls
```

Plus the 590 affected vitest frontend specs (`app.test.js`, `navbarFile.test.js`, `modalTemplateSelection.test.js`, `projectManager.test.js`) → all green. `node -c app/main.js` clean.

**Manual smoke-test plan** (I still can't run this myself — needs the Electron build):

1. Start the desktop app → Save → dialog proposes the project title. ✅
2. Save as `one.elpx` → Save again → dialog pre-fills with `one.elpx`. ✅
3. File > Open → pick `two.elpx` → Save → dialog pre-fills with `two.elpx` (**not** `one.elpx`). ✅
4. **Your exact repro:** Save as `documento-sin-titulo-1.elpx` → File > New → Save → dialog proposes the project title (Untitled / "Documento sin título"), **not** `documento-sin-titulo-1.elpx`. ✅
5. Export to HTML5 after saving a `.elpx` → dialog suggests `.zip`, not the `.elpx` (the extension-filter branch in `resolveEffectiveSaveName` is unchanged). ✅

If you spot another flow where the stale name leaks (drag & drop, "Recent files", double-clicking a `.elpx` in the OS), please let me know the exact steps and I'll wire those entry points to `setSavedPath` / `clearSavedPath` too.

ignaciogros (changes requested):
Hi @erseco,

This is still happening, at least on Windows:

- Open eXe, type a title and save the file. Example: "Documento sin título" is saved as "documento-sin-titulo-1.elpx".
- Click on File - New. Then click on save (with or without changing the project title).
- The predefined name is still "documento-sin-titulo-1.elpx".

Could you please review the solution?

Thanks.

erseco:
@ignaciogros thanks — your repro was right, my third round missed this exact code path. Just pushed `8986e018`.

**Root cause (round 3 was still incomplete)**

`createSession()` in `navbarFile.js` has two branches:

1. **Unsaved changes** → `modalSessionLogout` → `transitionToProject({action: 'new'})` → already calls `clearSavedPath()`. Round 2 wired this.
2. **Clean project** (your exact repro: save, then File → New) → falls into the static/Electron branch `window.newProject()` from `scripts/static-bundle/static-index.html`, which just does `location.reload()`. That branch was **never** calling `clearSavedPath`, so after the reload the global `currentFileSave` slot in Electron's `settings.json` still held `documento-sin-titulo-1.elpx`, and the round-2 `pickStoredSaveInfo` rule (global wins) dutifully pre-filled it.

**Fix**

One call to `window.electronAPI.clearSavedPath()` at the top of `createSession()`, **before** either branch runs, so both paths (clean and unsaved) clear the slot. Wrapped in try/catch so a missing electronAPI can't block creating a new project.

**TDD**

Three new cases in `navbarFile.test.js`:

- Regression for your exact scenario: `clearSavedPath` must fire before `window.newProject` on the static/Electron branch.
- Belt-and-braces: it also fires on the `transitionToProject` branch.
- Survives a missing `clearSavedPath` helper without throwing.

```
 Test Files  1 passed (1)
      Tests  189 passed (189)
```

All four related frontend suites still green (404 tests total), `bun test app/save-utils.spec.ts` 40/40, `make fix` clean.

**Manual smoke test for your exact Windows repro**

1. Open eXe → type title → Save → pick `documento-sin-titulo-1.elpx`. ✅
2. File → New → Save → dialog proposes the project title (Untitled / "Documento sin título"), **not** `documento-sin-titulo-1.elpx`. ✅ (this is the one that was failing)
3. Save `A.elpx` → make a change → Save again → pre-fills `A.elpx` (round-1 regression check). ✅
4. File → Open → pick `B.elpx` → Save → pre-fills `B.elpx`, not `A.elpx` (round-2 regression check). ✅
5. Export HTML5 after saving a `.elpx` → dialog suggests `.zip`. ✅

Please give it another spin on Windows when you have a moment. If any other File → … entry point leaks a stale name, let me know the exact steps.

ignaciogros (approved):
It works now. Thanks for those changes, @erseco.

Maybe this should be in a new issue, but I add this comment here just in case you prefer to do it now: Could the application remember the last path so you don't have to navigate to the folder again each time you open/create and save a project?

Example: I open a project from my desktop, then I click on "Save", but it opens as destination folder "Documents" instead of "Desktop".

erseco:
@ignaciogros the Save dialog now also remembers the **last folder** you used — not just the file name. Rules applied:

- Every successful Save pins a session-wide `lastUsedDir` in `settings.json`.
- `File > New` clears the current-file slot but keeps `lastUsedDir`, so the next project's Save dialog reopens at that folder instead of falling back to `~/Documents`.
- `File > Open`, the project file picker, and Export-to-folder all seed their `defaultPath` from `lastUsedDir` as well.
- Opening a project from disk via the Electron file picker stores the full path, so the follow-up pathless `setSavedPath(file.name)` from the import flow no longer clobbers the directory.

Scenario matches the one reported: open a project from the Desktop → click Save → dialog opens in Desktop, not Documents.

Covered by unit scenarios in `app/save-utils.spec.ts` (`save/open/new flow` and `resolveSaveDir`).

ignaciogros (changes requested):
Thanks, @erseco.

Now the folder's remembered, but there are still some file name-related issues:

- Save "Documento sin título X" as `documento-sin-titulo-x.elpx`
- New file: "Documento sin título Y" (saved as `documento-sin-titulo-y.elpx`).
- Open `documento-sin-titulo-x.elpx` and click on "Save".

The default name should be `documento-sin-titulo-x.elpx`, but it's `documento-sin-titulo-y.elpx` instead.

erseco:
Thanks for the repro, @ignaciogros  — root cause found and fixed.

In static/Electron mode, `File > Open` goes through `openFileInputStatic`'s hidden `<input type="file">`, not the native `electronAPI.openElp` IPC. That path never called `setSavedPath`, so the main-process "current file" slot still held the previously saved project's name — hence `documento-sin-titulo-y.elpx` leaking after you opened X.

Fix: call `setSavedPath` from the static file-input change handler before handing off to the unified open flow. When Electron exposes it, we also pass the absolute path via `webUtils.getPathForFile` (new `electronAPI.getFilePath` binding) so the folder is remembered too; otherwise we fall back to the basename and let the session-wide `lastUsedDir` handle the directory.

Covered by three new regression tests in `public/app/workarea/menus/navbar/items/navbarFile.test.js` (`openFileInputStatic` describe block).

ignaciogros (approved):
Thanks, @erseco!

ignaciogros (changes requested):
I thought it was fixed, but...

- Open the app and save `documento-sin-titulo-a.elpx`.
- Close the app.
- Open the app and type "Documento sin título 001" as project title.
- Click on "Save".

- Expected default file name: `documento-sin-titulo-001.elpx `
- Default file name: `documento-sin-titulo-a.elpx`

Could you please check the solution?

erseco:
Follow-up on the RC1 repro from the issue. The previous fix cleared `currentFileSave` on **File > New / File > Open**, but the **save → quit → relaunch → rename title → Save** path was still broken.

**Root cause**: `settings.json` persists `currentFileSave` (global slot) and `lastSaveName[uuid]` (per-project cache) across process lifetimes. After relaunch, `resolveEffectiveSaveName` picks the stored name over the fresh title-derived `suggestedName` whenever the extensions match — so the dialog kept proposing `documento-sin-titulo-a.elpx` even after the user renamed the project.

**Fix** (`app/main.js`): call `clearCurrentFileSaveInfo()` at the top of `app.whenReady()`. A fresh process has no file currently associated with the window; the argv / macOS `open-file` flow re-seeds the slot later via the renderer's `setSavedPath` call in `openFileFromPath`, so double-click-to-open keeps working.

**Tests** (`app/save-utils.spec.ts`): added 4 scenarios exercising the startup clear:
- exact repro: save → quit → relaunch → rename → Save uses the new title
- no-rename relaunch: Save still proposes the same title-derived name
- folder memory survives (`lastUsedDir` untouched, dialog opens at the last folder)
- argv-file launch: `setSavedPath` correctly re-seeds the slot after the startup clear

51/51 save-utils tests pass; 192/192 navbarFile tests pass. Commit c9d3a5cf.

ignaciogros (approved):

PR #1671, opened 2026-04-13
fix(editor): prevent TinyMCE from clobbering data-mce-src after blob upload (#1664, #1668)

### LINKED ISSUE #1664: Desktop version. Assets are not displayed.
Add an image, then add another one. Both from the File manager.

When you do that, the first one is not displayed, but the asset URL is OK:

Image

If you go to Tools - Edit source code and save without changes, both images are displayed again:

Image

### LINKED ISSUE #1668: Images disappear in editor when adding or modifying multiple images in an iDevice
When inserting multiple images into an iDevice, previously inserted images temporarily disappear during editing.

Specifically:

- When inserting a second image, the first image disappears
- When inserting a third image or resizing an existing one, the previous image disappears

However, after saving the iDevice, all images are displayed correctly.
This indicates that the issue only affects the editing view, not the actual content.

Steps to reproduce

- Insert an image into an iDevice
- Insert a second image → the first one disappears
- Insert a third image or resize an existing one → another image disappears
- Save the iDevice → all images appear correctly

https://github.com/user-attachments/assets/557fdcf9-601e-4496-a00c-6d5dda5e2f3f

erseco:
Fixes #1664, fixes #1668.

## Root cause

`automatic_uploads: true` makes TinyMCE's upload pipeline scan every `` and call our `images_upload_handler`. For cache hits the handler calls `success(blobUri, …)` so the image keeps rendering. The problem is that in `tinymce.min.js` the post-upload branch runs:

```js
e.status && l.getParam("images_replace_blob_uris", !0, "boolean")
  ? (m.removeByUri(u.src),
     l.$(r).attr({ src: o, "data-mce-src": l.convertURL(o, "src") }))
```

With `images_replace_blob_uris` defaulting to `true`, TinyMCE rewrites **both** `src` and `data-mce-src` to whatever we returned — i.e. the blob URI — clobbering the canonical `asset://` value previously written by `resolveAssetUrlsInEditor`.

Inserting a second asset image retriggers the scanner; the first image (already blob-sourced) gets re-queued and its `data-mce-src` is overwritten with a blob URL. The image then appears "broken" in the editor even though the persisted HTML stays correct (because `prepareHtmlForSync` converts blobs back to `asset://` on `GetContent`). The classic workaround — **Tools → Edit source code → Save** — works precisely because it forces a full reparse through `prepareContentForEditorLoad` + `resolveAssetUrlsInEditor`, which strips the corrupted `data-mce-src` and re-resolves cleanly.

## Fix

Set `images_replace_blob_uris: false` in the TinyMCE init config. TinyMCE still marks blobs as uploaded (`markUploaded` runs *before* the replace branch), so the scanner does not re-queue them, but it no longer rewrites the `` attributes. Our `resolveAssetUrlsInEditor` / `prepareHtmlForSync` pipeline already owns the blob ↔ `asset://` mapping end-to-end.

This is a 1-line config change; no logic changes.

## Test plan

- [x] `npx vitest run public/app/editor/tinymce_5_settings.test.js` — 120/120 pass (new failing test `init disables images_replace_blob_uris …` now green)
- [x] `make fix` — lint clean
- [ ] Manual: insert two images via File Manager → both render, both keep `data-mce-src="asset://..."`
- [ ] Manual: resize/replace an existing image → previous image stays visible
- [ ] Manual: undo/redo and dialog edit flows (regressions from #1607)

## Note

@ignaciogros reported after my initial analysis that the bug also reproduces in non-Electron Chromium browsers (not only the desktop build). The TinyMCE code path addressed here is browser-agnostic, so the fix applies regardless of the runtime. If any residual Chromium-specific symptom remains after this lands, please reopen with a fresh repro and I'll dig into blob-URL lifecycle differences as the next layer.

### DIFF
--- a/public/app/editor/tinymce_5_settings.js
+++ b/public/app/editor/tinymce_5_settings.js
@@ -351,6 +351,13 @@ var $exeTinyMCE = {
             },
             /* enable automatic uploads of images represented by blob or data URIs*/
             automatic_uploads: true,
+            // Prevent TinyMCE from rewriting src/data-mce-src after images_upload_handler
+            // resolves. Our handler returns the blob URI (for display) and we manage the
+            // blob ↔ asset:// mapping ourselves via resolveAssetUrlsInEditor /
+            // prepareHtmlForSync; letting TinyMCE clobber data-mce-src with a blob URL
+            // corrupts previously-resolved images when a second asset:// image is inserted
+            // (issue #1664).
+            images_replace_blob_uris: false,
             file_picker_types: 'file image media',
             /* and here's our custom image picker - opens Media Library modal */
             file_picker_callback: function (cb, value, meta) {
--- a/public/app/editor/tinymce_5_settings.test.js
+++ b/public/app/editor/tinymce_5_settings.test.js
@@ -245,6 +245,18 @@ describe('TinyMCE 5 Settings', () => {
       expect(wrapper.classList.contains('hidden-editor')).toBe(true);
     });
 
+    it('init disables images_replace_blob_uris so TinyMCE does not overwrite data-mce-src after upload (issue #1664)', () => {
+      // With the default `images_replace_blob_uris: true`, TinyMCE's upload
+      // pipeline overwrites both `src` and `data-mce-src` with whatever the
+      // handler returns via success(). Because our handler returns the blob
+      // URI for cache hits, that clobbers the canonical asset:// value we
+      // previously set in resolveAssetUrlsInEditor — corrupting the first
+      // image whenever a second asset image is inserted.
+      globalThis.$exeTinyMCE.init('multiple', '#editor', true);
+      const config = globalThis.tinymce.init.mock.calls[0][0];
+      expect(config.images_replace_blob_uris).toBe(false);
+    });
+
     it('init instance callback triggers toggler and editor hook', () => {
       globalThis.$exeTinyMCEToggler.documentWidth = 1000;
       const initSpy = vi.spyOn(globalThis.$exeTinyMCEToggler, 'init').mockImplementation(() => {});
--- a/test/e2e/playwright/specs/collaborative/file-manager.spec.ts
+++ b/test/e2e/playwright/specs/collaborative/file-manager.spec.ts
@@ -123,22 +123,37 @@ async function uploadFile(page: Page, fixturePath: string): Promise<void> {
 }
 
 /**
- * Helper to select the first file in the grid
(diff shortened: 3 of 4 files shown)


erseco:
Added an end-to-end Playwright regression spec that actually exercises the user-reported flow (not just the config contract): [`test/e2e/playwright/specs/idevices/text-multi-image.spec.ts`](../blob/1664-desktop-version-assets-are-not-displayed/test/e2e/playwright/specs/idevices/text-multi-image.spec.ts).

## What the spec does

1. Creates a project, adds a text iDevice, enters TinyMCE edit mode.
2. Opens the File Manager via the image toolbar button, uploads `test/fixtures/sample-2.jpg`, inserts it.
3. Collapses the caret to the end of the editor body (otherwise a second click on the image toolbar button would put TinyMCE's dialog into *edit selected image* mode and replace the first image instead of appending a new one).
4. Opens the File Manager again, uploads `test/fixtures/sample-3.jpg`, inserts it.
5. Reads every `` in the TinyMCE iframe and asserts, per image:
   - `src` is a live `blob:` URL, `naturalWidth > 0`, `complete === true` (the bitmap actually loaded)
   - `data-mce-src` is **never** a `blob:` URL — it is either absent or the canonical `asset://` URL (this is the exact regression from Case B in issue #1664)
   - an `asset://` reference still exists on either `data-mce-src` or `data-asset-src`

## Verified both ways

I temporarily reverted the 1-line fix and ran the spec against the unfixed code:

```
  ✘  inserting a second image keeps the first image canonical and visible
    Error: both images must finish loading their bitmap data
    Expected: 2
    Received: 1
```

Only one of the two images loads its bitmap data — exactly matching @ignaciogros's report ("the first one is not displayed"). With the fix back in, the same spec passes in ~6 s (and both images show `naturalWidth=536`, `complete=true`, canonical `asset://` references intact).

So the spec is a real regression guard: it would have caught this bug pre-#1607 too, and it will catch any future regression on the same TinyMCE code path.

Test plan (updated):
- [x] `npx vitest run public/app/editor/tinymce_5_settings.test.js` — 120/120
- [x] `make fix` — lint clean
- [x] `bun x playwright test --project=chromium test/e2e/playwright/specs/idevices/text-multi-image.spec.ts` — 1 passed on the fix, 1 failed on the unfixed code (manually verified)

ignaciogros (approved):
Thanks, @erseco.

It worked perfectly in all cases, even in collaborative edition.

PR #1676, opened 2026-04-14
docs(agents): enforce test + 90% patch coverage submission gates

erseco:
## Summary

- Add a top-level **Definition of Done** block to `AGENTS.md` listing the 7 non-negotiable submission requirements: `make fix`, unit tests, integration tests, E2E tests, patch coverage ≥ 90% on changed lines, colocated tests in the same PR, and no `.skip` / `.todo` tests.
- Rewrite `AGENTS.md` section 5.3 as **Coverage & Submission Gates**, clarifying that the 90% threshold is measured against the diff (patch coverage), not the global project average, and listing the exact pre-submit command sequence.
- Update the frontmatter `description:` of all 10 skills (`backend-service`, `backend-route`, `frontend-module`, `e2e-test`, `idevice`, `exporter`, `database-migration`, `api-v1`, `websocket-yjs`, `i18n`) so each one explicitly states the testing artifacts required for its domain and reminds the agent to run `make fix`, `make test-unit`, `make test-integration`, and `make test-e2e` (plus `make test-e2e-static` for the exporter) before submission.

## Why

Agents working on this repo have been landing changes without consistently running the full test matrix or adding tests for new lines. Global coverage can mask under-tested diffs, and the previous wording only described coverage as a soft \"target\". These edits make the expectation unambiguous at both the top-level instructions and at each skill entry point, so the requirement is visible wherever an agent starts work — not just buried in a single section. Patch coverage (not global average) is called out explicitly so new/changed lines cannot ride on pre-existing coverage.

## Test plan

- [ ] Review rendered `AGENTS.md` on GitHub and confirm the new \"Definition of Done\" and \"Coverage & Submission Gates\" sections read clearly.
- [ ] Spot-check a few `.agents/skills/*/SKILL.md` files to confirm the updated descriptions are well-formed YAML frontmatter.

### DIFF
--- a/.agents/skills/api-v1/SKILL.md
+++ b/.agents/skills/api-v1/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: api-v1
-description: Adding or modifying external REST API v1 endpoints in src/routes/api/v1/.
+description: Adding or modifying external REST API v1 endpoints in src/routes/api/v1/. Every change must ship with colocated `*.spec.ts` tests covering auth, success, and error paths, update `doc/development/rest-api.md`, keep patch coverage ≥ 90% on changed lines, and pass `make fix`, `make test-unit`, `make test-integration`, and `make test-e2e` before submission.
 ---
 
 # Skill: REST API v1 (External)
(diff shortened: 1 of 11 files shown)


ignaciogros (approved):
Running some of those Make commands before submission might be quite slow, but I think we should try to achieve that coverage to maintain and improve the code quality. We can always remove that part if we find any performance problems.

Thanks!

PR #1679, opened 2026-04-14
feat(idevice): in-button progress bar for Download Source File (#1677)

### LINKED ISSUE #1677: Add progress bar when downloading iDevice source file (ELPX)
Currently, when a user downloads the ELPX file from the iDevice using the "Download Source File" option, there is no visual feedback indicating the progress of the download.

It would improve the user experience to include a progress bar, similar to the one displayed when loading a project. This would help users understand that the download is in progress and provide an estimate of completion, especially for larger files.

erseco:
## Summary

Closes #1677.

Downloading the ELPX source file from the `download-source-file` iDevice had no visual feedback — the button text briefly switched to `Generating...` and then stayed there until the full fetch-and-zip pipeline finished. On larger projects with many assets this felt frozen.

This PR turns the existing download button into a Bootstrap-styled determinate progress bar **without replacing the button**, so users keep their customized background color, text color and font size.

### What changed

- `public/libs/exe_elpx_download/exe_elpx_download.js`
  - `showLoadingIndicator(true)` now:
    - Preserves the original `innerHTML` and `className` on `data-original-html` / `data-original-class` attributes.
    - Adds the Bootstrap utility classes `position-relative overflow-hidden` to the button so the overlay is clipped.
    - Injects a two-span structure — `.exe-elpx-progress-bar` (the animated bar) and `.exe-elpx-progress-label` (the `Processing... X%` text).
    - Sets `aria-busy="true"` and disables pointer events.
  - `showLoadingIndicator(false)` restores the original markup/classes and clears `aria-busy`.
  - `updateProgress(completed, total)` now computes a clamped percentage and updates the bar width + label. It's already wired into the existing fetch worker pool, so progress advances smoothly as files stream in.
- `public/app/common/common_i18n.js` — adds `elpxProcessing` key (`Processing...`) so the label is translatable in exported sites.
- Unit tests (`exe_elpx_download.test.js`) — updated the `showLoadingIndicator` suite to match the new Bootstrap overlay structure, added assertions for:
  - Bootstrap utility classes (`position-relative`, `overflow-hidden`, `position-absolute`, `top-0`, `start-0`, `h-100`, `bg-dark`, `bg-opacity-25`).
  - `aria-busy` set/cleared.
  - Percentage math (`Math.round((completed / total) * 100)` clamped to `[0, 100]`).
  - A full-workflow test verifying the button is fully restored (original text, no `data-original-html`, no `aria-busy`) after a successful download.

The produced markup mirrors the example in the issue:

```html
<a class="... position-relative overflow-hidden" aria-busy="true" aria-valuenow="65">
  <span class="exe-elpx-progress-bar position-absolute top-0 start-0 h-100 bg-dark bg-opacity-25" style="width: 65%;" aria-hidden="true"></span>
  <span class="exe-elpx-progress-label position-relative">Processing... 65%</span>
</a>
```

## Test plan

- [x] `make test-frontend` — 207 files, 11645 tests passing (including the 89 tests in `exe_elpx_download.test.js`).
- [x] `make fix` — clean (no lint changes).
- [ ] Manual: open an exported site with a `download-source-file` iDevice and click the download button — verify the bar fills from 0% to 100% while files are fetched and that the button is restored after the browser save dialog appears.
- [ ] Manual (file:// fallback): open an exported site via `file://` and trigger the download — verify the processing label shows while the folder picker finishes reading files.

### DIFF
--- a/public/app/common/common_i18n.js
+++ b/public/app/common/common_i18n.js
@@ -40,6 +40,7 @@ $exe_i18n = {
     "mode_toggler": c_("Light/Dark mode"),
     "teacher_mode": c_("Teacher mode"),
     "elpxGenerating": c_("Generating..."),
+    "elpxProcessing": c_("Processing..."),
     "elpxFolderPickerTimeout": c_("The folder picker did not respond. This may happen when opening exported files directly from the filesystem (file:// protocol). Try opening the file through a local web server instead."),
     "elpxFolderPickerEmpty": c_("No files were returned by the folder picker. This is a known limitation when opening exported files with the file:// protocol in some browsers. Try using a different browser or opening the file through a local web server."),
     "elpxFileProtocolWarning": c_("Local mode: Due to browser security policy, you will need to select the folder from which you opened this file. On a web server this will not be necessary."),
--- a/public/libs/exe_elpx_download/exe_elpx_download.js
+++ b/public/libs/exe_elpx_download/exe_elpx_download.js
@@ -411,40 +411,97 @@
     }
 
     /**
-     * Show/hide loading indicator
+     * CSS classes added to the host button so the absolutely positioned
+     * progress bar is clipped and stacks correctly (Bootstrap utility classes).
+     * @type {string[]}
+     */
+    var PROGRESS_HOST_CLASSES = ['position-relative', 'overflow-hidden'];
+
+    /**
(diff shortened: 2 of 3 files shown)


ignaciogros (approved):
I think it's a really good improvement, also with a simple but nice presentation.

Thank you!

PR #1686, opened 2026-04-16
fix(assets): recover missing blobs from server during export (#1685)

### LINKED ISSUE #1685: Cache Issues in the Online Version
In the online version of eXeLearning 4, there are still caching issues. A resource (I’ll send you the URL privately, Ernesto), which has been under development for quite some time, fails to download most of its assets when accessed via the browser typically used to edit it.
In the image, you can see the difference in size, and in eXe Desktop, we can see how the assets have disappeared.

Image

Image

Image

erseco:
## Summary

Fixes #1685 — Cache Issues in the Online Version.

When a project is edited in the browser for a long time, the browser may evict Cache API entries under storage pressure. Since blobs are immediately evicted from memory after being written to the Cache API, this leaves assets with Yjs metadata but no retrievable blob data. Exports then **silently drop those assets**, producing much smaller files with missing images.

### Root causes identified

1. **`getBlobForExport()` had no server fallback** — when both memory and Cache API missed, it returned `null` and the asset was silently skipped during export.
2. **`downloadMissingAssets()` didn't detect blob-less assets** — it checked `getAsset(id)` which returns non-null when Yjs metadata exists, even when the blob is `null`. So assets with metadata but evicted blobs were never re-downloaded.
3. **No `navigator.storage.persist()`** — the browser could freely evict Cache API entries under storage pressure without any protection.

### Changes

- **`AssetManager.js`**: `getBlobForExport()` now falls back to fetching individual assets from the server REST API when both memory and Cache API miss. Server credentials stored via `setServerConfig()`.
- **`AssetManager.js`**: `downloadMissingAssets()` now checks `!local.blob` in addition to `!local`, so assets with metadata but evicted blobs are re-downloaded on project open.
- **`AssetManager.js`**: `navigator.storage.persist()` requested during init to reduce browser-initiated cache eviction.
- **`YjsProjectBridge.js`**: Calls `setServerConfig()` during init to wire API credentials.
- **`BrowserAssetProvider.ts`**: `processMetadataAssets()` logs a warning with count and IDs of assets still missing after all fallbacks.

## Test plan

- [x] New unit tests for `getBlobForExport` server fallback (5 tests)
- [x] New unit test for `downloadMissingAssets` detecting blob-less assets
- [x] New unit tests for `BrowserAssetProvider` missing blob warning (2 tests)
- [x] New unit test for `setServerConfig`
- [x] All existing tests pass (`make test-unit` — 6806 tests, 0 failures)
- [x] 90% coverage threshold met
- [ ] Manual test: open a project in the online version, export as ELPX and HTML5 web zip, verify all assets are present


ignaciogros (approved):
Hi @erseco,

All assets were exported properly, and I saw the confirmation message for the first time:

Image


I think that we should merge this now, and wait for feedback, as some users might find strange that persistent storage permission request.

PR #1688, opened 2026-04-17
feat(workarea): add admin panel link to user dropdown for admin users

erseco:
## Summary

- Expose `user.isAdmin` (plus the parsed `user.roles` array) in the workarea view model so templates can branch on role.
- Add a new `admin_panel` translation key to the workarea `t` bag and extract it into all 11 locale XLF files, with targets filled for `es`, `ca`, `va`, `gl`, `pt`, `it`, `de`, `eu`, `ro` and `eo`.
- Render a conditional **Admin panel** entry in `views/workarea/menus/menuHeadTop.njk` that links to the existing `/admin` route. The entry is only emitted when `user.isAdmin` is true, so non-admin users see no change.

The existing `/admin` route (at `src/routes/pages.ts`) already enforces `ROLE_ADMIN` server-side and returns 403 otherwise, so this PR only adds discoverability — the authorization boundary is unchanged.

## Implementation notes

- `currentUser.roles` is stored as either a JSON string or an array; the workarea route now normalizes it the same way the `/admin` route does before computing `isAdmin`.
- The dropdown item uses `{{ basePath }}/admin` so installs mounted at a sub-path keep working, and carries a `data-testid="admin-panel-link"` for future E2E assertions.
- Translations were extracted with `make translations`; only the new trans-unit was added, no other keys were touched.

## Test plan

- [x] `bun test src/routes/pages.spec.ts` — 86 pass, including two new cases covering `isAdmin` false/true flows and the presence of the `admin_panel` translation key.
- [x] `make fix` — biome check clean.
- [ ] Manual smoke in the workarea: log in as a `ROLE_ADMIN` user, open the avatar dropdown, confirm the **Admin panel** entry renders and navigates to `/admin`.
- [ ] Manual smoke as a plain `ROLE_USER`: confirm the dropdown still shows only *Preferences* and *Logout* (no Admin panel entry).


ignaciogros (approved):
Hi @erseco,

It’s much better with the link...

I’ve made a few changes because the link was displayed but nothing happened when clicking on it.

I’ve also reverted (at least temporarily) the changes in the language files and left the link text as “Admin” for the following reasons:

* In some languages the layout breaks. This way we avoid long strings in that menu and having to adjust CSS right now.
* The admin panel interface is always in English. That’s why I added the `hreflang` attribute to the link.

Please confirm if this works for you.

erseco:
It works, go ahead! This was a suggestion by @ismagago

PR #1694, opened 2026-04-18
[codex] Fix LaTeX rendering in Select activity export

### LINKED ISSUE #1693: Select activity does not render LaTeX in exported web questions
## Summary
The **Select** activity (`quick-questions-multiple-choice`) shows LaTeX source such as `\(\oplus\)` instead of rendering it when the question is displayed dynamically.

## Steps to reproduce
1. Create a **Select** activity.
2. Enter a question containing inline LaTeX, for example `\(\oplus\)`.
3. Save the question and start/view the activity inside the editor.

## Actual result
The question displays the raw LaTeX source instead of rendered math.

## Expected result
The activity should render the LaTeX expression with MathJax when the question/word blocks are displayed dynamically.

## Scope
This is reproducible directly inside the editor when the activity view is rendered. Exporting is not required to trigger the bug.

## Root cause
The runtime detects LaTeX correctly, but the follow-up `updateLatex()` calls in `quick-questions-multiple-choice/export/quick-questions-multiple-choice.js` pass selectors without the `#` prefix:

- `seleccionaWordDiv-${instance}`
- `seleccionaQuestionDiv-${instance}`

Since `updateLatex()` expects a valid CSS selector, MathJax never receives the correct target node for those dynamic updates.

## Proposed fix
Update both calls to use ID selectors:

- `#seleccionaWordDiv-${instance}`
- `#seleccionaQuestionDiv-${instance}`

and add a regression test covering LaTeX rendering for the question/word blocks in the Select activity runtime.

jjdeharo:
## Summary
This fixes LaTeX rendering in the **Select** activity (`quick-questions-multiple-choice`) when questions and word/definition blocks are rendered dynamically.

Fixes #1693.

## What changed
- updated the dynamic MathJax refresh for the word/definition block to target `#seleccionaWordDiv-${instance}`
- updated the dynamic MathJax refresh for the question block to target `#seleccionaQuestionDiv-${instance}`
- added a runtime regression test that verifies `updateLatex()` receives valid ID selectors for both dynamic areas

## Root cause
The runtime already detected LaTeX correctly, but the follow-up `updateLatex()` calls used selector strings without the `#` prefix. Since `updateLatex()` expects a valid CSS selector, MathJax never received the correct node after the question/word content was updated dynamically.

## Impact
Inline LaTeX such as `\(\oplus\)` now renders correctly when the Select activity view is shown inside the editor, and the same fix also applies to exported web output because it uses the same runtime.

## Validation
- `npx vitest run public/files/perm/idevices/base/quick-questions-multiple-choice/export/quick-questions-multiple-choice.test.js`
- `make fix`
- `make test-unit` currently fails before the changed code is exercised due to an existing repo-level build issue in `css:node`/`sass` (`Positional and ":" arguments may not both be used.`)


ignaciogros (changes requested):
Thank you, @jjdeharo!

The fix is correct, but please adjust the test file before moving out of Draft:

The manual `$exeDevices` setup inside `loadExportIdevice` is redundant, because `vitest.setup.js` already provides that object preloaded. You can follow the pattern used in other files like `classify.test.js` or `casestudy.test.js`.

```
function loadExportIdevice(code) {
    let modifiedCode = code.replace(
        /var\s+\$quickquestionsmultiplechoice\s*=/,
        'global.$quickquestionsmultiplechoice ='
    );
    // ...
}
```

The `[\s\S]*?` pattern is too broad. Maybe you can use the exact iDevice name.

Thanks.

mnarvaezm:
Hi, and thanks for fixing it. I added a commit to your branch that fixes updateLatex calls in other iDevices that followed the same pattern as the one in Select.
@jjdeharo

jjdeharo:
Updated as requested.

The test now reuses the preloaded `$exeDevices` object from `vitest.setup.js`, and the auto-init removal uses the exact `$quickquestionsmultiplechoice.init()` pattern instead of the broader regex.

I also reran the focused Vitest file.

mnarvaezm (approved):

ignaciogros (commented):
Thank you!

PR #1698, opened 2026-04-19
fix(static): accept ?url= downloads without an .elpx/.elp/.zip pathname

erseco:
## Summary

- The static bundle's `?url=` auto-import rejected any URL whose pathname did not end in `.elpx`, `.elp`, or `.zip` *before* fetching, which breaks common download endpoints that return a ZIP payload under an extensionless path — e.g. `https://codeload.github.com/OWNER/REPO/zip/refs/heads/BRANCH`.
- `fetchUrlParam()` now drops the up-front pathname gate. After a successful fetch, it resolves the filename from `Content-Disposition` first, the URL path second, and a fallback third. If the resolved name still lacks a known extension, the payload is accepted when its first four bytes match the ZIP local-file-header signature (`PK\x03\x04`, `PK\x05\x06`, or `PK\x07\x08`). Non-archive payloads still produce the original `URL must point to an .elpx, .elp, or .zip file` error.
- Network errors (`fetch()` throwing `TypeError`) now surface a more actionable message that points users to the browser console for the underlying CORS/DNS details — CORS on cross-origin archive hosts is a browser-level restriction that the static build cannot work around on its own.

## Before / after

| URL | Before | After |
|-----|--------|-------|
| `https://codeload.github.com/.../zip/refs/heads/main` | ❌ rejected as non-`.zip` path | ✅ fetch attempted; if the host exposes CORS the file imports |
| `https://github.com/.../archive/refs/heads/main.zip` | ❌ CORS, generic "Failed to fetch" | ❌ CORS (unchanged — browser enforces), but error now hints at CORS/console |
| `https://somehost.example/download` returning a ZIP | ❌ rejected as non-`.zip` path | ✅ validated via `Content-Disposition` or ZIP magic bytes |
| `https://somehost.example/download` returning text | n/a (wouldn't have reached fetch) | ❌ rejected with the existing archive-only message |

Note: GitHub's archive endpoints do **not** expose `Access-Control-Allow-Origin` on either `github.com/.../archive/...zip` or `codeload.github.com/.../zip/...`, so GitHub-hosted archives still cannot be imported directly from the static app. That requires a server-side proxy, which is out of scope for this hotfix.

## Test plan

- [x] `make fix` — Biome clean.
- [x] `make build-static` — static bundle rebuilds; new helpers are visible in `dist/static/index.html`.
- [x] `bun x playwright test --project=static test/e2e/playwright/specs/url-param-static.spec.ts` — 3 passed (extensionless + `Content-Disposition`, extensionless + magic-byte fallback, non-archive rejected).
- [x] Manual smoke: open `http://localhost:8080/?url=<CORS-friendly .zip endpoint without file extension>` and confirm the project imports. like localhost:8080/?url=https://github-proxy.exelearning.dev/?repo=ateeducacion/exelearning-style-spectrum128k&branch=main


ignaciogros (approved):
Yes! It works perfectly. And I have Spectrum 128K installed!

ignaciogros (approved):

PR #1713, opened 2026-04-22
fix(editor): recover pasted images stripped by paste_as_text (#1712)

### LINKED ISSUE #1712: Paste images in TinyMCE
You can't paste images in TinyMCE. Pasted images should be treated as other assets: they should be included in the File manager, etc.

Just in case it's useful:

eXeLearning 2.9 (TinyMCE 4, not 5) handled it this way:

**TinyMCE settings .js file:**

```
// Drag and Drop
paste_data_images: true,
images_upload_handler : function(blobInfo, success, failure) {
    var editor = tinyMCE.activeEditor.getBody();
    var imgs = editor.getElementsByTagName("IMG");

    var n = imgs.length - 1;

    var blobType= blobInfo.blob().type;
    if(blobInfo.blob().name === undefined){
        if(blobType.includes('image/')){
            blobName='img'+n+'.'+blobType.substr(6);
        }else{
            blobName='img'+n+'.png';
        }
    }else if(blobInfo.blob().name == 'image.png'){
        blobName='img'+n+'.png';
    }else{
        blobName=blobInfo.blob().name;
    success(exe_tinymce.dragDropImage(
    }
    'previewTinyMCEimageDragDrop', this, '', this,
    this.name, blobName, blobInfo
            .base64()));
},
```

**Another .js file:**

```
dragDropImage : function(theTarget, node, evalAfterDone, win, win_name,
        blobName, blobBase64) {
    var local_imagePath = 'data:image/jpeg;base64,' + blobBase64;

    var unescaped_local_imagePath = unescape(local_imagePath);
    var oldImageStr = new String(blobName);

    exe_tinymce.uploaded_file_1_name = "";

    var RegExp1 = /[\ \\\/\:\%\&]/g;
    var ReplaceStr1 = new String("_");
    var newImageStr = oldImageStr.replace(RegExp1, ReplaceStr1);

    var early_preview_imageName = encodeURIComponent(newImageStr);
    var preview_imageName = early_preview_imageName.replace(RegExp1,
            ReplaceStr1);
    var full_previewImage_url = "/previews/" + preview_imageName;

    var previewTinyMCEDragDropImageDone = function(imgURL) {

        var alternativeText = function(button, input_alt_value) {

            var editor = tinyMCE.activeEditor.getBody();
            var imgs = editor.getElementsByTagName("IMG");

            var n = imgs.length - 1;

            // The last image in the editor (default image to update)
            var img = imgs[n];

            for (var i=0;i<imgs.length;i++) {
                if (imgURL && imgURL!="") {
                    var currImg = imgs[i];
                    if (typeof(currImg.src)=='string' && jQuery(currImg).attr("src")=="/previews/"+imgURL) {
                        // The image to update (see issue 479)
                        img = currImg;
                    }
                }
            }

            img.setAttribute('width', img.width);
            img.setAttribute('height', img.height);

            if (input_alt_value.length == 0) {

                parent.Ext.Msg
                        .confirm(
                                "",
                                _("Are you sure you want to continue without including an Image Description? Without it the image may not be accessible to some users with disabilities, or to those using a text browser, or browsing the Web with images turned off."),
                                function(button) {
                                    if (button == "yes") {
                                        img.setAttribute('alt', '');
                                    } else {
                                        Ext.Msg
                                                .prompt(
                                                        _('Image description'),
                                                        _('Please provide an image description (alternative text):'),
                                                        alternativeText);
                                    }
                                });
            } else {

                if (button === 'ok') {
                    img.setAttribute('alt', input_alt_value);
                } else {
                    img.setAttribute('alt', '');
                }
            }
        }

        if (exe_tinymce.forcePrompt==true) {
            Ext.Msg
                    .prompt(
                            _('Image description'),
                            _('Please provide an image description (alternative text):'),
                            alternativeText);
        } else {
            exe_tinymce.forcePrompt = true;
        }

        eXe.app.un('previewTinyMCEDragDropImageDone',
                previewTinyMCEDragDropImageDone);
    }
    eXe.app.on('previewTinyMCEDragDropImageDone',
            previewTinyMCEDragDropImageDone);

    window.parent.nevow_clientToServerEventPOST(theTarget, node,
            evalAfterDone, false, win, win_name, unescaped_local_imagePath,
            preview_imageName);

    return (full_previewImage_url);
},

```

erseco:
## Summary

Closes #1712.

**Problem.** TinyMCE's built-in text-paste path strips images, so four distinct flows were broken:

| # | Source | Was | Now |
|---|---|---|---|
| 1 | OS screenshot (only `image/png`) | ✓ built-in pipeline | ✓ unchanged |
| 2 | App copy with image + text (`kind:'file'` + `text/html`) | ✗ paste plugin skipped the file | ✓ our early handler stores it |
| 3 | Google Docs / web page (`` in HTML) | ✗ `paste_as_text:true` stripped the HTML | ✓ we fetch + store (fallback to external URL on CORS) |
| 4 | TinyMCE → TinyMCE / TinyMCE → external app | ✗ `` meant nothing to anyone | ✓ copy emits self-contained `data:` URLs; paste decodes them back (SHA-256 dedup) |

`paste_as_text: true` is preserved — no regressions to the "text-only by default" paste behaviour the project chose.

### Part A — Copy/cut

`ed.on('copy cut', …)` runs before the paste plugin. When the selection contains `` it:

1. Parses the selected HTML with `DOMParser`.
2. Resolves each image's bytes:
   - `blob:` → `fetch(src).then(r => r.blob())`
   - `asset://uuid` → `AssetManager.getBlob(uuid)`
   - `https?://` → left untouched (refetching is wasteful and breaks CORS)
3. `FileReader.readAsDataURL` → rewrites `src` to `data:image/…;base64,…`.
4. Strips `data-mce-src` / `data-asset-src` (internal attributes); keeps `data-asset-id` as a round-trip fast-path hint.
5. Writes to the clipboard via `navigator.clipboard.write([new ClipboardItem({ 'text/html': htmlPromise, 'text/plain': textBlob })])` — the async clipboard API accepts Promises so the FileReader work fits inside the copy user-gesture without pre-computing. `execCommand('Delete')` is issued for `cut`.

Browsers without `navigator.clipboard.write` → no-op; default copy path still runs.

### Part B — Paste (expanded)

`handleImagePaste` now covers three `` schemes on top of file items:

- **`asset://uuid`** — if `AssetManager.getAssetMetadata(uuid)` hits, the image is reinserted directly (no fetch, no re-hash). `resolveAssetUrlsInEditor` handles the blob: display URL on the next tick. Foreign UUIDs are logged and skipped.
- **`data:image/…`** — fetched + `AssetManager.insertImage(file)` which dedups by SHA-256 so the same bytes collapse to the same UUID. The `data-asset-id` attribute we left behind on copy short-circuits the decode when the hint is already local.
- **`https?://`** — existing branch (fetch + store; external URL fallback on CORS / 404 / non-image).

Pure-image paste (no accompanying text) still defers to TinyMCE's built-in `paste_data_images` pipeline to avoid double-inserting.

## Changes

- `public/app/editor/tinymce_5_settings.js` — `handleCopyCut` + `_buildCopyHtml` / `_getBlobForSrc` / `_encodeBlobAsDataURL`; `handleImagePaste` now splits HTML `` into `localAssets` / `dataImages` / `remoteImages`; new `_insertLocalAsset`, `_insertPastedDataURL`, `_insertAssetImgTag`.
- `public/app/editor/tinymce_5_settings.test.js` — 18 new unit tests (TDD, confirmed red first). Helper builder now supports HTML/text payloads.
- `test/e2e/playwright/specs/idevices/paste-image-tinymce.spec.ts` — 4 Playwright scenarios: file+text, remote URL, `data:` URL, full round-trip (verifies clipboard HTML is `data:` not `asset://`, and that the second paste reuses the first image's UUID via hash dedup).

## Test plan

- [x] `make fix` clean.
- [x] `make test-frontend` — **11796/11796 pass** (18 new).
- [x] `make test-unit` — **6857/6857 pass**.
- [x] `make test-integration` — **700/700 pass** (client-only change, untouched).
- [x] Playwright — **4/4** scenarios pass on Chromium.
- [ ] **Manual QA matrix:**
  - [ ] Copy image in TinyMCE → paste in same iDevice → 2 ``, same `data-asset-id`, File Manager shows 1 asset.
  - [ ] Copy image in TinyMCE → paste in another Text iDevice of the same project → same as above.
  - [ ] Copy image in TinyMCE → paste in a second eXeLearning tab → image stored as a local asset.
  - [ ] Copy image in TinyMCE → paste in Word / Gmail / Apple Notes → image renders from the `data:` URL.
  - [ ] Copy image from a Chrome web page → paste into TinyMCE → stored as asset.
  - [ ] Copy image from Google Docs → paste into TinyMCE → stored when CORS allows, external URL kept otherwise.
  - [ ] Copy image from Word → paste into TinyMCE → file-item path, stored as asset.
  - [ ] `PrtSc` screenshot → paste into TinyMCE → built-in pipeline stores it.


ignaciogros (commented):
Thanks for checking this issue, @erseco, but I've tested it again in `main` and now I realize that there's no need to change the current behavior. You can paste an image, and it's properly saved as an asset, just disabling the "Paste as Plain Text" option. I think it's just the expected behavior.

@cristinavaldera:

Please close this issue if you agree.

Thanks.

erseco:
There are more use cases that this PR fixes, copying from website and also copying from other apps like a text processor. Please check that changing the param you said also fixes it

ignaciogros:
I can paste from an iDevice, from another iDevice, from a website and even from Word. In my tests, it worked as expected in all cases, except when pasting a complete `.docx` document. But it that case, it had the same behavior in this branch and `main`.

That param in the TinyMCE configuration is intentional: paste as text should be the default paste behavior. That was decided to avoid unnecessary code in the content, because pasting from Word or websites is quite common, and the recommendation is to add format using TinyMCE's options instead.

PR #1739, opened 2026-04-29
fix(import): Keep indentation inside <pre> tags when parsing legacy .elp files

lrlopez:
This PR skips preprocessing content inside ```<pre>``` blocks inside text areas when parsing legacy .elp files as the current implementation corrupts indentation.

The issue is rather visible when importing files that contain source code text.

# Why
```<pre>``` blocks should keep spacing and indentation as-is when displaying their content. Current legacy parsing takes the imported content and does this:
```js
xml = xml.replace(/ {5}/g, '');
xml = xml.replace(/\t/g, '');
```
The intent of those lines is to fix encoding issues from eXe 2.x exports. The problem is that these replacements affect everything, including the content inside ```<pre>``` blocks, so they unintendedly remove tabs and 5 spaces groups which should be kept.

# What changed
To fix the problem we introduce two changes:

1. Before applying those replacement rules, we replace those ```<pre>``` blocks encoded into the ```value``` attribute of the XML ```<unicode>``` tag into a marker ```__LEGACY_PRE_BLOCK_id__```, where ```id``` is a incrementing counter.
2. After applying the existing preprocessing, we undo the markers replacement by reinstating the original content.

# Before
image

# After
image

### DIFF
--- a/src/shared/import/LegacyXmlParser.spec.ts
+++ b/src/shared/import/LegacyXmlParser.spec.ts
@@ -164,6 +164,13 @@ describe('LegacyXmlParser', () => {
             const result = parser.preprocessLegacyXml(input);
             expect(result).toContain('&#10;');
         });
+
+        it('should keep encoded pre content inside unicode value attribute', () => {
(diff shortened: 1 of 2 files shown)


ignaciogros (approved):
Thank you very much, @lrlopez!

PR #1740, opened 2026-04-29
fix(platform): generate ode package in the browser before forwarding to Moodle

erseco:
## Summary

Fixes the missing-images symptom reported on the Moodle plugins:

- exelearning/exelearning#2132
- exelearning/exelearning#2094

When a user finished editing on eXeLearning Online and clicked **Finish** to publish back to Moodle, the resulting webzip / SCORM package on the platform side did not contain the images referenced by the HTML pages. Downloading the same project as a regular SCORM (browser-side export) included `content/resources/<filename>` correctly; only the *Finish to Moodle* path produced an HTML referencing a literal `content/resources/<uuid>.<ext>` URL with no matching file in the ZIP, yielding 404s on the Moodle activity.

## Root cause

The browser already holds the source of truth for online editing sessions: the live Yjs document and the IndexedDB asset blobs uploaded during the session. The previous **Finish to Moodle** path, however, ran on the server (`platformPetitionSet`):

1. Fetched a Yjs snapshot from the `yjs_documents` table -- which can be stale or missing entirely (the user's report shows `/api/projects/uuid/.../yjs-document` returning 404 in their staging environment).
2. Read assets via `DatabaseAssetProvider`, which only sees what the server has persisted on disk -- not the IndexedDB blobs the user just uploaded.

Whenever the asset upload had not finished syncing to the server (or the project was created in a non-Yjs flow), `addFilenamesToAssetUrls` would fall back to a literal `content/resources/<uuid><ext>` URL while `addAssetsToZipWithResourcePath` skipped the asset entirely. Result: the ZIP shipped **with no `content/resources/` folder at all**.

User-provided ZIPs comparing the two flows (single image, original filename `400x200.png`, internal UUID `9dacde75-...`):

| Aspect | Direct SCORM download (works) | Finish to Moodle (broken) |
|---|---|---|
| `content.xml` reference | `asset://9dacde75-...png` | `asset://9dacde75-...png` |
| HTML reference | `content/resources/400x200.png` | `content/resources/9dacde75-...png` |
| File in ZIP | `content/resources/400x200.png` :white_check_mark: | **No `content/resources/` folder at all** :x: |

## Fix

The package is now generated **in the browser** -- where the data already lives -- and the server is reduced to a thin JWT-validating proxy that forwards the binary to Moodle's `set_ode.php`.

### Client (`public/app/workarea/menus/navbar/items/navbarFile.js`)

`uploadPlatformEvent` now:
1. Decodes the platform JWT to read `pkgtype` (validation still happens server-side).
2. Calls `SharedExporters.quickExport(format, documentManager, assetCache, resourceFetcher, {}, assetManager)` -- the same path the in-browser SCORM/HTML5 download already uses, which produces a correct package.
3. POSTs the resulting blob as multipart to a new server endpoint, along with `jwt_token` and `projectUuid`.
4. Falls back to the legacy `postFirstTypePlatformIntegrationElpUpload` server-side path when SharedExporters or the Yjs bridge are unavailable.

### Server

* `src/services/platform-integration.ts` -- new `platformPetitionSetForward(payload, jwtToken, projectUuid, packageData, filename)` helper. Base64-encodes the buffer, builds the standard `ode_data` JSON, POSTs to Moodle's `set_ode.php`, and links `project.platform_id` only on success.
* `src/routes/platform-integration.ts` -- new `POST /api/platform/integration/set_platform_new_ode_browser` route. Validates the JWT, reads the multipart `package` File, and delegates to `platformPetitionSetForward`.
* `src/routes/api-routes.ts` -- registers the new endpoint name.

### Defensive change to the existing exporter

`src/shared/export/exporters/BaseExporter.ts` -- `addAssetsToZipWithResourcePath` also writes each asset under its `<assetId><ext>` fallback path, mirroring the literal URL produced by `addFilenamesToAssetUrls` when an `asset://` reference cannot be resolved against the export path map. This protects the legacy server-side path as well: even if a stale reference slips through, Moodle's `pluginfile.php` can still resolve the URL instead of returning 404. A warning is logged for every unresolved reference so future upstream mismatches surface in diagnostics.

## How to verify

1. Open eXeLearning Online from a Moodle activity (mod_exeweb or mod_exescorm).
2. Insert an image whose original filename is something other than a UUID (`x.png`).
3. Click **Finish** to publish back to Moodle.
4. Open the activity in Moodle and check the network tab: the image URL should be `content/resources/x.png` and return 200, no longer `content/resources/<uuid>.png` returning 404.

## Tests

* `services/platform-integration.spec.ts` -- 7 new unit tests for `platformPetitionSetForward`: URL derivation, missing payload, base64 encoding, default filename, platform error responses, transport errors, and the `platform_id` linkage skip when `projectUuid` is empty.
* `navbarFile.test.js` -- 5 new frontend tests covering the new client-side flow, error surfacing on failed exports, the legacy fallback when SharedExporters is unavailable, and the JWT payload decoder.
* `BaseExporter.spec.ts` -- 3 new regression tests for the `<assetId><ext>` fallback path in `addAssetsToZipWithResourcePath`.

Coverage:

- [x] `bun run lint:src` -- no errors.
- [x] `bun run lint:public` -- no errors.
- [x] `bun test src/shared/export/` -- 1834 pass.
- [x] `bun test src/services/platform-integration.spec.ts src/routes/platform-integration.spec.ts` -- 61 pass.
- [x] `bun x vitest run public/app/workarea/menus/navbar/items/navbarFile.test.js` -- 197 pass.

Refs: exelearning/exelearning#2132, exelearning/exelearning#2094


erseco:
## Reproduction confirmed against real packages

User-provided ZIPs comparing the two flows on the same project (single image asset, original filename `400x200.png`, internal UUID `9dacde75-8cda-922f-348b-555ad92e70ff`):

| Aspect | Direct SCORM download (works) | Finish to Moodle (broken) |
|---|---|---|
| `content.xml` reference | `asset://9dacde75-8cda-922f-348b-555ad92e70ff.png` | `asset://9dacde75-8cda-922f-348b-555ad92e70ff.png` |
| HTML reference | `content/resources/400x200.png` | `content/resources/9dacde75-8cda-922f-348b-555ad92e70ff.png` |
| File in ZIP | `content/resources/400x200.png` :white_check_mark: | **No `content/resources/` folder at all** :x: |

The source data (`asset://` reference in `content.xml`) is identical between the two exports. The divergence happens during package assembly:

- The browser export resolves the UUID via `BrowserAssetProvider` (IndexedDB) and writes both the friendly-named file and the matching HTML reference.
- The server-side `platformPetitionSet` calls `DatabaseAssetProvider`, which returns no asset for that UUID (the upload only reached IndexedDB), so the export path map is empty for that asset. `addFilenamesToAssetUrls` falls back to a literal `<uuid><ext>` URL, while `addAssetsToZipWithResourcePath` skips the asset entirely.

This PR addresses both halves of that divergence:

1. Client-side: `uploadPlatformEvent` now flushes pending assets via `assetManager.uploadPendingAssets()` before triggering the server-side export, so the database has the asset by the time `platformPetitionSet` runs.
2. Server-side: `addAssetsToZipWithResourcePath` writes each asset under both the resolved path and a `<assetId><ext>` fallback path, so even if a stale reference slips through, the URL still resolves on the platform.

ignaciogros (dismissed):
When I click on "Finish" or "Send to Moodle" I see this message:

Image

ignaciogros (changes requested):
I tested with the same Moodle and eXeLearning online versions, and now I see a different message:

> Failed to forward package to platform: Unable to connect. Is the computer able to access the url?

But it doesn't go back to Moodle.

**There's also another problem:** 

I'm editing a previously published content, but when I click on "Edit" and go to eXe, that content's not loaded. I just see an empty content.

That happens both in `exeweb` and `exescorm`. It doesn't matter if you're previously logged in or not.

Thanks, @erseco.

ignaciogros:
@erseco,

> I'm editing a previously published content, but when I click on "Edit" and go to eXe, that content's not loaded. I just see an empty content.

I'm not sure about the behavior of the previous version. Maybe we should check that again before trying a different approach.

Thanks.

ignaciogros (approved):
Thanks, @erseco.

There are still some details to fix:

- Untranslated "Sending to platform" message.
- No "Exit without saving" when editing a platform package in the Team version.
- Clicking more than once in "Finish" triggers the publish event each time you click on it, and that provokes an error in the platform.
- Other usability issues.

We'll fix those problems in separated issues as soon as possible.

PR #1744, opened 2026-04-30
feat(file): advanced open/save folder mode for unpacked .elpx projects

erseco:
## Summary

Adds an advanced workflow for developers, integrators, QA and power users: open an eXeLearning project directly from an unpacked folder, and save it back to a folder, without packing/unpacking an `.elpx` archive in between. The folder layout is identical to the contents of an `.elpx`, so the same project can move between the two formats without data loss. **The default `.elpx` workflow is unchanged.**

## Why

Folder-backed projects make eXeLearning content easier to:

- track in Git
- review in pull requests
- inspect during debugging
- compare across versions
- manipulate with external tools
- integrate into automated workflows
- avoid repeated `.elpx` unpack/repack cycles during development

This is intended as an **advanced workflow** — the regular `.elpx` flow remains the recommended path for most users.

## UX

Two new entries appear in the offline File menu:

- **Open from folder**
- **Save to folder**

They are **hidden by default** (`d-none`) and revealed only when:

- the user holds **Alt**, **or**
- the runtime feature flag `window.eXeLearning.config.enableFolderProjects` is set (e.g. for development builds).

If neither runtime backing is available (Electron IPC bridge or `showDirectoryPicker`) **and** the flag is off, the entries stay hidden — no broken-feature menu items.

### New module — `public/app/core/ProjectFolderStorage.js`

A framework-free storage abstraction. Conversion happens in memory:

```
folder ↔ Map<relativePath, Uint8Array> ↔ zip Uint8Array ↔ File("project.elpx")
```

The existing `importFromElpxViaYjs(file)` / `exportToElpxViaYjs()` pipeline accepts a `File` / produces a `Uint8Array`, so this module plugs in at that boundary and never duplicates import/export logic.

Public surface:

- `entriesToZip(entries, deps?)` / `zipToEntries(zipBytes, deps?)` — pack and unpack via `fflate`
- `validateProjectEntries(entries)` — rejects folders without `content.xml` / `contentv3.xml`
- `zipToElpxFile(zipBytes, name?)` — wraps a zip as a `File` for the existing import flow
- `isPathSafe(relativePath)` — security guard against `..`, absolute paths, drive letters, NUL bytes
- `supportsFileSystemAccess()` — feature detection
- `openProjectFolderInBrowser()` / `saveProjectFolderInBrowser(zipBytes)` — File System Access API helpers
- `readDirectoryHandle(dir)` / `writeDirectoryHandle(dir, entries)` — recursive helpers

### Electron — `app/main.js` + `app/preload.js`

- New IPC handler `app:openProjectFolder` walks the selected directory, refuses symlinks and paths that escape the root, packs entries with `fflate`, and returns base64 + `suggestedName` to the renderer.
- Save reuses the existing `app:exportBufferToFolder` handler (already implemented). Exposed under a distinct preload name (`saveProjectFolder`) so the renderer can offer it as a separate menu entry without touching `.elpx` save behaviour.
- Folder validity (`content.xml` / `contentv3.xml`) is checked in main so a native error dialog can surface invalid choices.

### Browser (static) path

When Electron is not available, the same flows go through `window.showDirectoryPicker`:

- **Open** — picks a directory, recursively reads it into a `Map`, validates, packs to a zip, wraps as a `File`, hands it to `eXeLearning.app.modals.openuserodefiles.largeFilesUpload(file)`.
- **Save** — runs `SharedExporters.quickExport('ELPX')`, unpacks the resulting zip in memory, writes each entry to the chosen `FileSystemDirectoryHandle`.

### Renderer wiring — `public/app/workarea/menus/navbar/items/navbarFile.js`

- `setFolderProjectActionsEvents()` registers Alt-key visibility, feature-flag override, and the two click handlers. Skips registration entirely when no backing runtime is available.
- `openProjectFromFolderEvent()` — Electron path first, browser path as fallback.
- `saveProjectToFolderEvent()` — Electron path first, browser path as fallback. Marks the document clean on success.

### Template — `views/workarea/menus/menuNavbar.njk`

Two new `<li class=\"d-none exe-folder-project-action\">` entries injected into the offline File-menu block (next to Open/Save).

## Security

- All folder entries are validated with `isPathSafe()` on both the read and write sides:
  - rejects absolute POSIX paths (`/foo`)
  - rejects Windows drive-letter paths (`C:\\foo`)
  - rejects UNC paths (`\\\\server\\share`)
  - rejects parent-directory traversal (`..`) at any depth
  - rejects empty strings and NUL bytes
- The Electron walker refuses to follow symlinks and double-checks that resolved absolute paths stay under the user-selected root.
- `.DS_Store` and `Thumbs.db` are skipped on save so OS noise does not bleed into the project.
- All permission prompts are user-initiated (`showDirectoryPicker`, native dialog).

## Compatibility

- Existing `.elpx` open/save flow is **untouched**.
- Existing exports are **unchanged**.
- Static/offline behaviour does not regress — entries simply stay hidden when the backing runtime is absent.
- No collaboration changes.

## Tests

- `public/app/core/ProjectFolderStorage.test.js` — **26 new tests** covering:
  - path safety (well-formed, empty, absolute, drive letters, traversal, NUL bytes, normalization)
  - zip ↔ entries round-tripping with real `fflate`
  - rejection of unsafe entries on both pack and unpack
  - `validateProjectEntries` modern/legacy/empty/no-content
  - `File` wrapper extension handling
  - File System Access helpers via an in-memory `FakeDirectoryHandle`:
    - recursive read
    - skipping unsafe entry names
    - recursive write with auto-created subfolders
    - `openProjectFolderInBrowser` happy path + invalid folder rejection
    - `saveProjectFolderInBrowser` round-trip
    - clear errors when the API is missing
- Existing `navbarFile.test.js` (200 tests) **still passes** unchanged.
- Wider sweep across `public/app/core/`, `public/app/workarea/menus/navbar/`, `public/app/yjs/` → **3628 / 3628 tests pass**.
- Biome lint clean (no fixes applied).
- `node -c` syntax check passes for `app/main.js` and `app/preload.js`.

## Test plan

- [ ] Run `make fix && make test-unit` — all green
- [ ] In the desktop build, hold **Alt** with the File menu open → **Open from folder** / **Save to folder** appear
- [ ] **Save to folder** writes the same set of files an `.elpx` would contain
- [ ] **Open from folder** on the folder produced by Save loads the project identically to opening the original `.elpx`
- [ ] In a Chromium/Edge static build, the same entries appear under Alt and use the File System Access API picker
- [ ] In Firefox or Safari (no File System Access API), the menu entries stay hidden when the feature flag is off
- [ ] A folder without `content.xml` / `contentv3.xml` shows the "not a valid eXeLearning project" error


pabloamayab:
Hi @erseco good idea!!!! 
It works fine, although there is a delay when loading/saving resources that doesn't occur when you open/save them the traditional way

I tested it with a 90MB resource in the Electron app

cristinavaldera (approved):

pabloamayab (approved):
LGTM!

PR #1748, opened 2026-05-05
fix(dev): align local env setup and upload-session validation

andresito87:
## Summary

This branch fixes the local setup problems I hit while preparing my contribution and includes the websocket validation work that surfaced while getting the change through the repository quality gates.

- switch the local Docker image from Alpine to Debian/glibc and freeze Bun installs so native packages such as `@napi-rs/canvas` resolve consistently
- separate local development and local production flows with `.env.development`, `make up-local`, `make up-production`, and explicit Docker `.env.dist` + `.env` precedence
- tighten `asset-coordinator` upload-session validation by covering the missing edge cases, removing an unused helper, and fixing the project user reference to use `Project.owner_id`

## Why

I could not run the project reliably in Docker with the previous image/runtime combination because native canvas-related dependencies were failing to resolve correctly. While aligning the local dev/prod environment story, the validation pass also exposed two websocket issues that were directly blocking the repo gates for this branch:

1. `src/websocket/asset-coordinator.ts` was still under the coverage threshold required by `make test-unit`.
2. The upload-session code was reading `project.user_id`, but `Project` does not expose that property; the correct field is `owner_id`.

Because those websocket fixes were discovered while validating this exact branch, I kept them in the same PR instead of leaving the branch red.

## Scope and rationale

Most of the branch is about local Docker/dev-prod setup. The websocket changes are intentionally limited to the minimum needed to make the affected upload-session path type-correct and fully covered under the existing repository rules.

## Changes

| File | Change |
|------|--------|
| `.gitignore` | Ignore `.env.development` while allowing the development example file to be committed. |
| `.env.development.example` | Add the local development environment template used by `make up-local`. |
| `Dockerfile` | Move to Bun Debian base image, install runtime packages with `apt`, and use `bun install --frozen-lockfile` for reproducible native dependency installs. |
| `docker-compose.yml` | Document local Docker env precedence, add explicit build args/image name, keep Apple Silicon fallback note, and clarify demo credentials/comments. |
| `Makefile` | Add `check-env-development`, keep `up-local`/`up-production` behavior explicit, and restore the related helper targets consistently. |
| `package.json` | Run `start:local` with `.env.development`. |
| `README.md` | Document the updated local commands and Docker env behavior. |
| `doc/development/environment.md` | Document which env file each local mode uses. |
| `src/websocket/asset-coordinator.ts` | Remove an unused broadcast helper and use `project.owner_id` when creating upload sessions. |
| `src/websocket/asset-coordinator.spec.ts` | Cover upload-session and collaboration edge cases that were leaving `asset-coordinator` under the coverage gate. |
| `bun.lock` | Record the dependency resolution used by the Debian/native runtime setup. |

## Validation

- [x] `make fix`
- [x] `make test-unit`
- [x] `make test-integration`
- [ ] `make test-e2e`
  - One persistent failure remains in `test/e2e/playwright/specs/idevices/image-gallery.spec.ts` (`should load images from different folder depths in preview`), which appears unrelated to the Docker/dev-prod and upload-session scope of this PR.

## Notes

- I did not find an existing upstream issue in `exelearning/exelearning` covering this Docker/native-runtime + env-separation work, so this draft PR documents the context directly.
- If maintainers want a strict all-green E2E gate before review, I can investigate the remaining `image-gallery` failure in a follow-up focused on that spec.

### DIFF
--- a/.env.development.example
+++ b/.env.development.example
@@ -1,0 +1,17 @@
+# Local development environment for `make up-local`
+
+APP_ENV=dev
+APP_DEBUG=1
+APP_ONLINE_MODE=1
+
+APP_PORT=8080
+PORT=8080
+
+FILES_DIR=data/
+
+DB_DRIVER=pdo_sqlite
+DB_PATH=data/exelearning.db
+DB_SERVER_VERSION=3.32
+
+TEST_USER_EMAIL=user@exelearning.net
+TEST_USER_PASSWORD=1234
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,9 @@
 /.env.local
 /.env.local.php
 /.env.*.local
+.env.development
+!.env.dist
+!.env.development.example
 /config/secrets/prod/prod.decrypt.private.php
 /public/bundles/
 /var/*
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,24 +1,31 @@
 # syntax=docker/dockerfile:1
+
 ARG BUN_VERSION=1.3
 ARG VERSION=v0.0.0-alpha
 
 ################################################################################
 # Base image for reuse
 ################################################################################
-FROM oven/bun:${BUN_VERSION}-alpine AS base
+# Debian/glibc avoids native dependency issues with packages that expect *-gnu
+# binaries, such as @napi-rs/canvas on some platforms.
+FROM oven/bun:${BUN_VERSION}-debian AS base
+
 WORKDIR /app
 
 ################################################################################
-# Dependencies stage - separated for cache efficiency
+# Dependencies stage
 ################################################################################
 FROM base AS deps
+
 COPY package.json bun.lock ./
-RUN bun install
+
+RUN bun install --frozen-lockfile
 
 ################################################################################
-# Build stage - Compile TypeScript and assets
+# Build stage
 ################################################################################
 FROM deps AS builder
+
 COPY tsconfig.json ./
 COPY src/ ./src/
 COPY scripts/ ./scripts/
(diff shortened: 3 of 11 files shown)


erseco:
Thanks for the contribution and for taking the time to look into this.

I’m a bit concerned about the scope of this PR, especially as a first contribution to the project. Switching the Docker base image from Alpine to Debian is a significant runtime change, and right now it does not seem clearly justified by the issue being addressed.

Could you explain why moving away from Alpine is required here, and what specific problem it fixes? Ideally, this should include the exact failure you found with the current image and why it cannot be solved with a smaller change.

The `owner_id` fix may make sense as a focused change, but the Docker/runtime changes should probably be discussed and reviewed separately.

andresito87:
Thanks, that is a fair question.

In my case this was not a theoretical change, Alpine was the image that failed during the Docker build. The failure happened inside `bun install`, in the native dependency chain around `@napi-rs/canvas`, with this error:

`Fail extracting tarball from @napi-rs/canvas-linux-arm64-gnu`

Screenshot 2026-05-05 at 10 39 10

After switching the Docker image to the Bun Debian variant, that same build path completed consistently.

So the reason for proposing Debian was simple: Alpine was not working reliably in the Docker setup I was using, while Debian did.

That said, I understand your concern about changing the project runtime base in a broader way. If you prefer, I can narrow this PR and leave the Debian/Alpine runtime discussion for a separate issue or focused PR with the exact reproduction details and logs.

erseco:
Thanks for the clarification and for the willingness to help.

I understand that Alpine failed in your local Docker setup and that switching to Debian made the build work for you. That said, changing the project Docker base image from Alpine to Debian is a significant runtime change, and I do not think we should merge it as part of this PR without a prior discussion and a more focused scope.

For this kind of change, it would be better to open an issue first with the exact reproduction steps, platform details, logs, and possible alternatives. Then we can discuss whether changing the base image is the right solution, or whether there is a smaller fix that keeps the current runtime.

The `owner_id` fix may make sense as a separate, focused PR, but this PR currently combines several unrelated changes: Docker runtime, environment handling, documentation, Makefile targets, lockfile changes, and websocket tests/fixes.

I’m going to close this PR for now because the scope is too broad and the main Docker/runtime change needs discussion before being proposed as an implementation.

Thanks again for the contribution. If you want to continue with this, please open focused issues first, and then smaller PRs linked to those issues.

PR #1750, opened 2026-05-05
fix(http): encode non-ASCII filenames in Content-Disposition (#1749)

### LINKED ISSUE #1749: Asset download fails with 500 when filename contains non-ASCII characters (Content-Disposition header)
## Summary

`GET /api/projects/:projectId/assets/:assetId` (and `/assets/by-client-id/:id`) returns **HTTP 500** when the stored asset filename contains non-ASCII characters (e.g. `á`, `ñ`, `ó`, `ú`). The browser cannot fetch the asset, so images/files referenced from a project fail to load.

## Environment

- Online dev environment (Docker, MariaDB 11.8.6)
- Backend: Bun + Elysia
- Affected route handlers: `src/routes/assets.ts:623` and `src/routes/assets.ts:676`
- Likely also affected: `src/routes/api/v1/assets.ts:313`, `src/routes/admin-templates.ts:230`, `src/routes/admin.ts:1277`, `src/routes/convert.ts:365,458`, `src/routes/config.ts:369`, `src/routes/api/v1/export.ts:278,288`

## Error from server logs

```
[ERROR] GET /api/projects/264f2889-.../assets/9d716adb-... - 500: Header '17' has invalid value: 'attachment; filename="San Marcial de Rubicón.png"'
TypeError: Header '17' has invalid value: 'attachment; filename="San Marcial de Rubicón.png"'
    at Response (unknown)
    at mapResponse2 (/app/dist/index.js:125768:28)
    at handle (file:///app/dist/index.js:9:19)
```

Other failing filenames seen in the same session:

- `40. Actividad 1 - Matemáticas- Basico.jpg`
- `San Marcial de Rubicón.png`
- `Cenobio de Valerón.png`
- `El Julán.png`
- `Cercanía a pastos.jpg`
- `cueva bencomo santa úrsula.jpg`
- `Chasogo molinos de mano cañadas teide.png`

## Root cause

HTTP header values must be ISO-8859-1 / ASCII-safe. Bun's `Response` constructor (used internally by Elysia's `mapResponse2`) validates header values and rejects any byte outside that range, throwing the `TypeError: Header '17' has invalid value: ...` shown above.

The asset routes build the header by interpolating the raw filename:

```ts
// src/routes/assets.ts:623 and :676
set.headers['content-disposition'] = `attachment; filename=\"${asset.filename}\"`;
```

When `asset.filename` contains UTF-8 characters this produces an invalid header value and the request fails with 500.

## Fix

The export route already implements the correct RFC 5987 pattern (`src/routes/export.ts:662-672`):

```ts
const safeFilename = rawFilename.replace(/[^\x20-\x7E]/g, '_');
const encodedFilename = encodeURIComponent(rawFilename)
    .replace(/['()]/g, escape)
    .replace(/\*/g, '%2A');

set.headers['content-disposition'] =
    `attachment; filename=\"${safeFilename}\"; filename*=UTF-8''${encodedFilename}`;
```

This needs to be applied to every place that sets `content-disposition` from a user-controlled filename. Suggested approach:

1. Extract a shared helper (e.g. `buildContentDisposition(filename: string)`) — single source of truth, per the project's no-duplication rule.
2. Replace all current call sites listed above with the helper.
3. Likewise sanitise `x-filename` and other custom headers in `src/routes/assets.ts` (lines 626-628) — they will hit the same validator.
4. Add a unit test with a non-ASCII filename for the assets route, and an E2E spec that imports an `.elp` containing an image with accented filename and verifies it loads in the preview.

## Reproduction

1. Create or import a project containing an asset whose filename has non-ASCII characters (e.g. `San Marcial de Rubicón.png`).
2. Open the project in the workarea so the client requests the asset.
3. Server logs show `TypeError: Header '17' has invalid value`, the request returns 500, and the asset fails to load in the UI.

## Side observation (separate issue)

The same logs show `php: not found` during the docker-entrypoint post-configure step:

```
/app/docker-entrypoint.sh: eval: line 97: php: not found
[entrypoint] Some post-configure commands failed
```

The Bun-based image no longer ships PHP, so these legacy post-configure commands should be removed or guarded. Worth tracking as a separate ticket.

erseco:
## Summary

Closes #1749.

`GET /api/projects/:projectId/assets/:assetId` (and friends) returned **HTTP 500** whenever an asset filename contained accented characters such as `San Marcial de Rubicón.png`. Bun's `Response` constructor rejects header values with bytes outside printable ASCII and throws `TypeError: Header '17' has invalid value: 'attachment; filename="…"'`, so the asset never streamed to the browser.

## What changed

- New helper `src/shared/http/headers.ts` (single source of truth):
  - `buildContentDisposition(filename, disposition?)` emits the RFC 5987 pair `filename="…ascii…"; filename*=UTF-8''…percent-encoded…`, strips control characters from the legacy form, and replaces `"` with `'` to keep the quoted-string valid.
  - `encodeHeaderValue(value)` percent-encodes only non-ASCII inputs (passthrough for plain ASCII).
- Replaced every duplicated inline `attachment; filename="…"` site with the helper:
  - `src/routes/assets.ts` (the two endpoints in the bug report, plus the `X-Filename` / `X-Folder-Path` metadata headers used by the WebSocket prefetch).
  - `src/routes/api/v1/assets.ts`, `admin-templates.ts`, `admin.ts`, `convert.ts` (×2), `config.ts`, `api/v1/export.ts` (×2).
  - `src/routes/export.ts` — folded the previously inline RFC 5987 implementation into the helper.
- `public/app/yjs/AssetManager.js` and `AssetWebSocketHandler.js` now `decodeURIComponent` the `X-Filename` / `X-Folder-Path` values they receive, with a defensive `try/catch` that falls back to the raw value if the input is malformed.

## Tests

- `src/shared/http/headers.spec.ts` — 13 unit tests for the helpers, including a guard that the value is accepted by Bun's `Response` constructor (the exact regression from #1749) and that CR/LF cannot smuggle extra headers.
- `src/routes/assets.spec.ts` — two regression tests exercising the `/assets/:id` and `/assets/by-client-id/:id` paths with filenames containing accents.

```
$ bun test src/shared/http/headers.spec.ts src/routes/assets.spec.ts
74 pass / 0 fail / 174 expect() calls
$ make test-unit
6915 pass / 0 fail across 190 files
```

(The `make test-unit` exit code reports a pre-existing coverage gate on `src/websocket/asset-coordinator.ts` at 88.20%, unrelated to this PR — verified on `main` at the same level.)

## Test plan

- [x] Unit tests for the helper functions
- [x] Regression tests for `/assets/:id` and `/assets/by-client-id/:id` with non-ASCII filenames
- [x] Full backend + frontend unit suite (`make test-unit`)
- [x] `make fix` passes
- [ ] Manual verification on the dev environment by importing a project with assets named `San Marcial de Rubicón.png` and confirming they render without 500s in the network tab


pabloamayab (approved):
I think I ran the right tests, but I'm not entirely sure.
I started a Docker instance locally with `docker compose up --build -d` and then ran the logs with `docker compose logs -f`.
Although I think that even if I hadn't started Docker in daemon mode, it would show the same log output in the terminal.
I opened eXe, loaded images with accents in their names, and they loaded and previewed correctly, with no errors appearing in the logs.

PR #1767, opened 2026-05-06
fix(save): generate screenshot before encoding Yjs state on first save

### LINKED ISSUE #1766: screenshot.png missing from .elpx on first save (saved Yjs state lacks pre-rendered screenshot)
## Problem

The very first save of a brand-new project produces an `.elpx` whose archive **does not include `screenshot.png`** at the root, even though `ElpxExporter` is fully wired to embed it. Subsequent saves include it, but the first one is always missing — and any consumer that reads the very first export (Omeka S import, ODE catalog ingest, third-party tooling) gets no preview.

This is a pure timing/wiring bug, not a missing feature: every piece of the screenshot pipeline already exists.

## Why this matters / context

The companion Omeka S module ships a feature ([exelearning/exelearning#2053](https://github.com/exelearning/exelearning/issues/2053), PR [#15](https://github.com/exelearning/omeka-s-exelearning/pull/15)) that uses the bundled `screenshot.png` as the real Omeka thumbnail. That feature is only useful if the editor reliably ships `screenshot.png` inside the saved/exported `.elpx` — including the very first save. Right now it works only after the second save, which is a confusing UX.

## Reproduction

1. Open the eXeLearning editor on a fresh project.
2. Add some content.
3. Click **Save** once.
4. Trigger an `.elpx` export (download, or "Save to Omeka").
5. Open the resulting `.elpx` as a ZIP — `screenshot.png` is missing.
6. Save again, re-export — now `screenshot.png` is present.

## Root cause

Walking the flow in `public/app/yjs/SaveManager.js#save()`:

1. **Line 1176** — `encodedYjsState = window.Y.encodeStateAsUpdate(documentManager.ydoc)` snapshots the Yjs CRDT for upload. At this point, on a first save, `metadata.get('screenshot')` is empty.
2. **Step 1 (line 1210)** uploads that snapshot to the server.
3. **Steps 2–3** finish the rest of the save.
4. **Line 1287** — *after* the save resolves, `bridge.generateScreenshotFromFirstPage()` runs **non-blocking**. It renders the first page in a hidden iframe via `html2canvas`, then sets `metadata.set('screenshot', dataUrl)` in the **local** Yjs document.

So the screenshot only enters the CRDT *after* the encoded state has already left the client. The next save (or autosave) ships it; the first save never can.

`src/shared/export/exporters/ElpxExporter.ts` (lines 365–387) does its job perfectly: if `meta.screenshot` is set, it embeds `screenshot.png`. The fallback `options.generateScreenshot` hook is also defined, but `src/routes/export.ts` (line 573) builds `exportOptionsWithHooks` with only `preRenderLatex` / `preRenderDataGameLatex` — it does **not** wire `generateScreenshot`, and it cannot easily, because server-side has no DOM/canvas (`html2canvas` is browser-only).

So neither the saved Yjs state nor the server-side export pipeline can produce the screenshot on a first save — the only place that can render it is the browser, and the browser does it too late.

Confirmation:
- `public/app/yjs/YjsProjectBridge.js#generateScreenshotFromFirstPage()` (line 2924) does the rendering and is gated by `if (metadata.get('screenshot')) return;` — fast no-op once a screenshot exists, so calling it earlier costs nothing on subsequent saves.
- `SaveManager.test.js:580` already verifies the post-save call ("calls generateScreenshotFromFirstPage on bridge after successful save") — confirming the post-save timing is the intended behavior, not a bug we missed.

## Proposed fix

Generate the screenshot **before** encoding the Yjs state for upload, with a generous timeout fallback so a slow render never wedges save:

1. In `SaveManager.save()`, after `metadata` is available and before `encodeStateAsUpdate` (around line 1170), call `bridge.generateScreenshotFromFirstPage()` with a `Promise.race` timeout (e.g. 5–8s).
2. Keep the existing post-save call as a recovery fallback in case the pre-save attempt fails or times out — it'll catch up on the *next* save.
3. The existing `metadata.get('screenshot')` short-circuit inside `generateScreenshotFromFirstPage` makes pre-save calls free on subsequent saves.

Impact:
- First save costs ~the html2canvas render time (~hundreds of ms on a typical project) — acceptable for the once-per-project cost.
- Every save produces an `.elpx` with `screenshot.png`.
- No server-side changes needed; no new dependencies.

## Out of scope

- Refreshing the screenshot when content changes substantially (current behavior keeps the first generated screenshot forever; that's a separate UX issue).
- Headless server-side screenshot generation (would require puppeteer/playwright; not needed for this fix).

## Acceptance criteria

- [ ] First save of a brand-new project produces an `.elpx` whose root contains a valid `screenshot.png`.
- [ ] No regression in save latency on subsequent saves (pre-save call short-circuits).
- [ ] Save still succeeds when `html2canvas` or the iframe render times out.
- [ ] Existing `SaveManager.test.js` post-save assertion still passes (post-save call kept as fallback).
- [ ] New unit test covers the pre-save generation path.

## References

- `public/app/yjs/SaveManager.js` lines 1176, 1210, 1286–1291
- `public/app/yjs/YjsProjectBridge.js` lines 2924–3055
- `src/shared/export/exporters/ElpxExporter.ts` lines 365–387
- `src/routes/export.ts` lines 570–580
- Companion: exelearning/exelearning#2053, exelearning/omeka-s-exelearning#15

erseco:
## Summary

Closes #1766.

The very first save of a new project produced an `.elpx` without `screenshot.png` at the root because `bridge.generateScreenshotFromFirstPage()` ran *after* the Yjs snapshot was already encoded and uploaded. Subsequent saves shipped the screenshot, but the first export — including the first round-trip through the Omeka S module ([exelearning/exelearning#2053](https://github.com/exelearning/exelearning/issues/2053), PR [#15](https://github.com/exelearning/omeka-s-exelearning/pull/15)) — never had a thumbnail.

## Fix

`SaveManager.js`:

- New `ensureScreenshotBeforeSave(metadata, timeoutMs = 8000)` helper.
- Called *before* `Y.encodeStateAsUpdate(...)` in `save()`, so the persisted CRDT snapshot already carries the screenshot.
- Awaits `bridge.generateScreenshotFromFirstPage()` wrapped in a `Promise.race` against an 8 s timeout — a slow `html2canvas` render can never wedge save.
- The bridge method already short-circuits when `metadata.get('screenshot')` is set, so subsequent saves pay zero cost.
- Existing post-save call retained as a recovery fallback for the cases where the pre-save attempt fails or times out — the next save catches up.

## Tests

`SaveManager.test.js` — adds five tests, keeps the existing post-save assertion:

- ordering: pre-save screenshot precedes `saveYjsState` upload.
- helper short-circuits when `metadata.screenshot` already exists.
- helper is a no-op when the bridge has no generator.
- save still resolves successfully when the pre-save generator rejects.
- helper returns within its timeout when the generator hangs.

Result: **150/150 passing** in `vitest run public/app/yjs/SaveManager.test.js`. `bun run lint:public` and `bun run lint:test` clean.

## Test plan

- [x] Open the editor on a brand-new project, add content, click Save once.
- [x] Trigger an `.elpx` export (download or "Save to Omeka").
- [x] Open the resulting archive — `screenshot.png` is at the root.
- [x] Save a second time — no measurable extra latency vs. before (helper short-circuits).
- [ ] Force a render error in `bridge.generateScreenshotFromFirstPage` — save still succeeds; existing post-save fallback runs on the next save.
- [x] Companion: confirm the Omeka S module shows the screenshot as the media thumbnail on the first round-trip (issue exelearning/exelearning#2053, PR omeka-s-exelearning#15).

## References

- `public/app/yjs/SaveManager.js` — new `ensureScreenshotBeforeSave`, called before encode.
- `public/app/yjs/YjsProjectBridge.js` lines 2924–3055 — unchanged renderer.
- `src/shared/export/exporters/ElpxExporter.ts` lines 365–387 — unchanged consumer of `meta.screenshot`.


cristinavaldera:
It works right but only when saving as elpx, not in exportations. Is this right?

cristinavaldera:
The screenshot is generated the first time you click Save, but it is not updated on subsequent saves. Is this the expected behavior? I think you could change the screenshot in Properties, but  the tab does not appear.

pabloamayab:
> The screenshot is generated the first time you click Save, but it is not updated on subsequent saves. Is this the expected behavior? I think you could change the screenshot in Properties, but the tab does not appear.

I agree; IMHO the image should be rendered every time the ELP is downloaded.

erseco:
Hi @cristinavaldera.

1. Yes, only the .elpx have the screenshot.png
2. Correct, only the first time creates the screenshot.png, on the subsequent the user must "regenerate" or upload it. (currently the UI for this is hidden, but @ignaciogros will enable it soon)

cristinavaldera (approved):

ignaciogros (approved):
> Force a render error in bridge.generateScreenshotFromFirstPage — save still succeeds; existing post-save fallback runs on the next save.

It saves properly, but I'm not sure about that "render error". It saves, and the image is not overwritten, so I think we can merge this into `main` now.

I'll enable the screenshot editor as soon as possible.

Thanks.

PR #1773, opened 2026-05-09
fix(api): use project.uuid from createProject instead of crypto.randomUUID

nubelistadotcom:
## Summary

The route handler for `POST /api/v1/projects` was generating its own UUID via `crypto.randomUUID()` and passing it to `createProject()`, but `createProject()` internally generates its own UUID via `uuidv4()` and ignores the passed uuid (its type is `Omit<NewProject, 'uuid'>`).

This caused a mismatch: `ensureDocument()` was called with the crypto UUID while the project was stored with the `uuidv4()` UUID, leading to Yjs document initialization failures for newly created projects created via the API.

## Changes

- Removed `crypto.randomUUID()` generation from the route handler
- Use `project.uuid` from `createProject()` result instead
- Also fixed response to use actual database values (`project.id`, `project.created_at`, `project.updated_at`) instead of `Number(project.insertId)`, `Date.now()`, and `null`

## Root Cause

`createProject()` at `src/db/queries/projects.ts:300` has signature `Omit<NewProject, 'uuid'>` — the uuid parameter is explicitly omitted from the data argument because the function generates its own UUID. The route handler was unknowingly sending a uuid that was silently ignored.


erseco (changes requested):
Hi @nubelistadotcom, thanks a lot for digging into this and identifying the UUID mismatch issue. This looks like a real bug and the fix makes sense.

I also checked the timestamp changes, and those look fine too since the project schema already stores `created_at` / `updated_at` as Unix timestamps in milliseconds across all supported databases.

One thing I noticed: the duplicate-project route seems to follow the same pattern and may still have the same UUID mismatch problem. Could you also apply the same fix there?

It would also be great to add a regression test covering project creation (and ideally duplication too), verifying that:

* the returned UUID matches the persisted UUID;
* `ensureDocument()` is initialized with the correct UUID.

Thank you for your help :)

erseco (approved):
Thanks @nubelistadotcom! I've pushed two additional commits directly to your branch to address the feedback:

**1. Same fix applied to the duplicate route** (`POST /:uuid/duplicate`)

The route was also generating `crypto.randomUUID()` and passing it to `createProject()`, then calling `ensureDocument()` with the wrong UUID and returning stale values (`insertId`, `Date.now()`, `null`) in the response. Now it uses `newProject.uuid`, `newProject.id`, `newProject.created_at`, and `newProject.updated_at` from the `createProject()` result — same pattern as your fix on the create route.

**2. Regression tests** for both routes

Added two tests in `projects.spec.ts` that verify:
- The UUID returned in the response exists in the database (i.e., response UUID == persisted UUID)
- The `id` in the response matches the database record
- `ensureDocument` is called with the persisted UUID (captured via the doc-manager's `findProjectByUuid` mock)

All 29 tests pass. Approving — great catch on the original bug!

PR #1774, opened 2026-05-09
fix(api): fix updateComponent call signature in route handler

nubelistadotcom:
## Summary

The `PUT /:uuid/components/:componentId` route handler was calling `updateComponent()` with the wrong argument structure, causing `"Component [object Object] not found"` errors on every component update attempt.

## Root Cause

The function signature is:
```typescript
updateComponent(ydoc: Y.Doc, componentId: string, updates: UpdateComponentInput): OperationResult
```

But the route was passing a single object as the second argument:
```typescript
updateComponent(ydoc, {
    componentId: params.componentId,
    htmlContent: body.htmlContent,
    // ...
})
```

## Fix

Extract `params.componentId` as the second argument and pass the remaining fields as the third argument:
```typescript
updateComponent(ydoc, params.componentId, {
    htmlContent: body.htmlContent,
    // ...
})
```

Note: `createComponent()` takes `(ydoc, input)` as a single input object, but `updateComponent()` takes `(ydoc, componentId, updates)` — different patterns in the same module.

### DIFF
--- a/src/routes/api/v1/components.spec.ts
+++ b/src/routes/api/v1/components.spec.ts
@@ -340,6 +340,41 @@ describe('Components API v1', () => {
             // Should return 404 or 400 when component not found
             expect([400, 404]).toContain(response.status);
         });
+
+        it('should call updateComponent with correct argument order (regression)', async () => {
+            // Regression test: updateComponent(ydoc, componentId, updates)
+            // Previously the route passed an object as second arg:
+            //   updateComponent(ydoc, { componentId, ... })
+            // which caused "Component [object Object] not found" errors.
+            await db
+                .insertInto('projects')
+                .values({
+                    uuid: 'regression-update-comp',
+                    title: 'Regression Test Project',
+                    owner_id: userId,
+                    created_at: now(),
+                })
+                .execute();
+
+            const response = await app.handle(
+                new Request('http://localhost/projects/regression-update-comp/components/some-comp-id', {
+                    method: 'PUT',
+                    headers: {
+                        Authorization: `Bearer ${userToken}`,
+                        'Content-Type': 'application/json',
+                    },
+                    body: JSON.stringify({ title: 'Updated Title' }),
+                }),
+            );
+
+            // If the fix is correct, the endpoint should NOT return an error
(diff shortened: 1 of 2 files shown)


erseco (changes requested):
Hi @nubelistadotcom thanks for your contribution, 

The `updateComponent()` fix looks correct and addresses a real issue.

`updateComponent()` expects:

```ts
updateComponent(ydoc, componentId, updates)
```

but the route is currently passing an object as the second argument, which explains the `Component [object Object] not found` error.

That said, I don't think the route change for component creation should be included in this PR.

Even if blocks are internally scoped by page during creation, component updates/searches are already resolved globally by ID, so the current API can continue working as-is without introducing `pageId` into the route.

Changing:

```http
POST /api/v1/projects/:uuid/blocks/:blockId/components
```

to:

```http
POST /api/v1/projects/:uuid/pages/:pageId/blocks/:blockId/components
```

would also introduce a breaking API change unrelated to the reported bug.

I would suggest keeping this PR focused only on the `updateComponent()` fix, preserving API compatibility, and opening a separate issue if there is interest in redesigning the component creation routes around page-scoped blocks.

Please also add a regression test for the `PUT /projects/:uuid/components/:componentId` endpoint.

nubelistadotcom (commented):
Thanks for the thorough review, @erseco!

I've addressed your feedback:

- **Reverted the route change** —  is back to its original signature. No breaking API change.
- **Kept the `updateComponent()` fix** — passing `componentId` as the second argument instead of wrapping it in an object.
- **Added a regression test** for `PUT /projects/:uuid/components/:componentId` that verifies the endpoint doesn't return the `[object Object] not found` error.

Ready for another look when you have a moment.

erseco (approved):
Thanks @nubelistadotcom.

I reviewed the updated changes and they address the points from my previous review:

- the component creation route change has been reverted, so there is no unrelated breaking API change;
- the `updateComponent()` call now passes `componentId` as the second argument, matching the expected function signature;
- a regression test was added for the `PUT /projects/:uuid/components/:componentId` endpoint.

The PR is now focused on the reported bug and looks good to me.

Approved.

PR #1775, opened 2026-05-09
feat(admin): allow re-uploading site themes to update them in place (#1755)

### LINKED ISSUE #1755: Admin panel. Update style.
Upgrading a style will be quite common. It'll be good to have an option in the admin panel for that.

erseco:
Closes #1755.

## Summary

- Admins can now upgrade a custom site theme by re-uploading a ZIP with the
  same internal `<name>`. The upload endpoint is idempotent on the
  slugified `dir_name`: built-in themes are still rejected, site themes are
  replaced in place, brand-new uploads keep the existing create flow.
- The replace flow preserves `is_default`, `is_enabled`, and `sort_order`,
  so admins do not have to reconfigure the theme after the upgrade.
- Site-theme URLs now embed the theme's `updated_at` as a cache-busting
  suffix (`/v{appVersion}-{updatedAtMs}/site-files/themes/{dir}`), so end
  users automatically pull the new CSS/icons next time they load the
  workarea — no manual cache reset needed.
- The admin upload UI shows a distinct "Theme replaced" notification when
  the backend reports `replaced: true`, so the upgrade flow is
  self-documenting.

## How an admin upgrades a theme now

1. Open **Admin panel → Styles**.
2. Click **Upload Theme** and pick the new ZIP.
3. As long as the new ZIP's `<name>` matches the existing theme's
   `dir_name`, the server overwrites the on-disk files, refreshes the DB
   metadata, keeps the default/enabled flags, and returns
   `{ replaced: true, message, theme }`.
4. The admin sees a "Theme replaced" toast. End users get the new files on
   the next page load thanks to the cache-busting URL — no admin action
   needed on their browsers.

## API change

`POST /api/admin/themes/upload`:

- New theme → `201 Created` + serialized theme (unchanged).
- Existing site theme with same `dir_name` → `200 OK` +
  `{ replaced: true, message, theme }`.
- Existing **built-in** theme name → `400 Bad Request` (unchanged behavior).

## Files

| File | What changed |
|------|--------------|
| `src/routes/admin-themes.ts` | Replace-on-upload flow with built-in protection. |
| `src/utils/site-theme-url.ts` (new) | Pure helper that builds cache-busted site-theme URLs. |
| `src/routes/themes.ts`, `src/routes/project.ts`, `src/routes/config.ts` | Use the new helper so every API path returns the same cache-busted URL. |
| `views/admin/index.njk` | "Theme replaced" notification when backend reports `replaced: true`. |
| `src/routes/admin-themes.spec.ts` | New tests for replace, built-in conflict, and DB-update-failure paths. |
| `src/utils/site-theme-url.spec.ts` (new) | Coverage for the URL helper. |
| `src/routes/themes.spec.ts` | Updated URL assertion to match cache-busted format. |

## Test plan

- [x] `bun test src/routes/admin-themes.spec.ts` — 64 pass
- [x] `bun test src/routes/themes.spec.ts` — 41 pass
- [x] `bun test src/utils/site-theme-url.spec.ts` — 6 pass
- [x] `make test-unit` — 6922 backend tests pass; coverage on changed files
      ≥ 99% (the only sub-90% file flagged by the global threshold,
      `src/websocket/asset-coordinator.ts`, is pre-existing and unrelated).
- [x] `make fix` — clean.
- [x] Manual smoke: upload a fresh theme, then upload a modified ZIP with
      the same `<name>` and confirm the workarea picks up the new CSS
      after a reload (the `?`-less URL changes because of the
      `updated_at` segment).


ignaciogros (dismissed):
Hi @erseco,

Thanks!

ignaciogros (dismissed):
All suggested tests work, but when you add or update a style, no screenshot is displayed in the Admin Panel (with custom `BASE_PATH=`).

``

`/202605211316/site-files/themes/my-style/screenshot.png` throws a 404 error.

The screenshot is properly displayed when `BASE_PATH=` is empty in the `.env` file.

erseco:
@ignaciogros thanks for catching this — fixed in d7cc572.

**Root cause:** the Admin Panel emits screenshot URLs as `{BASE_PATH}/site-files/themes/{dir}/screenshot.png`. That path was only served by a **root-mounted** route, while the `onRequest` hook (which strips `BASE_PATH` and serves `public/`, `/files/tmp`, and versioned `/v{…}/site-files/*` assets) had **no branch for the plain `/site-files/themes/*` path**. With `BASE_PATH=` empty the root route matched, which is exactly why it only 404'd behind a subdirectory proxy.

**Fix:** extracted the path matching, traversal guard and file read into a tested `serveSiteThemeFile()` helper (handles both the versioned cache-busted and the plain URL shapes) and call it from both the `onRequest` hook and the fallback route. Because the hook runs *after* `BASE_PATH` is stripped, theme files now resolve for both root and subdirectory installs.

While here, I also extracted `buildDefaultThemePayload()` from the project-create route so its site/base branches are unit-testable, closing the patch-coverage gaps codecov flagged in `config.ts` and `project.ts`.

**Tests added:**
- `src/utils/site-theme-file.spec.ts` — resolver + server, including traversal attempts
- `test/integration/site-theme-files-basepath.spec.ts` — serves the screenshot with `BASE_PATH` set (200) and without (200), 404 for missing files, refuses `../` traversal
- `config.spec.ts` — `GET /api/config/default-theme` returns the cache-busted site URL
- `project.spec.ts` — `buildDefaultThemePayload` site/base branches

**Verification:** `make fix` clean; `make test-unit` (7057) and `make test-integration` (709) green; coverage threshold met (`site-theme-file.ts` 100%).

Could you re-test with a custom `BASE_PATH` and re-review when you have a moment? 🙏

pabloamayab:
Hi @erseco . After updating a style via the admin panel, the style didn't update in either Chrome or Firefox until I cleared the site data (I tried logging out and logging back in). In incognito mode, the style update took effect immediately.

erseco:
@pabloamayab great catch, thanks for testing this! Fixed in d219368.

**Root cause.** The server already cache-busts the static asset URL with the theme's `updated_at` (i.e. `/v{appVersion}-{updatedAt}/site-files/themes/{dir}`), so `<link rel="stylesheet">` references and screenshots invalidate fine. But the workarea pulls theme files through `ResourceFetcher`, which keyed its **IndexedDB** (`exelearning-resources-v1`) cache only by app version — `('theme', themeName, appVersion)` — and fetched the bundle at a stable URL `/api/resources/bundle/theme/{name}`. So after a re-upload:

- Memory cache: wiped by reload ✅
- IndexedDB cache: still keyed by the unchanged `appVersion` → **returned stale files** ❌
- Browser HTTP cache: bundle URL hadn't changed → **also stale** ❌

Incognito worked because IndexedDB starts empty per session. Logout/login didn't help because the DB is scoped to the origin, not the user.

**Fix.**
1. `/api/themes/installed` now exposes `updatedAt` (the `themes.updated_at` from DB) on each site theme alongside the existing URL cache-buster.
2. `ResourceFetcher.init()` loads a `dirName → updatedAt` map once and caches it on the instance.
3. `fetchTheme()` uses that as part of the IndexedDB cache key (`{appVersion}-{updatedAt}`) and appends `?v={updatedAt}` to the bundle URL, so the HTTP cache invalidates too. Base themes keep using just `{appVersion}` (they version with the app).

**Tests.**
- `src/routes/themes.spec.ts` — asserts `updatedAt` is present and matches the suffix encoded in `url`.
- `public/app/yjs/ResourceFetcher.test.js` — covers (a) the cache key contains the `updated_at` for site themes, (b) re-uploading produces a different key, (c) the bundle URL gets `?v=…`, (d) `loadSiteThemeVersions()` is idempotent and skipped in static mode.

Could you re-test the re-upload flow without clearing site data when you have a moment? 🙏

pabloamayab (approved):
LGTM!!!!! Thnanks @erseco

ignaciogros (approved):
Thanks!

PR #1777, opened 2026-05-09
feat(preferences): user-level default style for new projects (#1754)

### LINKED ISSUE #1754: New user preference. Default style.
Allow the user to choose the default style.

erseco:
Closes #1754.

## Summary
- New **Default style for the new documents** dropdown in the User Preferences modal, placed between *License for the new documents* and *Default AI Assistant*. The first option is an empty *Use the default style of the site* entry that opts back into the admin/site default.
- The user value takes precedence over the admin/site default when creating a new resource, and is honored in both the cloud build and the static (offline) build.
- The dropdown options are derived from the enabled themes (base + admin-uploaded site themes) per request, so newly installed/disabled themes show up immediately.

## How it works
- `src/routes/config-params.ts`: `buildConfigParams` accepts an optional `THEMES` dependency and exposes a `defaultTheme` `select`. Empty value = "use site default".
- `src/routes/config.ts`: builds the theme map at request time from `getEnabledThemes(db)`.
- `scripts/build-static-bundle.ts`: feeds the same map from the bundled `themes/base/` directory so the static distribution exposes the dropdown.
- `src/routes/project.ts`: `POST /api/project/create-quick` resolves the theme via the new pure helper `resolveDefaultThemeForNewProject` (`userPref → adminDefault → 'base'`). It validates the referenced theme still exists and is enabled before honoring the preference.
- `public/app/workarea/project/projectManager.js`: `initialiceProject()` prefers the new `defaultTheme` preference, then the legacy hidden `theme` key, then `eXeLearning.config.defaultTheme`. An empty value falls back to the site default.
- The legacy hidden `theme` preference key is kept for backward compatibility.

## Test plan
- [x] `bun test src/routes/config-params.spec.ts src/routes/project-default-theme.spec.ts` — new precedence + options tests.
- [x] `bun test src/routes/config.spec.ts src/routes/project.spec.ts` — full backend route specs (with `DB_PATH=data/exelearning.db`).
- [x] `npx vitest run public/app/workarea/user/preferences/userPreferences.test.js` — round-trip of the new key in `localStorage` and via the server API.
- [x] `npx vitest run public/app/workarea/project/projectManager.test.js` — `initialiceProject` precedence rules.
- [x] `make fix` — lint clean.
- [x] `make test-unit` — 6928/6928 unit tests pass; coverage threshold (90%) holds for every file.
- [x] `make test-integration` — 700/700 integration tests pass.
- [x] Manual smoke: open Preferences in the UI, choose a theme, create a new project and verify it opens with that style (cloud + static).


pabloamayab:
Hi @erseco !!! It works fine, but if I open a file whose style isn't installed, should it open using the system's “default” style or the one chosen by the user?

erseco:
Hi @pabloamayab — good catch, that case wasn't covered. Until now, when a project referenced a theme that wasn't installed, `ThemesManager.selectTheme` fell straight back to the site/admin default and ignored the user's preference.

Pushed [63fdb21](https://github.com/exelearning/exelearning/commit/63fdb21f) which extends the fallback chain in `public/app/workarea/themes/themesManager.js` to:

1. The theme requested by the document.
2. The user's **"Default style for the new documents"** preference (if it points to an installed theme).
3. The site/admin default theme.
4. `themeRegistryOverride.fallbackTheme` (host integrations).
5. `base` as last resort.

That keeps the client behavior consistent with `resolveDefaultThemeForNewProject` on the server, so the user's choice is honored both when creating new documents and when opening existing files whose style isn't installed. Covered with three new unit tests in `themesManager.test.js`.

pabloamayab:
Hi @erseco . It still isn't working; I've tried both the cloud and desktop versions.

pabloamayab:
Hi @erseco !! When you have opportunity, please take a look at this PR because it still doesn't work when opening ELPx files with styles that aren't installed using the user-selected style.

erseco:
Hi @pabloamayab — fixed in e843fd7. The real cause was that the file-open path never reached the fallback chain: `_checkAndImportTheme` called `selectTheme(config.defaultTheme, ...)` directly when a theme wasn't installed, so the user-preference branch in `selectTheme` was skipped. It now passes the original `themeName`, letting `selectTheme` run its chain (user default → admin default → base).

Verified in the browser (`make up-local`): set the user default style to **Zen**, then opened an `.elpx` whose style (`chiquito`) is not installed and has no `/theme/` folder. The document now opens with **Zen** instead of the site default, and the console confirms:

> `[ThemesManager] Theme 'chiquito' unavailable; falling back to user default 'zen'.`

The document keeps its original theme reference (`chiquito`) so it still round-trips and renders natively if that style is ever installed. Updated `YjsProjectBridge.test.js` and added a regression test. Since this is client-side code, it applies equally to cloud, static and desktop.

ignaciogros (approved):
Tested in these versions:

- Cloud
- Static
- Electron

Thanks, @erseco.

Issue #1769, opened 2026-05-07
Duplicate assets when saving and exporting


pabloamayab:
When an asset is added in Tiny or in File Manager, it appears correctly in the file manager, but when exported or saved, that asset is duplicated—one with its original name and another with a UUID format.
This has been tested in eXe 4 on the desktop and cloud versions, as well as on app.exelearning.net

<img width="945" height="90" alt="Image" src="https://github.com/user-attachments/assets/402005d8-f8cf-4fee-b90a-8829ff755921" />

PR #1778, opened 2026-05-09
fix(platform): require asset provider before client-side platform export and warn on legacy fallback (#1770)

### LINKED ISSUE #1770: Export to Moodle does not package the assets
I am opening this issue because, following the release of eXe (cloud) version 4.0 and the Moodle plugins, publishing to Moodle continues to fail; the assets are not included in the ZIP file that is sent.
The PR was #1740 
The problem seems to be with the generation (or caching) of the eXe image in the pool when generating the 4.0 release, and not with the solution provided in that PR. If I run `docker compose -f docker-compose-mariadb.yml up -d --pull always --force-recreate`, eXe-to-Moodle publication fails. 

However, if I modify the YAML file with
`build: https://github.com/exelearning/exelearning.git#main` (instead of using TAG:-latest) so that it generates the main image locally, it does work correctly.

The test environment details are:
- Debian 6.1.170-1 x86_64
- docker-compose-mariadb.yml provided in the eXe repository
- Chrome and Firefox

erseco:
## Summary

Investigation + fix for [#1770](https://github.com/exelearning/exelearning/issues/1770) "Export to Moodle does not package the assets". The reporter sees the activity end up with HTML referencing `content/resources/<file>` URLs but no matching ZIP entries — the same symptom [#1740](https://github.com/exelearning/exelearning/pull/1740) was meant to fix.

[#1740](https://github.com/exelearning/exelearning/pull/1740) does the right thing on the happy path: it generates the package in the browser (where the live Yjs document and IndexedDB asset blobs live) and forwards it to the platform via the new `/api/platform/integration/set_platform_new_ode_browser` endpoint. But TDD investigation surfaced **two reachable code paths that still produce an asset-less package and silently call them a success**, both of them inside `navbarFile.uploadPlatformEvent`. Either one is enough to reproduce the missing-images symptom on a deployed Docker image.

### 1. `canExportClientSide` did not require an asset provider

```js
const canExportClientSide =
    !!yjsBridge?.documentManager && typeof SharedExporters?.quickExport === 'function';
```

When `_yjsBridge.assetManager` *and* the legacy `_yjsBridge.assetCache` are **both null** (e.g. `AssetManager` init was racing the click on "Finish to Moodle", or the wiring failed for any reason), this gate still passes. `SharedExporters.quickExport` is then invoked with `null, null` for the asset providers. `BrowserAssetProvider` falls through to `createNullAssetProvider()` (see `src/shared/export/browser/index.ts`), and the resulting ZIP has **no `content/resources/` folder at all** — exactly the symptom on Moodle.

### 2. The legacy fallback was silent

When the gate evaluates to **false** (e.g. `exporters.bundle.js` failed to expose `SharedExporters`, or the project never ran a Yjs bridge), `uploadPlatformEvent` silently calls the old `postFirstTypePlatformIntegrationElpUpload` server-side route — the very flow [#1740](https://github.com/exelearning/exelearning/pull/1740) was retiring because it builds from a stale Yjs snapshot and an empty `DatabaseAssetProvider`. There was **no signal in the operator console** that the wrong path had been taken; the user got a "redirect to Moodle on success" indistinguishable from a healthy export.

Either branch matches the user-facing report on `:latest`, where the same code that works locally with `build: …#main` produces a broken package on Moodle.

## Fix

`public/app/workarea/menus/navbar/items/navbarFile.js`:

1. **Require an asset provider on the gate.** `canExportClientSide` now also checks `!!(yjsBridge.assetManager || yjsBridge.assetCache)`. If neither is available we no longer pretend the browser-side path is healthy — we fall through to the legacy route on purpose.
2. **Make the legacy fallback noisy.** Whenever the gate fails, log a `console.warn` with the per-dependency boolean breakdown (`hasDocumentManager`, `hasAssetProvider`, `hasSharedExporters`) so the missing-asset symptom is diagnosable from a deployed image's console.

The two new TDD regression tests in `navbarFile.test.js` pin both contracts.

## Tests

- `navbarFile.test.js` — two new regressions added inside `describe('platform upload')`:
  - "client-side branch must not silently produce an empty-asset package when yjsBridge.assetManager is null (#1770)" — fails on `main` (quickExport runs with `null, null`); passes after the gate change.
  - "legacy fallback (no SharedExporters) must surface a warning that the asset-less server path is being used (#1770)" — fails on `main` (no console signal); passes after the warn is added.

Both tests fail without the `navbarFile.js` change and pass with it.

## Verification

- `make fix` ✅
- `make test-unit` — 6915 pass, coverage ≥ 90% ✅
- `npx vitest run public/app/workarea/menus/navbar/items/navbarFile.test.js` — 202 pass ✅

## Relationship to #1769

Independent issue. [#1769](https://github.com/exelearning/exelearning/issues/1769) is about saved/exported packages bundling each asset twice (UUID + friendly path) and is fixed in [#1776](https://github.com/exelearning/exelearning/pull/1776). #1770 is about Moodle exports producing **no** assets in some runtime configurations — covered here. The fixes do not overlap.

## Test plan

- [x] TDD: failing tests reproduce both code paths that produce the empty-asset package
- [x] Both tests pass after the gate + warn change
- [x] Existing 200+ navbarFile tests still pass
- [x] Full unit suite green
- [x] Manual: with `_yjsBridge.assetManager = null`, the client-side path is now skipped and the operator sees the diagnostic `console.warn`


pabloamayab (approved):
I'm OK with this PR being merged into the main branch so we can test it on the production system where I detected the bug

Issue #1786, opened 2026-05-11
Stable identifiers across import/export — unify ID generation and preserve project identity


erseco:
## Summary

eXeLearning regenerates identifiers that should remain stable across a project's life cycle. This causes:

- **Loss of SCORM/IMS tracking** in LMSs (Moodle, Canvas, …) because `manifest@identifier` changes on every export even though the project is the same.
- **Pages, blocks and iDevices with different IDs** after a simple `open .elpx → download .elpx` round-trip, with no content changes.
- **Useless diffs** between two successive downloads of the same project: the entire `content.xml` looks different.
- **External links** that target pages by ID break after every round-trip.
- **Two parallel implementations of `generateId(prefix)`** with different shapes, producing heterogeneous IDs inside the same `content.xml`.

## Steps to reproduce

1. Import `project.elpx`.
2. Download as `elpx-1` without changing anything.
3. Close and re-import `project.elpx`.
4. Download as `elpx-2`.
5. Diff the `content.xml` files.

**Actual**: no `<odePageId>` matches. Same for `<odeBlockId>`, `<odeIdeviceId>`, `<odeId>`, `<odeVersionId>`, and `imsmanifest.xml@identifier`.
**Expected**: identifiers are preserved across the round-trip.

### Observed evidence

Two consecutive SCORM 1.2 exports of the same project (`recreando-a-los-austrias-nuevo-1_scorm (4)` and `(5)`):

| Element | Export 4 | Export 5 |
|---|---|---|
| Page "Guía didáctica y fichero fuente" | `page-mp0fppwf-71v64kl4r` | `page-moyxsz3m-clizazesi` |
| `manifest@identifier` | `eXe-MANIFEST-mp0gpahppfpj4` | `eXe-MANIFEST-mp113xnrkpjwk` |
| `organization@identifier` | `eXe-mp0gpahppfpj4` | `eXe-mp113xnrkpjwk` |

Zero ID matches between the two packages.

## Root causes (four independent bugs)

| # | Where | What |
|---|---|---|
| 1 | `public/app/yjs/YjsStructureBinding.js:2824` vs `src/yjs/structure-binding.ts:38` + others | Two `generateId(prefix)` functions with different formats (decimal vs base36 timestamp) coexist in the same project's `content.xml` |
[...]

PR #1788, opened 2026-05-11
fix(import): preserve page/block/iDevice IDs in ELPX v4 round-trip

### LINKED ISSUE #1783: Preserve page/block/iDevice IDs across ELPX v4 import (stop regenerating on every import)
## Summary

The ELPX v4 importer unconditionally regenerates every `<odePageId>`, `<odeBlockId>` and `<odeIdeviceId>` it reads from `content.xml`, even when there is no collision and even when the project is being imported into a clean Y.Doc. This breaks identity-stability across `import → export` round-trips.

This is sub-issue **B** of the *Stable identifiers* tracking issue.

## Steps to reproduce

1. Import `project.elpx`.
2. Download as `elpx-1` without changing anything.
3. Close and re-import the same `project.elpx`.
4. Download as `elpx-2`.
5. `diff elpx-1/content.xml elpx-2/content.xml` — none of the page/block/iDevice IDs match.

| Page (same title) | Export 4 | Export 5 |
|---|---|---|
| "Guía didáctica y fichero fuente" | `page-mp0fppwf-71v64kl4r` | `page-moyxsz3m-clizazesi` |

## Root cause

`src/shared/import/ElpxImporter.ts::buildFlatPageList`:

```ts
const newPageId = this.generateId('page'); // line 969 — unconditional
...
idRemap.set(originalPageId, newPageId);    // line 972
```

Same pattern at:

- line `1059` — `this.generateId('block')` (every block)
- line `1099` — `this.generateId('idevice')` (every component)

The `idRemap` is only useful in **merge mode** (importing into a Y.Doc that already has pages with the same IDs). For the common case — fresh import into an empty Y.Doc — the remap rewrites every ID for no reason.

The XML already contains the IDs the user authored (the `<odePageId>` tag is read at line 1516 via `getPageId`), and they are valid Format-A IDs (assuming sub-issue A also lands). They should be kept.

## Proposal

In `buildFlatPageList`:

1. Try to read `originalPageId` from the XML (`<odePageId>` or `odeNavStructureId` attribute) — already done.
2. If `clearExisting === true` **or** `originalPageId` does not collide with any ID currently in `navigation`, use `originalPageId` verbatim. No remap.
3. Otherwise (merge mode with real collision, or malformed/empty ID), generate a fresh ID and record the remap so `remapInternalPageLinks` can fix internal `exe-node:` links.
4. Same logic for block IDs and iDevice IDs.

Additionally:

- The **legacy v3 path** (`convertLegacyPagesToPageData`, line 644) keeps its current behaviour: legacy `.elp` files use short non-namespaced IDs like `page-4`, which collide trivially and must be regenerated to be safe. Only the modern v4 path changes.
- Update the existing tests that assert \"importer assigns new IDs\" — they encode the current bug. New tests should pin: *same XML in → same IDs out*.

## Why this is safe

Format-A IDs (`page-<8 base36>-<9 base36>`) carry ~70 bits of entropy in the random suffix plus a timestamp. Collisions across independent projects are negligible. Within the same project they are already unique by construction. The only realistic collision is the user importing the same `.elpx` twice into the same workspace — which is exactly the merge-mode case we still handle.

## Acceptance

- Round-trip test: import `test/fixtures/<v4-elpx>`, export, parse the exported `content.xml`, assert that every `<odePageId>`, `<odeBlockId>`, `<odeIdeviceId>` matches the input.
- Merge-mode test: import the same fixture twice into one Y.Doc; assert that the second import remaps colliding IDs and that internal links inside the remapped pages point to the new IDs.
- Existing legacy `.elp` import tests continue to pass.

## Related

Part of the *Stable identifiers across import/export* tracking issue.

erseco:
## Summary

`ElpxImporter.buildFlatPageList` regenerated every `<odePageId>`, `<odeBlockId>` and `<odeIdeviceId>` it read from content.xml — even on fresh import into an empty Y.Doc. A simple `open .elpx → download .elpx` round-trip produced a fully different ID set.

Fix: keep the original IDs verbatim when they do not collide with existing Y.Doc navigation. A new `collectExistingIds()` helper pre-seeds a `usedIds` Set at the start of each import:

- **Fresh import (`clearExisting=true`)**: set starts empty → all original IDs from content.xml flow through unchanged.
- **Merge-mode import (`clearExisting=false`)**: set is pre-seeded from the current navigation → real collisions are detected, the colliding ID is regenerated, and `idRemap` rewrites any internal `exe-node:` links as before.

The same logic applies to `buildBlockData` and `buildComponentData` for block and iDevice IDs.

The legacy v3 path (`convertLegacyPagesToPageData`) keeps its current behavior unchanged: legacy IDs (`page-4`, `idevice-2`) are short non-namespaced strings that collide trivially across documents and must still be regenerated.

## Test plan

- [x] New test: fresh v4 import preserves explicit `<odePageId>` values verbatim (`importFromZipContents`, `clearExisting: true`)
- [x] New test: block + iDevice IDs also preserved in the same fresh import
- [x] New test: merge-mode collision still remaps and updates internal `exe-node:` links
- [x] New test: legacy v3 (`contentv3.xml`) path still regenerates IDs — pins the v3/v4 boundary
- [x] Updated existing `anchors.zip` test: now asserts that cross-page links resolve to the preserved imported IDs (no dangling references)
- [x] Updated existing anchor-fragment test: asserts the imported page ID is preserved, link fragment survives
- [x] `make test-unit` → 6919 pass, 0 fail
- [x] `make test-integration` → 700 pass, 0 fail
- [x] `make fix && make lint` → clean

Closes #1783

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -11,7 +11,7 @@
         "@elysiajs/static": "^1.4.7",
         "@elysiajs/swagger": "^1.3.1",
         "@sinclair/typebox": "^0.34.48",
-        "@xmldom/xmldom": "^0.8.11",
+        "@xmldom/xmldom": "^0.9.9",
         "bcryptjs": "^3.0.3",
         "dotenv": "^17.3.1",
         "elysia": "^1.4.27",
@@ -29,8 +29,8 @@
         "mime-types": "^3.0.2",
         "mysql2": "^3.18.2",
         "nunjucks": "^3.2.4",
-        "pdfjs-dist": "5.5.207",
-        "uuid": "^13.0.0",
+        "pdfjs-dist": "5.7.284",
+        "uuid": "^14.0.0",
         "ws": "^8.19.0",
         "y-websocket": "^3.0.0",
         "yjs": "^13.6.29",
@@ -39,7 +39,7 @@
         "@babel/core": "^7.29.0",
         "@babel/preset-env": "^7.29.0",
         "@biomejs/biome": "^2.4.5",
-        "@codecov/bundle-analyzer": "^1.9.1",
+        "@codecov/bundle-analyzer": "^2.0.1",
         "@electron/notarize": "^3.1.1",
         "@playwright/test": "^1.58.2",
         "@types/bcryptjs": "^3.0.0",
@@ -52,14 +52,14 @@
         "@vitest/coverage-v8": "^4.0.18",
         "@vitest/ui": "^4.0.18",
         "cross-env": "^10.1.0",
-        "electron": "^41.0.0",
+        "electron": "^42.0.0",
         "electron-builder": "^26.8.1",
-        "esbuild": "^0.27.3",
+        "esbuild": "^0.28.0",
         "happy-dom": "^20.8.0",
-        "http-proxy-middleware": "^3.0.5",
+        "http-proxy-middleware": "^4.0.0",
         "kill-port": "^2.0.1",
         "sass": "^1.97.3",
-        "typescript": "^5.9.3",
+        "typescript": "^6.0.2",
         "vite": "^8.0.0",
         "vitest": "^4.0.18",
         "wait-on": "^9.0.4",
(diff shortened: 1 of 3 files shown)


mnarvaezm (changes requested):
src/shared/import/ElpxImporter.ts (line 1484): remapInternalPageLinks() can partially remap IDs when one ID is a prefix of another. For example, with page-1 and page-10, a link like exe-node:page-10 may be remapped using the mapping for page-1, leaving the trailing 0. This affects both the legacy path and the new v4 path when page IDs collide by prefix. A common fix is to sort IDs by descending length and enforce an ID boundary condition in the replacement logic.
@erseco

erseco:
Thanks for the catch on the partial-prefix remap — confirmed and fixed in `0eaff14e`.

**TDD proof:** added a failing test `ElpxImporter - remapInternalPageLinks prefix-collision safety` that imports a v4 fixture with `page-1` and `page-10` (the link targets `exe-node:page-10`) twice in merge mode so both ids collide and end up in `idRemap`. Before the fix the test red-lined with:

```
Expected to contain: "exe-node:page-mp1njfjn-57gdtheow"
Received:            "exe-node:page-mp1njfjn-s0kfa0l690"
```

The trailing `0` is exactly the stray you described: the regex matched `page-1` inside `page-10` and replaced it with the new-page-1 id.

**Fix in `remapInternalPageLinks`:**

1. Sort `idRemap.keys()` by descending length so longer ids (`page-10`) are tried before shorter prefixes (`page-1`) in the alternation.
2. Add an ID-boundary lookahead `(?![A-Za-z0-9_-])` so even if a shorter id appears first it cannot match inside a longer one.

```ts
const sortedKeys = Array.from(idRemap.keys()).sort((a, b) => b.length - a.length);
const escapedIds = sortedKeys.map(id => id.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'));
const pattern = new RegExp(
    `(exe-node:)(${escapedIds.join('|')})(?![A-Za-z0-9_-])(#[^"'\\s]*)?`,
    'g',
);
```

Covers both the v4 merge-mode path and the legacy v3 path since both feed through the same helper. Full importer suite: 63 pass / 0 fail; `make fix` clean.

mnarvaezm (approved):

erseco:
Superseded by #1791 — single landable bundle for the four parallel fixes. This PR's scope (preserve page/block/iDevice IDs on ELPX v4 import + prefix-collision fix (#1783)) is included verbatim via a `Merge #1788` commit so attribution stays intact. The bundle also closes the remaining coverage gaps the reviewer flagged across this PR and the others. Branch `fix/1786-stable-identifiers-bundle`.

erseco:
Superseded by #1791. Scope (preserve page/block/iDevice IDs in ELPX v4 round-trip + prefix-collision boundary, #1783) merged verbatim via `Merge #1788`. The unified bundle is the simplest path to land #1786 in one shot.

Issue #1785, opened 2026-05-11
Stable SCORM/IMS manifest identifier (derive from `odeIdentifier`, drop per-export regeneration)


erseco:
## Summary

Every SCORM 1.2 / SCORM 2004 / IMS CP export generates a fresh `manifest@identifier` via `Date.now() + Math.random()`, ignoring the project's persistent `odeId`. LMSs (Moodle, Canvas, …) key tracking off `manifest@identifier`, so **every re-upload is treated as a different course** and existing learner progress, grades and completion are orphaned.

This is sub-issue **D** of the *Stable identifiers* tracking issue.

## Evidence

Three identical bugs in the export layer:

- `src/shared/export/exporters/Scorm12Exporter.ts:47` + `:390`
- `src/shared/export/exporters/Scorm2004Exporter.ts:47` + `:378`
- `src/shared/export/exporters/ImsExporter.ts:47` + `:353`

All three:

```ts
const projectId = this.generateProjectId(); // ignores meta.odeIdentifier / meta.scormIdentifier
// ...
generateProjectId(): string {
    return Date.now().toString(36) + Math.random().toString(36).substring(2, 7);
}
```

The result becomes the manifest root identifier, organization identifier, and the LOM `catalog/entry`.

### Observed in the wild

Two consecutive SCORM 1.2 exports of the same project (`recreando-a-los-austrias-nuevo-1_scorm (4)` and `(5)`):

| Element | Export 4 | Export 5 |
|---|---|---|
| `manifest@identifier` | `eXe-MANIFEST-mp0gpahppfpj4` | `eXe-MANIFEST-mp113xnrkpjwk` |
| `organization@identifier` | `eXe-mp0gpahppfpj4` | `eXe-mp113xnrkpjwk` |

Zero matches between two unchanged exports. From the LMS's point of view, that's two unrelated courses.

### Approach: derive deterministically from `odeIdentifier`

`scormIdentifier` is already declared in `metadata-properties.ts:240-246` and `interfaces.ts:79,482` but is never read or written anywhere. Repurpose it as an **optional user override** instead of a generated value, and compute the default deterministically:

```ts
protected getManifestIdentifier(): string {
    const meta = this.getMetadata();
    if (meta.scormIdentifier) return meta.scormIdentifier;            // user override
    if (meta.odeIdentifier)   return 'eXe-MANIFEST-' + meta.odeIdentifier;
[...]

PR #1789, opened 2026-05-11
fix(import): preserve odeId/odeVersionId across ELPX import

### LINKED ISSUE #1784: Preserve `odeId` and `odeVersionId` across ELPX import (read `<odeResources>`)
## Summary

The ELPX importer never reads the project's `odeId` or `odeVersionId` from the imported `content.xml`. Both values are regenerated on every export, so two consecutive `import → export` cycles on the same project produce different identifiers and a noisy diff.

This is sub-issue **C** of the *Stable identifiers* tracking issue.

## Evidence

`src/shared/import/ElpxImporter.ts`: a grep for `odeResource`, `odeId` or `odeVersionId` returns **zero** parsing logic — the importer reads `odeProperties` but skips the entire `<odeResources>` block.

`src/shared/export/generators/OdeXmlGenerator.ts:44-45`:

```ts
const odeId      = options?.odeId      || meta.odeIdentifier || generateOdeId();
const versionId  = options?.versionId  || generateOdeId();
```

The export side **already respects** `meta.odeIdentifier` — it just never gets populated on import. `odeVersionId` has no metadata counterpart at all, so it is regenerated unconditionally.

`metadata-properties.ts:212-218` already declares `odeIdentifier` as an internal metadata field (`excludeFromXml: true`). All the wiring is in place; the importer just has to read the value.

## Proposal

1. In `ElpxImporter`, after parsing `<odeProperties>`, also parse `<odeResources>` and extract:
   - `<odeResource><key>odeId</key><value>…</value></odeResource>` → `metadata.set('odeIdentifier', value)`
   - `<odeResource><key>odeVersionId</key><value>…</value></odeResource>` → `metadata.set('odeVersionId', value)`
2. Add `odeVersionId` to `metadata-properties.ts` as a sibling of `odeIdentifier` (internal, `excludeFromXml: true` so the exporter still writes it through `<odeResources>`, not `<odeProperties>`).
3. Extend `OdeXmlGenerator` so the version-id fallback chain mirrors `odeId`:

   ```ts
   const versionId = options?.versionId || meta.odeVersionId || generateOdeId();
   ```

4. **Policy for `odeVersionId`: preserve always.** A re-export without content changes should produce the same `odeVersionId`. Bumping it when content actually changes is out of scope for this issue — file a follow-up if we want save-time bumping later. For now: maximum diff stability over the simple round-trip case is the goal.
5. Legacy `.elp` import path (`importLegacyFormat`): `contentv3.xml` has no `<odeResources>`. Generate a fresh `odeId` / `odeVersionId` once and store in metadata, so subsequent re-exports of the converted project are stable.

## Why this matters

- **Stable diffs**: `diff content.xml content2.xml` for two unchanged round-trips becomes near-empty.
- **External links / catalogs** that target a project by its `odeId` survive re-exports.
- **Unblocks sub-issue D** — the SCORM manifest identifier can be derived deterministically from `odeIdentifier`.

## Acceptance

- Round-trip test: import a v4 `.elpx`, export, re-import the exported file, export again — `odeId` and `odeVersionId` match across all four documents.
- Legacy `.elp` test: import once, export, re-import the exported `.elpx`, export — the `odeId`/`odeVersionId` generated during the first import survive the rest of the chain.
- Unit tests on `OdeXmlGenerator` for the new fallback order.

## Related

Part of the *Stable identifiers across import/export* tracking issue. Sub-issue D depends on this landing first.

erseco:
## Summary

The importer never read `<odeResources>` from content.xml, so a project's stable identity (`odeId`, `odeVersionId`) was lost on every round-trip. The export side already honours `meta.odeIdentifier` — this PR just populates it on the import side and adds the matching versionId fallback.

Unblocks #1785 (stable SCORM manifest derived from `odeIdentifier`).

## Changes

- `ElpxImporter.extractOdeResources()` parses `<odeResources>` in the v4 path; writes `odeId` → `metadata.odeIdentifier` and `odeVersionId` → `metadata.odeVersionId`.
- Legacy `.elp` v3 path (`importLegacyFormat`): contentv3.xml has no `<odeResources>`, so generate both once via `generateOdeId()` at import time. Subsequent re-exports of the converted project are stable.
- `OdeXmlGenerator.generateOdeXml()` extends the versionId fallback chain: `options.versionId > meta.odeVersionId > generateOdeId()`.
- `metadata-properties.ts` declares `odeVersionId` as a sibling of `odeIdentifier`, internal (`excludeFromXml: true` — written via `<odeResources>`, not `<odeProperties>`).
- `interfaces.ts` (both export and import) gains the `odeVersionId?: string` field.

## Test plan

- [x] v4 import populates `metadata.odeIdentifier` and `metadata.odeVersionId` from `<odeResources>`
- [x] Legacy v3 import generates both once (pattern `^\d{14}[A-Z0-9]{6}$`) so re-exports are stable
- [x] Missing or empty `<odeResources>` tolerated (export-side fallback runs)
- [x] `OdeXmlGenerator` honours `options.versionId` > `meta.odeVersionId` > `generateOdeId()`
- [x] Round-trip: import → `generateOdeXml` → re-parse → identifiers match
- [x] `bun test src/shared/import/ElpxImporter.spec.ts src/shared/export/generators/OdeXmlGenerator.spec.ts` → 116 pass
- [x] `make fix` / `make lint` clean

Closes #1784

### DIFF
--- a/doc/elpx-format/ids.md
+++ b/doc/elpx-format/ids.md
@@ -69,7 +69,7 @@ All three forms are accepted by the importer. The modern generator always produc
 | Identifier | Set when | Changes when | Notes |
 |---|---|---|---|
 | `odeId` | Project is created for the first time | Never (stable for the project's lifetime) | Stored in `odeResources`. Retrieved from `meta.odeIdentifier` if already set, otherwise `generateOdeId()` is called once (`OdeXmlGenerator.ts:44`). |
-| `odeVersionId` | Every export / save | Every export / save | Always a freshly generated `generateOdeId()` call (`OdeXmlGenerator.ts:45`). Acts as a change cursor. |
+| `odeVersionId` | Project is created for the first time | Preserved across round-trips | Read from `meta.odeVersionId` when present, otherwise a freshly generated `generateOdeId()` (`OdeXmlGenerator.ts:45`). Imported from `<odeResources>` so an open/save round-trip is byte-stable. |
 | `odePageId` | Page is created | On import (reassigned — see below) | Stable within a project; changes only when the file is re-imported. |
 | `odeBlockId` | Block is created | On import (reassigned) | Stable within a project; changes only when the file is re-imported. |
 | `odeIdeviceId` | iDevice is added | On import (reassigned) | Stable within a project; changes only when the file is re-imported. |
--- a/doc/elpx-format/metadata.md
+++ b/doc/elpx-format/metadata.md
@@ -75,7 +75,7 @@ function generateOdeResourcesXml(odeId: string, versionId: string): string {
 | Key | Description |
 |---|---|
 | `odeId` | Stable project identifier. Retrieved from `meta.odeIdentifier` on export; generated once via `generateOdeId()` if absent. |
-| `odeVersionId` | Regenerated on every export/save via `generateOdeId()`. Acts as a change cursor. |
+| `odeVersionId` | Stable per-version identifier. Read from `meta.odeVersionId` on export; generated once via `generateOdeId()` only when absent. Imported from `<odeResources>` so a round-trip without content changes preserves the value. |
 | `exe_version` | The ODE format version string, currently `'3.0'` (constant `ODE_VERSION` from `constants.ts:1000`). |
 
 **Legacy keys observed in older fixtures:** the v4 generator writes exactly three resources (`odeId`, `odeVersionId`, `exe_version`). Some older fixtures additionally carry `odeVersionName`, `isDownload`, or the misspelling `eXeVersion`. The importer accepts all of these for backward compatibility, but **`generateOdeResourcesXml()` never emits them in v4** — treat them as informational legacy keys when reading older `.elpx` files.
--- a/src/shared/export/adapters/YjsDocumentAdapter.spec.ts
+++ b/src/shared/export/adapters/YjsDocumentAdapter.spec.ts
@@ -1154,3 +1154,59 @@ describe('YjsDocumentAdapter', () => {
         });
     });
 });
+
+describe('YjsDocumentAdapter.getMetadata - stable identifiers (#1784)', () => {
+    it('forwards odeIdentifier, odeVersionId and scormIdentifier from the Y.Map (mocked)', () => {
+        const manager = new MockYjsDocumentManager({
+            title: 'Stable IDs',
+            odeIdentifier: '20251201123456ABCDEF',
+            odeVersionId: '20251201123456FEDCBA',
+            scormIdentifier: 'CUSTOM-SCORM-XYZ',
+        });
+        const adapter = new YjsDocumentAdapter(
+            manager as unknown as ConstructorParameters<typeof YjsDocumentAdapter>[0],
+        );
+        const meta = adapter.getMetadata();
+        expect(meta.odeIdentifier).toBe('20251201123456ABCDEF');
+        expect(meta.odeVersionId).toBe('20251201123456FEDCBA');
+        expect(meta.scormIdentifier).toBe('CUSTOM-SCORM-XYZ');
+    });
+
+    it('returns undefined for unset stable identifiers (legacy projects)', () => {
+        const manager = new MockYjsDocumentManager({ title: 'Legacy' });
+        const adapter = new YjsDocumentAdapter(
+            manager as unknown as ConstructorParameters<typeof YjsDocumentAdapter>[0],
+        );
+        const meta = adapter.getMetadata();
+        expect(meta.odeIdentifier).toBeUndefined();
+        expect(meta.odeVersionId).toBeUndefined();
+        expect(meta.scormIdentifier).toBeUndefined();
+    });
+
+    it('reads stable identifiers from a real Y.Doc through YjsDocumentAdapter (no mocks)', async () => {
+        const Y = await import('yjs');
+        const doc = new Y.Doc();
+        const meta = doc.getMap('metadata');
+        meta.set('title', 'Real Y.Doc');
+        meta.set('odeIdentifier', '20251201123456ABCDEF');
+        meta.set('odeVersionId', '20251201123456FEDCBA');
+        // Minimal navigation so the adapter is happy.
+        doc.getArray('navigation');
+
+        const manager = {
+            getMetadata: () => meta,
+            getNavigation: () => doc.getArray('navigation'),
+            getDoc: () => doc,
+            projectId: 'real-yjs-project',
+        };
+        const adapter = new YjsDocumentAdapter(
+            manager as unknown as ConstructorParameters<typeof YjsDocumentAdapter>[0],
+        );
+
+        const exportMeta = adapter.getMetadata();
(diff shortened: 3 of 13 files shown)


mnarvaezm (changes requested):
The real Yjs -> export path still drops the identifiers.

`YjsDocumentAdapter.getMetadata()` does not return `odeIdentifier` or `odeVersionId`. This branch updates `BaseExporter.getManifestIdentifier()` to derive a stable manifest identifier from `meta.odeIdentifier`, but that only works if the exporter actually receives that field. In the browser/Yjs export path it currently does not, so the code falls back to a newly generated random identifier.

Local reproduction:

```text
Y.Map contains odeIdentifier = 20251201123456ABCDEF
adapter.getMetadata() does not contain odeIdentifier
generated manifest id = eXe-MANIFEST-20260511211310JCUVEK
```

The new tests pass because they inject metadata directly through mocks, for example:

```ts
new MockDocument({ odeIdentifier: '20251201123456ABCDEF' }, samplePages)
```

That verifies `BaseExporter` behaves correctly when `odeIdentifier` is already present in `ExportMetadata`, but it does not cover the real path from `Y.Doc` through `YjsDocumentAdapter`.

I also confirmed that `fix/1785-stable-scorm-manifest` does not currently include `fix/1783-preserve-page-ids-on-import` or `fix/1784-preserve-odeid-on-import`; it is based on `main`. Even if this PR is applied after those two, the adapter still needs to forward the imported identifier for the stable manifest logic to work.

Suggested fix in `YjsDocumentAdapter.getMetadata()`:

```ts
odeIdentifier: (meta.get('odeIdentifier') as string) || undefined,
odeVersionId: (meta.get('odeVersionId') as string) || undefined,
```

Please also add coverage that uses a real `Y.Map` through `YjsDocumentAdapter`, not only `MockDocument`, and validates that SCORM/IMS exports generate `eXe-MANIFEST-<odeIdentifier>` from the Yjs metadata.

erseco:
Thanks — the adapter gap is real and now fixed in `58c57593`.

**TDD proof:** added two failing tests in `YjsDocumentAdapter.spec.ts`:

```ts
describe('YjsDocumentAdapter.getMetadata - stable identifiers (#1784)', () => {
    it('forwards odeIdentifier, odeVersionId and scormIdentifier from the Y.Map (mocked)', ...);
    it('returns undefined for unset stable identifiers (legacy projects)', ...);
    it('reads stable identifiers from a real Y.Doc through YjsDocumentAdapter (no mocks)', ...);
});
```

The "real Y.Doc" test builds an actual `Y.Doc`, writes `odeIdentifier` / `odeVersionId` on its metadata Y.Map, wraps it in `YjsDocumentAdapter`, and asserts the fields appear in `ExportMetadata`. Before the fix it red-lined with `Expected: "20251201123456ABCDEF" / Received: undefined`.

**Fix in `YjsDocumentAdapter.getMetadata()`** — added the three missing fields right next to the existing `screenshot` pass-through:

```ts
odeIdentifier:   (meta.get('odeIdentifier')   as string) || undefined,
odeVersionId:    (meta.get('odeVersionId')    as string) || undefined,
scormIdentifier: (meta.get('scormIdentifier') as string) || undefined,
```

With this in place the chain from `Y.Doc` → adapter → `ExportMetadata` → `BaseExporter.getManifestIdentifier()` carries the imported identifier end-to-end. Full adapter suite: 61 pass / 0 fail; `make fix` clean.

mnarvaezm (commented):
```md
The main import/export flow looks fixed now: a normal `content.xml` with both `<odeResources>` and `<odeProperties>` preserves `odeId` and `odeVersionId` through import -> Yjs metadata -> `YjsDocumentAdapter` -> `generateOdeXml()`.

However, there is still one edge case worth fixing before closing this PR.

`<odeProperties>` is optional according to the DTD:

```xml
<!ELEMENT ode (userPreferences?, odeResources?, odeProperties?, odeNavStructures)>
```

But `ElpxImporter.importStructure()` only writes metadata when `<odeProperties>` exists:

```ts
if (odeProperties && clearExisting) {
    this.setMetadata(metadata, metadataValues);
}
```

That means a valid `content.xml` containing `<odeResources>` but no `<odeProperties>` will parse `odeId` / `odeVersionId`, place them into `metadataValues`, and then never write them into the Y.Doc metadata.

I reproduced this locally:

```text
content.xml contains <odeResources> with odeId / odeVersionId
content.xml does not contain <odeProperties>
resulting Yjs metadata: {}
```

Suggested fix: decouple writing stable identifiers from the existence of `<odeProperties>`. For example, call `setMetadata()` when either `<odeProperties>` exists or `metadataValues` contains imported resource identifiers:

```ts
const hasStableIdentifiers = Boolean(metadataValues.odeIdentifier || metadataValues.odeVersionId);

if (clearExisting && (odeProperties || hasStableIdentifiers)) {
    this.setMetadata(metadata, metadataValues);
}
```

Please also add a regression test for a v4 `content.xml` with `<odeResources>` and no `<odeProperties>`, verifying that `metadata.odeIdentifier` and `metadata.odeVersionId` are still populated.

Minor follow-up: the format docs still say `odeVersionId` is regenerated on every export/save, but this PR now preserves `meta.odeVersionId` when present. Those docs should be updated to match the new behavior.
```

**Comment by erseco** [comment]
Both edge cases addressed in `7dc3f24b`.

### 1. Stable identifiers when `<odeProperties>` is absent

You're right that the DTD makes `<odeProperties>` optional, and the gate at line 491 was tied to it. A minimal v4 `content.xml` with `<odeResources>` alone would extract `odeId` / `odeVersionId` into `metadataValues` and then drop them on the floor.

**TDD proof** — failing test added before the fix:

```
Expected: "20251201123456ABCDEF"
Received: undefined
(fail) odeResources preservation > should still populate metadata.odeIdentifier
       when <odeProperties> is absent (DTD edge case)
```

**Fix** — decouple the gate from `<odeProperties>`:

```ts
const hasStableIdentifiers = Boolean(metadataValues.odeIdentifier || metadataValues.odeVersionId);
if (clearExisting && (odeProperties || hasStableIdentifiers)) {
    this.setMetadata(metadata, metadataValues);
}
```

### 2. Format docs updated

Updated the two lines that still described `odeVersionId` as regenerated every export/save:

- `doc/elpx-format/ids.md:72` — set-when/changes-when columns now reflect "Project is created for the first time" / "Preserved across round-trips", with a note that it's read from `meta.odeVersionId` when present.
- `doc/elpx-format/metadata.md:78` — description rewritten to match the new behaviour and the import path.

### Verification

- `bun test src/shared/import/ src/shared/export/generators/` → **810 pass / 0 fail** (4 new tests in this PR cycle).
- `make fix` clean.

erseco:
Superseded by #1791. Scope (preserve `odeId`/`odeVersionId` on import + `YjsDocumentAdapter` forwarding + `<odeProperties>`-absent DTD edge + doc updates, #1784) merged verbatim via `Merge #1789`. The bundle also adds the end-to-end bridge test (`test/integration/stable-identifiers.spec.ts`) that catches a regression in either the importer OR the adapter — which the per-file specs could not.

PR #1790, opened 2026-05-11
fix(export): stable SCORM/IMS manifest identifier

### LINKED ISSUE #1785: Stable SCORM/IMS manifest identifier (derive from `odeIdentifier`, drop per-export regeneration)

erseco:
## Summary

SCORM 1.2, SCORM 2004 and IMS CP exports each called `Date.now() + Math.random()` to build `manifest@identifier`, breaking LMS tracking on every re-upload.

`BaseExporter.getManifestIdentifier()` now derives a stable identifier:
1. `meta.scormIdentifier` if set (user override),
2. `'eXe-MANIFEST-' + meta.odeIdentifier` (default — same project identity used in `content.xml`),
3. `'eXe-MANIFEST-' + generateOdeId()` (fallback for legacy projects).

The same root identifier flows through the manifest, the organization id and the LOM `catalog/entry`. Manifest generators accept an optional `manifestIdentifier` override and use it verbatim; otherwise they fall back to the legacy `eXe-MANIFEST-<bareId>` shape for backward compatibility.

Full user-visible win (LMS tracking survives open/save) depends on #1784 landing too, since that PR is what populates `meta.odeIdentifier` on import. This PR is self-contained and tested in isolation by setting `meta.odeIdentifier` directly in fixtures.

## Test plan

- [x] Existing `id1 !== id2` tests flipped to assert stability with `meta.odeIdentifier` set
- [x] User override via `meta.scormIdentifier` wins
- [x] Fallback when neither is set still produces a valid manifest identifier (`/^eXe-MANIFEST-\d{14}[A-Z0-9]{6}$/`)
- [x] Cross-consistency: LOM `catalog/entry` and `manifest@identifier` share the same id root
- [x] Sanity check: two consecutive exports of the same project produce byte-identical `imsmanifest.xml`
- [x] `make test-unit` green (6927 pass, 0 fail)
- [x] `make test-integration` green (700 pass, 0 fail)
- [x] `make fix` / `make lint` clean
- [x] Patch coverage ≥ 90% on changed lines (BaseExporter 98.72% lines, manifest generators ≥ 94%)

Closes #1785

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -11,7 +11,7 @@
         "@elysiajs/static": "^1.4.7",
         "@elysiajs/swagger": "^1.3.1",
         "@sinclair/typebox": "^0.34.48",
-        "@xmldom/xmldom": "^0.8.11",
+        "@xmldom/xmldom": "^0.9.9",
         "bcryptjs": "^3.0.3",
         "dotenv": "^17.3.1",
         "elysia": "^1.4.27",
@@ -29,8 +29,8 @@
         "mime-types": "^3.0.2",
         "mysql2": "^3.18.2",
         "nunjucks": "^3.2.4",
-        "pdfjs-dist": "5.5.207",
-        "uuid": "^13.0.0",
+        "pdfjs-dist": "5.7.284",
+        "uuid": "^14.0.0",
         "ws": "^8.19.0",
         "y-websocket": "^3.0.0",
         "yjs": "^13.6.29",
@@ -39,7 +39,7 @@
         "@babel/core": "^7.29.0",
         "@babel/preset-env": "^7.29.0",
         "@biomejs/biome": "^2.4.5",
-        "@codecov/bundle-analyzer": "^1.9.1",
+        "@codecov/bundle-analyzer": "^2.0.1",
         "@electron/notarize": "^3.1.1",
         "@playwright/test": "^1.58.2",
         "@types/bcryptjs": "^3.0.0",
@@ -52,14 +52,14 @@
         "@vitest/coverage-v8": "^4.0.18",
         "@vitest/ui": "^4.0.18",
         "cross-env": "^10.1.0",
-        "electron": "^41.0.0",
+        "electron": "^42.0.0",
         "electron-builder": "^26.8.1",
-        "esbuild": "^0.27.3",
+        "esbuild": "^0.28.0",
         "happy-dom": "^20.8.0",
-        "http-proxy-middleware": "^3.0.5",
+        "http-proxy-middleware": "^4.0.0",
         "kill-port": "^2.0.1",
         "sass": "^1.97.3",
-        "typescript": "^5.9.3",
+        "typescript": "^6.0.2",
         "vite": "^8.0.0",
         "vitest": "^4.0.18",
         "wait-on": "^9.0.4",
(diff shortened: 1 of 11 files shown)


mnarvaezm (changes requested):
This PR improves the manifest identifier logic, but there are still two problems that can break the intended behavior.

## 1. The real Yjs → export path drops the identifiers

`BaseExporter.getManifestIdentifier()` now derives the manifest identifier from `meta.odeIdentifier`, or from `meta.scormIdentifier` when the user override is set.

That logic is correct **only if** the exporter receives those metadata fields.

In the real browser export path, metadata comes from `YjsDocumentAdapter.getMetadata()`. That adapter currently does **not** return:

- `odeIdentifier`
- `odeVersionId`
- `scormIdentifier`

So even when the project has those values stored in the Yjs document, the exporter never sees them and falls back to generating a new random manifest identifier.

### Local reproduction

```text
Y.Map contains          odeIdentifier = 20251201123456ABCDEF
adapter.getMetadata()   does not contain odeIdentifier
generated manifest id   = eXe-MANIFEST-20260511211310JCUVEK
```

### Consequences

- Browser/Yjs SCORM and IMS exports can still get a new manifest identifier on each export.
- LMS tracking may still break on re-upload, because the LMS may treat the updated package as a different course.
- `meta.scormIdentifier` user overrides will not work in the real Yjs export path if they are stored in project metadata.

The new tests can pass while the application path remains broken, because the tests inject metadata directly through mocks:

```ts
new MockDocument({ odeIdentifier: '20251201123456ABCDEF' }, samplePages)
```

That verifies `BaseExporter` when metadata is already correct, but it does not test the actual route from `Y.Doc` through `YjsDocumentAdapter`.

### Proposed fix

Forward the missing metadata fields from `YjsDocumentAdapter.getMetadata()`:

```ts
odeIdentifier:   (meta.get('odeIdentifier')   as string) || undefined,
odeVersionId:    (meta.get('odeVersionId')    as string) || undefined,
scormIdentifier: (meta.get('scormIdentifier') as string) || undefined,
```

Also add a test that uses a real `Y.Map` through `YjsDocumentAdapter` (not only `MockDocument`) and verifies that SCORM/IMS exports generate:

```text
eXe-MANIFEST-<odeIdentifier>
```

from Yjs metadata.

---

## 2. The fallback path can generate internally inconsistent IDs

The exporters now do this:

1. Calculate `manifestIdentifier`.
2. Calculate `projectId`.

But both values are derived through separate calls. When neither `odeIdentifier` nor `scormIdentifier` exists, each call can generate a different fallback ID.

### Local reproduction (SCORM 1.2, no IDs in metadata)

```text
manifest:     eXe-MANIFEST-20260511212158K1JVQA
organization: eXe-20260511212158U6OWBT
LOM entry:    ODE-20260511212158U6OWBT
```

The manifest uses one generated root, while the organization and LOM entry use another.

Before this PR, a single `projectId` was used consistently for all those values.

### Consequences

- A single exported package can contain multiple unrelated project identifiers.
- The manifest identifier, organization identifier, and LOM catalog/entry may no longer describe the same package identity.
- LMSs or validation tools that expect consistent identifiers may behave unpredictably.

The current fallback tests do not catch this because they only verify that `manifest@identifier` has a valid format.

### Proposed fix

Generate the fallback identity once and derive both values from that same root.

For example, replace the two independent calls with a single helper that returns both:

```ts
{
  manifestIdentifier: 'eXe-MANIFEST-' + rootId,
  projectId: rootId,
}
```

Where `rootId` is chosen once from:

1. `meta.scormIdentifier` if set.
2. `meta.odeIdentifier` if set.
3. One generated fallback ID.

Alternatively, make `getBareProjectIdentifier()` accept the already computed `manifestIdentifier` instead of calling `getManifestIdentifier()` again.

Add fallback coverage that verifies all related identifiers share the same root when no metadata identifier is available:

- `manifest@identifier`
- `organization@identifier`
- LOM catalog/entry

### Expected shape

```text
manifest:     eXe-MANIFEST-<root>
organization: eXe-<root>
LOM entry:    ODE-<root>
```

erseco:
Both issues fixed:

**1. Adapter forwarding** — separate concern, fixed in PR #1789 (`YjsDocumentAdapter.getMetadata()` now forwards `odeIdentifier`/`odeVersionId`/`scormIdentifier` to `ExportMetadata`, with a real-`Y.Doc` test). This unblocks the Yjs → export path that previously dropped the stable identifier before it ever reached `BaseExporter.getManifestIdentifier()`.

**2. Fallback path inconsistency** — fixed in this PR at `5657991b`.

**TDD proof:** added a failing fallback-consistency test for each exporter (SCORM 1.2, SCORM 2004, IMS CP). The Scorm12 case red-lined with:

```
Expected: "20260511202900VRKW22"
Received: "20260511202900S7VMYL"
```

Two unrelated roots: the manifest used one and the organization another, because `getBareProjectIdentifier()` invoked `getManifestIdentifier()` a second time and got a fresh `generateOdeId()`.

**Fix in `BaseExporter`** — memoize the resolved manifest identifier on the instance:

```ts
private _manifestIdentifier: string | undefined;

protected getManifestIdentifier(): string {
    if (this._manifestIdentifier !== undefined) return this._manifestIdentifier;
    const meta = this.getMetadata();
    if      (meta.scormIdentifier) this._manifestIdentifier = meta.scormIdentifier;
    else if (meta.odeIdentifier)   this._manifestIdentifier = 'eXe-MANIFEST-' + meta.odeIdentifier;
    else                            this._manifestIdentifier = 'eXe-MANIFEST-' + generateOdeId();
    return this._manifestIdentifier;
}
```

`getBareProjectIdentifier()` still strips the prefix off the memoized value, so the entire export observes one consistent root — manifest, organization (`eXe-<root>`) and LOM `catalog/entry` (`ODE-<root>`) all match.

New tests pin the shared-root invariant on the fallback path:

```ts
// SCORM 1.2 / 2004
expect(orgMatch![1]).toBe(root);
expect(lomEntryMatch![1]).toBe(root);

// IMS CP
expect(orgMatch![1]).toBe(manifestMatch![1]);
```

Full exporter suite: 632 pass / 0 fail; `make fix` clean.

mnarvaezm (approved):

erseco:
Superseded by #1791. Scope (stable SCORM/IMS manifest identifier derived from `odeIdentifier` + memoized fallback consistency, #1785) merged verbatim via `Merge #1790`. The bundle resolves the cross-PR concern you flagged in the review: with #1784's adapter forwarding now alongside this fix, the full Yjs → export path delivers a stable `eXe-MANIFEST-<odeId>` so LMS course tracking survives re-uploads.

PR #1787, opened 2026-05-11
fix(ids): unify generateId(prefix) into shared module

erseco:
## Summary

Collapse the four parallel `generateId(prefix)` implementations into a single canonical module at `src/shared/ids.ts`. Eliminates mixed-format IDs inside the same `content.xml`.

- `src/yjs/structure-binding.ts` re-exports from the shared module to preserve the public API.
- `src/shared/import/ElpxImporter.ts` imports directly; private method removed.
- The two browser-side sites (`public/app/yjs/YjsStructureBinding.js`, `public/app/yjs/ComponentImporter.js`) keep byte-equivalent mirror copies of Format A with a `// keep in sync` comment. Format B is gone.
- Deprecated `String.prototype.substr` replaced with `.substring(2, 11)`.

## Test plan

- [x] New `src/shared/ids.spec.ts` (5 cases, 100% coverage on the new file)
- [x] Existing `bun test src/yjs/structure-binding.spec.ts` passes (92 tests)
- [x] Existing `bun test src/shared/import/ElpxImporter.spec.ts` passes (58 tests)
- [x] Frontend specs (`YjsStructureBinding.test.js`, `ComponentImporter.test.js`) pass (281 tests)
- [x] `make fix` and `make lint` clean
- [x] `make bundle` regenerates `public/app/yjs/importers.bundle.js` with the canonical helper (gitignored)

Closes #1782

PR #1791, opened 2026-05-12
fix(ids): unify and harden the four stable-identifier PRs (#1786 bundle)

### LINKED ISSUE #1782: Unify the four parallel `generateId(prefix)` implementations into a single shared module
## Summary

The codebase ships **two different runtime formats** for `generateId(prefix)` used for pages, blocks and iDevices, plus two other near-duplicates that use deprecated APIs. The format that ends up in `content.xml` depends on which code path created the entity, so a single project can carry mixed-format identifiers.

This is sub-issue **A** of the *Stable identifiers* tracking issue.

## Evidence

**Format A** — `${prefix}-${Date.now().toString(36)}-${Math.random().toString(36).substring(2, 11)}` → `page-mp0fppw7-8djmx3rm8`

- `src/yjs/structure-binding.ts:38` (server CRUD)
- `src/shared/import/ElpxImporter.ts:1824` (ELPX import, TS)
- `public/app/yjs/ComponentImporter.js:473` (browser, uses deprecated `String.prototype.substr`)
- `public/app/yjs/importers.bundle.js:13601` (compiled bundle of the TS importer)

**Format B** — `${prefix}-${Date.now()}-${Math.random().toString(36).substr(2, 9)}` → `page-1778458980537-3ekb41j4d`

- `public/app/yjs/YjsStructureBinding.js:2824` (browser, fires when the user clicks *"+ Add page"* in the workarea)

Two unrelated `generateId` symbols exist for different purposes and are **not** duplicates of the above — leave them alone:

- `src/utils/id-generator.util.ts:37` — Symfony-compatible `YYYYMMDDHHmmss + 6 letters`, used for project/session IDs and file-system layout
- `src/shared/export/exporters/BaseExporter.ts:359` — uppercase no-dash variant used by manifest/LOM generators

## Why this matters

Concrete reproduction: a project imported from `.elpx` and then extended via *"+ Add page"* ends up with both formats in the same `content.xml`:

```xml
<odePageId>page-mp0fppwf-71v64kl4r</odePageId>      <!-- A: imported -->
<odePageId>page-1778458980537-3ekb41j4d</odePageId> <!-- B: added later in UI -->
```

Consequences:

- **No single source of truth.** A future change to the ID format (e.g. UUIDv7, stronger randomness) has to be made in four places at once. The next person who forgets to update one will silently re-introduce divergence.
- **Inconsistent lengths** (8 vs 13 chars in the timestamp segment) break any regex, parser, or URL helper that assumes a fixed shape.
- **Deprecated `String.prototype.substr`** in two of the call sites is a latent bug — the spec is gone, behaviour is preserved by convention only.
- **Tests are fragile**: a test that pins one format passes for one code path and silently misses the other.

## Proposal

1. Create `src/shared/ids.ts` exporting a single canonical `generateId(prefix: string): string` using **Format A** (already the dominant form, 3 of 4 sites).
2. Update the four call sites to import from the shared module:
   - `src/yjs/structure-binding.ts` → import and re-export, keep public API
   - `src/shared/import/ElpxImporter.ts` → drop the private method, import directly
   - `public/app/yjs/YjsStructureBinding.js` → import (or, if bundling pipeline forbids, ship `public/app/shared/ids.js` as the single source the bundle also resolves to)
   - `public/app/yjs/ComponentImporter.js` → same as above
3. Verify the importers bundle (`public/app/yjs/importers.bundle.js`) is rebuilt from the updated TS so the two stay in sync — that file is generated, not edited.
4. Replace `.substr(2, 9)` with `.substring(2, 11)` in the unified function (matches Format A; same entropy, removes the deprecated call).
5. Add a unit test that asserts the same prefix pattern from both the server and the browser path: `/^page-[a-z0-9]{8}-[a-z0-9]{9}$/`.

## Scope guard

This issue **only unifies the function**. It does **not**:

- Change page/block/iDevice IDs that already exist in saved projects (they remain valid — same prefix, same alphabet).
- Switch to UUIDv7 or any other scheme. The shared module makes that a one-line change later, but it is out of scope here.
- Touch `BaseExporter.generateId` or `id-generator.util.ts` — different concerns.

## Acceptance

- All four call sites import from `src/shared/ids.ts`.
- `grep -rn "Date.now().toString(36).*Math.random()" public/app src` returns only the canonical module.
- `make test-unit` passes; the new shared-module test passes from both the Bun side and the Vitest side.
- `make fix && make lint` clean.

## Related

Part of the *Stable identifiers across import/export* tracking issue (to be filed).

### LINKED ISSUE #1783: Preserve page/block/iDevice IDs across ELPX v4 import (stop regenerating on every import)

### LINKED ISSUE #1784: Preserve `odeId` and `odeVersionId` across ELPX import (read `<odeResources>`)

### LINKED ISSUE #1785: Stable SCORM/IMS manifest identifier (derive from `odeIdentifier`, drop per-export regeneration)

### LINKED ISSUE #1786: Stable identifiers across import/export — unify ID generation and preserve project identity

erseco:
## Summary

Single landable bundle for the four parallel fixes that close issue #1786 (and its four sub-issues #1782–#1785). Replaces the four open PRs #1787–#1790 with one self-contained branch, plus the coverage gaps each PR could not test in isolation.

The original bug, end-to-end: opening and re-downloading a `.elpx` rewrote every page/block/iDevice id and the SCORM `manifest@identifier`, so LMS course-tracking broke on every re-upload and content.xml diffs were unusable.

### Bundled fixes (one commit per upstream PR)

| Closes | Area | Fix |
|---|---|---|
| #1782 | `src/shared/ids.ts` (new), `structure-binding.ts`, `ElpxImporter.ts`, `ComponentImporter.js`, `YjsStructureBinding.js` | Single canonical `generateId(prefix)`. Two browser sites keep byte-equivalent mirror copies tagged `// keep in sync`. Deprecated `.substr` replaced with `.substring(2, 11)`. |
| #1783 | `ElpxImporter.buildFlatPageList`, `remapInternalPageLinks` | Preserve `<odePageId>` / `<odeBlockId>` / `<odeIdeviceId>` verbatim when no collision. Merge-mode remaps only on real collisions. Boundary lookahead `(?![A-Za-z0-9_-])` + descending-length sort fix the `page-1` / `page-10` partial-prefix remap. |
| #1784 | `ElpxImporter.extractOdeResources` (new), `metadata-properties.ts`, `OdeXmlGenerator.ts`, `YjsDocumentAdapter.getMetadata`, doc updates | Read `<odeResources>` on import; forward `odeIdentifier` / `odeVersionId` / `scormIdentifier` from the Yjs metadata Y.Map; legacy v3 generates the pair once; `versionId` fallback chain mirrors `odeId`. Gate also fires when `<odeProperties>` is absent (valid per DTD). |
| #1785 | `BaseExporter.getManifestIdentifier` (memoized), `Scorm12 / Scorm2004 / Ims` exporters + generators | `manifest@identifier` derives deterministically from `meta.scormIdentifier` > `'eXe-MANIFEST-' + meta.odeIdentifier` > `'eXe-MANIFEST-' + generateOdeId()`. The same root flows into `organization@identifier` (`eXe-<root>`) and the LOM `catalog/entry` (`ODE-<root>`). Memoization prevents the fallback path from generating three unrelated roots. |

Each merge commit (`Merge #1787…#1790`) preserves the original commit history so review attribution stays intact.

### Coverage gaps closed in this bundle

The four parallel PRs covered their own files but could not test the chain between them. This PR adds:

1. **`public/app/yjs/YjsProjectManagerMixin.js`** — three additional sites that minted ids in Format B (decimal timestamp + deprecated `.substr`) discovered while diffing against the existing `hotfix/id-mismatch-issues` branch. Routed through a canonical helper that mirrors `src/shared/ids.ts`. Test added in `YjsProjectManagerMixin.test.js`.

2. **`test/integration/stable-identifiers.spec.ts`** — end-to-end bridge test (XML → `ElpxImporter` → Y.Doc metadata + navigation → `YjsDocumentAdapter` → `ExportMetadata`). This is the gap the reviewer flagged on #1789 / #1790: a regression in import or adapter previously made the per-exporter unit tests pass while the real Yjs export path silently fell back to a random id.

## Verification

- `make test-unit-ci` → **6956 pass / 0 fail**
- `bun test test/integration/` → **704 pass / 0 fail** (4 new bridge tests included)
- Frontend `YjsProjectManagerMixin.test.js` → **107 / 107 pass**
- `make fix` / `make lint` → clean

## End-to-end invariant after merge

1. Round-trip a `.elpx` twice without changes → diff of `content.xml` is empty (modulo legitimate timestamps).
2. Two SCORM 1.2 / 2004 / IMS exports of the same project → byte-identical `imsmanifest.xml`.
3. Re-uploading the updated SCORM package to Moodle / Canvas preserves learner tracking.
4. Mixed-format ids in already-saved projects continue to work; new ids are uniform.

## Out of scope

- Migrating existing saved projects to UUIDv7 or any other scheme.
- Bumping `odeVersionId` automatically when content changes (separate follow-up).
- Rewriting legacy `.elp` v3 ids — short collision-prone, still regenerated on import.

Closes #1782
Closes #1783
Closes #1784
Closes #1785
Closes #1786

Supersedes #1787
Supersedes #1788
Supersedes #1789
Supersedes #1790

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -11,7 +11,7 @@
         "@elysiajs/static": "^1.4.7",
         "@elysiajs/swagger": "^1.3.1",
         "@sinclair/typebox": "^0.34.48",
-        "@xmldom/xmldom": "^0.8.11",
+        "@xmldom/xmldom": "^0.9.9",
         "bcryptjs": "^3.0.3",
         "dotenv": "^17.3.1",
         "elysia": "^1.4.27",
@@ -29,8 +29,8 @@
         "mime-types": "^3.0.2",
         "mysql2": "^3.18.2",
         "nunjucks": "^3.2.4",
-        "pdfjs-dist": "5.5.207",
-        "uuid": "^13.0.0",
+        "pdfjs-dist": "5.7.284",
+        "uuid": "^14.0.0",
         "ws": "^8.19.0",
         "y-websocket": "^3.0.0",
         "yjs": "^13.6.29",
@@ -39,7 +39,7 @@
         "@babel/core": "^7.29.0",
         "@babel/preset-env": "^7.29.0",
         "@biomejs/biome": "^2.4.5",
-        "@codecov/bundle-analyzer": "^1.9.1",
+        "@codecov/bundle-analyzer": "^2.0.1",
         "@electron/notarize": "^3.1.1",
         "@playwright/test": "^1.58.2",
         "@types/bcryptjs": "^3.0.0",
@@ -52,14 +52,14 @@
         "@vitest/coverage-v8": "^4.0.18",
         "@vitest/ui": "^4.0.18",
         "cross-env": "^10.1.0",
-        "electron": "^41.0.0",
+        "electron": "^42.0.0",
         "electron-builder": "^26.8.1",
-        "esbuild": "^0.27.3",
+        "esbuild": "^0.28.0",
         "happy-dom": "^20.8.0",
-        "http-proxy-middleware": "^3.0.5",
+        "http-proxy-middleware": "^4.0.0",
         "kill-port": "^2.0.1",
         "sass": "^1.97.3",
-        "typescript": "^5.9.3",
+        "typescript": "^6.0.2",
         "vite": "^8.0.0",
         "vitest": "^4.0.18",
         "wait-on": "^9.0.4",
@@ -69,15 +69,15 @@
   "packages": {
     "7zip-bin": ["7zip-bin@5.2.0", "", {}, "sha512-ukTPVhqG4jNzMro2qA9HSCSSVJN3aN7tlb+hfqYCt3ER0yWroeA2VR38MNrOHLQ/cVj+DaIMad0kFCtWWowh/A=="],
 
-    "@actions/core": ["@actions/core@1.11.1", "", { "dependencies": { "@actions/exec": "^1.1.1", "@actions/http-client": "^2.0.1" } }, "sha512-hXJCSrkwfA46Vd9Z3q4cpEpHB1rL5NG04+/rbqW9d3+CSvtB1tYe8UTpAlixa1vj0m/ULglfEK2UKxMGxCxv5A=="],
+    "@actions/core": ["@actions/core@3.0.1", "", { "dependencies": { "@actions/exec": "^3.0.0", "@actions/http-client": "^4.0.0" } }, "sha512-a6d/Nwahm9fliVGRhdhofo40HjHQasUPusmc7vBfyky+7Z+P2A1J68zyFVaNcEclc/Se+eO595oAr5nwEIoIUA=="],
 
-    "@actions/exec": ["@actions/exec@1.1.1", "", { "dependencies": { "@actions/io": "^1.0.1" } }, "sha512-+sCcHHbVdk93a0XT19ECtO/gIXoxvdsgQLzb2fE2/5sIZmWQuluYyjPQtrtTHdU1YzTZ7bAPN4sITq2xi1679w=="],
+    "@actions/exec": ["@actions/exec@3.0.0", "", { "dependencies": { "@actions/io": "^3.0.2" } }, "sha512-6xH/puSoNBXb72VPlZVm7vQ+svQpFyA96qdDBvhB8eNZOE8LtPf9L4oAsfzK/crCL8YZ+19fKYVnM63Sl+Xzlw=="],
 
-    "@actions/github": ["@actions/github@6.0.1", "", { "dependencies": { "@actions/http-client": "^2.2.0", "@octokit/core": "^5.0.1", "@octokit/plugin-paginate-rest": "^9.2.2", "@octokit/plugin-rest-endpoint-methods": "^10.4.0", "@octokit/request": "^8.4.1", "@octokit/request-error": "^5.1.1", "undici": "^5.28.5" } }, "sha512-xbZVcaqD4XnQAe35qSQqskb3SqIAfRyLBrHMd/8TuL7hJSz2QtbDwnNM8zWx4zO5l2fnGtseNE3MbEvD7BxVMw=="],
+    "@actions/github": ["@actions/github@9.1.1", "", { "dependencies": { "@actions/http-client": "^3.0.2", "@octokit/core": "^7.0.6", "@octokit/plugin-paginate-rest": "^14.0.0", "@octokit/plugin-rest-endpoint-methods": "^17.0.0", "@octokit/request": "^10.0.7", "@octokit/request-error": "^7.1.0", "undici": "^6.23.0" } }, "sha512-tL5JbYOBZHc0ngEnCsaDcryUizIUIlQyIMwy1Wkx93H5HzbBJ7TbiPx2PnFjBwZW0Vh05JmfFZhecE6gglYegA=="],
 
-    "@actions/http-client": ["@actions/http-client@2.2.3", "", { "dependencies": { "tunnel": "^0.0.6", "undici": "^5.25.4" } }, "sha512-mx8hyJi/hjFvbPokCg4uRd4ZX78t+YyRPtnKWwIl+RzNaVuFpQHfmlGVfsKEJN8LwTCvL+DfVgAM04XaHkm6bA=="],
+    "@actions/http-client": ["@actions/http-client@4.0.1", "", { "dependencies": { "tunnel": "^0.0.6", "undici": "^6.23.0" } }, "sha512-+Nvd1ImaOZBSoPbsUtEhv+1z99H12xzncCkz0a3RuehINE81FZSe2QTj3uvAPTcJX/SCzUQHQ0D1GrPMbrPitg=="],
 
-    "@actions/io": ["@actions/io@1.1.3", "", {}, "sha512-wi9JjgKLYS7U/z8PPbco+PvTb/nRWjeoFlJ1Qer83k/3C5PHQi28hiVdeE2kHXmIL99mQFawx8qt/JPjZilJ8Q=="],
+    "@actions/io": ["@actions/io@3.0.2", "", {}, "sha512-nRBchcMM+QK1pdjO7/idu86rbJI5YHUKCvKs0KxnSYbVe3F51UfGxuZX4Qy/fWlp6l7gWFwIkrOzN+oUK03kfw=="],
 
     "@antfu/install-pkg": ["@antfu/install-pkg@1.1.0", "", { "dependencies": { "package-manager-detector": "^1.3.0", "tinyexec": "^1.0.1" } }, "sha512-MGQsmw10ZyI+EJo45CdSER4zEb+p31LpDAFp2Z3gkSd1yqVZGi0Ebx++YTEMonJy4oChEMLsxZ64j8FH6sSqtQ=="],
 
@@ -301,9 +301,9 @@
 
     "@chevrotain/utils": ["@chevrotain/utils@11.1.1", "", {}, "sha512-71eTYMzYXYSFPrbg/ZwftSaSDld7UYlS8OQa3lNnn9jzNtpFbaReRRyghzqS7rI3CDaorqpPJJcXGHK+FE1TVQ=="],
 
-    "@codecov/bundle-analyzer": ["@codecov/bundle-analyzer@1.9.1", "", { "dependencies": { "@codecov/bundler-plugin-core": "^1.9.1", "micromatch": "^4.0.8", "yargs": "^17.7.2" }, "os": [ "linux", "darwin", ], "bin": { "bundle-analyzer": "dist/cli.cjs" } }, "sha512-WVWZIV0v09RueL0L2x8pgaWzGsJsKMsMFjsFgzfjeApq9cE6bZ+0fQC3f/pTQ8h40Y3D5Gk9o4W6jd/if3ohSA=="],
+    "@codecov/bundle-analyzer": ["@codecov/bundle-analyzer@2.0.1", "", { "dependencies": { "@codecov/bundler-plugin-core": "^2.0.1", "micromatch": "^4.0.8", "yargs": "^17.7.2" }, "os": [ "linux", "darwin", ], "bin": { "bundle-analyzer": "dist/cli.cjs" } }, "sha512-SkL0+Ccs6VuHIA9Bxb8/J110CffRXpqtKt7m18FTWC0P6gDKEsoieUerDdcd6K9BFKujj6gaUdTt5i4I/v8GTA=="],
 
-    "@codecov/bundler-plugin-core": ["@codecov/bundler-plugin-core@1.9.1", "", { "dependencies": { "@actions/core": "^1.10.1", "@actions/github": "^6.0.0", "chalk": "4.1.2", "semver": "^7.5.4", "unplugin": "^1.10.1", "zod": "^3.22.4" } }, "sha512-dt3ic7gMswz4p/qdkYPVJwXlLiLsz55rBBn2I7mr0HTG8pCoLRqnANJIwo5WrqGBZgPyVSMPBqBra6VxLWfDyA=="],
+    "@codecov/bundler-plugin-core": ["@codecov/bundler-plugin-core@2.0.1", "", { "dependencies": { "@actions/core": "^3.0.0", "@actions/github": "^9.0.0", "chalk": "4.1.2", "semver": "^7.5.4", "unplugin": "^1.10.1", "zod": "^3.22.4" } }, "sha512-TkdKn/rEwZQ723M7DDUmHe5r0IJa23rUT4TAx5jXmg12wGZGAHGWWU7LKeQsYCsKdLMxK7bLaGk9M++4wSRD5w=="],
 
     "@csstools/color-helpers": ["@csstools/color-helpers@6.0.2", "", {}, "sha512-LMGQLS9EuADloEFkcTBR3BwV/CGHV7zyDxVRtVDTwdI2Ca4it0CCVTT9wCkxSgokjE5Ho41hEPgb8OEUwoXr6Q=="],
 
@@ -323,7 +323,7 @@
 
     "@electron/fuses": ["@electron/fuses@1.8.0", "", { "dependencies": { "chalk": "^4.1.1", "fs-extra": "^9.0.1", "minimist": "^1.2.5" }, "bin": { "electron-fuses": "dist/bin.js" } }, "sha512-zx0EIq78WlY/lBb1uXlziZmDZI4ubcCXIMJ4uGjXzZW0nS19TjSPeXPAjzzTmKQlJUZm0SbmZhPKP7tuQ1SsEw=="],
 
-    "@electron/get": ["@electron/get@2.0.3", "", { "dependencies": { "debug": "^4.1.1", "env-paths": "^2.2.0", "fs-extra": "^8.1.0", "got": "^11.8.5", "progress": "^2.0.3", "semver": "^6.2.0", "sumchecker": "^3.0.1" }, "optionalDependencies": { "global-agent": "^3.0.0" } }, "sha512-Qkzpg2s9GnVV2I2BjRksUi43U5e6+zaQMcjoJy0C+C5oxaKl+fmckGDQFtRpZpZV0NQekuZZ+tGz7EA9TVnQtQ=="],
+    "@electron/get": ["@electron/get@5.0.0", "", { "dependencies": { "debug": "^4.1.1", "env-paths": "^3.0.0", "graceful-fs": "^4.2.11", "progress": "^2.0.3", "semver": "^7.6.3", "sumchecker": "^3.0.1" }, "optionalDependencies": { "undici": "^7.24.4" } }, "sha512-pjoBpru1KdEtcExBnuHAP1cAc/5faoedw0hzJkL3o4/IJp7HNF1+fbrdxT3gMYRX2oJfvnA/WXeCTVQpYYxyJA=="],
 
     "@electron/notarize": ["@electron/notarize@3.1.1", "", { "dependencies": { "debug": "^4.4.0", "promise-retry": "^2.0.1" } }, "sha512-uQQSlOiJnqRkTL1wlEBAxe90nVN/Fc/hEmk0bqpKk8nKjV1if/tXLHKUPePtv9Xsx90PtZU8aidx5lAiOpjkQQ=="],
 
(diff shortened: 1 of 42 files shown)


erseco:
Copilot review acknowledged — 6/6 comments evaluated, all valid, all applied in `a61e8b61`. Summary of the verdicts:

| # | File / line | Verdict | Action |
|---|---|---|---|
| 1 | `ElpxImporter.ts:432` — `Record<string,string>` lies about missing keys | Valid (soft) | Return type → `Partial<Record<string,string>>`. The previous shape compiled but typed missing keys as defined `string`, masking real `undefined` access downstream. |
| 2 | `ElpxImporter.spec.ts:1952` — link-scan regex omits `_` | Valid | Regex now matches the importer's own id boundary class: `[a-zA-Z0-9_-]+`. Keeps the dangling-link assertion accurate for fixtures with underscore ids. |
| 3 | `YjsStructureBinding.js:2836` — empty-prefix guard missing | Valid | The canonical `src/shared/ids.ts::generateId` _does_ throw on empty prefix (verified the source). Added the same guard to the mirror so behaviour is consistent across runtimes. |
| 4 | `ComponentImporter.js:480` — same | Valid | Same guard; new test case in `ComponentImporter.test.js` pins it. |
| 5 | `YjsProjectManagerMixin.js:637` — same on the canonical helper | Valid | Same guard inside `generateCanonicalId`. |
| 6 | `ElpxImporter.ts:1181` — embedded `ideviceId` in `jsonProperties` stale after collision-remap | Valid | After parsing `jsonProperties`, if `originalComponentId !== componentId` and the payload's `ideviceId` equals the old XML id, it's rewritten to the fresh id. New regression test in `ElpxImporter.spec.ts` (`rewrites embedded ideviceId in jsonProperties when the component is regenerated on collision`). |

### Verification
- Targeted: 145 backend pass / 0 fail across the four touched specs (`ids`, `ElpxImporter`, `YjsDocumentAdapter`, `stable-identifiers` integration).
- Frontend: `ComponentImporter.test.js` 46 pass, `YjsProjectManagerMixin.test.js` 107 pass — 153 / 153.
- `make fix` clean.

The points are subtle but each one closes a real footgun. Thanks for the careful pass.

erseco:
External AI review evaluated. Verified each finding against the actual code paths before applying. 4/4 valid items applied in `319f65db` (plus `6e325886` to untrack `.omc/` that leaked into the previous push).

### Applied

| # | Claim | Verdict | Where |
|---|---|---|---|
| 1 | `buildStructureFromYjs()` drops stable identifiers; server-side export-from-client falls back to a random manifest@identifier | **Confirmed** — `apiCallManager.js:1180`, `request-payloads.ts:128`, `export.ts:215` all omit the 3 fields | Type extended, browser forwards, server writes |
| 2 | New blank projects have no stable identity → first SCORM export already breaks LMS tracking | **Confirmed** — `YjsDocumentManager.createBlankProject` + `yjs-initializer.createBlankYjsDocument` set everything except `odeIdentifier`/`odeVersionId` | Mint both at create-time in browser + server |
| 3 | `scormIdentifier` override doesn't round-trip through ELPX | **Confirmed** — `metadata-properties.ts:240` declares it `excludeFromXml: true` and `OdeXmlGenerator` never emitted it | Emit in `<odeResources>` alongside `odeId`; importer reads from `extractOdeResources` |
| 4 | `ComponentImporter` regenerates compId but doesn't follow the embedded `ideviceId` in `jsonProperties` (same shape as the ELPX bug already fixed) | **Confirmed** in `ComponentImporter.js:99-107` | Loop now rewrites `comp.properties.ideviceId` when it matches the original id |

### Skipped (verified as out of scope or no current bug)

| Claim | Reason |
|---|---|
| Boundary regex should also reject `.`, `:`, `%`, `/` | Pure defensive; canonical IDs (`Format A` and v3 legacy) are alphanumeric + dash only. No current id format triggers it. |
| `scormIdentifier` sanitization | No current bug; downstream XML escaping already protects against `<`, `&`, etc. Validation would be a separate UX-level decision. |
| Missing `<odeResources>` should mint identifiers at import | Tradeoff — preferred policy is "don't invent IDs without user intent". Already handled by the export-side fallback. |
| `generateUnusedId` retry loop in `generateId` | 70 bits entropy in the random suffix; collision under non-mocked time is negligible. |
| `LegacyXmlParser.ts:1245` still has `.substr` | Different code path (legacy v3 `reference` attribute fallback); IDs it produces get regenerated downstream by `convertLegacyPagesToPageData`. |
| IMS should emit a LOM identifier/catalog entry | IMS CP spec doesn't require it; the project identity flows through `content.xml`. The PR description's wording on IMS is accurate. |
| Doc inconsistencies elsewhere | Pre-existing, not introduced by this PR. |

### Cleanups also done

- `src/shared/ids.ts` header now lists all three browser mirrors (was two — third is `YjsProjectManagerMixin.js::generateCanonicalId`).
- `YjsStructureBinding.test.js`: flipped the "generates IDs without prefix" test to assert the throw guard added in `a61e8b61` for parity with the canonical TS contract.

### Verification

- Backend: `make test-unit-ci` → **6971 pass / 0 fail**.
- Frontend: `npx vitest run` → **11928 pass / 0 fail**.
- Lint: `make fix` clean.

New tests added in this round: 11 (yjs-initializer, populateYDocFromStructure, scormIdentifier round-trip and OdeXmlGenerator emission, embedded-ideviceId rewrite, throw-guard parity).

erseco:
Status across all Copilot inline comments on this PR:

| # | File · Line | Comment | Status |
|---|---|---|---|
| 1 | `ElpxImporter.ts:432` | `extractOdeResources` return type lies about missing keys | ✅ Fixed in `a61e8b61` (now `Partial<Record<string,string>>`) |
| 2 | `ElpxImporter.spec.ts:1954` | dangling-link regex omits `_` | ✅ Fixed in `a61e8b61` (now `[a-zA-Z0-9_-]+`) |
| 3 | `YjsStructureBinding.js:2839` | browser mirror missing empty-prefix guard | ✅ Fixed in `a61e8b61` + spec flipped in `319f65db` |
| 4 | `ComponentImporter.js:491` | same browser mirror gap | ✅ Fixed in `a61e8b61` + test pinned |
| 5 | `YjsProjectManagerMixin.js:641` | same on `generateCanonicalId` | ✅ Fixed in `a61e8b61` |
| 6 | `ElpxImporter.ts:1187` | embedded `ideviceId` in `jsonProperties` not rewritten on collision-remap | ✅ Fixed in `a61e8b61` + `319f65db` for the parallel browser path |
| **7** | **`odeId.spec.ts:21`** | **probabilistic 100-unique-from-100 assertion** | **✅ Just fixed in `f0c72ae9`** |

Last one (#7) is valid: the assertion was mathematically safe (~5e-10 collision probability for two calls; ~5e-7 for 100) but the design technically allowed a flake. Replaced with a deterministic two-call distinctness check — same coverage of the "produces different ids" invariant, no remaining flake surface.

`bun test src/shared/export/utils/odeId.spec.ts` → 4 pass / 0 fail.

erseco:
E2E (chromium + firefox) was failing on `yjs-binary-integrity.spec.ts:226` — *"should verify Yjs document binary size is consistent after save/load"*. Investigated, confirmed real cause, fixed in `689d62ac`.

### Root cause

The test creates a blank project, opens `test/fixtures/basic-example.elp` (legacy v3, no `<odeResources>`), then asserts `|beforeSaveSize − afterReloadSize| ≤ 5%`.

This branch introduced two mint paths for `odeIdentifier` / `odeVersionId`:
1. **Server**, in `createBlankYjsDocument` — fires at project creation.
2. **Legacy importer**, in `setLegacyMetadata` — fired once at import time so legacy `.elp` files (which lack `<odeResources>`) still end up with a stable pair.

Both ran in this flow. Yjs is a CRDT so it kept *both* writes per key as history; save → snapshot → reload garbage-collects the earlier ones, shrinking the persisted state by ~80 bytes per overwritten key. With 2 keys overwritten, the delta lands at **362 bytes** vs the **325-byte** (5%) tolerance — exactly what the failing run reported (deterministic across all 3 retries).

### Fix

`setLegacyMetadata` now guards each mint with `metadata.get(...)`:

```ts
if (!metadata.get('odeIdentifier')) {
    metadata.set('odeIdentifier', generateOdeId());
}
if (!metadata.get('odeVersionId')) {
    metadata.set('odeVersionId', generateOdeId());
}
```

This codifies the conceptually correct rule: **identity belongs to the PROJECT, not to the imported file**. A blank-project mint already established the identity; opening a legacy `.elp` into it preserves it. This is exactly the semantic SCORM LMS tracking needs across re-uploads.

The v4 import path is **unchanged**: v4 files carry their own identity in `<odeResources>` and that explicitly overrides — expected when opening a file with its own identifier.

### Tests

- Existing `'should generate odeIdentifier and odeVersionId once on legacy v3 import'` still passes (fresh `Y.Doc`, no pre-existing identity → mint happens).
- Added `'should NOT overwrite an existing odeIdentifier when opening a legacy .elp into a blank-but-minted Y.Doc (#1786)'` to pin the new guard.
- `make test-unit-ci` → **6972 pass / 0 fail**. Lint clean.

### Why not just raise the tolerance?

The 5% threshold is doing real work — it catches state bloat from unintended persistence patterns. Loosening it would mask exactly the kind of issue this PR introduced. The right answer is to *not* double-write.

### Why static e2e passed

`e2e (static)` succeeded because static mode runs without the server-side Y.Doc init (no `createBlankYjsDocument`), so only the legacy importer minted — single write, no GC delta.

erseco:
**Second attempt at fixing the e2e** — previous patch in `689d62ac` was based on the wrong assumption.

### What was wrong

I assumed `test/fixtures/basic-example.elp` was a legacy v3 fixture (it has the `.elp` extension) and patched `setLegacyMetadata`. Re-inspected the fixture: it contains `content.xml` (v4) with full `<odeResources>` carrying odeId / odeVersionId. The v4 path runs, not the legacy path. My v3 guard was correct in spirit but never executed for this test.

So the original failure cause — double-write of `odeIdentifier` in the same Y.Doc — kept happening through the v4 path: server mint + v4 import overwriting = 2 CRDT ops per key → GC delta on save/load.

### The fix (`aa998fdc`)

Remove the server-side mint entirely.

- `createBlankYjsDocument` no longer writes `odeIdentifier` / `odeVersionId`.
- `YjsDocumentManager.createBlankProjectStructure` ditto. Unused `_generateOdeId` helper removed.
- `setLegacyMetadata` reverts to unconditional mint (v3 fixtures genuinely lack `<odeResources>`, so the mint is the *first* and only write — no overwrite, no GC delta).

### Why this is the right semantic

The file's identity now wins on import. That's actually what users expect: opening a `.elpx` means "I'm working on this document, with its own identity". A blank project that never imports anything inherits identity from the first import (or relies on `BaseExporter.getManifestIdentifier` fallback for export).

### Tradeoff

Brand-new project that never imports AND never saves AND exports twice → two consecutive SCORM exports get different `manifest@identifier`. This is the degenerate corner-case the external reviewer's claim #2 flagged. I'd argue the LMS workflow never hits this in practice (users always save before exporting), and the alternative — minting at create + accepting the GC overhead — has a real cost the binary-integrity test legitimately catches.

If we want strict guarantee for this degenerate case in a follow-up, the right design is to **persist the fallback on first export** (write back into the Y.Doc) rather than to mint at create. Out of scope for this PR.

### Tests

- Flipped both `yjs-initializer.spec.ts` mint-at-create tests into a single "does NOT mint on create" assertion that pins the new contract.
- Removed the legacy-import "don't overwrite" test from `689d62ac` (was guarding the wrong scenario; v3 legacy never has odeResources so there's nothing to overwrite).

`make test-unit-ci` → **6970 pass / 0 fail**. Lint clean.

mnarvaezm (approved):

PR #1804, opened 2026-05-14
fix(deploy): prepend BASE_PATH on server-emitted icon URLs and TinyMCE iframe URLs (#1802)

### LINKED ISSUE #1802: Subdirectory deployment (BASE_PATH): icon picker, codemagic editor and exemindmap editor 404 because BASE_PATH is not prepended
### Summary

When eXeLearning is deployed in a subdirectory behind a reverse proxy that only forwards the `BASE_PATH` namespace to the Bun backend (i.e. **the realistic production shape**, not a flat dev box), several client-side surfaces emit URLs **without `BASE_PATH`** and 404 because the reverse proxy never forwards them.

The bug is invisible in local dev because the Bun server registers every route twice — at root **and** under `${BASE_PATH}` (`src/index.ts:574–657` and `:280–329`). That dual-mount makes a flat `bun run dev` happy with either URL shape, so client-side BASE_PATH bugs are masked. As soon as something between the browser and Bun (Apache, nginx, an Ingress) only mounts the BASE_PATH namespace, the unprefixed URLs hit the proxy's root namespace and 404 — Bun never sees them.

Reproduced on a deployment with `BASE_PATH=/subfolder1/subfolder2/exelearning` (replace with your own subdirectory). Diagnostic snapshot from a working production-shape proxy chain (Apache at the public edge, nginx as the immediate reverse proxy in front of Bun):

```
curl -sI https://<host>/                                                     -> Server: Apache
curl -sI https://<host>/subfolder1/subfolder2/exelearning/                   -> Server: nginx
```

**Verified live from the deployed page (DevTools console on the workarea URL):**

- `window.eXeLearning.app.composeUrl('/api/codemagic-editor/codemagic.html')` returns the BASE_PATH-prefixed URL — the helper is working correctly.
- `window.eXeLearning.config.basePath` is populated correctly.
- `themes.selected.icons['experiment']` is `{id, title, type, value: '/v<version>/files/perm/themes/base/base/icons/experiment.png'}` — note `value` is bare-rooted, no `_relativePath`, no `blob:` cache.
- `HEAD /v<version>/files/perm/themes/base/base/icons/objectives.png` → **404** (hits proxy root, no proxy match).
- `HEAD /subfolder1/subfolder2/exelearning/v<version>/files/perm/themes/base/base/icons/objectives.png` → **200 image/png** (proxy → Bun root-mount serves it via the dual-mount shim).
- Same shape for codemagic: bare 404, BASE_PATH-prefixed 200.

So the helper, config, and server are all healthy — the bug is that the icon emission (server-side) and two TinyMCE plugin iframes (client-side) never go through `composeUrl()` / `getBasePath()`.

#### 1. iDevice icon picker (Block iDevice "Change icon" modal)

Every icon 404s. URLs start at root (e.g. `/v<version>/files/perm/themes/base/base/icons/objectives.png`) — the BASE_PATH prefix is missing. The proxy never forwards them.

Origin: **server-side** in `src/routes/themes.ts:181, 212, 300` — the icon `value` is built as `'${themeUrl}/icons/${filename}'` where `themeUrl` is `/${version}/files/perm/themes/...`, with no `BASE_PATH` prefix. The icon picker renders that verbatim in `public/app/workarea/project/idevices/content/blockNode.js:508–525` (`makeIconValueElement` only adjusts the path for `isStaticMode` / `isOfflineInstallation`, never for server-mode behind a proxy, and never calls `composeUrl()`). The blob-cache fallback path through `_relativePath` (`public/app/workarea/themes/theme.js:417, 444`) is not active for this user — `_relativePath` is undefined, so even that secondary code path wouldn't help.

#### 2. exemindmap TinyMCE plugin (mind-map editor)

`GET /api/exemindmap-editor/index.html` → 404 (BASE_PATH missing).

Origin: **client-side**, hardcoded in `public/libs/tinymce_5/js/tinymce/plugins/exemindmap/plugin.js` (and `plugin.min.js`). The intended behaviour the test documents is the bare path itself (`exemindmap-plugin.test.js:12–17` asserts `/api/exemindmap-editor/index.html` as the server-mode URL), so this is specified-as-buggy: the plugin only branches on static-mode for an override, never on `BASE_PATH`. Compare with codemagic below, which DOES branch in its test.

#### 3. codemagic TinyMCE plugin (HTML source editor)

`GET /api/codemagic-editor/codemagic.html` → 404. Then every relative asset inside that page 404s too: `includes/codemirror/codemirror.css`, `includes/codemagic.css`, `includes/codemirror/codemirror.js`, `includes/i18n.js`, `includes/batched_scripts.js`, `includes/codemagic.js`, `images/icons/{undo,redo,lens,file}.png`.

Origin: `public/libs/tinymce_5/js/tinymce/plugins/codemagic/plugin.js`. Unlike exemindmap, this one's unit test (`codemagic-plugin.test.js:26`) asserts the **correct** behaviour (`basePath + '/api/codemagic-editor/codemagic.html'`). The test passes, but the deployed plugin emits the bare path anyway — and we know the config is healthy (`config.basePath` returns the right value at the same DevTools console). So the runtime plugin code path is **not** the path the test exercises — it's reading `basePath` from a different (empty) source, or skipping the prefix entirely. Needs a maintainer to align the runtime emission with the test.

### Why this is invisible in dev

`src/index.ts:574–657` registers all routes (including `/api/codemagic-editor`, `/api/exemindmap-editor`, the static-file handler at `/v<version>/...`) at root **and** at `${BASE_PATH}`. A flat `bun run dev` therefore answers either URL shape, and the icon picker / TinyMCE plugins look fine. The bug only surfaces when a real proxy enforces the namespace boundary.

### How to reproduce locally

See the **"How to test this works locally"** section in PR #1804 — it ships a self-contained 3-file nginx-in-Docker recipe that puts a reverse proxy in front of the local Bun backend, forwarding only the BASE_PATH namespace and 404'ing everything else. Run it on `main` to reproduce all three failure modes byte-for-byte; run it on the PR branch to verify the fix.

### Solution

1. **Icons** — in `src/routes/themes.ts:181, 212, 300` prefix the icon `value` with `getBasePath()` from `src/utils/basepath.util.ts`. Single source of truth, fixes all icons in one place.
2. **exemindmap iframe** — in `public/libs/tinymce_5/js/tinymce/plugins/exemindmap/plugin.js` (and `plugin.min.js`), replace the hard-coded `/api/exemindmap-editor/index.html` with `window.eXeLearning.app.composeUrl('/api/exemindmap-editor/index.html')`. Update `exemindmap-plugin.test.js` to assert the BASE_PATH-prefixed URL.
3. **codemagic iframe** — figure out why the deployed plugin emits the bare path despite the test asserting the BASE_PATH-prefixed one. Likely the runtime reads `basePath` from a source that's empty at TinyMCE-init time. Align the runtime with the test (also via `composeUrl`).
4. **(Recommended cleanup)** Drop or feature-flag the dual-mount in `src/index.ts:574–657` and `:280–329`. It's what hides 1., 2. and 3. in `bun run dev`. With it gone, the dev server would 404 the same URLs as production, so these regressions would fail in CI.

### Related prior issues

- #1042 — "TinyMCE issues when using a different BASE_PATH" (closed). Same shape; the codemagic / exemindmap plugins are exactly this class of regression.
- #229 — "Different BASE_PATH problems" (closed).
- #369 — "Test eXe with a different BASE_PATH and improve documentation" (closed). Worth revisiting — there's still no automated test that exercises a non-flat deploy, which is exactly why this came back.

erseco:
## Summary

Fixes the BASE_PATH 404s documented in #1802. When eXeLearning is deployed in a subdirectory behind a reverse proxy that only forwards the BASE_PATH namespace (the realistic production shape), three client-facing surfaces emitted URLs without BASE_PATH and 404'd:

- **iDevice icon picker** (~40 broken icons) — `src/routes/themes.ts` emitted bare `/${version}/files/perm/themes/.../icons/*.png`. Fixed by routing every emitted URL through `prefixPath()` from `src/utils/basepath.util.ts`.
- **codemagic TinyMCE plugin** (HTML source editor iframe 404) — read `basePath` from a runtime-stale config field; on the affected deployment it ended up empty even though `eXeLearning.config.basePath` was correctly populated. Fixed by routing through `window.eXeLearning.app.composeUrl(...)` (the canonical client helper, defined in `public/app/app.js`).
- **exemindmap TinyMCE plugin** (mind-map iframe 404) — hardcoded `/api/exemindmap-editor/index.html`. Fixed the same way.

Plus a new `STRICT_BASE_PATH_ROUTES` env (default off, opt-in) that disables the root dual-mount in `src/index.ts` when `BASE_PATH` is set, so `bun run dev` / CI behaves like a real proxy and surfaces these regressions instead of masking them. Documented in `.env.dist`.

## Why this was invisible in dev (and how this PR prevents regressions)

The Bun server registered every route at root AND under `BASE_PATH`, so a flat `bun run dev` answered either URL shape. Client-side BASE_PATH bugs only surfaced behind a real proxy. Every fix in this PR ships with a regression test:

- `src/routes/themes.spec.ts` — new `BASE_PATH handling (regression for issue #1802)` describe block with 3 specs that toggle `process.env.BASE_PATH` and assert URLs are (or aren't) prefixed accordingly. Outer `beforeEach` now clears `process.env.BASE_PATH` so the rest of the suite is deterministic regardless of the developer's local `.env`.
- `codemagic-plugin.test.js` and `exemindmap-plugin.test.js` rewritten to mirror the new `composeUrl` contract: server mode with the host app present → prefixed URL; without → bare fallback; static-mode override still wins.

With `STRICT_BASE_PATH_ROUTES=true` set in CI, any future regression of this class would 404 immediately in unit/integration test fixtures that issue HTTP requests against the embedded server.

## How to test this works locally (behind a real reverse proxy)

A flat `bun run dev` will not catch this — the server's dual-mount answers both URL shapes. To reproduce the production failure mode, put a real reverse proxy in front that only forwards the BASE_PATH namespace. The minimum viable nginx-in-Docker recipe (3 files):

**`docker-compose.yml`**

```yaml
services:
  nginx:
    build: .
    ports: ["80:80"]
    extra_hosts:
      - "host.docker.internal:host-gateway"
```

**`Dockerfile`**

```dockerfile
FROM alpine:3.20
RUN apk add --no-cache nginx \
    && mkdir -p /run/nginx \
    && rm -f /etc/nginx/http.d/default.conf
COPY nginx.conf /etc/nginx/http.d/exelearning.conf
EXPOSE 80
CMD ["nginx", "-g", "daemon off;"]
```

**`nginx.conf`** (only forwards the BASE_PATH namespace; everything else 404s — same boundary the production proxy enforces)

```nginx
server {
    listen 80 default_server;
    server_name _;
    location /subfolder1/subfolder2/exelearning/ {
        proxy_pass http://host.docker.internal:81/subfolder1/subfolder2/exelearning/;
        proxy_set_header Host $host;
    }
}
```

Steps:

1. Add `127.0.0.1 machine.local` to `/etc/hosts` (or use any local hostname you prefer; just keep it consistent below).
2. Set `BASE_PATH=/subfolder1/subfolder2/exelearning` in `.env` (any path works; just keep both sides consistent).
3. Start eXeLearning on the host at port `81` (`make up` or `make up-local`).
4. Save the three files above in a folder, then `docker compose up -d --build`.

Sanity probes:

```sh
# bare-root: nginx 404s (the bug pattern)
curl -I http://machine.local/v0.0.0-alpha/files/perm/themes/base/base/icons/objectives.png   # -> 404

# BASE_PATH-prefixed: nginx forwards to Bun -> 200 (what the client should be emitting)
curl -I http://machine.local/subfolder1/subfolder2/exelearning/v0.0.0-alpha/files/perm/themes/base/base/icons/objectives.png   # -> 200 image/png
```

Then open `http://machine.local/subfolder1/subfolder2/exelearning/workarea` in a browser, log in, and paste this in DevTools:

```js
const t = window.eXeLearning?.app?.themes?.selected;
const allIcons = Object.values(t?.icons || {});
({
    total: allIcons.length,
    withBasePath: allIcons.filter(i =>
        i.value.startsWith('/subfolder1/subfolder2/exelearning/'),
    ).length,
    codemagicUrl: window.eXeLearning?.app?.composeUrl?.(
        '/api/codemagic-editor/codemagic.html',
    ),
    exemindmapUrl: window.eXeLearning?.app?.composeUrl?.(
        '/api/exemindmap-editor/index.html',
    ),
})
```

On `main` (without this PR): `withBasePath` is `0`, codemagic / exemindmap URLs are bare, and the icon picker / both iframes 404. After this PR: `withBasePath === total`, both URLs are BASE_PATH-prefixed, all three flows work.

## Test plan

- [x] `make fix` clean.
- [x] `bun test src/routes/themes.spec.ts` → 44/44 pass (3 new BASE_PATH specs).
- [x] `bun test public/libs/tinymce_5/js/tinymce/plugins/{codemagic,exemindmap}/*-plugin.test.js` → 35/35 pass.
- [x] Full `bun test` → same pre-existing 21 fails as `main` (env / DB_PATH issues from local `.env`, not introduced by this PR — verified via stash + diff).
- [x] End-to-end against the local nginx proxy recipe above: 50/50 icons load, codemagic + exemindmap iframes return 200 (verified live).
- [x] `make test-e2e` — to be run by reviewer; my changes don't touch any flow not already exercised by existing E2E.

Closes #1802

### DIFF
--- a/.env.dist
+++ b/.env.dist
@@ -23,6 +23,13 @@ APP_PORT=8080
 #   BASE_PATH=/web/exelearning
 BASE_PATH=
 
+# When STRICT_BASE_PATH_ROUTES=true AND BASE_PATH is set, routes are
+# registered ONLY under BASE_PATH (no dual-mount at root). This makes
+# `bun run dev` behave like a real production reverse proxy that only
+# forwards the BASE_PATH namespace, surfacing client-side BASE_PATH bugs
+# (issue #1802) instead of masking them. Default: false (compat mode on).
+STRICT_BASE_PATH_ROUTES=false
+
 # Allow users to import/install styles
 ONLINE_THEMES_INSTALL=0
 
--- a/public/libs/tinymce_5/js/tinymce/plugins/codemagic/codemagic-plugin.test.js
+++ b/public/libs/tinymce_5/js/tinymce/plugins/codemagic/codemagic-plugin.test.js
@@ -8,150 +8,107 @@
 
 describe('codemagic plugin - Path Handling', () => {
     describe('Dialog URL selection', () => {
-        // Extract the logic from plugin.min.js lines 69-102
-        function getCodemagicUrl(config, capabilities, staticModeGlobal) {
-            const basePath = (config && config.basePath) || '';
-
-            // Check if we're in static mode
+        // Mirror plugin.min.js: in server mode, route through composeUrl() so that
+        // BASE_PATH is prepended (issue #1802). Static mode still wins when its
+        // flags are set. Reading basePath directly from config is gone — composeUrl
+        // is the single source of truth, identical to the rest of the client.
+        function getCodemagicUrl(config, capabilities, staticModeGlobal, app) {
             let isStaticMode = config?.isStaticMode || config?.isOfflineInstallation;
             if (!isStaticMode) {
                 isStaticMode = capabilities ? !capabilities.storage.remote : staticModeGlobal;
             }
 
-            // In static mode, use relative file path
-            // In server mode, use API endpoint
             if (isStaticMode) {
                 return './libs/tinymce_5/js/tinymce/plugins/codemagic/codemagic.html';
-            } else {
-                return basePath + '/api/codemagic-editor/codemagic.html';
             }
+            return app && typeof app.composeUrl === 'function'
+                ? app.composeUrl('/api/codemagic-editor/codemagic.html')
+                : '/api/codemagic-editor/codemagic.html';
         }
 
-        describe('Online mode (server available)', () => {
-            it('should return API URL with no basePath', () => {
-                const result = getCodemagicUrl({ basePath: '' }, null, false);
-                expect(result).toBe('/api/codemagic-editor/codemagic.html');
+        // Mirror window.eXeLearning.app.composeUrl from public/app/app.js:849-858
+        function makeApp(basePath) {
+            return {
+                composeUrl(p) {
+                    const normalized = !basePath || basePath === '/' ? '' : basePath.replace(/\/+$/, '');
+                    const path = p.startsWith('/') ? p : `/${p}`;
+                    return `${normalized}${path}`;
+                },
+            };
+        }
+
+        describe('Server mode', () => {
+            it('returns bare API URL when no host app is available (back-compat)', () => {
+                expect(getCodemagicUrl(null, null, false)).toBe('/api/codemagic-editor/codemagic.html');
+                expect(getCodemagicUrl(undefined, null, false)).toBe('/api/codemagic-editor/codemagic.html');
+                expect(getCodemagicUrl({}, null, false)).toBe('/api/codemagic-editor/codemagic.html');
             });
 
-            it('should return API URL with basePath prefix', () => {
-                const result = getCodemagicUrl({ basePath: '/web/exe' }, null, false);
-                expect(result).toBe('/web/exe/api/codemagic-editor/codemagic.html');
-            });
-
-            it('should return API URL with deep basePath prefix', () => {
-                const result = getCodemagicUrl({ basePath: '/deep/nested/path' }, null, false);
-                expect(result).toBe('/deep/nested/path/api/codemagic-editor/codemagic.html');
-            });
-
-            it('should return API URL when config is null', () => {
-                const result = getCodemagicUrl(null, null, false);
-                expect(result).toBe('/api/codemagic-editor/codemagic.html');
-            });
-
-            it('should return API URL when config is undefined', () => {
-                const result = getCodemagicUrl(undefined, null, false);
-                expect(result).toBe('/api/codemagic-editor/codemagic.html');
-            });
-
-            it('should return API URL when config is empty object', () => {
-                const result = getCodemagicUrl({}, null, false);
-                expect(result).toBe('/api/codemagic-editor/codemagic.html');
-            });
-        });
-
-        describe('Static mode via config flags', () => {
-            it('should return relative URL when isStaticMode is true', () => {
-                const result = getCodemagicUrl({ isStaticMode: true }, null, false);
-                expect(result).toBe('./libs/tinymce_5/js/tinymce/plugins/codemagic/codemagic.html');
-            });
-
-            it('should return relative URL when isOfflineInstallation is true', () => {
-                const result = getCodemagicUrl({ isOfflineInstallation: true }, null, false);
-                expect(result).toBe('./libs/tinymce_5/js/tinymce/plugins/codemagic/codemagic.html');
-            });
-
-            it('should return relative URL when both flags are true', () => {
-                const result = getCodemagicUrl({ isStaticMode: true, isOfflineInstallation: true }, null, false);
-                expect(result).toBe('./libs/tinymce_5/js/tinymce/plugins/codemagic/codemagic.html');
-            });
-
-            it('should ignore basePath in static mode', () => {
-                const result = getCodemagicUrl({ basePath: '/web/exe', isStaticMode: true }, null, false);
-                expect(result).toBe('./libs/tinymce_5/js/tinymce/plugins/codemagic/codemagic.html');
-            });
-        });
-
-        describe('Static mode via capabilities fallback', () => {
-            it('should return relative URL when capabilities.storage.remote is false', () => {
-                const capabilities = { storage: { remote: false } };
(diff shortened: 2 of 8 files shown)


ignaciogros (approved):
I haven’t been able to run it on Windows with nginx following the instructions, but `make up` and `make run-app` worked correctly. 

I think we can merge the solution into `main` to test it under real conditions. We could revert the commit if necessary. 

The usual deployments are not broken, and the test route (CodeMagic) is the expected one.

PR #1815, opened 2026-05-17
fix(auth): block password login for disabled users (#1814)

### LINKED ISSUE #1814: Disabled users can still log in with password authentication
## Summary

Users marked as inactive (`is_active = 0`) can still log in using password authentication.

## Environment

This was reproduced in the multi-user/server setup described in the documentation as **Server deployment for multiple users** / **production Docker deployment**, with real-time collaborative editing available through Yjs WebSocket.

Runtime details from the tested instance:

- eXeLearning v4.0.0 Docker image: `ghcr.io/exelearning/exelearning:v4.0.0`
- Docker Compose deployment
- `APP_ONLINE_MODE=1`
- SQLite database (`DB_DRIVER=pdo_sqlite`, `DB_PATH=/mnt/data/exelearning.db`)
- Password authentication enabled (`APP_AUTH_METHODS=password`)

## Steps to reproduce

1. Create a local user with `ROLE_USER`.
2. Log in successfully with that user.
3. As admin, edit the user and set Status to inactive.
4. Confirm in the database that the row has `is_active = 0`.
5. Try to log in again with the same email and password.

## Expected behavior

Inactive users should not be able to authenticate.

## Actual behavior

The inactive user can still log in. In the tested installation, the database row had `is_active = 0`, but password login still succeeded.

## Technical notes

In `src/routes/auth.ts`, both `/api/auth/login` and `/login_check` look up the user with `findUserByEmail()` and validate the password, but do not appear to check `user.is_active`.

Also, existing JWT cookies remain valid after disabling the user. That may be a separate design decision, but it means disabling a user does not immediately revoke current access.

## Impact

The admin UI suggests that disabling a user blocks access, but it does not prevent password login. This can leave temporary or revoked accounts able to access the instance.

erseco:
## Summary

Fixes #1814.

`POST /api/auth/login` and `POST /login_check` looked up the user with `findUserByEmail()` and validated the password, but never checked `user.is_active`. Users marked as inactive in the admin UI could still authenticate with their password.

The disabled-user check runs **after** password verification, so the specific "Account deactivated" message is only ever shown to someone who already knows the correct credentials — there is no extra account-state leak versus the existing "Invalid credentials" path. That trade-off lets the UX favor the legitimate user: they get told to contact the admin instead of wasting time on a password reset.

- `POST /api/auth/login` → `403 Forbidden` with `{ error: 'Forbidden', message: 'Account deactivated' }`
- `POST /login_check` → `302` redirect to `/login?error=Account%20deactivated`

Note: as the issue mentions, existing JWT cookies are not revoked by this change — that is a separate concern and is out of scope here.

## Test plan

- [x] New unit test: `Auth Routes > POST /api/auth/login > should return 403 with "Account deactivated" for disabled user with correct password`
- [x] New unit test: `Auth Routes > POST /login_check > should redirect to login with "Account deactivated" error for disabled user`
- [x] Full `src/routes/auth.spec.ts` suite green (`bun test src/routes/auth.spec.ts` → 102 pass / 0 fail)
- [x] `make fix` clean
- Manual reproduction (matches the steps in the issue):
  1. Create a `ROLE_USER`, log in successfully.
  2. As admin set the user's status to inactive (`is_active = 0`).
  3. Try to log in again with the same email and password → now blocked with `Account deactivated` instead of succeeding.


pabloamayab (approved):
Great work, @erseco . It works well; I've only tested it locally with manually created users. I assume it will work just as well with users created automatically via OpenID or a similar service.

PR #1820, opened 2026-05-18
Preserve escaped script-like text on ELPX import

### LINKED ISSUE #1816: Text after "<script>" disappears after reopening .elpX project
**Bug description**

In any of the three versions of eXeLearning (desktop, online, or cloud), if I create a text block and include the word <script> (either as plain text or as HTML), the editor initially accepts it without any issues. The text displays correctly in both editing mode and preview mode.

However, after saving or exporting the .elpX file and reopening it in any of the three versions, the behavior changes:

- When reopening the project, all text written after the word <script> disappears from both editing mode and preview mode.
- If I inspect the HTML source, the missing text is still present, but an unexpected closing tag </script> has been automatically inserted.

**Additional behavior**

If I replace the word script with any other word using the same format (for example <example>):

- During creation, the behavior is the same: the editor accepts it and displays it correctly.
- After saving and reopening:

    - The text written after that tag is no longer removed.
    - Only the custom tag itself disappears in editing and preview mode.
    - In the HTML source, an automatic closing tag is also inserted (for example </example>).

**Steps to reproduce**

1. Open any version of eXeLearning (desktop, online, or cloud).
2. Create a text block.
3. Insert text containing <script> (either as HTML or plain text).
4. Save or export the .elpX project.
5. Reopen the project.

**Expected behavior**

The editor should preserve literal text or escaped HTML tags without altering subsequent content or automatically inserting unexpected closing tags.

**Actual behavior**

The parser interprets the tag incorrectly, inserts a closing tag automatically, and in the case of <script>, all following text becomes hidden in edit and preview modes.

Creation in any version 
Image 
After reopening in any version 
Image 
Creation in any version
Image
After reopening in any version
Image

erseco:
## Summary
- Preserve escaped literal tags in text iDevice JSON properties when importing ELPX files.
- Keep legacy exe-text wrapper decoding for old encoded content.
- Add a regression test for escaped `<script>` text followed by visible content.

## Root cause
The ELPX importer decoded HTML entities across all `jsonProperties` values. That turned literal `&lt;script&gt;` text into a real `<script>` element after reopening, so following content disappeared in edit and preview.

## Test steps
1. Create a text iDevice with `Text before <script> text after`.
2. Save or export the project as ELPX.
3. Reopen the ELPX.
4. Confirm `<script>` remains visible as text and the following text is still shown.


mnunezcedec (approved):
Thanks

PR #1835, opened 2026-05-21
fix(export): remove deprecated unload event from SCORM packages (#1831)

### LINKED ISSUE #1831: SCORM 1.2 export uses unloadPage() on onunload, causing Chrome Permissions Policy violation and lost scores in Moodle
## Description

SCORM 1.2 packages exported from eXeLearning include `unloadPage()` as an unload handler in the generated HTML pages.

In our case, the exported pages contain markup like:

```html
<body ... onunload="unloadPage()">
```

When these SCORM packages are uploaded to Moodle and opened with recent versions of Google Chrome / Chromium, Chrome logs the following error:

    [Violation] Permissions policy violation: unload is not allowed in this document.

As a result, some SCORM tracking data is not reliably saved in Moodle, especially scores/completion data when the learner leaves or closes the activity.

**Environment**

-   Authoring tool: eXeLearning
-   Export format: SCORM 1.2 ZIP
-   LMS: Moodle 4.5
-   Browser affected: Google Chrome / Chromium
-   Browser not affected or less affected: Firefox
-   Problem: SCORM scores/completion are not always saved in Chrome

**Why this is a problem**

Chrome is deprecating the unload event. According to Chrome documentation, unload handlers are gradually being stopped unless pages explicitly opt in, and the recommended replacements are visibilitychange and/or pagehide.

This affects SCORM packages generated by eXeLearning because the SCORM session finalization currently depends on unloadPage() being called from the page unload lifecycle.

**Current eXeLearning behavior**

In the current SCORM 1.2 exporter, the generated page options still include:

```js
onLoadScript: 'loadPage()',

onUnloadScript: 'unloadPage()',
```
This leads to generated SCORM HTML pages using unloadPage() on page unload.

**Expected behavior**

SCORM exports should avoid using the deprecated/unreliable unload event.

Instead, eXeLearning should finalize or commit the SCORM session using modern lifecycle events, for example:
```js
(function () {

var scormPageClosed = false;

function scormSafeUnload() {

if (scormPageClosed) {

return;

}

scormPageClosed = true;

if (typeof unloadPage === "function") {

unloadPage();

}

}

if (window.addEventListener) {

window.addEventListener("pagehide", scormSafeUnload, false);

window.addEventListener("visibilitychange", function () {

if (document.visibilityState === "hidden") {

scormSafeUnload();

}

}, false);

}

}());
```

Or another equivalent approach that does not rely on the unload event.

**Workaround tested**

We manually patched one exported SCORM package by:

1.  Removing `onunload="unloadPage()"` from all generated HTML pages.
2.  Adding a guarded listener using pagehide / beforeunload to call `unloadPage()` once.

After this manual patch, the SCORM package stopped triggering the Chrome Permissions Policy violation and scores were saved correctly in Moodle.

However, this is not sustainable because we have hundreds of SCORM packages generated with eXeLearning.

**Suggested fix**

Please update the SCORM 1.2 exporter so that generated packages do not use `onunload="unloadPage()"`.

Possible approach:

-   remove `onUnloadScript: 'unloadPage()'` from the generated HTML body attributes;
-   register SCORM finalization through pagehide and/or visibilitychange;
-   ensure unloadPage() / LMSFinish() is only called once;
-   preserve compatibility with Moodle and other SCORM 1.2 LMS platforms.

**References**

Chrome unload deprecation:  
[https://developer.chrome.com/docs/web-platform/deprecating-unload](https://developer.chrome.com/docs/web-platform/deprecating-unload)

Relevant eXeLearning source:  
[https://github.com/exelearning/exelearning/blob/main/src/shared/export/exporters/Scorm12Exporter.ts](https://github.com/exelearning/exelearning/blob/main/src/shared/export/exporters/Scorm12Exporter.ts)

erseco:
## Problem

Closes #1831.

SCORM 1.2 (and 2004) packages exported by eXeLearning rendered every page with:

```html
<body ... onload="loadPage()" onunload="unloadPage()" onbeforeunload="unloadPage()">
```

Chrome/Chromium are deprecating the `unload` event and block it entirely under a
restrictive **Permissions Policy** — which is exactly what Moodle applies to the
iframe that runs a SCORM activity:

```
[Violation] Permissions policy violation: unload is not allowed in this document.
```

Because SCORM session finalization (`scorm.quit()` → `LMSFinish`) was wired to that
blocked `onunload` path, scores/completion were not reliably saved in Chrome
(Firefox was less affected). Affected: Moodle 4.5 + Chrome.

## Solution

Move SCORM finalization off the deprecated `unload` event onto modern lifecycle events:

- **`pagehide`** → finalize the session **once** (the recommended `unload` replacement;
  fires on close, navigation and bfcache, and is not gated by the Permissions Policy).
- **`visibilitychange` → `hidden`** → **commit progress only** (`scorm.save()`, no quit),
  so switching tabs does not terminate the session prematurely while still persisting
  data if `pagehide` is never delivered.

`unloadPage()` is still called at most once (guarded), preserving existing completion
semantics for Moodle and other SCORM 1.2/2004 LMSs. `onload="loadPage()"` is unchanged
(`load` is not deprecated).

### Changes

- `public/app/common/scorm/SCOFunctions.js`: new exported `registerScormLifecycleHandlers()`,
  auto-invoked in the browser. This is the canonical runtime shipped for both 1.2 and 2004.
- `src/shared/export/renderers/PageRenderer.ts` + `interfaces.ts`: stop emitting
  `onunload`/`onbeforeunload`; removed the now-unused `onUnloadScript` render option.
- `src/shared/export/exporters/Scorm12Exporter.ts` / `Scorm2004Exporter.ts`: stop passing
  `onUnloadScript`; their fallback SCO functions self-register the same handlers.

## Tests

- **Unit (bun):** `Scorm12Exporter.spec.ts`, `Scorm2004Exporter.spec.ts`,
  `PageRenderer.spec.ts` — assert the generated page has `onload` but **no**
  `onunload`/`onbeforeunload`, and that fallback SCO functions register
  `pagehide`/`visibilitychange` (not `unload`).
- **Unit (vitest):** `SCOFunctions.test.js` — `registerScormLifecycleHandlers()` registers
  the modern events, finalizes once on `pagehide`, and commits (without quitting) on
  `visibilitychange → hidden`. 100% line coverage on the changed file.
- **Integration (bun):** `test/integration/export/scorm12-exporter.spec.ts` — the actual
  exported SCORM ZIP's `index.html` contains no `onunload`/`onbeforeunload`.

## Verification

- `make fix` — clean.
- `make test-unit` — 7009 backend + frontend unit tests pass. (The coverage gate flags
  `src/websocket/asset-coordinator.ts` at 88.2%; this is **pre-existing on `main`** and
  unrelated to this change.)
- `make test-integration` — SCORM export integration tests green.
- `make test-e2e` (page-export-import) — export/import flows green. The fix only changes
  generated file bytes (no UI flow change); the exported artifact is validated end-to-end by
  the integration test, so no dedicated SCORM E2E was added.

> Note: `public/app/yjs/exporters.bundle.js` is a gitignored build artifact regenerated by
> `make bundle`; it is not committed.


ignaciogros (approved):
I tested with SCORM content published in Moodle 4.5 using the latest version of the `mod_exescorm` plugin. The results were recorded correctly.

I think it would be good to run some additional tests to ensure that the changes made do not affect tracking.

pabloamayab (approved):
I've tested it in Moodle 5 with the latest version of the mod_scorm plugin, using a wide variety of games, and it has tracked the activities correctly

PR #1839, opened 2026-05-21
feat(workarea): open .elpx/.elp files via drag-and-drop (#1836)

### LINKED ISSUE #1836: Drag-and-drop to open .elpx files in the editor window
This was suggested by @montesdeocaleocadia-sys

## Summary

Allow users to open a project by dragging an `.elpx` (or `.elp`) file directly onto the eXeLearning window. When a file is dropped, the app should prompt the user to confirm opening it. This should work consistently across the **cloud**, **online (self-hosted)**, and **desktop (Electron)** versions.

## Motivation

Today, opening an existing project requires going through the file picker (File → Open / Import). Dragging a file onto the window is a faster, more intuitive interaction that most users expect from a document-based application. It lowers friction for educators who manage many `.elpx` files locally.

## Proposed behavior

1. User drags an `.elpx`/`.elp` file from their file manager onto the eXeLearning window.
2. A drop overlay/indicator is shown while hovering with a valid file.
3. On drop, the app shows a confirmation dialog (e.g. "Open *filename.elpx*? Unsaved changes in the current project will be lost.").
4. On confirm, the file is imported into the current session.
   - Reuse the existing browser-side import path (`importElpDirectly` → `importFromElpxViaYjs`) so the server is not involved in parsing.
   - For large files, fall back to the chunked-upload path if needed (consistent with the current import flow).

## Scope / platforms

- **Online / Cloud (browser):** handle the native `dragover`/`drop` events on the document, validate file extension, prevent the browser default of navigating to the file.
- **Desktop (Electron):** ensure the renderer receives the dropped file path and routes it through the same import logic; prevent Electron's default of opening the file in the window.

## Acceptance criteria

- [ ] Dropping a `.elpx`/`.elp` file shows a confirmation prompt before opening.
- [ ] Confirming imports the project via the existing client-side Yjs import flow.
- [ ] Cancelling leaves the current project untouched.
- [ ] Non-supported file types are rejected with a clear message.
- [ ] Works in browser (online + cloud) and Electron desktop builds.
- [ ] Covered by unit tests and a Playwright E2E spec for the drag-and-drop flow.

erseco:
Closes #1836 This was suggested by @montesdeocaleocadia-sys

## What

Lets users open a project by **dragging an `.elpx`/`.elp` file onto the editor window**. Works in all three runtimes: **online** (server), **cloud / static PWA**, and **Electron desktop**.

When a project is already open with unsaved changes, dropping a file shows the **same "save the current project?" dialog** as opening a project the normal way — no new dialog, no duplicated logic.

## How it works

The existing open paths already converge on `modalOpenUserOdeFiles.largeFilesUpload(file)`, which:
- runs the unsaved-changes check and shows `ModalSessionLogout` (`Save` / `Open without saving` / `Cancel`), and
- runs the in-memory Yjs import (with chunked-upload fallback online).

So drag-and-drop only needs to capture the dropped file and hand it to the same funnel. The new `FileDropHandler`:

- installs a global drop target on `window`, prevents the browser/Electron from navigating to the file, and shows a translucent drop overlay;
- validates the extension (`.elpx` / `.elp`) and alerts on anything else;
- routes the file through the existing per-runtime entry points so behaviour matches a normal open exactly:
  - **Electron** → `App.openFileFromPath(file.path)` (preserves save-path behaviour),
  - **static / cloud PWA** → `App.openStaticFile(file)`,
  - **online** → `largeFilesUpload(file)`.

Disabled in **embedded mode** so an LMS iframe can't have its project replaced by a stray drop. Only the first dropped file is opened (opening a project is single-document).

### Files

- `public/app/common/fileDropHandler.js` — new handler (+ pure `isOpenableProjectFile` validator)
- `public/app/common/fileDropHandler.test.js` — Vitest unit tests
- `public/app/app.js` — instantiate + `bind()` in `init()`
- `assets/styles/components/_filedrop.scss` + `main.scss` — drop overlay
- `test/e2e/playwright/specs/drag-drop-open.spec.ts` — E2E

No backend or Electron main-process changes: in-window drops are normal renderer DOM events, and `will-navigate` in `app/main.js` already blocks file navigation.

## How to verify

**Automated**
- `npx vitest run public/app/common/fileDropHandler.test.js` — 18 tests (extension validation, per-runtime routing, overlay show/hide, embedded skip).
- `bun x playwright test --project=chromium test/e2e/playwright/specs/drag-drop-open.spec.ts` — drop opens the project; dropping with unsaved changes shows the save prompt.
- `bun x playwright test --project=static test/e2e/playwright/specs/drag-drop-open.spec.ts` — open works against the static build.
- `make fix` clean; `public/app/app.test.js` (168 tests) still green.

**Manual**
1. **Online** (`make up-local`): open a project, drag an `.elpx` from the file manager onto the window → overlay appears → drop → project opens.
2. **Unsaved prompt**: edit something, then drag+drop another `.elpx` → "save current project?" dialog appears (same as File → Open).
3. **Cloud / static PWA** (`make build-static`): repeat — opens via in-memory import.
4. **Desktop** (`make run-app`): drag an `.elpx` onto the Electron window → opens, and the next Save pre-fills the dropped file's path.
5. Drop a non-`.elpx` file → friendly "only .elpx/.elp" alert, current project untouched.


erseco:
Fixed an issue found in manual testing: dropping a file on a deeply-nested element (e.g. the properties form or the content area) still let the browser open/download the file. The cause was that an intermediate element calls `stopPropagation()`, so the `drop` event never reached the window-level handler.

The handler now attaches its drag listeners in the **capture phase**, so it always runs before any descendant `stopPropagation()` and cancels the browser's native open everywhere. It also defers to an open modal (e.g. the File Manager's own dropzone) instead of hijacking those file drops.

The E2E now dispatches the drop on a deep nested element and asserts `defaultPrevented`, so the propagation path is covered. Verified live in the browser across `#node-content`, `.form-properties input`, the title heading and the workarea — all now prevent the native open and import the project.

ignaciogros (approved):
I was expecting it to work with exported content too, but It works perfectly with `.elpx` and `.elp` files. The warning message for other formats is properly displayed.

Tested in these versions:

- Cloud.
- Online (static), in Firefox and G. Chrome.
- Desktop.

Thanks, @erseco!

PR #1849, opened 2026-05-24
fix(build): exclude node_modules and dev files from static build

### LINKED ISSUE #1848: make build-static copies unnecessary files and directories
## Summary
Running `make build-static` includes unnecessary files and directories in the final static editor build, such as the `node_modules` folder and other development-related files. These should be excluded from the production artifact.

## Root cause
The build process currently copies the directory contents without filtering out development dependencies or configuration files that are only needed during the build stage.

## Proposed solution
Implement a filtering mechanism during the static editor generation step to explicitly exclude `node_modules/` and other non-essential files/folders from being copied into the final build.

franmate:
## Description
This PR fixes the issue where unnecessary files, such as `node_modules`, were being included in the static editor build. 

Now, the build process filters the contents to ensure only production-ready files are included in the final artifact.

## Related Issue
Fixes #1848 

## Changes Made
* Updated the `build-static-bundle.ts` script to implement a filtering mechanism.

## How to Test
1. Run the static build command: make build-static

### DIFF
--- a/scripts/build-static-bundle.ts
+++ b/scripts/build-static-bundle.ts
@@ -1205,7 +1205,7 @@ async function buildStaticBundle() {
     console.log('  Copied bundles/');
 
     // Copy files/perm (themes, iDevices, favicon)
-    copyDirRecursive(path.join(projectRoot, 'public/files/perm'), path.join(outputDir, 'files/perm'));
+    copyDirRecursive(path.join(projectRoot, 'public/files/perm'), path.join(outputDir, 'files/perm'),['src', 'node_modules','bun.lock','package.json']);
     console.log('  Copied files/perm/');
 
     // Copy images folder (default-avatar.svg, logo.svg, etc.)


franmate:
Independent of the 'Slide' fix, keeping this filter is important to prevent this bug on other new IDevices. It should be implemented as a universal solution.

erseco:
I don't think so. Keeping that filter is exactly the kind of patch-on-patch fix that grows technical debt without anyone noticing.

The bug in #1848 isn't that `copyDirRecursive` is too permissive — it's that a `package.json` + `node_modules/` lived inside `public/files/perm/idevices/`, which is something the project's vendoring model already forbids (every other iDevice and every other library is built from the root). #1850 removes the nested toolchain and brings the Slide iDevice in line with the rest of the repo. There is no longer anything to filter.

If we add the filter "just in case a future iDevice does it again", we

- give the next contributor implicit permission to drop a nested install in `public/`, because the build no longer complains;
- silently swell the dev `node_modules` (the ~38 MB Fabric tree was already there during development, just hidden from `dist/static/`);
- and force every future build script that walks `public/files/perm/` to remember to exclude the same paths (the Electron packager, `scripts/build-resource-bundles.js`, etc.). Load-bearing arbitrary state.

If the goal is genuinely "prevent this bug on other new iDevices", the right shape is a guard that **fails loudly** rather than copies-but-filters — for example a small check in `build:all` or in CI that errors out if anything under `public/files/perm/idevices/**` contains a `package.json`, `bun.lock`, or `node_modules/`. That makes the problem visible in the PR of the offending iDevice, exactly where it should be fixed. I'm happy to add that on top of #1850 if you want — but please not as a hidden filter at copy time. We already carry enough technical debt; let's not add more by masking root causes with ad-hoc patches.

PR #1827, opened 2026-05-19
Add Spanish state-level LOMLOE curriculum dataset

erseco:
## Summary

This PR replaces the placeholder `lomloe-ES.json` dataset with the Spanish state-level LOMLOE curriculum data and enables the `ES` dataset in the LOMLOE iDevice. It also adds a second dataset `lomloe-ES-EFP.json` for the Ministry-managed territory (Ceuta and Melilla), built from the three Órdenes EFP that operate as the curriculum reference inside that ámbito de gestión.

Builds on the initial LOMLOE iDevice work from #1370 and closes #1823.

## What changed

- Added the national Spanish LOMLOE curriculum dataset at:
  `public/files/perm/idevices/base/lomloe/data/lomloe-ES.json`
- Added the Ministry-managed-territory dataset at:
  `public/files/perm/idevices/base/lomloe/data/lomloe-ES-EFP.json`
- Enabled both `ES` and `ES-EFP` datasets in the LOMLOE iDevice dataset registry (`available: true`).
- Kept compatibility with the existing dataset schema used by the Canary Islands dataset; the `ES-CN` entry is untouched and still loads.
- Documented the BOE sources, the cycle-to-year duplication rule, and the Infantil-only-on-ES exception in the iDevice README.
- Added Vitest tests that validate both new JSONs parse, expose the expected etapas/niveles, follow the iDevice schema, and that the registry keeps `ES`, `ES-EFP` and `ES-CN` available.

## Data sources

All datasets are extracted from official BOE sources only.

State Royal Decrees (`lomloe-ES.json`):

- `BOE-A-2022-1654` — RD 95/2022 (Educación Infantil)
- `BOE-A-2022-3296` — RD 157/2022 (Educación Primaria)
- `BOE-A-2022-4975` — RD 217/2022 (ESO)
- `BOE-A-2022-5521` — RD 243/2022 (Bachillerato)

Ministry Orders (`lomloe-ES-EFP.json`):

- `BOE-A-2022-12066` — Orden EFP/678/2022 (Primaria, ámbito MEFP)
- `BOE-A-2022-13172` — Orden EFP/754/2022 (ESO, ámbito MEFP)
- `BOE-A-2022-13173` — Orden EFP/755/2022 (Bachillerato, ámbito MEFP)

## Mapping decisions

[...]

PR #1850, opened 2026-05-24
fix(build): close #1848 at the source — remove Slide iDevice nested toolchain, vendor fabric/dompurify, slim editor bundle to 79 KB

### LINKED ISSUE #1848: make build-static copies unnecessary files and directories

erseco:
## Summary

Closes #1848. Replaces #1849.

This PR fixes the root cause of #1848 (`make build-static` shipping `node_modules` and dev files): the Slide iDevice carried its own nested `package.json` + `bun.lock` + `node_modules/` + build script. That nested toolchain was the only iDevice in the whole repo doing this — every other iDevice is built and bundled from the root.

The fix removes the nested install entirely. `fabric` and `dompurify` move into the existing `public/libs/` vendoring path (next to `jquery`, `bootstrap`, `fflate`, `yjs`, `abcjs`, `tinymce`, `pdfjs`, etc.) and the editor TypeScript source is compiled by a single root-level script (`scripts/build-slide-editor.ts`) that marks them as externals pointing at `window.fabric` / `window.DOMPurify` — the same shim pattern already used in `scripts/build-importers-bundle.js` for `yjs`.

Side effect: the editor bundle drops from **~376 KB to ~79 KB (−79%)** because fabric and dompurify are no longer inlined per-build.

## Why not #1849

#1849 adds a filter to `copyDirRecursive(...)` in `scripts/build-static-bundle.ts` so the nested `node_modules/` does not end up in `dist/static/`. That is treating the symptom, not the cause. The nested install is the bug. Filtering it out at copy time:

- Leaves the duplicated toolchain in the repo (two `node_modules` trees, two `bun install` runs, two lockfiles to keep aligned).
- Forces every future build script that walks `public/files/perm/` to remember to exclude the same paths — `scripts/build-resource-bundles.js`, the Electron packager, any future tooling. That's load-bearing arbitrary state.
- Hides the fact that fabric is being inlined into the editor bundle on every rebuild (~300 KB of avoidable churn that this PR removes).
- Does nothing to bring the Slide iDevice in line with the rest of the codebase's vendoring story (`public/libs/`).

The original `package.json` inside `public/files/perm/idevices/base/slide/` landed when fabric was in active development and not yet cleaned up. **It should never have stayed in the normal build flow** — having a sub-`package.json` inside `public/` defeats the single-source-of-truth model the repo already follows for every other library. Future build scripts have to special-case it; dependency updates require manual intervention; the static distribution silently swells.

A quick `git log -- scripts/build-static-bundle.ts` and a look at the already-open #1593 would have shown both that the vendoring story is solved in design and that the right fix is "don't put `package.json` inside `public/`", not "let's filter the output". Reviewing prior PRs before adding lines to a build script is part of the job.

## Related work this aligns with

- **#1593** (still open) — centralises *all* vendored frontend libs via `bundle:vendor` + `scripts/copy-vendor-libs.js`. This PR adopts the exact same model for the Slide iDevice (`fabric` and `dompurify` move into `public/libs/{fabric,dompurify}/`). When #1593 lands, the follow-up is mechanical: stop committing the two `.min.js` files and add two lines to `copy-vendor-libs.js`. Already prepared on the `feature/add-yjs-and-others-dependencies-to-npm` branch.
- **#1827** — the same kind of static-build hygiene work: it gzip-compresses the LOMLOE and DigCompEdu JSON datasets, saving **−82%** (~30 MB) on every static build. This PR removes another **~297 KB** per editor load (slide bundle: 376 → 79 KB) and the ~38 MB of `fabric/` `node_modules` that #1849 was filtering at copy time. Same direction, same goal: stop shipping waste.

## Comparison

| Aspect | main today | #1849 (filter) | This PR (root cause) |
|---|---|---|---|
| Nested `package.json` in iDevice | yes | yes | **removed** |
| Nested `bun.lock` in iDevice | yes | yes | **removed** |
| Nested `node_modules/` in iDevice | yes (~38 MB) | yes (~38 MB) | **removed** |
| `bun install` invocations | 2 | 2 | **1** |
| `slide-editor.bundle.js` size | ~376 KB | ~376 KB | **~79 KB (−79%)** |
| Fabric copies in repo | 2 (root + slide `node_modules`) | 2 | **1** (root only) |
| `node_modules` leaks into `dist/static/` | yes | no (filtered) | no (doesn't exist) |
| Per-script filter lists to maintain | 0 | 1 (and counting) | 0 |
| Reusable `window.fabric` for other iDevices | no (inlined) | no (inlined) | **yes** (vendored in `public/libs/`) |
| Alignment with #1593 vendoring story | no | no | **yes** (same pattern) |
| Aligned with #1827 dist hygiene | n/a | n/a | **yes** (less to ship) |

## Changes

- Delete `public/files/perm/idevices/base/slide/{package.json,bun.lock,.gitignore,src/build.ts}` and `node_modules/`.
- Add `public/libs/fabric/fabric.min.js` (+ `LICENSE`) — vendored UMD, exposes `window.fabric`.
- Add `public/libs/dompurify/purify.min.js` (+ `LICENSE`) — vendored UMD, exposes `window.DOMPurify`.
- Update `public/libs/README.md` (third-party manifest).
- New `scripts/build-slide-editor.ts` — `Bun.build` with externals plugin pointing at the globals (same pattern as `scripts/build-importers-bundle.js`).
- `package.json`:
  - Move `fabric ^7.4.0` and `dompurify ^3.2.4` into `devDependencies`.
  - `bundle:slide-editor` → `bun scripts/build-slide-editor.ts` (no more `cd … && bun install && bun run build`).
  - `postinstall` regenerates the editor bundle after each install.
- `.gitignore`: drop the obsolete `/public/files/perm/idevices/base/slide/node_modules/` line; keep the rule for `edition/slide-editor.bundle.js` (still a build artifact).
- `public/files/perm/idevices/base/slide/edition/slide.js`: lazy-load fabric + dompurify from `public/libs/` before the editor bundle. Path resolved via `resolveLibsBase(idevicePath)`. Loading stays lazy — no extra network on workarea startup.
- `public/files/perm/idevices/base/slide/edition/slide.test.js`: pre-fills `window.fabric` / `window.DOMPurify` in `beforeEach` so the bridge only injects the bundle script in the concurrency test.

## Test plan

- [x] `bun install` triggers `postinstall` → bundle generated at `public/files/perm/idevices/base/slide/edition/slide-editor.bundle.js` (~79 KB).
- [x] `ls public/files/perm/idevices/base/slide/` → only `config.xml`, `slide-icon.svg`, `src/`, `edition/`, `export/`.
- [x] `npx vitest run public/files/perm/idevices/base/slide` → **121/121** tests pass.
- [x] `make fix` clean.
- [x] `make build-static` → `dist/static/files/perm/idevices/base/slide/` contains the editor bundle and the iDevice source but no `node_modules`, no `package.json`, no `bun.lock`.
- [x] Manual E2E: `make up-local`, open a project, insert a Slide iDevice — `fabric.min.js` + `purify.min.js` + `slide-editor.bundle.js` load in that order (Network tab) and **only** when the iDevice opens (not on workarea startup); draw / undo / save / reload work.


franmate:
I don't think we should close #1848 just yet. The main goal here is to prevent unnecessary files from being copied into the static build in the first place.
The #1849 solution is specific to one iDevice, whereas #1848 provides a universal fix that prevents the same issue from happening on new or future iDevices. Keeping this filter makes the entire build process more robust and future-proof.

erseco:
"Universal in scope" and "correct" aren't the same thing. A filter in `copyDirRecursive` is universal, yes — universally **silent**. It accepts the bad input (a nested `package.json` / `node_modules/` under `public/`) and erases it from the output. The next contributor adds the same mistake on a new iDevice, the build is green, the dev `node_modules` quietly grows by tens of MB again, and nobody notices until somebody runs `du` on the repo. That's not future-proofing; that's burying the problem.

The repo *already* has a universal rule, and #1849 contradicts it: every iDevice and every third-party library is built from the root and vendored through `public/libs/` (the model #1593 is finishing). The Slide iDevice was the only one breaking that rule. #1850 puts it back in line — there is nothing left to filter, and "filter just in case" sets the precedent that nested installs under `public/` are tolerable as long as the output stays clean. They are not.

If the worry is "what if a future iDevice does it again", the answer isn't a hidden filter — it's review enforcing the rule the repo already follows. A silent filter doesn't enforce anything; it covers for whoever broke the rule.

ignaciogros (commented):
Thanks, @erseco.

If the iDevice will use common .js files, maybe it could use some of eXe's functions too, like `eXe.app.loadScript` instead of a custom `loadScript` function (it's not important, because it's a really simple function). 

Maybe `resolveLibsBase` can become a common function too. There might be a function for that already, but it'd be useful to have it if not.

pabloamayab:
Hi @erseco . 
I have tested next commands and it's ok.

> bun install triggers postinstall → bundle generated at public/files/perm/idevices/base/slide/edition/slide-editor.bundle.js (~79 KB).
>  ls public/files/perm/idevices/base/slide/ → only config.xml, slide-icon.svg, src/, edition/, export/.
>  npx vitest run public/files/perm/idevices/base/slide → 121/121 tests pass.
>  make fix clean.

  But the next one still generates the specified directory with over 80 MB; the nodes module is generated

> make build-static → dist/static/files/perm/idevices/base/slide/ contains the editor bundle and the iDevice source but no node_modules, no package.json, no bun.lock.

I have tested in Windows 11+GitBASH

image

erseco:
@pabloamayab That `node_modules/` is leftover junk from installs done before the hotfix. Commit `83374c1` removed `slide/`'s `package.json` and its `postinstall`, so it no longer regenerates — but the directory created earlier is still in your working copy and `make build-static` drags it into `dist/`.

Delete it and try again:

```bash
rm -rf public/files/perm/idevices/base/slide/node_modules
make build-static
```

After that, `dist/static/files/perm/idevices/base/slide/` should only contain `config.xml`, `slide-icon.svg`, `src/`, `edition/` and `export/`.

And this is exactly why merging #1849 is a mistake: it just hides the bloat with a `.gitignore`/cleanup band-aid instead of fixing the root cause. You'd end up carrying junk files in your branch that shouldn't be there — a latent problem waiting to bite us later. This PR fixes it at the source.

pabloamayab:
Hi @erseco !!! Sorry, after running `rm -R public/files/perm/idevices/base/slide/node_modules` and then `make build-static`, everything is working properly now.

pabloamayab (approved):
OK!!!

erseco:
Thanks @mnarvaezm for the thorough review! 🙏

All 9 findings are confirmed and fixed in `1482ca7`:

| # | Finding | Fix |
|---|---------|-----|
| 1 | Save from Code mode produced a stale SVG | Textarea is mirrored back into the canvas on `input` (debounced ~200 ms), so `getFabricJSON()` and `getSvgString()` stay in sync at save time. |
| 2 | Dimensions changed on reopen | Extracted `clampDimensions()` in `serializer.ts` as the single source of truth; used by both the editor (on input) and the serializer (on load). Width/Height inputs now use per-axis MIN/MAX bounds. |
| 3 | `_scrubSvg` let unquoted handlers through | Added `replace(/\son[a-z]+\s*=\s*[^\s>]+/gi, '')` with regression tests for `onload=alert(1)` and variants. |
| 4 | Future-version payload → blank canvas | `parsePrevious` now returns `unsupportedVersion: true`. The editor shows a banner, disables editing, and the bridge re-emits the original payload untouched (`getUnreadPayload()`), so an older bundle can never overwrite a slide authored by a newer one. |
| 5 | Resize/background outside Undo | New `src/snapshot.ts` serializes `{scene, width, height, background}`; `setDimensions` and `setBackground` push/commit to history; `applySnapshot` restores all three fields on Undo/Redo. Legacy scene-only snapshots are accepted for forward compatibility. |
| 6 | `fullscreenchange` touched other slides | Only the wrapper entering/leaving native fullscreen is updated; `body.style.overflow` is reset only when no slide remains active (native or CSS-simulated). |
| 7 | Broken bundle stayed cached | The `__slideEditorInit.mount` validation now runs inside the `loadBundle` chain, so its `.catch` clears `window.__slideBundlePromise` and the next init can retry. |
| 8 | Dead `_state` in export | Removed the assignment and the tests that pinned it down; `renderView` already re-normalizes its own argument. |
| 9 | `opts.background` no-op | Removed from `CanvasAdapterOptions` and the editor; the background is always painted via CSS from `--slide-canvas-bg`. |

Coverage: iDevice vitest suite **143 passing (was 121)** — new tests cover `clampDimensions`, the snapshot codec, the `unsupportedVersion` flag, the unquoted-handler scrub, the bundle-cache reset, and the unread-payload save path.

Out of scope for this commit: Playwright E2E coverage for the Code→Save flow and the future-version banner — those land in the follow-up PR to keep this hotfix focused.

erseco:
Thanks @ignaciogros for the suggestion! 🙏 I looked into both before touching the bridge — here's where I landed:

### `loadScript` — keeping the bridge's own version (for now)

`eXe.app.loadScript` does exist (`public/app/workarea/project/idevices/idevicesEngine.js:2772`), but routing the bridge through it would be a small net regression on three fronts that aren't obvious from the outside:

1. **Cache-busting.** It unconditionally appends `?t=${Date.now()}` to every URL. Fabric.js (~330 KB) and DOMPurify (~22 KB) are loaded once per session in the current bridge and then cached by the browser across iDevice opens — going through `eXe.app.loadScript` would force a re-download every time, which is noticeable on slow links.
2. **Callback vs Promise.** The bridge reads as a Promise chain (`loadGlobal(fabric) → loadGlobal(DOMPurify) → loadScript(bundle) → validate`); with a callback-only loader we'd end up wrapping each step in `new Promise(...)` and the code would actually get longer than the 9-line helper it replaces.
3. **Host independence.** The bridge has to keep working outside the workarea — in the vitest suite (`edition/slide.test.js` mocks `document.createElement('script')` directly) and in the static editor / embedding flows where `eXe.app` isn't guaranteed to be present.

The one genuine benefit of `eXe.app.loadScript` — cleanup via `ideviceScriptsElements` — is actually undesirable here, since we *want* Fabric/DOMPurify to stick around between opens.

### `resolveLibsBase` — fully agree, just not in this hotfix

You're right that this shouldn't live inside one iDevice. Today `preview-sw.js` hardcodes `/libs/pdfjs/...` and the slide bridge derives `/libs/` on its own — both should be reading from one `eXe.app.paths.libs` (or similar) seeded with `BASE_PATH` from `projectManager.js`.

That work belongs alongside #1593 (which is making Bun the source of truth for vendored libs) rather than inside this hotfix. I've opened **#1857** to track it as a `v4.1.0` enhancement, covering both the centralized base path and a Promise-friendly, cache-respecting `loadScript` that the slide bridge can migrate to once it's in place.

erseco:
Thanks @mnarvaezm for the second-pass review and for your patch — it caught five real issues that the unit tests on the previous round missed (they were on the editor surface, not the bridge). Fixed in 3a3d537.

1. **`getUnreadPayload()` not on `mount()` (critical).** The bridge already checked for it, but `src/index.ts` never exposed it — so v4 payloads were quietly downgraded to v3 in production. Now passed through.
2. **Saving from Code mode could emit stale SVG (critical).** The textarea→canvas mirror is debounced + async, but `$exeDevice.save()` is sync. Added `canSave()` on the editor; mid-edit Code mode now shows an inline error and the bridge bails out instead of mixing new JSON with old SVG.
3. **Visual/Code toggle bypassed the unsupported-version lockout (high).** Exiting Code mode always stripped `--disabled`. `setCodeMode` / `finishExitCodeMode` now honor `unreadPayload`, the toggle is `disabled` + `aria-disabled`, and `setDimensions` / `setBackground` are no-ops while the banner is up.
4. **`aria-label` wrong for slides in CSS-fallback fullscreen (medium).** Decision now uses `slide-fs-active` alongside `document.fullscreenElement`.
5. **Resize coalesced into the next history entry (medium).** `setDimensions` used `push()` (debounced); a quick resize + add-shape produced a single undo step. Switched to `commit()`.

Tests: 144/144 vitest passing; added 3 Playwright specs (code-save guard, future-payload lockout, dimension-undo) and unit tests for `canSave` and the fullscreen aria-label.

PR #1855, opened 2026-05-26
docs: improve CLI usage with docker run examples

erseco:
## Summary

- Document running the CLI directly against the published `exelearning/exelearning` image with `docker run`, so users who only want to use the CLI (e.g. for export/convert automation) don't need to clone the repo or use Docker Compose.
- Add `elp:export` and `elp:convert` examples that mount host files on `/data`, with per-shell tabs for macOS / Linux / Git Bash, Windows PowerShell, and Windows CMD.
- Warn explicitly against `-w /app` and mounting on top of `/app` (would hide `docker-entrypoint.sh`), and add a troubleshooting note for the resulting `[dumb-init] /app/docker-entrypoint.sh: No such file or directory` error.
- Keep the existing `docker compose exec` instructions for development, clearly distinguishing the two use cases.

## Test plan

- [x] Render `doc/development/environment.md` with MkDocs (Material) and confirm the `pymdownx.tabbed` tabs display correctly.
- [x] Run `docker run --rm exelearning/exelearning bun cli` and confirm the command list is shown.
- [ ] Run the `elp:export` and `elp:convert` examples with a sample `project.elpx` / `project.elp` in `\$PWD` and confirm the output appears on the host.


pabloamayab (approved):
That's right, @erseco . I had to make a few small changes because there was some incorrect Markdown code.

ignaciogros (dismissed):
I tested this on Windows.

The PowerShell and CMD commands work correctly, but the Git Bash version does not. Git Bash/MSYS automatically rewrites POSIX-style paths passed to Docker, so paths like `/data/elpx.elpx` are transformed into something like `C:/Program Files/Git/data/elpx.elpx`, which causes the container to fail with “Input file not found”.

Using `MSYS_NO_PATHCONV=1` fixes the issue in Git Bash.

**Example:**

```
MSYS_NO_PATHCONV=1 docker run --rm \
  -v "/c/myPath/Downloads/cli:/data" \
  exelearning/exelearning \
  bun cli elp:export /data/project.elpx /data/exported --format=html5
```

erseco:
Thanks a lot for testing this on Windows, @ignaciogros — great catch on the MSYS path-rewriting issue.

I've split the docker-run snippets into four shell tabs so users copy the command that matches their shell:

- `macOS / Linux`
- `Windows Git Bash`
- `Windows PowerShell`
- `Windows CMD`

The new `Windows Git Bash` tab includes a short note explaining the MSYS path-rewriting behaviour and prefixes every command with `MSYS_NO_PATHCONV=1`, so the `/data/...` paths reach the container untouched. The same treatment has been applied to the three blocks (export, legacy conversion, troubleshooting `-w /app`).

Could you re-test on Git Bash when you have a minute? Thanks again for the observation!

Captura de pantalla 2026-05-28 a las 11 11 50

ignaciogros (approved):
Tested on Windows Git Bash.

Thanks, @erseco!

PR #1865, opened 2026-05-29
fix(workarea): tolerate whitespace in cssClass to prevent render crash

### LINKED ISSUE #1864: Page stuck loading: invalid cssClass value crashes render via classList.add('')
## Summary

When a block or iDevice has a **CSS class** property that contains leading/trailing/double whitespace — for example when a user pastes a whole CSS rule into the "CSS class" field — opening the page throws a fatal error and the page gets stuck loading.

## Error

```
Uncaught (in promise) SyntaxError: Failed to execute 'add' on 'DOMTokenList': The token provided must not be empty.
    at setPropertiesClassesToElement
    at generateBlockContentNode
    at loadComponentsPage
```

(The many report-only CSP warnings in the console are unrelated noise.)

## Root cause

`setPropertiesClassesToElement()` in both `public/app/workarea/project/idevices/content/blockNode.js` and `ideviceNode.js` splits `cssClass.value` on a single space and adds every token to `classList` without filtering empty strings:

```js
let cssClasses = this.properties.cssClass.value
    ? this.properties.cssClass.value.split(' ') // ' a  b ' -> ['', 'a', '', 'b', '']
    : [];
cssClasses.forEach((cls) => { el.classList.add(cls); }); // classList.add('') throws
```

A value with leading/trailing/consecutive whitespace yields empty tokens, and `classList.add('')` throws a `SyntaxError`. The throw aborts `generateBlockContentNode` -> `loadComponentsPage`, so the page never finishes rendering.

## Reproduction

1. Edit a block or iDevice **CSS class** property and enter a value with extra spaces, or paste a CSS rule such as `.foo {  color: red;  }`.
2. Reopen / reload the page that contains it.
3. The page stays stuck loading; the console shows the DOMTokenList `SyntaxError`.

## Proposed fix

Parse the class string defensively (split on any whitespace, drop empty tokens) before calling `classList.add`. Page rendering must tolerate any stored value. Validating the field on input is a worthwhile follow-up but is out of scope for the hotfix.

erseco:
## Problem

Users reported that opening certain pages leaves them stuck loading. The console shows a flood of report-only CSP warnings (unrelated noise) and one fatal error:

```
Uncaught (in promise) SyntaxError: Failed to execute 'add' on 'DOMTokenList': The token provided must not be empty.
    at setPropertiesClassesToElement
    at generateBlockContentNode
    at loadComponentsPage
```

Captura de pantalla 2026-05-29 a las 12 47 41

## Root cause

`setPropertiesClassesToElement()` in `blockNode.js` and `ideviceNode.js` split the block/iDevice `cssClass` value on a single space and added every token to `classList` without dropping empty strings. A value with leading/trailing/double whitespace (for example when a user pastes a whole CSS rule into the "CSS class" field) produces empty tokens, and `classList.add('')` throws. The throw aborts the page render, so the page never finishes loading.

Reproduced live in Chrome against a real project: running the original logic against the stored value threw after 2 classes with the exact production error; the fixed logic adds all tokens with no throw.

## Fix

Extract a shared pure helper `parseCssClassList()` (single source of truth) that splits on any whitespace and drops empty tokens, and use it in both `blockNode.js` and `ideviceNode.js`. Page rendering now tolerates any stored value.

Input-side validation of the field is a possible follow-up and is intentionally out of scope for this hotfix.

## Tests (TDD)

- `cssClassHelper.test.js` — unit tests for the helper (empty, null/undefined, non-string, whitespace-only, normal, double/leading/trailing spaces, tabs/newlines, pasted CSS rule).
- Regression tests in `blockNode.test.js` and `ideviceNode.test.js` using a strict `classList` that throws on empty tokens (like a real browser, unlike jsdom). Confirmed RED before the fix, GREEN after.
- E2E spec `block-css-class-whitespace.spec.ts` reproducing the page-load crash. Confirmed RED (block never renders) before the fix, GREEN after.

## Verification

- `make fix` clean.
- Frontend unit suite: 229 files, 12577 tests passing.
- New E2E spec passing on chromium.

Fixes #1864


ignaciogros (approved):
Detecting that was not easy! Thanks.

The solution works, although it feels a bit complicated. Maybe the following line could have been added below `let cssClasses = this.properties.cssClass.value;`

```javascript
let cssClass = cssClasses.trim().replace(/\s+/g, ' ');
```

ignaciogros (commented):
I suggest merging the proposed solution now so content creators can continue working with the problematic contents.

Please send an additional PR if you think it would be better to simplify the solution.

Thanks.

erseco:
Thanks @ignaciogros! Agreed that the field handling is conceptually a one-liner.

One note on the inline `trim().replace(/\s+/g, ' ')` version: it fixes the reported pasted-CSS case, but a whitespace-only value (e.g. `"   "`) trims to `""`, which `split(' ')` turns into `['']`, so `classList.add('')` would still throw — it needs an extra `.filter(Boolean)` to be fully safe.

That's why the helper uses `value.split(/\s+/).filter(Boolean)`: it covers leading/trailing/multiple spaces, tabs/newlines and whitespace-only values in one place. I extracted it into `cssClassHelper.js` rather than inlining because the same logic runs in both `blockNode.js` and `ideviceNode.js` — following AGENTS.md's "single source of truth / extract testable pure functions" — so it stays deduplicated and unit-tested.

Happy to switch to the inline style if you'd prefer it. Thanks for the quick review and merge!

erseco:
Also confirming: the fix works perfectly. Verified live in Chrome against the affected project and green across all CI E2E projects (chromium, firefox and static), so the previously stuck pages now render correctly. 🎉

ignaciogros:
You're right :-)

Thanks, @erseco!

Issue #1838, opened 2026-05-21
"Selecciona Multimedia" iDevice plays question audio automatically on first page, hides speaker icon


greynoa:
[errormultimedia.elpx.zip](https://github.com/user-attachments/files/28094231/errormultimedia.elpx.zip)

<h3>Description</h3>
<p>There is a bug in the <strong>Selecciona Multimedia</strong> (Select Multimedia) activity/iDevice. When you add an audio file to the main question/statement (enunciado), the audio automatically triggers and plays as soon as the page loads.</p>

<p>Furthermore, once it finishes playing, there is no way to replay it because the speaker icon does not appear at all.</p>

<blockquote>
  <strong>Note:</strong> This issue <strong>only happens with the first multimedia element / first page</strong> of the activity. On subsequent pages or items within the same iDevice, the activity functions correctly (the speaker icon is visible and it doesn't autoplay).
</blockquote>

<h3>Steps to Reproduce</h3>
<ol>
  <li>Add a <strong>Selecciona Multimedia</strong> iDevice to a page.</li>
  <li>Add an audio file to the main statement field of the first item.</li>
  <li>Save and view the page (or preview it).</li>
  <li><strong>Result:</strong> The audio plays automatically on load, and the speaker icon is missing for the statement. Subsequent pages work as intended.</li>
</ol>

<h3>Expected Behavior</h3>
<p>The audio should not play automatically upon entering the page. A speaker icon should be visible next to the text, allowing the user to click and play/replay the audio at will, just like it happens on the rest of the pages.</p>

<h3>Additional Context</h3>
<p>I have attached a sample <code>.elpx</code> file where you can directly test and verify this behavior.</p>

franko-c:
from what I can see, the speaker icon on the first question is hidden because `select-media-files.js:289` gates both `playSound()` and the icon display behind `if (num > 0)`. on init (line 111), `showPhrase(0, i)` fires with `num=0`, so line 286 hides the icon and nothing re-shows it.

splitting the icon from the autoplay gate:

```js
if (
    mOptions.phrase.audioDefinition &&
    mOptions.phrase.audioDefinition.length > 4
) {
    if (num > 0) {
        $exeDevices.iDevice.gamification.media.playSound(
            mOptions.phrase.audioDefinition
        );
    }
    $(`#slcmpAudioDef-${instance}`).css('display', 'block');
}
```

[...]

Issue #1763, opened 2026-05-06
Remove the JavaScript player


ignaciogros:
eXe's `mediaelement` library is no longer needed. It just provides a common UI, but the important functionality (improve compatibility) is not needed any more, because it provides a Flash fallback that will no longer work.

PR #1868, opened 2026-05-30
feat(file-manager): centralized asset metadata foundation (description, alt, title, license, author)

### LINKED ISSUE #1087: Feature Request: New "Resources" iDevice
The current methods to add materials in eXeLearning (such as "Attached Files" or "External website") lack a unified interface to list multiple items with their respective educational metadata. Currently, there is no automatic relationship between the resources used across different pages and a final centralized list.

**Proposed Feature:** A new iDevice in the "Information and presentation" category called "Resources". This iDevice would allow users to create a collection of materials (files or links) where each entry is treated as an independent object with its own metadata.

Key Functional Requirements (per resource entry):

* General Info: Title and a brief Description.

* Source: * Local File: Upload/Select a file to be embedded in the .elpx. or External Link: URL for online resources.

* Metadata (especially for local files):

- Author/Creator: Field to specify who created the resource.

- License: A dropdown menu with standard licenses (Creative Commons, Public Domain, etc.) + a free-text field for custom licenses.

Advanced Integration (Direct Linking): To improve workflow, I suggest adding a "Insert Resource" button in the standard editing toolbar (TinyMCE) or the iDevice menus.

This button would allow the creator to "call" or link to any resource already defined in the "Resources" list.

This creates a dynamic relationship: if a resource is updated in the central list, the links across the different pages stay consistent.

Proposed UI Logic: The iDevice should allow an "Add another resource" action to build the list. In the final export, this should be rendered as a clean, accessible list where license information is clearly visible (ideally via icons).

Visual Concept: To better illustrate the workflow and UI integration I have in mind, here is a conceptual mockup. In my head, the idea would be the following:

Image

Image

Image

This is how the final list would look, centralizing all project resources in a single, clean, and professional layout:

Image

### LINKED ISSUE #1243: Feature Request: Centralized "Global" Metadata for Project Images to avoid data entry redundancy
Currently, when an author inserts an image into an iDevice (e.g., using the TinyMCE text editor), they are presented with dialog boxes to enter information regarding the image.

* In the "General" tab, they must enter the "Alternative description" and "Image title".

* In the "Title and recognition" tab, they must enter crucial OER information such as "Source/Author" and "License".

**The Problem:** This metadata is currently tied to the specific instance of the image within that specific iDevice, rather than to the image file itself within the project.

If an author needs to reuse the exact same image in a different iDevice on another page, eXeLearning presents blank fields again. The author is forced to re-type the alternative text, author name, and re-select the license. This redundancy is inefficient, tedious, and increases the risk of inconsistent attribution across a project.

**Proposed Feature:** I propose implementing a centralized system where metadata is associated globally with the image file upon its first upload or definition within the project.

**Desired Workflow:**

When an image is first uploaded to the project (or perhaps managed via a central "Image Library"), the author defines its metadata once: Title, Alt Text, Author, and License.

Subsequently, whenever that specific image file is selected for insertion into any iDevice throughout the project, the fields in the "Insert/Edit Image" dialogs  should automatically pre-fill with this stored global data.

(Optional refinement: The author could still be allowed to override these pre-filled fields for a specific instance if absolutely necessary, but the default behavior should be inheritance from the global definition).

**User Benefits:**

**Efficiency**: Significantly reduces time spent re-entering the same data for commonly used assets (like logos, icons, or recurring diagrams).

**OER Compliance & Consistency**: Ensures that crucial authorship and licensing information is attached to the image wherever it appears, reducing attribution errors.

**Easier Maintenance**: If an author's name needs to be updated, it only needs to be done once centrally, rather than finding every instance of the image in the project.

Image
Image

### LINKED ISSUE #1244: Enhancement: Add Searchable Image Descriptions in File Manager
The File Manager (Gestor de archivos) currently displays basic metadata: filename, type, size, and date. While it includes a search bar, it only filters by the filename.

**The Problem:** In large projects with many assets, filenames are often insufficient to identify or locate specific images. A filename like perro.png tells the author very little about the actual content of the image. Currently, authors must rely on visual previews or memory to find the right asset, which is inefficient.

**Proposed Feature:** I propose adding a "Description" field for image files within the File Manager. This field should be:

**Editable:** Authors can add a detailed description after uploading the file.

**Searchable:** The existing search bar should query both the filename and this new description field.

Specific Example:

Filename: perro.png

Proposed Description: "A white toy poodle sitting on a wooden floor, looking directly at the camera with a bright, smiling expression."

In this scenario, if a user searches for "poodle" or "smiling", the File Manager would instantly surface perro.png, even if those words aren't in the title.

**UI Integration (based on attached mockups):**
**In the List View**: The description could appear as a new column or as sub-text under the name.

**Automatic Alt-Text**: Ideally, if an image has a description in the File Manager, this text could pre-fill the "Alternative description" field when the image is inserted into an iDevice, saving even more time.

**User Benefits:**
**Improved Searchability**: Find assets based on content, not just nomenclature.

**Better Organization**: Helps teams or individual authors maintain a clear overview of their media library.

**Accessibility Workflow**: Encourages authors to write descriptive text early in the process, which translates to better Alt-Text for the final OER.

Image

### LINKED ISSUE #1817: Improve File Gallery usability and asset management
## Description

While using the file gallery, I noticed several areas that could significantly improve usability and resource management, especially in projects with many reused assets.

### 1. Sort by number of references

Have you considered adding an option to sort files by the number of references/usages in iDevices?

This would make it much easier to:

* identify the most reused assets,
* detect unused files,
* manage large projects more efficiently.

### 2. Incorrect file date displayed

The date shown in the gallery does not seem to reflect the actual file date.

In my case, all files appear to have been updated today, even when they were not recently modified.

### 3. Replace image/file button

It would be extremely useful to add a button to replace an existing image/file while keeping all current references intact.

This would simplify updating assets that are used across multiple iDevices without needing to manually reinsert them everywhere.

### 4. Show the iDevices where an image is used

Currently, the gallery only shows the number of references/usages.

Would it be possible to also list the specific iDevices where the selected image is being used?

This would help to:

* quickly locate asset usage,
* review related content,
* avoid accidental deletions.

### 5. Increase the width of the “Sort by” dropdown

The “Sort by” dropdown needs a bit more horizontal space, since the text is currently truncated and difficult to read.

## Screenshot

The attached screenshot highlights some of the issues mentioned above:

* “Sort by” dropdown text being cut off,
* incorrect file date,
* proposed “Replace” button.

Thanks for the great work on eXeLearning 🙂

Image

### LINKED ISSUE #1860: About the new 3D Mol iDevice
Regarding the new iDevice 3D Mol, it would be nice if a license could be attributed to the 3D image, just as you can already do with still images.

Greetings and thanks for all the work and improvements being done.

erseco:
## Summary

Makes the **File Manager the single source of truth for reusable asset metadata** (`description`, `alt text`, `title`, `license`, `author`) and builds the first two consumers on top of it: a **Resource Report iDevice** and a set of **File Manager usability improvements** (usage insight, references sorting, and file replacement).

> **This PR now consolidates three previously separate, stacked PRs into one.**
> It supersedes and **includes the full work of #1869 and #1870**, which have been closed in favour of this single thread. See the rationale in the pinned comment below.

This is a deliberately broad but coherent line of work around centralized resource/metadata/licensing management. Because several open design questions span all three pieces (see *Open questions* at the bottom), keeping them in one PR makes the discussion and review tractable instead of fragmenting it across a stack.

cc @greynoa @cristinavaldera @pabloamayab @celestebonillaTF @mnunezcedec

### 1. Centralized asset metadata foundation (originally this PR)

**Data model (additive, backward compatible)**
- Migration `008` adds nullable `description` / `alt_text` / `title` / `license` / `author` columns to `assets` (portable `TEXT`, runs on SQLite/PostgreSQL/MariaDB).
- Mirrored in the Yjs `assets` map (the client source of truth) and preserved across rename / move / import.

**API**
- Asset serializers expose the new fields (internal API + external API v1).
- `PATCH /api/projects/:projectId/assets/by-client-id/:clientId/metadata` to update metadata (trims strings, allows empty, ignores unknown keys).
- `?search=` on the list endpoint matches filename **+ description + alt text** (case-insensitive).

**File Manager UI**
- Editable metadata in the existing properties panel — available for **all file types** (images, audio, video, PDF, 3D models…). The image-specific *Alternative text* row is shown only for images.
- Search now finds assets by description / alt text, not just filename.
- License reuses the **existing** license vocabulary + `resolveLicenseKey()` — **no new/duplicate license registry**.

**Insertion pre-fill (exeimage / TinyMCE)**
- Picking an asset pre-fills `alt`, `title`, *Source/Author* and *License* in the image dialog from the asset's centralized metadata, **without overwriting** values the user already typed.

**ELPX round-trip**
- The exporter writes a tolerant `content/asset-metadata.json` sidecar (only when metadata exists); the importer merges it back and ignores it when absent, so older `.elpx` files import unchanged.

### 2. Resource Report iDevice (was #1869, closes #1087)

- New **Resource report** / *Informe de recursos* iDevice (`json` component type), auto-discovered via `config.xml`.
- **Automatically lists project resources** (no manual data entry): all assets by default, with a **type filter** (images / audio / video / documents / other) and a **"used in this project"** filter.
- Per resource: thumbnail for images (or inline-SVG file-type icon), **title** (filename fallback), optional description, author and license (shown only when present), and **View** + **Download** links.
- Links use `asset://` URLs resolved by the preview resolver and rewritten to packaged paths by the export pipeline, so the report works in **preview and export**.
- Moves asset-reference scanning into `AssetManager` (single source of truth) so the File Manager usage badges and the iDevice "used" filter share it.

### 3. File Manager usability improvements (was #1870, closes #1817)

- **Sort by references/usage** — *most used first* / *least used first*, computed in a single project traversal (`AssetManager.getAllAssetReferenceCounts()`).
- **"Used in"** — the selected-asset panel lists where the asset is used (Page · iDevice · Block, first 10 + "+ N more"), or **"Not used in this project"**, via `AssetManager.getAssetUsageLocations()` reusing the same traversal.
- **Replace file** — overwrites an asset's binary **under the same id/clientId**, so existing `asset://id` references and centralized metadata + the *Added* date are **preserved**. Validates same broad type, refreshes the preview/thumbnail, persists on save/sync (idempotent by clientId — no new endpoint).
- **"Date" → "Added"** relabel (the value is `createdAt`, the date the asset was added; the original file mtime is not preserved, so it is not faked).
- Single shared traversal (`_forEachComponent`) feeds count, referenced-ids, per-id counts and usage-locations.

## Compatibility
- All fields optional/nullable — existing projects, assets and `.elpx` files load unchanged.
- No new dependencies; no new asset model or duplicated reference scanning.

## Tests
- **Backend:** migration, queries, routes, providers, exporter.
- **Frontend:** AssetManager (field preservation + update, reference scanning, usage locations, one-pass counts, `replaceAssetContent`), File Manager (edit/save, search, references sort, "Used in", replace handler), Resource Report edition + export, shared license options, TinyMCE/exeimage attribution mapping.
- **E2E:** `file-manager-metadata.spec.ts`, `resource-report-idevice.spec.ts`, `file-manager-usage-replace.spec.ts`.
- `make fix`, `make test-unit`, `make test-integration` and the affected E2E specs pass on the consolidated branch.

## Related issues

Closes #1087
Closes #1817
Closes #1244
Closes #1243

Also relates to #1817, #1087, #1860 (licensing extended to non-static-image resources) and #1838.

## Open questions for reviewers (carried over from #1868/#1869/#1870)

These are the design decisions this line of work needs to settle — please weigh in here:

1. **Cross-iDevice centralization** (@cristinavaldera): should inserting a resource in *any* iDevice (Image, Video, games…) read metadata from the File Manager and reconsider the per-iDevice "Title and attribution" / authorship-licence fields to avoid duplication? How costly is that iDevice review?
2. **Real-time propagation**: when metadata is edited in the File Manager, should already-inserted instances update automatically (similar to the Source File iDevice), instead of only on iDevice edit/save? (@cristinavaldera, @pabloamayab)
3. **Backwards-compat import**: should legacy per-instance metadata (the "traditional" way) be imported into the centralized store on load?
4. **Resource Report layout** (@cristinavaldera, @pabloamayab): add a **table / condensed** layout in addition to list and cards; fix the **card-layout text overflow**.
5. **Search scope** (@greynoa): extend File Manager search beyond description/alt to **author, license and title**.
6. **Panel ergonomics** (@cristinavaldera): make the metadata panel larger / more prominent (above technical details), avoid the explicit "Save metadata" button, and clarify the "title" vs "caption" semantics; optionally store resource/author **URL** as metadata too.

## Fix — Resource Report snapshots inflated asset usage counts

A Resource Report iDevice (default `resourceMode:'all'`) stores an `asset://` link for every listed asset, so the reference scanner counted every asset as used — breaking the "sort by references" ranking and the "not used in this project" filter. The scan now skips `resource-report` components, so only genuine references count. Commit `b0d273e8`; test added.

## Fix — 3 minors (SQLite LIKE escape, i18n headers, neutral labels)

1) Asset search now emits an explicit `ESCAPE '\'` on its `LIKE` clauses, so the `%`/`_`/`\` escaping is actually applied on SQLite (default driver) — searching e.g. `report_2024` matches the literal filename instead of nothing. 2) The Resource Report table headers use literal `this.t('…')` calls so the i18n extractor picks them up. 3) The File Manager metadata labels use asset-type-neutral wording (`Name`, `Source URL`) instead of `Image name`/`Image URL`, which were misleading for audio/video/document assets. Commit `93c074ba`; ESCAPE covered by a new test.


erseco:
@cristinavaldera @greynoa maybe can be interesting searching for more fields as well as the description like author.... what do you think about?

cristinavaldera (changes requested):
Thanks, Ernesto!! We really needed this!

I’ve noticed a few things I wanted to comment on:

- The metadata requires a lot of scrolling to reach. Maybe it should be more visible, for example by placing it above the technical details such as type, size, date, usage, URL, etc.

- Could we avoid having to click the “Save metadata” button?

- The "image title" field is being interpreted as the accessibility title, not as the image name. The image name is actually in the “Title and caption” tab, in the “Caption text” field. It should appear there instead.

Thanks!

cristinavaldera (commented):
This change represents a very important improvement for eXe as an OER authoring tool. Being able to add authorship and licence information directly from the file manager is a fantastic feature for OER.

That said, I think the change is deeper than it may seem at first sight, because it opens the door to centralising resource metadata management in the file manager.

** Implications for iDevices**

If the idea is to centralise metadata in the file manager, when inserting a resource in any iDevice —whether it is Text, Image, Video or any other type of content— eXe should retrieve its metadata and display them below the resource.

This would make it necessary to review some current iDevice features. For example, the following might no longer make sense, or might need to be reconsidered (to avoid duplicate data):

1. The “Title and attribution” tab in the Image or Video plugins.
2. The authorship and licence fields in games and interactive activities.
3. Other similar fields that are currently managed independently within different iDevices.

In other words, this change would probably affect not only the file manager, but also the behaviour of iDevices.

Would the necessary changes in the iDevices be very costly?

**Backwards compatibility**

This new feature should maintain backwards compatibility with resources that have already been created.

In this regard, it would seem logical that, if an image or resource already has associated metadata (in teh "traditional way"), those metadata should be imported into this new centralised way of storing and managing them.

**Update**

It would also be necessary that, if the metadata are modified in the file manager, those changes are updated in the places in the project where the resource has already been inserted, in a similar way to how some fields in the Source File iDevice work.

**Other adjustments**

In addition to the above, I see a few minor adjustments that could improve the feature:

1. Make the panel larger so that metadata can be edited more comfortably.
2. Also include the link to the resource and to the author as metadata, since both are currently available when inserting an image or video.

In summary, I think this is a very powerful and necessary improvement, but it would be worth considering its overall scope so that metadata management becomes truly centralised and consistent throughout eXe.

greynoa:
Thanks @erseco — this is exactly the direction a few of us were hoping for. It nicely brings together several threads (#1838 among them), so it feels like the natural convergence of ideas that have come up from different angles. Big +1 to extending search beyond the description: matching on author (and ideally license and title too) would make the File Manager much more useful as assets pile up, especially for OER reuse where you often look for "everything by X" or "everything under license Y".

I also completely agree with the points raised by @Cristinavaldera. Having the information uploaded just once and fully centralized is, without a doubt, the most sensible approach.

From an authoring perspective, this would be a massive quality-of-life improvement for teachers and creators. It would save a lot of time by eliminating the need to repeatedly enter the exact same data for the same resource across different iDevices. Moreover, centralizing this prevents human error, ensuring we don't end up with conflicting data or different licenses assigned to the exact same asset within a project.

This PR is a fantastic and much-needed foundation to achieve that true consistency.

pabloamayab:
Hi @erseco !! I've tested it on the desktop and in the cloud, and everything works fine.

As @cristinavaldera says, I miss the ability to update metadata for items that are already inserted, or at least for the metadata to update when editing/saving an iDevice. Although in my opinion, if this complicates development, it’s a minor issue, since the user can see how many references there are for that resource and search for them to modify.

In the Image Gallery, each image can include a title, link title, author, author's URL, and license. I think this iDevice could benefit from this improvement. Although we might need to create a separate issue for that.

erseco:
## 🔀 Consolidation note: #1869 and #1870 folded into this PR

I've merged the two stacked PRs into this one and closed them:

- **#1869** — *Resource Report iDevice* → now section **2** here (still **Closes #1087**).
- **#1870** — *File Manager usability improvements* → now section **3** here (still **Closes #1817**).

The branch `feature/file-manager-metadata-management` now contains the full combined work (foundation + iDevice + usability) and is up to date with `main`. The `Closes` for #1087 and #1817 were moved here, and the closing keywords were removed from #1869/#1870 so issue tracking points to a single place.

### Why consolidate

This started as a clean 3-step stack, but the reviews made it clear the work is more interdependent than a linear stack suggests — and **it has decisions to make that span all three pieces** (centralizing metadata across iDevices, real-time propagation of metadata edits, backwards-compat import, report layout). Reviewing those across three PRs meant re-reading the same overlapping `AssetManager` / File Manager code three times and splitting the conversation. One thread keeps the discussion and the diff coherent while we settle the design.

Nothing is lost: every commit from #1869 and #1870 is in this branch's history, and all unit/integration tests plus the three feature E2E specs pass on the merged result.

### Your feedback is carried over

I've gathered the open points from all three PRs into an **"Open questions for reviewers"** section in the description above, and would love your input there:

- @cristinavaldera — cross-iDevice centralization, real-time propagation, backwards-compat import, panel ergonomics, table layout.
- @pabloamayab — table/condensed layout + the card-layout text overflow, real-time update.
- @greynoa — extending search to author / license / title.
- @celestebonillaTF, @mnunezcedec — thanks for the interest and the original request (#1817); this is where it continues.

Apologies for the cross-PR shuffle — please continue the discussion here from now on. 🙏

erseco:
### Code Quality findings — dismissed as false positives

The Code Quality bot raised **10 _"Property access on null or undefined"_ findings**, all in `public/app/workarea/modals/modals/pages/modalFileManager.test.js`. I've reviewed each and **dismissed all 10 as false positives** (with a per-finding note).

**Why they're false positives:**
- `modal` is built in the top-level `beforeEach` (`modal = new ModalFilemanager(mockManager)`), and `modal.assetManager` is assigned right after from the test mock. Vitest runs `beforeEach` before **every** test, so the base is never `undefined` at runtime.
- The **full frontend suite passes** (241 files) — an undefined base would throw `TypeError`.
- The analyzer doesn't model the Vitest `beforeEach` lifecycle and can't resolve `modal.assetManager` because it's read back through a mutable global (`window.eXeLearning.app.project._yjsBridge.assetManager`). It flags only ~10 of dozens of identical accesses, consistent with a heuristic, not a real null-deref.

No test code was changed: adding defensive guards/optional chaining to correct tests purely to satisfy a misfiring analyzer would be a workaround. The named **CodeQL check passes** and there are no code-scanning alerts on this branch.

erseco:
### Follow-up improvements pushed (commit 5772d360b)

Implemented a focused batch of reviewer-requested improvements on this branch:

**Resource Report iDevice**
- **Table / condensed layout** — a third layout alongside *List* and *Cards*; columns honor the existing show-* toggles. (@cristinavaldera, @pabloamayab)
- **Clickable license URLs** — the license now links to its canonical URL (CC 4.0 / CC0 / GNU GPL). Resolution is self-contained in the iDevice export JS (the same approach as image-gallery's `getLinkLicense`, so it works in exported packages); the values mirror the canonical `LICENSE_REGISTRY` in `src/shared/export/constants.ts`. Public Domain / Copyright / custom licenses remain plain text.
- **License icons** — an inline Creative Commons mark is shown next to CC licenses (self-contained SVG, like the existing file-type icons), plus the `cc cc-by-sa…` class for theme styling.
- **Card text overflow fix** — long titles/descriptions now wrap inside the card instead of overflowing. (@pabloamayab)

**Attribution prefill from the centralized File Manager metadata**
- **Image gallery** — picking an image now seeds its *title / author / license* from the asset's centralized metadata, filling only empty fields so values you already typed are never overwritten. The gallery's license vocabulary matches the centralized labels, so it maps across directly.
- **Hidden image** — the *Authorship* field is pre-filled from the picked asset's author when empty (via `getAssetMetadata`, mirroring the 3D viewer).

**Tests / DoD:** new colocated specs for the license link/icon URLs, the table markup + column toggles, and both prefill behaviors (seed + no-overwrite + no-op). `make fix`, `make test-unit` (7162 backend + 13164 frontend), `make test-integration` (709) and the Resource Report E2E all pass. No new dependencies, no DB changes.

Remaining triaged items (real-time propagation, source/author URL metadata fields, auto-extract on upload, full cross-iDevice attribution dedup) are still open for discussion — see the triage in the earlier comment.

erseco:
### Resource Report: license icons + live refresh (commit 05bf7ed60)

Fixed the two issues you spotted, both following the **download-source-file** iDevice.

**License icons (the empty boxes)** — the shared `cc cc-<code>` markup only gets an icon from theme CSS scoped to `#packageLicense` (the footer), so in the page body the icon `<span>` was an empty box. The report now ships its **own** self-contained sprite (`resource-report-licenses.png`, the same asset the download-source-file iDevice uses) plus CSS for `.resource-report-IDevice .exe-prop-license a.cc span`, so the CC icons render in preview and in exported packages.

**Always up to date** — the report no longer shows only the snapshot taken at save time. In the workarea and the browser-side export the list is now resolved **live** from the AssetManager (`getAllAssetsMetadata` / `getReferencedAssetIds` / `getAssetUrl`), and `renderBehaviour` observes the assets `Y.Map` and re-renders the instance (debounced, deduped per `ideviceId`) whenever assets are added/removed or their metadata changes — exactly the pattern download-source-file uses for project metadata. The saved snapshot stays as the fallback for static/server (CLI/API) exports where no AssetManager exists.

_Minor limitation:_ the "used in this project" filter depends on content references in the structure map (not the assets map), so it refreshes on asset changes but not on unrelated text edits elsewhere; reopening/saving reconciles it.

**Tests:** export unit tests for live resolution, snapshot fallback, `data-idevice-id`, and the observer (re-render + dedupe), plus a new E2E that uploads a second asset and asserts the report updates **without** re-editing the iDevice. `make fix`, `make test-unit` (7162 backend + 13171 frontend), `make test-integration` (709) and the Resource Report E2E all pass.

cristinavaldera:
Thanks for the proposal, Ernesto.

We're going to test a few changes and see how they work. The general approach would be similar to something we've already discussed before:

FILE MANAGER

On the one hand, the image metadata that should not be modified on each insertion should be centralized in the file manager: author, author link, image name, image URL, and license. These fields would only be editable from the file manager. In addition, any changes made to the metadata in the file manager should be propagated to every instance where the image has been inserted in the resource, just as currently happens with the fields in the “Download source file” iDevice.

This information would automatically appear in the image caption whenever the image is used throughout the resource, whether it is inserted from the TinyMCE editor or from the games.

INSERT IMAGE FROM TINYMCE

In the image plugin, the “Title and Attribution” tab would be removed, since that information would be retrieved directly from the file manager.

In the General tab, the source and dimensions would be displayed, and a new Caption section would be added with the following fields:

- Heading.
- Notes/Comments or other name, in case additional information needs to be added to the caption, since the caption itself is not editable.
- A checkbox to hide the caption when necessary, for example, in the case of an icon.

There would also be an Accessibility section to include the alternative text and title. The alternative text should not be stored in the file manager because it must be unique: it describes the context of the image at a specific point in the resource, and if the image is used multiple times, the alternative text may be different in each case.

I've attached a proposal.

plugin_imagen


OTHER IDEVICES

As for the other iDevices, the fields that are already centralized in the file manager should be removed, while keeping the alternative text field and perhaps also the title and notes fields. However, this would be a broader change that we can evaluate later.

I suggest we start with the changes to the file manager and the TinyMCE image plugin.

erseco:
@cristinavaldera thanks for the mock-up — very helpful. I've implemented your proposal on this branch:

**File Manager — single source of truth**
- The metadata that should not be re-entered on each insertion is centralized and editable **only** here: image name, author, author link, image URL and license (I added the two missing fields, author link + image URL).
- Editing any of these now **propagates to every place the image is inserted**, live, the same way the *Download source file* iDevice propagates project metadata.
- Alt text is no longer here — it is per-instance (see below).

**Insert/Edit image dialog**
- Removed the *Title and attribution* tab; the dialog now has only **General + Advanced**, matching your mock-up.
- General tab: Source, Dimensions, an **"Image header and caption"** section (Heading, Notes, *Hide image caption*) and an **"Accessibility"** section (Alternative text, Title).
- The caption (name / author / links / license) is **auto-derived** from the File Manager metadata and is not editable in the dialog. Heading/Notes and the accessibility alt/title are per-instance.

**Captions everywhere**
- The caption renders live while authoring and is **baked into exported packages** (HTML5/SCORM/EPUB/ELPX) from the current metadata, so exports always reflect the File Manager.

**Two points for your input**
1. The General tab is now fairly **condensed**. We could move the **Accessibility** fields (alt text + title) to the **Advanced** tab to lighten it — happy to do that if you prefer.
2. This same rework should be **replicated in the Insert/Edit media dialog** (video/audio), which still carries the old attribution fields. I'll mirror it there next.

I'll add screenshots of the new dialog shortly.

mnunezcedec:
Hi @erseco , great work on this PR!. I wanted to share some of the tests I've been running and the results I've found:

- If an image is added and all fields are completed through the File Manager, everything is displayed and works as expected (Caption, Notes, Alternative Text, Title, and the File Manager fields).

- However, after doing this, if we edit the image, modify the Caption and Notes fields, and save the changes, the updated values are not displayed. If we edit the image again, the changes are still there, even though they are not shown in the fields where they were originally entered.

- Additionally, the **"Hide image caption"** checkbox does not remain checked after saving, and therefore does not perform the function it is intended for.

Thanks!

erseco:
Thanks a lot @mnunezcedec for the thorough testing — really helpful! Both points are now addressed:

- **"Hide image caption" not persisting / not hiding** — fixed. The figure was being unwrapped when the caption had no visible content (which is the case when it's hidden), so the hidden state was lost. Now an image with caption metadata always keeps its `<figure>`, the checkbox round-trips, and the caption is actually hidden.

- **Editing Caption / Notes not showing the updated values** — this was happening on the earlier commit; it was resolved by the dialog rework in a later commit. Editing the heading/notes of an existing image now updates the rendered caption immediately. I also added an end-to-end test covering both this and the hide round-trip so they don't regress.

Both changes are on the branch now. Could you please pull the latest and give it another try? 🙏

mnunezcedec:
Hi @erseco, thanks for your reply. Here are a few more improvements

When adding a header and a footer, everything looks correct in the preview. However, when editing again, the content can be modified directly in TinyMCE (it allows it), but after saving, the original content (the source of truth) is displayed. If you edit it again, the modified text is still present in the editor, even though it is neither displayed nor reflected in the originally entered content.

It seems somewhat confusing that these fields can be edited in TinyMCE when those changes are not actually rendered. Perhaps these two fields should allow modifications to the original source content as well?

erseco:
Thanks again @mnunezcedec — good catch, and you're right that it was confusing.

You spotted the core issue: the caption (the **header** and the **footer/figcaption**) is *auto-derived* — the name/author/license come from the centralized File Manager metadata, and the heading/notes come from the image dialog. So TinyMCE letting you type into it was misleading: those inline edits were never the source of truth and got discarded on save/re-render.

I went with the option you hinted at — **making the caption non-editable inline** rather than letting TinyMCE edits write back. Writing back would fight the single-source-of-truth design (the centralized name/author/license can't be back-propagated per instance, and the heading/notes already have a home in the dialog). So now:

- The header `div.figcaption.header` and the `figcaption.figcaption` are marked `contenteditable="false"` inside the editor, so you can't type into them by mistake.
- The attribute is stripped on serialize, so the stored/exported markup stays byte-for-byte identical to the shared caption builder (no `contenteditable` leaks into exports).
- Editing still happens where it should: **heading/notes** in the image dialog, and **name/author/links/license** in the File Manager — both propagate live to every inserted instance.

Coverage added: unit tests for the caption-lock helpers (parser/serializer filters + the DOM lock applied right after a figure is built) and an e2e assertion that both caption nodes are non-editable. Full `image-caption-centralized` e2e spec is green.

Pushed to the branch — please pull and give it another try. 🙏

cristinavaldera (changes requested):
I like how it works and I have found no errors. 

Maybe @ignaciogros can take a look  and improve the visual layout in both the File manager and the Insert/edit image window.

erseco:

Reviewing the current draft head.

- **Important:** `public/app/yjs/AssetManager.js:818,842` — the reference scanners walk every component's `htmlContent` + `jsonProperties`, and a Resource Report iDevice snapshots an `asset://` URL for every listed asset (default `resourceMode:'all'`; see `resource-report/edition/resource-report.js` `save()` → `resources[].assetUrl` and the rendered `<a href="asset://…">`). So once one default Resource Report exists, every asset is counted as referenced: "sort by references" ranks all assets ≥1, "Not used in this project" never fires, and the report's own "Used in this project" filter becomes self-referential. Exclude `resource-report` components from the traversal (no test exercises a resource-report in the scan).
- **Minor:** `src/db/queries/assets.ts:159` — the `?search=` LIKE pattern escapes `%`/`_`/`\` but the query emits no `ESCAPE '\'` clause; on SQLite (the default driver) backslash is not the default escape, so the escaping is a no-op that also injects a literal `\`. Searching a term containing `_` (common in filenames, e.g. `report_2024`) matches nothing on SQLite. Add an explicit `ESCAPE` clause and a test with `_`/`%`.

Verified: static inspection at f11e6ba204; CI: all checks green on f11e6ba ✅ (no local build/test run).
Residual risk: (1) the exeimage "Title and Attribution" tab was removed — a plain re-edit/resize is safe (an existing figcaption is preserved at `plugin.min.js:2352`), but a legacy image with a caption header, and external non-`asset://` images, can lose or can no longer receive per-instance attribution; this overlaps the PR's own open question #3 and needs a decision. (2) `figure-caption.ts` builds `sourceUrl`/`authorUrl` hrefs with no scheme allowlist (`javascript:`/`data:` pass through) — defense-in-depth only under the existing author-trusted export model. (3) Two confirmed minors not bulleted: resource-report table headers use `this.t(<var>)`/object-literal labels that key extraction won't pick up (always English), and the File Manager metadata "Image name"/"Image URL" labels show for non-image assets.

---
**Update — resolved (`b0d273e8`):** `resource-report` components are skipped in the reference traversal, so their generated links no longer inflate usage counts; test added.

---
**Update — 3 minors resolved (`93c074ba`):** `ESCAPE '\'` on the asset LIKE search (+test); resource-report headers use extractable `this.t('…')`; File Manager `Image name`/`Image URL` → neutral `Name`/`Source URL`.

erseco:
## Metadata centralization follow-ups: foundation fixes, exemedia centralized captions, game author prefill

This batch lands three coordinated workstreams on top of the File Manager redesign (`2656d8b7`), plus a reconciliation merge (`97cecafb3`) integrating the parallel A1/A2/A4 fixes pushed to this branch last night.

### Workstream A — foundation fixes
- **Reference scanner** (`0eaccd0a`): Resource Report snapshots no longer count as asset references (one default report used to make *every* asset rank as "used", breaking sort-by-references, "Not used in this project" and the report's own used-filter). After the merge, the exclusion lives once in `_forEachComponent`.
- **SQLite search** (`7db7d1b3`): `?search=` now emits `LIKE ? ESCAPE '!'` — previously the `%`/`_` escaping was a no-op on SQLite (searching `report_2024` matched nothing). `!` chosen because `\` is escape-prone on MariaDB and invalid as `ESCAPE '\\'` on PostgreSQL. 3 of the new tests fail against the old code.
- **Caption URL hardening** (`1b5d89ef`): the shared figure-caption builder twins now allowlist `http:`/`https:`/relative URLs for `sourceUrl`/`authorUrl` (`javascript:`/`data:` render as plain text, incl. control-char smuggling variants). exemedia (Workstream B) consumes the hardened builder.
- **Root-cause fix** (`075ce214`, scope addition): the "Unsaved changes detected" alert fired spuriously from an async page-select continuation (`menuStructureBehaviour`), breaking save-after-add flows deterministically. Split silent `hasOpenIdevice()` from alerting `checkOpenIdevice()`.
- **resource-report i18n** (`e3b20ed8`): export labels moved to extractable literal keys.

### Workstream B — exemedia centralized captions
- For `asset://` media the "Title and Attribution" tab is gone; title/source/author/license come from the File Manager's centralized metadata, and the General tab gains per-instance "Media header and caption" (heading, notes, hide) mirroring exeimage (`90b1701b`). Figures persist as `figure.exe-media[data-asset-id][data-caption-*]`, wrapped only when a caption would render (#1664/#1668 guard), captions locked `contenteditable=false` in-editor and stripped on serialize.
- Export baking extended to media figures (video/audio/iframe/embed) so HTML5/SCORM/EPUB reflect current metadata (`0db59493`); File Manager edits propagate live via the existing assets observer.
- Two latent dialog bugs fixed en route: a positional checkbox hack and a deferred `SetDefaulData()` that wiped quickly-reopened dialogs.
- New E2E: `media-caption-centralized.spec.ts` (`f304f61d`).

### Workstream C — game author prefill
- Picking a file in a game edition now seeds the adjacent **Authorship** field from the asset's centralized author — only when the field is empty, resolved at selection time, never touching alt (`0cf6c89c`).
- Opt-in via `data-author-target` on the picker; wired across **31 pickers in 24 games** (`deb96be7`). Skipped: `download-source-file` (project-level author), games whose Authorship is not per-media, audio pickers, activity-level author fields. No game exposes a license select matching the shared vocabulary, so no `data-license-target` was wired.
- New E2E: `game-author-prefill.spec.ts` (`f096e31c`).

### Test evidence (local, macOS)
- `make fix` clean · backend **7562/7562** (90% coverage gate ✓) · integration **721/721** · frontend Vitest **13922/13922**.
- Targeted chromium E2E (11 specs): **81 passed / 2 pre-annotated skips / 0 failed**; the 3 merge-affected specs re-run green (5/5).
- Full `make test-e2e`: 467 passed / 33 failed / 6 pre-annotated skips — **all 33 failures are parallel-load flakes on the local machine (28/33 are two-browser collaborative specs) and pass 33/33 when re-run serially**, including all 7 `collaborative/file-manager` metadata-sync tests. CI is the authoritative parallel signal.
- Full `make test-e2e-static`: **399 passed / 116 skipped / 0 failed** — skips are the expected `skipInStaticMode` set; the new caption/prefill specs all ran and passed against the static build.

### exeaudio finding (read-only, requested)
`exeaudio` **is reachable**: listed in `plugins` and placed on `buttons3` in `public/app/editor/tinymce_5_settings.js`, shown via the `toggletoolbars` button. It therefore remains a live consumer of the legacy MediaElement player for #1763 planning.

### Chosen defaults pending team confirmation
- **External media attribution stays per-instance**, shown only for external (YouTube/Vimeo/URL) sources; auto-importing external media into the File Manager is out of scope.
- **Legacy per-instance attribution import** into the centralized store: deferred. Note: saving a legacy `asset://` media figure through the new dialog adopts the centralized caption (save-triggered boundary).
- **Game author/license field removal**: deferred — this batch only adds prefill.
- **Alt text is never prefilled** from asset metadata (per-instance by design).
- **Scope additions made to un-block acceptance** (flagged for review): the unsaved-changes-alert root-cause fix in `projectManager`/`menuStructureBehaviour` (A), and a guarded seeding call in `ideviceNode.js`'s file-picker `onSelect` (C) — the delegated `common_edition.js` handler specified originally is dormant in the live workarea (game pickers bind directly in `ideviceNode`), which was E2E-proven before extending it.
- **Interpretation note (B)**: "per-instance Accessibility fields" for exemedia = the existing Subtitles tab; exemedia has no alt/title fields and none were invented.
- **Reconciliation merge decisions** (`97cecafb3`): kept single-point scanner exclusion in `_forEachComponent`, kept `ESCAPE '!'`, unified escaping tests (assets.spec.ts: 80), kept `this.trans('…')` for extractability, adopted the remote's File Manager label renames ("Name", "Source URL").

erseco:
### Follow-up decision recorded: ADR-0043 (Proposed)

The "game author/license field removal" item listed above as a *chosen default pending team confirmation* now has a dedicated ADR on this branch for review: [`doc/architecture/adr/ADR-0043-centralized-attribution-ownership.md`](../blob/feature/file-manager-metadata-management/doc/architecture/adr/ADR-0043-centralized-attribution-ownership.md) (`402bce404`).

**Proposal in one line:** make the File Manager asset the single owner of attribution (author, author link, license, title, source URL) for `asset://` media, drop the editable per-media author/license fields from game iDevices and the image gallery, and replace them with a per-activity **"Show attribution"** toggle — attribution resolved at render/export time (the `bakeFigureCaptions` pattern), consistent with what `exeimage`/`exemedia` already do. External (non-`asset://`) sources keep per-instance fields, shown only when the source is external.

**Why an ADR here, not an issue yet:** so reviewers of this PR can accept/reject the direction before we commit to it. The prefill shipped in this batch (`0cf6c89c`, `deb96be7`) only fills empty fields, which leaves a permanent dual source of truth — this ADR is the path to closing that gap.

**Scope:** the ADR is `Proposed` and documentation-only. It explicitly calls for opening a dedicated implementation issue + an SDD (two phases: image gallery pilot, then the game family via a shared attribution-line builder) and implementing on a **follow-up branch** — nothing in this PR changes stored formats or export renderers. Please review the ADR's *Decision*, *Risks* (legacy attribution replaced on next save) and *Follow-up work* sections.

cristinavaldera (dismissed):
Thanks! Better organised!

I like the window is wider, and the tabs to structure all the information.

Change proposals:

1) Maybe the view of the tabs should be more similar to other tabsin the app:

Image

2) In the Details, could the field name be in line with the data? I mean:

"TIPO: image/png"

instead of

"TIPO:
image/png"

3) Translations: In Spanish, change "Autor" for "Autoría"

Thank you!

erseco:
Thanks! Addressed in `65cb9ca6`:

1. ✅ **Tabs** — now use the app's standard underline tab style (teal active label + underline), matching the project-properties tabs from your screenshot, instead of the pill/segmented look.
2. ✅ **Details** — fields render inline now: `TYPE: image/png` on a single line instead of the label stacked above the value. (The editable *Metadata* form keeps its stacked label/input layout, which reads better for inputs.)
3. **"Autor" → "Autoría"** — this is a translation-target change. The `translations/**` XLF files are owned by the dedicated translation process and can't land in a code PR, so this one needs to go through that flow rather than this branch. Flagging it here so it's not lost.

Re-requesting your review 🙏

cristinavaldera (changes requested):
More "cositas":

- maybe make bigger teh window, so that you can see all teh fields in teh first tab without scrolling.
- change "Notas" for "additional information" , it conveys the meaning of the field better.

Works perfectly!

erseco:
Traje `main` a esta rama y resolví el conflicto en `src/shared/export/renderers/IdeviceRenderer.ts`: conservé el rename de iconos de tema de main (`resolveRenamedThemeIcon`) y el bake de captions de este PR (`bakeFigureCaptions` / `assetCaptionMetadataMap`).

cristinavaldera:
Hi!

I really like how this is working now. Some comments:

- I would change "Notas" for "additional information" or similar. "Notes" feels more like internal information, but it is a field that will appear in the caption, for example, to add info about pictures generated with AI.

- I have some concerns about what information should be in the fle manager as unique, like licence, and which one should be filled in every time the resource is used, like ALT. I have concers with "title", "description" and "Notes/additional info". Here we need to make decisions.

- Once that is defined, we should decide whether the metadata fields that appear in activitiy idevices should be skipped or modified.

-  Besides, we still need to figure out how to manage old metadata information. Also a decision to make.

I think the functionality works good, but we would need to make all these decisions to be able to move forward.

PR #1871, opened 2026-05-30
Focused full-workarea iDevice edit mode

### LINKED ISSUE #1411: UX Issue: Confusing save behavior when editing an iDevice
While editing an iDevice, it is very frustrating to click the **Save** button and receive the message indicating that the iDevice is still being edited.

This situation happens because there are two different Save buttons:

* **Save iDevice**
* **Save Project**

This creates confusion, especially when:

* The iDevice Save button is pushed down due to large content and is no longer visible.
* The user clicks the global “Save Project” button expecting the current iDevice to be saved.
* The interface does not clearly communicate that the user is still in *iDevice edit mode*.

The result is a disruptive workflow and potential loss of time, particularly for new users or when working with long iDevice content.

### Root Cause

* Two separate Save actions coexist in the same UI context.
* The iDevice Save button may not remain visible when the editor content is long.
* The global Save button does not adapt to the current editing context.
* There is no strong visual indication that the user is in iDevice edit mode.

### Proposed Alternatives (Short-Term UX Improvements)

Until automatic iDevice saving can be evaluated or implemented, the following alternatives could be explored:

#### Option 1: Floating iDevice Save Button

Make the **Save iDevice** button sticky/floating so it always remains visible while editing.

**Benefit:**
Reduces accidental clicks on the wrong Save button and keeps the correct action accessible at all times.

#### Option 2: Blur the Background Interface During iDevice Editing

When entering iDevice edit mode:

* Blur or visually disable the rest of the interface.
* Clearly signal that the user is focused on editing a specific iDevice.

**Benefit:**
Improves focus and reduces the likelihood of interacting with unrelated controls.

#### Option 3: Context-Aware Global Save Button

If the user clicks **Save Project** while editing an iDevice:

* The system should either:

  * Automatically save the current iDevice first, or
  * Prompt the user: “You are editing an iDevice. Do you want to save it before saving the project?”

In other words, the global Save button should respect the current editing context.

**Benefit:**
Prevents confusion and aligns the Save behavior with user expectations.

### Why This Matters

This issue negatively impacts usability and generates frustration during content creation. The current behavior feels unintuitive and can interrupt the authoring flow.

Even a small UX adjustment could significantly improve the editing experience.

### Feedback Requested

@erseco @cristinavaldera @ignaciogros
Do any of these options seem reasonable as an interim solution until iDevice autosave can be properly evaluated?

### LINKED ISSUE #1811: Keep iDevice editing toolbar visible and auto-save open iDevice when saving resource
## Proposal

The iDevice editing toolbar should remain visible while editing an iDevice.

Currently, the toolbar with the actions **Save**, **Delete**, and **Discard changes** is only visible in its current position. When editing long iDevices, users need to scroll back to the toolbar to save or discard changes, which slows down the workflow.

## Suggested improvement

Keep the iDevice editing toolbar always visible during editing, for example as a sticky toolbar.

This would make the main editing actions immediately available at any point:

- Save
- Delete
- Discard changes

## Additional related improvement

When saving the whole resource while an iDevice is still open for editing, eXeLearning currently does not allow the resource to be saved until the iDevice itself has been saved first.

This adds an unnecessary extra step.

If the user chooses to save the resource while an iDevice is open, eXeLearning should automatically save the open iDevice first and then save the resource, instead of requiring the user to perform both actions manually.

## Expected behavior

- The iDevice editing toolbar remains visible while editing.
- The user can save the whole resource even if an iDevice is still open.
- If an iDevice has unsaved changes, saving the resource automatically saves that iDevice first.
- The user should not be forced to save the iDevice and then save the resource as two separate steps.

erseco:
## Summary

This Draft PR experiments with a focused full-workarea iDevice edit mode requested by @cristinavaldera.

When an iDevice enters edit mode it becomes a focused editor that fills the available workarea, keeps the iDevice action toolbar visible, and prevents the page/workarea behind it from scrolling. The goal is to reduce confusion between **Save iDevice** and **Save Project**, make the editing state more explicit, and improve the workflow for long iDevices.

This is the **final, always-on** editing behaviour — it is not configurable (no opt-in/opt-out).

## Why

Related to:

- #1811 — Keep iDevice editing toolbar visible and auto-save open iDevice when saving resource
- #1411 — UX Issue: Confusing save behavior when editing an iDevice

Both describe the same problem: users can lose access to the iDevice toolbar while editing long content, and the coexistence of global save and iDevice save actions is confusing.

## What it does

- The edited iDevice fills the content workarea via an opaque overlay (`position:absolute; inset:0` on the wrapping `.box`); the block header and other blocks are hidden behind it, so the surface shows **only the iDevice**.
- **Save / Delete / Discard** iDevice actions stay pinned at the top of the focused surface.
- Only the editor body scrolls; the outer workarea scroll is locked.
- The bottom quick-toolbar is hidden.
- Global **Save / Download / Preview / Styles / Settings / Share** are visually + accessibly disabled (`aria-disabled` + tooltip *"Save or discard the open iDevice before saving the project."*). The existing `checkOpenIdevice()` guard remains the authoritative block.
- The dev/beta warning banner is narrowed in focus mode so it no longer covers the iDevice toolbar.
- Enter/exit are announced via a polite `aria-live` region; outer scroll position is saved/restored. The editor manages its own focus (we don't steal it on enter, to avoid disrupting TinyMCE init).

## How it works (isolation)

- A single `FocusedEditMode` module (`public/app/workarea/interface/focusedEditMode.js`) listens via **one `MutationObserver`** on `#node-content`'s `mode` attribute. The engine already sets `mode="edition"` there on every enter/exit, so the feature hooks **save, discard, delete and edit** with **zero changes** to `idevicesEngine.js`, `ideviceNode.js`, the TinyMCE config, or any button module.
- All layout lives in `assets/styles/layout/_idevice-focus.scss`, scoped to `body.exe-idevice-focus-editing`.

## Test adaptations

Because focus mode shows only the iDevice (the block header is hidden during edition), block-level actions are performed outside edition mode. `block-icons.spec.ts` now saves the iDevice (leaving edition) before opening the block icon modal. Sibling blocks are kept in the DOM (covered by the overlay, not `display:none`) so existing "first iDevice" queries still resolve.

## What it does NOT do

- Does not implement full iDevice autosave (Option C investigated — follow-up).
- Does not redesign every iDevice editor.
- Does not use the browser Fullscreen API as the main solution (TinyMCE's own fullscreen still works, with a defensive CSS guard).
- Does not change export/rendered content.
- Does not remove the existing iDevice save/discard/delete lifecycle.

## Tests

- **Unit (Vitest):** `public/app/workarea/interface/focusedEditMode.test.js` — enter/exit state, `aria-disabled` (never the `disabled` attribute), live region, scroll save-restore, no focus-steal on enter, idempotency, single-focus invariant, observer teardown. `interfaceManager.test.js` updated.
- **E2E (Playwright):** `test/e2e/playwright/specs/idevices/focused-edit-mode.spec.ts` — entering applies focused state + disables global save, editor fills the workarea, save/discard/delete each exit focused mode, and the outer scroll is locked while editing.

## Manual QA checklist

- [ ] Long Text iDevice
- [ ] Text iDevice with images
- [ ] TinyMCE dialogs (CodeMagic, media, mermaid)
- [ ] Advanced fields / fieldsets
- [ ] Save / Discard / Delete
- [ ] Global Save while editing (disabled + tooltip; alert still blocks)
- [ ] Download/Export while editing
- [ ] Scroll inside editor body; page itself does not scroll; scroll restored after exit
- [ ] TinyMCE fullscreen button
- [ ] Narrow viewport
- [ ] Keyboard navigation + screen reader announcement
- [ ] Chrome, Firefox, Safari

## Related issues

Refs #1811.
Refs #1411.


greynoa:
This looks like a major improvement for the editing experience, @erseco. The confusion between saving an iDevice versus saving the entire project has been a long-standing UX hurdle, especially when dealing with extensive content blocks.
Thanks for the great work on this!

mnunezcedec (commented):
A very good improvement, without a doubt. Would it be possible to keep both the page title and the iDevice title visible, as in the current version? Is that compatible with this enhancement? Thank you very much

erseco:
> A very good improvement, without a doubt. Would it be possible to keep both the page title and the iDevice title visible, as in the current version? Is that compatible with this enhancement? Thank you very much

I tried to add it, but it used too much space and the flow was not very clear.

Since this is edit mode, I think we should first decide whether this should be shown only as informational content or also be editable there. Even with this “focus mode”, I’m not sure it fits well without making the interface heavier.

mnunezcedec:
> > A very good improvement, without a doubt. Would it be possible to keep both the page title and the iDevice title visible, as in the current version? Is that compatible with this enhancement? Thank you very much
> 
> I tried to add it, but it used too much space and the flow was not very clear.
> 
> Since this is edit mode, I think we should first decide whether this should be shown only as informational content or also be editable there. Even with this “focus mode”, I’m not sure it fits well without making the interface heavier.

At least the iDevice title, as the page appears in the navigation sidebar.

pabloamayab:
Great contribution, @erseco. 
I've found a bug that you can reproduce using the attached ELPx file.
1. I edit the fourth iDevice and switch TinyMCE to full-screen mode. I save at that point.
2. I edit the third iDevice, and a positioning issue occurs that prevents the edit from being displayed. If I try to perform another action, eXe indicates that an iDevice is currently being edited.
[biologia-y-geologia.zip](https://github.com/user-attachments/files/28541706/biologia-y-geologia.zip)

pabloamayab (changes requested):
Hi @erseco 
IMHO, it could move from an experiment to new feature. 
I've tested it again, and the only bugs I've found are the ones I mentioned to you at https://github.com/exelearning/exelearning/pull/1871#issuecomment-4610243849

erseco:
@pabloamayab thanks for the precise repro — fixed in b0930f373.

**Root cause.** The focused editor is an overlay (`position:absolute; inset:0`) inside `#node-content-container`, so it only lines up with the viewport when that container is not scrolled. Two things left it scrolled:

1. Saving an iDevice while TinyMCE was in **fullscreen** destroyed the editor *before* the fullscreen plugin's own teardown ran, so a `tox-fullscreen` class stayed stuck on `<html>`/`<body>`. That scroll-locked the document and, via the defensive CSS guard, suspended the focus layout for the *next* iDevice — exactly your step 2.
2. More generally, scrolling down and editing a lower iDevice on a multi-iDevice page pushed the overlay off the top of the page (measured `boxTop: -1826`), so the editor looked blank and *"an iDevice is already being edited"* blocked everything.

**Fix** (entirely inside `FocusedEditMode`, no engine changes):

- On entering edit mode, pin `#node-content-container.scrollTop = 0` (the previous scroll is still restored on exit) so the overlay always aligns with the viewport.
- On enter/exit, strip a stale `tox-fullscreen` class from `<html>`/`<body>` — but only when no editor is genuinely in fullscreen (`.tox-tinymce.tox-fullscreen` present), so real fullscreen is never disturbed.

**Verified** against your `biologia-y-geologia` ELP on a local build: edit the 4th iDevice → TinyMCE fullscreen → save → edit the 3rd → the editor now renders correctly, no positioning glitch, no blocking alert. Added unit coverage (scroll pin, stale-fullscreen cleanup on enter & exit, and a guard that never strips a live fullscreen) plus a deterministic E2E regression test.

@mnunezcedec re: keeping the title visible — in this focus mode the box icon + title belong to the *box* header and are intentionally hidden while editing, so I'm leaving that as-is for now.

fix-2-pabloamayab-repro-3rd-visible

pabloamayab (approved):
LGTM!!! Thanks @erseco !!!

cristinavaldera (approved):

PR #1874, opened 2026-06-01
feat(crossword): rebuild auto-layout to prevent fake merged words and improve mobile readability

### LINKED ISSUE #1851: Crossword auto-layout places horizontal words flush against vertical words, producing fake merged words
## Summary

The crossword iDevice auto-layout solver can place a word immediately next to a perpendicular word without an actual intersection or separator cell.

This happens in:

```text
public/files/perm/idevices/base/crossword/export/crossword.js
```

Specifically, `generateCrossword` / `canPlaceWord` may accept a horizontal placement where the cell immediately before the first letter or immediately after the last letter is already occupied by a vertical word. The same issue can occur symmetrically for vertical placements.

Visually, this produces fake merged words even though the validator still treats the intended word cells correctly.

Observed examples with the word set:

```text
ARIZONA
NEWYORK
ATHENS
PARIS
AGRA
```

Examples:

* `AGRA` is placed horizontally with its final `A` immediately before the `T` of vertical `ATHENS`, so the rendered row reads `AGRAT`.
* `ARIZONA` is placed horizontally with its first `A` immediately after the `O` of vertical `NEWYORK`, so the rendered row reads `OARIZONA`.

There is no actual letter intersection in these cases. The words are simply adjacent in the same row or column.

I can attach a screenshot from a freshly generated activity if useful.

## Steps to reproduce

1. Open the crossword editor.

2. Add these five words, using arbitrary clues:

   ```text
   ARIZONA
   NEWYORK
   ATHENS
   PARIS
   AGRA
   ```

3. Save and preview the activity.

4. Observe the rendered grid.

Depending on the random shuffle, one or more horizontal words may be placed flush against the side of a vertical word, producing visual artefacts such as:

```text
AGRAT
OARIZONA
```

The issue also reproduces deterministically when the crossword JSON is generated externally and the words are fed in this order:

```text
ARIZONA
NEWYORK
ATHENS
PARIS
AGRA
```

This is the kind of length-descending order a third-party converter would typically use to maximise intersections.

## Expected behaviour

The auto-layout solver should reject placements where a candidate word touches another word longitudinally without a valid intersection.

For horizontal placements, the solver should treat these cells as forbidden when already occupied:

* the cell immediately before the first letter;
* the cell immediately after the last letter.

For vertical placements, the same rule should apply to:

* the cell immediately above the first letter;
* the cell immediately below the last letter.

This is consistent with the spacing constraint already enforced between parallel vertical words through `occupiedColumns`, where `col - 1` and `col + 1` are reserved in `placeVerticalWords`.

## Actual behaviour

`canPlaceWord` only checks the cells that the candidate word would write into.

Current relevant logic:

```js
for (let i = 0; i < word.length; i++) {
    const cell = grid[row][col + i];
    if (typeof cell === 'undefined' || cell === null) continue;
    if (cell.letter && cell.letter !== word[i]) return false;
}
```

This prevents overwriting incompatible letters, but it does not check the longitudinal edge cells:

```js
grid[row][col - 1]
grid[row][col + word.length]
```

As a result, a horizontal word can be placed directly beside a vertical word, creating a visually merged row.

The vertical branch has the same kind of missing check for:

```js
grid[row - 1][col]
grid[row + word.length][col]
```

There is an early check involving `grid[row][col].vi`, but that only covers whether the starting cell itself is already part of a vertical word. It does not prevent a neighbouring occupied cell immediately outside the word bounds.

## Suggested fix

Before accepting a horizontal placement in `canPlaceWord`, reject it if either longitudinal edge cell is already occupied:

```js
// Cell immediately before the word.
if (
    col > 0 &&
    grid[row][col - 1] &&
    grid[row][col - 1].letter
) {
    return false;
}

// Cell immediately after the word.
if (
    col + word.length < mOptions.boardSize &&
    grid[row][col + word.length] &&
    grid[row][col + word.length].letter
) {
    return false;
}
```

Add the symmetric checks for vertical placements:

```js
// Cell immediately above the word.
if (
    row > 0 &&
    grid[row - 1][col] &&
    grid[row - 1][col].letter
) {
    return false;
}

// Cell immediately below the word.
if (
    row + word.length < mOptions.boardSize &&
    grid[row + word.length][col] &&
    grid[row + word.length][col].letter
) {
    return false;
}
```

This may reduce the number of placements accepted by the solver. If that makes some layouts harder to generate, the existing retry loop may need to be increased, or the fallback path should log a clear warning so authors know they need to enlarge the board or remove a word.

## Why this matters

The rendered crossword looks incorrect to learners.

For example, if a row visually reads:

```text
OARIZONA
```

learners may reasonably assume that the expected answer is `OARIZONA`, even though the validator expects only `ARIZONA` in the intended cells.

The validation logic may still be correct, but the activity UX is confusing.

This especially affects tools that auto-generate crossword content from external sources, such as:

* H5P.Crossword imports;
* quiz bank imports;
* other third-party converters.

In those workflows, the author may not have an easy way to manually adjust the generated layout.

## Environment

* Project: `exelearning/exelearning`

* Branch: `main`

* Affected file:

```text
public/files/perm/idevices/base/crossword/export/crossword.js
```

* Trigger: small word sets with shared letters that fit inside the default `16×16` board.

* Reproducible example word set:

```text
ARIZONA
NEWYORK
ATHENS
PARIS
AGRA
```

mnarvaezm:
- New generator that grows the crossword from a central seed: each word is  placed perpendicular to an already-placed one, crossing it on a matching  letter, so the grid stays a single connected component.
- Word distribution is optimized to **maximize the number of crossings** and  to **prevent words from sitting flush next to each other** horizontally or  vertically (mandatory empty separator before/after each word and no  sideways contact without a real crossing).
- Several randomized attempts are run and the best result (most words placed,  then most crossings) is kept.
- Added tests covering the new layout rules (crossings, separation, no   overlapping/merged words) and board sizing.

closes #1851 

**How to test it**
1. Create a crossword with 10 words.
2. Save the iDevice and open the preview.
3. Reload the page several times to check that the generator creates different board layouts.
4. On each reload, verify that:
> - Words only cross on matching letters.
> - Words are not placed flush against each other horizontally or vertically without separation.
> - No fake words are formed by adjacent cells touching.
> - All placed words belong to the original word list.
> - Cell and number sizes remain readable.


erseco (approved):
I tested with many words as well with the words from the issue

```
ARIZONA
NEWYORK
ATHENS
PARIS
AGRA
```

and no collision or adjacent words, great job!

greynoa:
Thank you so much for this contribution! This is a fantastic enhancement for the crossword iDevice. Implementing a seed-based growth algorithm with randomized attempts will definitely result in much tighter, cleaner, and more professional-looking puzzles.

The optimization rules you've set up—especially the mandatory empty separators and preventing accidental sideways contact—are spot on for avoiding invalid or "ghost" words. It's also great that you included robust tests for the layout logic and board sizing.
Awesome work on this!

PR #1900, opened 2026-06-04
1898 feat add LOMLOE Galicia json dataset

### LINKED ISSUE #1898: feat: Add LOMLOE Galicia JSON dataset
### Description
This issue is for the submission of the LOMLOE dataset for Galicia in JSON format.

Note: For this dataset, 'descriptorsPerCriterion: true' must be enabled in the DATASETS configuration.

[lomloe-ES-GA.json](https://github.com/user-attachments/files/28593873/lomloe-ES-GA.json)

franmate:
This PR adds the new LOMLOE dataset file for Galicia in JSON format. #1898 

Close #1898


erseco (changes requested):
Thanks for the contribution.

I have reviewed the PR against the linked issue #1898 and I have one question before approving it.

The issue says that, for this Galicia dataset, `descriptorsPerCriterion: true` must be enabled in the `DATASETS` configuration. I see that this was added in one commit, but later removed/restored to the previous configuration.

Could you confirm the expected behaviour for Galicia?

In the attached screenshot, the operational descriptors are shown per criterion, similar to how we handle them in the Canary Islands dataset. Should those descriptors be auto-assigned from the dataset/configuration, or should they be manually selected by the user in the UI?

Also, please review the identifier format. For example:

`ES-GA-EP2-CS-OBX1-CA1.1`

In the Canary Islands dataset we shortened some codes to make them more readable. This is not a blocker if the official Galicia codes should remain as-is, but please confirm whether these identifiers are intended to be kept in the current long format.

Apart from that, the dataset integration and the updated tests look aligned with the purpose of the issue.

Image

Thanks!

franmate:
Hi,
It should be set to "descriptorsPerCriterion: true". However, I ran into an issue: the "descriptores operativos" texts are currently fetched from "lomloe.js", which does not match the official phrasing and prevents the use of the Galician language.

Additionally, Galicia assigns mandatory "descriptores operativos"—which is solved by setting "descriptorsPerCriterion: true"—but it also allows the option to append different ones, a behavior that is currently not supported by the iDevice.


~~The 'criterios de evaluación' and 'saberes básicos' coding has the right format~~

The 'criterios de evaluación' and 'saberes básicos' **key** coding has the right format.

Thanks

franmate:
I have updated the dataset to include the official identifier directly within the description of both "CA" and "saberes básicos". 

Please let me know if you need any further adjustments!

erseco (changes requested):
Thanks for the update. I would not approve this yet — the final approach is still inconsistent with the linked issue and with your own comment in this PR.

Issue #1898 states that Galicia must enable `descriptorsPerCriterion: true`, and you confirmed in the discussion that it should be enabled. The flag was added in `f1f971f` and then removed in `cc979e9`. The final `ES-GA` entry in `DATASETS` only has `available: true`:

​```js
{
    id: 'ES-GA',
    ...
    file: '../data/lomloe-ES-GA.json',
    available: true
}
​```

This is not cosmetic. The dataset carries non-empty `competencias_clave` on the criteria. With the flag off, `toggleCriterio()` puts those descriptors into `descriptorOptions` and starts `competenciasClave` empty (checkbox mode), so the mandatory descriptores operativos are never auto-assigned. That contradicts the intended behaviour for Galicia.

There's also the nuance you raised: Galicia has mandatory descriptors but also allows appending others. The iDevice currently supports only two modes — fixed per-criterion badges (`descriptorsPerCriterion: true`) or empty + manual checkboxes. It does not support "mandatory + optional additional". We need an explicit decision:

1. Behave like Canarias → add `descriptorsPerCriterion: true` to `ES-GA` and a regression test asserting it.
2. Stay in checkbox mode → update the issue/PR rationale and add a test documenting that choice.
3. Mandatory + optional → this needs a new hybrid mode; it should be implemented or explicitly documented as unsupported for now.

Either way, `descriptorsPerCriterion` is currently asserted nowhere in `lomloe.test.js`, so whatever we decide should be locked by a test.

There's also a second blocker you mentioned: even with `descriptorsPerCriterion: true`, the descriptor labels are still pulled from `lomloe.js`, so they would render in Spanish instead of the official Galician phrasing. That needs solving regardless of which mode we pick.

On the identifiers: in af61793 the short id (e.g. `CA1.1.`) was prepended to the `descripcion` text, while `codigo` keeps the long form (`ES-GA-EI4-CRR-OBX1-CA1.1`). That duplicates the identifier (once in `codigo`, once inside the visible description) and doesn't answer the original question, which was whether the long `codigo` format should stay or be shortened as we did for Canarias. The identifier belongs in `codigo`, not embedded in the pedagogical text shown to the teacher. Please decide the canonical `codigo` format and drop the id prefix from `descripcion`.

One more: the Galicia section in the LOMLOE README is now out of sync with the final dataset. It documents nivel keys like `Primeiro ciclo (0-3 anos)`, `1º de educación primaria`, `1º de ESO`, but the JSON ships `4º Educación infantil`, `1º Educación primaria`, `1º ESO`. Please align the README with the final keys.

The dataset and tests look structurally fine; requesting changes until the descriptor behaviour, Galician descriptor labels and README alignment are resolved.

ignaciogros (changes requested):
Hi @franmate,

Could you please take another look as this issue? The possibility of translating the descriptors is already merged into `main`, so probably a new version of the GA translation will be enough.

Thank you.

franmate:
@ignaciogros 
I have updated the Galicia dataset to include the 'descriptors' key in the JSON file.

@erseco 
    I understand the concern about data cleanliness, but in the case of Galicia, the "criterios de evaluación" identifier is not merely a technical key it is an integral part of the criterion as defined in the official curriculum.
    As shown in the attached extract from an official Galician teaching programme document, the identifier always explicitly accompanies the criterion text. Since teachers work directly with this regulatory structure, removing the ID from the description would break alignment between eXeLearning and the official documentation they use on a daily basis.
    I am not aware of whether the same convention applies in other autonomous communities.

criterios-ga

ignaciogros (commented):
Thanks, @franmate.

Before reviewing the PR, I'd like to confirm if this is the final proposed solution, because I was only expecting changes in the language JSON file, but I see other modifications too.

I thought that in this case the descriptions would be more detailed, while the approach would remain the same for all languages.

Please confirm if we should proceed with the review of the current proposal.

franmate:
@ignaciogros
Yes, this is the final proposed solution. Initially, the idea was only to change the JSON file, but during development, changes were requested in other files (README, lomloe.js, and test). The text for the "descriptores operativos" was added according to the instructions, along with the change introduced in "main" to align it with other datasets. Once the handling of "descriptores operativos" in the iDevice "LOMLOE" is improved, the JSON will be modified accordingly. 
Please feel free to proceed with the review.

erseco (changes requested):
Thanks for the update, @franmate. I've gone through the whole thread and the final diff again. The foundation is solid — the JSON is valid, the suite is green (117/117), Codecov holds coverage (91.84%, all modified lines covered), the operational-descriptor labels now ship in Galician via the dataset's `descriptors` catalog, and the mode decision is locked with tests. That said, I'm not approving it yet. Here is everything I found, ordered by importance.

### 1. Operational-descriptor behaviour (product blocker)

The PR sets `descriptorsPerCriterion: true` (Canarias-style: fixed badges, no picker). That settles the "auto-assigned mandatory" part, but it does not match Galicia's actual rule, which is *mandatory descriptors + the option to append others*. You raised this yourself on 5 Jun, and in my earlier review (8 Jun) I listed it as option 3 — "mandatory + optional → needs a new hybrid mode; implement it or explicitly document it as unsupported for now." Right now it's neither implemented nor documented as a known limitation.

There's also a side effect that isn't mentioned anywhere: all 398 Educación infantil criteria ship `competencias_clave: []` (empty). With the flag on, `toggleCriterio()` assigns the verbatim list and renders no picker, so Infantil shows neither badges nor checkboxes — the teacher can't see or assign any descriptor in that stage (Primaria/ESO/Bacharelato all have them populated). We need to confirm whether that's faithful to the Galician curriculum (Infantil with no per-criterion descriptors) and, either way, write it down.

**Request:** document explicitly in the README (and/or the issue) that (a) the "mandatory + optional" mode isn't supported yet and this dataset is an interim step, and (b) Infantil carries no per-criterion descriptors. Alternatively, implement the hybrid mode. I'd rather not merge a behaviour you've described as temporary without leaving that on record.

### 2. Scope alignment between reviewers

@ignaciogros noted (22 and 24 Jun) that he expected changes to the language JSON only ("I was only expecting changes in the language JSON file"), whereas my 8 Jun review explicitly asked to also touch `lomloe.js`, the README and the tests. The PR follows the latter. Let's agree on the intended scope before merging — as it stands, the two of us were expecting different things. (For traceability: the flag was added in `f1f971f`, removed in `cc979e9`, and is active again in the final delivery; that part is resolved.)

### 3. README out of sync with reality

- The heading `### Build strategy (full Galician extraction)` contradicts its own body: the intro now says "There is no extraction from the DOG" with just "Sourced from Proens DB" underneath, yet the title still claims "full Galician extraction", and the `### Base curriculum decrees` table keeps the "DOG decrees" framing. Please rewrite that section to be consistent with the new provenance (Proens DB).
- The "Nivel labels" table still lists `Educación Infantil` and `Educación Secundaria Obrigatoria` in the *Etapa* column, but the JSON keys are `Educación infantil` and `ESO`. Please align it with the real keys.

### 4. Identifier duplicated inside `descripcion` (open since my 8 Jun review)

`codigo` already carries the identifier (`ES-GA-ESO1-BiXe-OBX1-CA1.1`), and it's repeated at the start of `descripcion` ("CA1.1. Analizar…"). I asked to drop it from the pedagogical text earlier; you defended it as integral to the Galician curriculum, but we never closed it out. Let's settle it: either accept it and document it as a deliberate choice, or strip the prefix from `descripcion`.

### 5. Cleanup / Definition of Done

- `git diff --check` flags trailing whitespace on 5 added lines (`README.md:349`; `lomloe.test.js:13, 621, 2058, 2062`). We require a clean `make fix` before merge — please run it.
- The test "datasets without a per-course distribution … are not filtered" was removed rather than updated to Galicia's new codArea values, leaving a blank gap in the block. That drops the explicit assertion for that path; better to rewrite it with the new codes.

### Summary

The dataset and tests are structurally fine and the integration is correct. Requesting changes until we resolve point 1 (decide and document the descriptor behaviour + the Infantil case), align scope with @ignaciogros (point 2), fix the README (point 3), close the identifier question (point 4), and the cleanup in point 5. Thanks again for the contribution and for bearing with the back-and-forth on the descriptor mode.

franmate:
@ignaciogros 
@erseco 
Thank you for the detailed review. We have internally analyzed your observations, and while they are relevant at the data level, we have identified that the underlying issue does not lie within the JSON, but rather in the **iDevice LOMLOE** engine.

After contrasting the feedback with the current state of the architecture, we have found that the engine assumes a homogeneity across educational stages that does not align with the regulations for key competencies and operational descriptors. Specifically, the system lacks distinct modeling per stage, overlooking the fact that certain early stages feature a native decoupling of these curricular elements, which the current model is unable to process.

For this reason, we have decided to place this PR on an ***on-hold*** status. Forcing the integration of the Galicia JSON at this point would introduce technical debt into the persistence layer, whereas the solution must be addressed within the core of `lomloe.js`.

To resolve this, two lines of engine refactoring should be prioritized in the backlog:

* **Implementation of a hybrid architecture for operational descriptor injection:** Moving past the current *hard-coding* of the restrictive 'Canarias-style' model to enable a dynamic and asynchronous behavior that supports the optionality required by regulations such as Galicia's (mandatory + optional descriptors).
* **Abstraction of the regulatory filtering logic and stage-specific treatment:** Ensuring that the engine—acting in a stage-agnostic manner—determines via metadata which elements are applicable according to the legal context of the educational stage, rather than relying on the data layer to do so.

These improvements are outside the scope of a data PR. Therefore, the most efficient decision is to pause this deployment until the iDevice LOMLOE includes the technical enhancements required to process the data in compliance with the applicable legislation

As soon as the engine (`lomloe.js`) pipeline is updated and validated, we will resume the integration of this dataset without any issues.

We remain at your disposal for any further questions or suggestions.

PR #1901, opened 2026-06-04
Implement OIDC discovery from OIDC_ISSUER

### LINKED ISSUE #1715: OIDC: implement discovery via OIDC_ISSUER (documented but not wired in auth.ts)
## Problem

`doc/development/authentication.md` states that the OpenID userinfo endpoint is *discovered automatically* from `OIDC_ISSUER` via the OIDC Discovery document (`/.well-known/openid-configuration`):

> The UserInfo endpoint URL is discovered automatically from `OIDC_ISSUER` using OIDC Discovery.
> — `doc/development/authentication.md:65`

In practice, `src/routes/auth.ts` never reads `OIDC_ISSUER`. It uses the explicit per-endpoint variables directly:

- `OIDC_AUTHORIZATION_ENDPOINT`
- `OIDC_TOKEN_ENDPOINT`
- `OIDC_USERINFO_ENDPOINT`

`OIDC_ISSUER` currently exists only in admin scaffolding (`src/routes/admin.ts`, `src/routes/pages.ts`, `views/admin/index.njk`) for display/storage — it has no runtime effect.

Surfaced during the `.env.dist` audit in #1652: the variable was flagged admin-scaffolding-only but deliberately kept because it is a legitimate OIDC configuration field and removing it would break operator expectations.

## Options

1. **Implement discovery** — fetch `${OIDC_ISSUER}/.well-known/openid-configuration` at startup (or first login), derive the three endpoints, and fall back to the explicit env vars only when discovery is unavailable or the issuer is empty. Matches the docs.
2. **Fix the docs** — drop the discovery claim from `authentication.md`, keep `OIDC_ISSUER` purely as a descriptive/display field, and document that all three endpoints must be configured explicitly.

Option 1 is the behavior most OIDC operators expect and would shrink the amount of config they need to write (issuer + client id + client secret should be enough for any standards-compliant IdP).

## Acceptance criteria

Whichever option is chosen:

- `src/routes/auth.ts` and `doc/development/authentication.md` agree on how `OIDC_ISSUER` is used.
- If discovery is implemented: cache the discovery document per issuer, respect the existing explicit endpoint vars as overrides, and add tests covering (a) discovery success, (b) discovery failure with explicit endpoints, (c) missing issuer, (d) malformed discovery document.
- If the docs are fixed instead: `OIDC_ISSUER` stays in `.env.dist` with an updated comment making clear it is informational only.

Refs #1652

erseco:
## Summary

Implements OpenID Connect Discovery for OIDC authentication configuration.

`OIDC_ISSUER` was documented as the source for OIDC endpoint discovery (`doc/development/authentication.md` claimed the UserInfo endpoint was "discovered automatically from `OIDC_ISSUER`"), but `src/routes/auth.ts` never read it — endpoints came only from the explicit `OIDC_*_ENDPOINT` variables, so `OIDC_ISSUER` had no runtime effect.

Now, when `OIDC_ISSUER` is set, the backend fetches the provider metadata from `${OIDC_ISSUER}/.well-known/openid-configuration` and uses it to fill any endpoint left blank. Explicit endpoint configuration keeps priority, so existing deployments are unaffected.

Closes #1715. Refs #1652.

## Behavior

- Explicit endpoint settings override discovered values (discovery never overrides them).
- Discovery only fills missing endpoints.
- Discovery is skipped when `OIDC_ISSUER` is empty or every endpoint is already explicit.
- The discovered `issuer` must match the configured one (a trailing-slash difference is tolerated).
- HTTPS is required for discovery, except local development issuers (`localhost`, `127.0.0.1`, `[::1]`).
- A short request timeout is used and metadata is cached in memory per issuer.
- On discovery failure the app falls back to the explicit configuration; the route's existing "OpenID Connect is misconfigured" error still fires if a required endpoint is genuinely missing.
- `end_session_endpoint` is optional: used when discovered and not explicitly configured (explicit wins).

## Changes

- **New** `src/services/oidc-discovery.ts` — `normalizeIssuer`, `discoverOidcMetadata` (fetch + validate + cache), `resolveOidcEndpoints` (precedence/merge), `resetOidcDiscoveryCache`.
- `src/routes/auth.ts` — new `getResolvedOidcConfig(db)` helper; `/login/openid`, `/login/openid/callback`, and `/logout` now resolve endpoints through it.
- `doc/development/authentication.md` — new "OIDC endpoint resolution (Discovery)" section.
- `.env.dist` — comments explaining discovery, precedence, and a minimal discovery-only sample.

Covered by `src/services/oidc-discovery.spec.ts` (new) and additions to `src/routes/auth.spec.ts`. Patch coverage: `oidc-discovery.ts` 100% funcs / ~99% lines; all modified `auth.ts` lines covered.

### 1. Automated (one command)

```
make test-unit
```

Expected: green. Or just the new code:

```
BASE_PATH="" DB_PATH=:memory: ELYSIA_FILES_DIR=/tmp/exelearning-test bun test src/services/oidc-discovery.spec.ts src/routes/auth.spec.ts
```

### 2. Manual — discovery works (the whole point)

Put this in `.env` (real Duende demo IdP, **no endpoints set** — they get discovered):

```
APP_AUTH_METHODS=password,openid
OIDC_ISSUER=https://demo.duendesoftware.com
OIDC_CLIENT_ID=interactive.confidential
OIDC_CLIENT_SECRET=secret
OIDC_SCOPE="openid email"
```

Then `make up-local` and:

- [x] On `/login`, click **OpenID** → you land on the Duende login page (authorize endpoint was discovered)
- [x] Log in (`bob` / `bob`) → you come back signed in (token + userinfo endpoints were discovered)
- [x] Click logout → you are redirected through the provider logout and back (end_session endpoint was discovered)

> Before this PR the same config fails with "OpenID Connect is misconfigured", because `OIDC_ISSUER` was ignored.

### 3. Manual — explicit config still wins (no regression)

- [x] Add `OIDC_AUTHORIZATION_ENDPOINT=https://demo.duendesoftware.com/connect/authorize` on top of the config above → login still works and uses that exact URL (discovery does not override it)
- [x] Set a broken `OIDC_ISSUER=https://nope.invalid` but fill **all** `OIDC_*_ENDPOINT` explicitly → login still works; server log shows a discovery-failure warning, not a hard error
- [x] Empty `OIDC_ISSUER` **and** empty `OIDC_AUTHORIZATION_ENDPOINT` → `/login/openid` returns 500 "misconfigured" (as today)

## Commands run

```
make fix
# Biome clean on changed files; only pre-existing warning in src/routes/export.ts:633 (untouched)

BASE_PATH="" DB_PATH=:memory: ELYSIA_FILES_DIR=/tmp/exelearning-test bun test ./src ./test/helpers
# 7145 pass, 0 fail, 201 files
```

Notes:
- The backend unit suite needs the project's test env (`ELYSIA_FILES_DIR`/`DB_PATH=:memory:`, as `make test-unit` sets); without it ~19 unrelated api/v1 + config tests fail on a clean tree.
- Integration and E2E suites were not run: backend-only change with no frontend/UI flow modification, and there is no E2E harness for a live OIDC provider.


ignaciogros (approved):
I suggest merging this PR into `main` as soon as possible so it can be properly tested in the environments where the need for it was identified.

PR #1903, opened 2026-06-05
fix(ci): harden Windows installer test against silent NSIS install failures

erseco:
## Follow-up to #1894

#1894 tried to stop the flaky Windows `test-install` step by adding a **240s poll**
for the installed `eXeLearning.exe`, on the theory that electron-builder's NSIS
installer extracts the app in a child process that can outlive the launcher
(a slow "extraction race").

That hypothesis turned out to be wrong — or at least incomplete. After #1894
merged, the step **still failed on `main`**, and now it fails *after the full
240s elapse* (job ~4m17s):

```
Installed EXE not found to launch after 240 s (NSIS silent install never produced eXeLearning.exe).
```

A full 240s timeout means the EXE **never appears at all** on those runs — the
silent install is genuinely **failing**, not racing. And the step captured zero
signal about why: the installer's exit code was discarded (`Start-Process -Wait`
without `-PassThru`) and nothing was dumped on failure, so every failure looked
like a blind timeout.

Flake rate on `main`: ~10% (26 success / 3 failure over the last 30 runs).

## Most likely root cause

On `windows-latest`, **Windows Defender** real-time scanning quarantines/locks
the freshly extracted Electron binary mid-install — the classic Electron-on-CI
flake. NSIS silent mode swallows the resulting file-copy error, leaving no EXE.

## Changes

Hardens only the Windows install portion of the `Test install on Windows
(deterministic)` step:

- **Capture the installer exit code** — `Start-Process … -PassThru`, logged.
- **Windows Defender exclusions** for the install dir + temp **before**
  installing (best-effort, `try/catch`) to rule out the AV quarantine.
- **Force the install dir** with NSIS `/D=` for a deterministic location.
- **One automatic retry** of the whole install if the first attempt produces no
  EXE.
- **Kill any app auto-launched** by `runAfterFinish:true` between attempts
  (CI-only; `package.json` keeps `runAfterFinish` so shipped installs still
  auto-open).
- **Rich diagnostics on total failure** (best-effort, never throws): recursive
  listing of `%LOCALAPPDATA%\Programs`, `eXeLearning*` under Program Files,
  any NSIS log, processes matching `eXeLearning|Setup`, `Get-MpThreatDetection`
  and the Defender Operational event log — so the next failure is debuggable
  instead of a blind timeout.

No `package.json` / build-config change. The Windows test stays a **required**
check on `main` (the `fail-fast: false` from #1894 is preserved, so a flake on
one OS still doesn't cancel the others).

## Notes

- `/D=` is passed last and unquoted (NSIS requirement); the install path has no
  spaces. `Add-MpPreference` needs admin + the Defender module, both present on
  `windows-latest` today, and is wrapped in `try/catch` so a restricted future
  image degrades gracefully.
- Worst case before failing is now 2 × `INSTALL_POLL_TIMEOUT`; if the Defender
  exclusion proves to be the fix, `INSTALL_POLL_TIMEOUT` (currently 240s) can be
  lowered in a follow-up since a successful install lands the EXE within seconds.

## Verification

- `workflow_dispatch` run on this branch with `test-install (windows-latest)`
  green; log shows `Installer exit code: …` and `Found installed EXE: …`.
- Other OS matrix legs unaffected.

### DIFF
--- a/.github/workflows/build-electron-installers.yml
+++ b/.github/workflows/build-electron-installers.yml
@@ -280,48 +280,147 @@ jobs:
           $installTimeout = [int]($env:INSTALL_POLL_TIMEOUT ?? 240)
           $installPoll    = [int]($env:INSTALL_POLL_INTERVAL ?? 2)
 
-          # -------- Find installer and install silently --------
+          # -------- Find installer --------
           $installer = Get-ChildItem installers\*Setup*.exe -ErrorAction SilentlyContinue | Select-Object -First 1
           if (-not $installer) { throw "No Windows installer found in 'installers'." }
 
-          # /S = silent for NSIS oneClick; avoids UI + autolaunch races.
-          # NOTE: electron-builder's NSIS installer extracts the app in a child
-          # process that can outlive the launcher it spawns, so -Wait sometimes
-          # returns before eXeLearning.exe exists on disk. Checking for the EXE
-          # only once right after this call is the root cause of the flaky
-          # "Installed EXE not found to launch." failures, so we poll below.
-          Start-Process $installer.FullName -ArgumentList '/S' -Wait -NoNewWindow
+          # Deterministic per-user install dir (NSIS perMachine:false default base).
+          $installDir = Join-Path $env:LOCALAPPDATA 'Programs\eXeLearning'
 
-          # -------- Locate installed EXE (poll to absorb the NSIS extraction race) --------
           $candidatePaths = @(
-            (Join-Path $env:LOCALAPPDATA 'Programs\eXeLearning\eXeLearning.exe'),
+            (Join-Path $installDir 'eXeLearning.exe'),
             (Join-Path ${env:ProgramFiles} 'eXeLearning\eXeLearning.exe'),
             (Join-Path ${env:ProgramFiles(x86)} 'eXeLearning\eXeLearning.exe')
-          )
+          ) | Select-Object -Unique
 
-          $exePath = $null
-          $installDeadline = (Get-Date).AddSeconds($installTimeout)
-          do {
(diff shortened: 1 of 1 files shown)


erseco:
@pabloamayab @ignaciogros — stress-tested this fix before requesting review.

I triggered the `Build Electron Desktop Offline Installers` workflow **20 times** on this branch via `workflow_dispatch` (4 rounds of 5). **All 20 runs passed**, including `test-install (windows-latest)` in every single one — versus the previous ~10% flake rate on `main` (≈3 failures in the last 30 runs before this PR).

**Log evidence from the Windows job** (consistent across runs):
- `=== Install attempt 1 of 2 ===` → installs on the **first** attempt, the automatic retry is never needed.
- `Installer exit code: 0` → exit code is now captured (was discarded before).
- `/D=C:\Users\runneradmin\AppData\Local\Programs\eXeLearning` → install dir forced deterministically.
- `Found installed EXE: …\eXeLearning.exe` → EXE appears within ~13–15s, not the old 240s timeout.

Most likely root cause was Windows Defender quarantining/locking the freshly extracted Electron binary mid-install; the `Add-MpPreference` exclusion before install removes it. And if it ever flakes again, the new diagnostics block dumps the real cause (Defender events / exit code / dir contents) instead of a blind timeout.

Run IDs (all ✅): 26999754799, 26999758361, 26999761635, 26999765085, 26999768721, 27000851294, 27000855143, 27000859008, 27000862824, 27000866803, 27001627220, 27001631132, 27001635279, 27001639366, 27001643382, 27002814079, 27002818127, 27002822499, 27002826818, 27002831107.

ignaciogros (approved):

ignaciogros (approved):

PR #1828, opened 2026-05-19
Add Extremadura LOMLOE curriculum dataset

erseco:
## Summary

This PR adds the Extremadura LOMLOE curriculum dataset (`lomloe-ES-EX.json`) to the LOMLOE iDevice as a new `ES-EX` concretion. It builds on the state-level work from #1827 and follows the same JSON schema used by the existing `ES`, `ES-EFP`, and `ES-CN` datasets.

> **Base branch:** this PR targets the `#1827` branch so its diff is purely additive (the Extremadura work only). Once #1827 lands on `main`, this PR can be rebased onto `main` without changes.

Related to #1827 (state datasets), #1370 (initial LOMLOE iDevice).

## What changed

- Added the Extremadura curriculum dataset at:
  `public/files/perm/idevices/base/lomloe/data/lomloe-ES-EX.json`
- Registered the `ES-EX` dataset in the LOMLOE iDevice DATASETS registry (`available: true`), placed between `ES-EFP` and `ES-CN`.
- Kept compatibility with the existing iDevice schema (`Etapa → Nivel → Materia`) and the per-year duplication convention used by the other ES datasets.
- Documented the DOE sources, the inheritance rule, the modification decrees reviewed, and the build strategy in the iDevice README.
- Added Vitest tests for `lomloe-ES-EX.json` (8 new test cases) and extended the registry regression guard.

## Build strategy (hybrid, per planning agreement)

Extremadura's curriculum decrees adopt the Spanish state RDs *verbatim* for competencias específicas and criterios de evaluación — the LOMLOE framework mandates this and the decrees state it explicitly. Only saberes básicos receive a regional concretion, delivered as multi-column ciclo tables in the DOE PDFs.

The dataset is therefore assembled in two layers:

1. **Competencias específicas + criterios de evaluación**: inherited verbatim from `lomloe-ES.json` with code prefixes re-emitted as `ES-EX-…` so selection IDs stay unique vs the state dataset.
2. **Saberes básicos**: extracted directly from the DOE PDF tables with `pdfplumber` where the regional concretion is available (Decreto 107/2022 uses an explicit `A.1.1.1.` saber-code scheme that maps cleanly). Where DOE table extraction does not yield content for an `(etapa, area)`, the state saberes are used as the fallback per the inheritance rule.

## Data sources

All sources are official DOE / Junta de Extremadura curriculum decrees:

| BOE / DOE ID | Norma | Etapa |
|---|---|---|
[...]

Issue #1832, opened 2026-05-20
Issues and improvements detected in iDevice LOMLOE


mnunezcedec:
While testing the iDevice LOMLOE, I have identified several improvements and a couple of issues that should be addressed:

### Improvements
**1. Filter label text update**

The current label:

"Filtrar materias"

should be changed to:

"Filter areas/subjects"

In Primary Education, the correct term is “areas”, so this change would improve terminology accuracy.

**2. Filtering subjects by course**

Currently, all subjects/areas from all courses or stages are displayed, which may lead to selecting subjects that do not correspond to the selected course.

It would be advisable to filter the list dynamically based on the selected course or educational stage to avoid potential selection errors.

### Issues
**3. Ceuta and Melilla curriculum duplication**

The curriculum for Ceuta and Melilla appears as a separate option, although it is identical to the national curriculum.

This duplication should be removed to avoid confusion and simplify the selection process.

**4. Incorrect mapping of competency descriptors**

A significant issue has been detected in the mapping of key competency descriptors.

Currently, only in the Canary Islands are operational descriptors directly associated with evaluation criteria. In the rest of the autonomous communities, they are linked to specific competencies instead, and the mapping has not been fully extended to evaluation criteria.

As a result, when selecting an evaluation criterion, operational descriptors may be assigned even if they are not actually related to that criterion.

This is a critical issue, as it may affect curricular consistency and the correctness of competency relationships.

_Proposed solution_

Since there is currently no explicit mapping between Assessment Criteria and Operational Descriptors of Key Competencies (except in the Canary Islands), it is proposed to replace the current automatic or predefined assignment system with a manual linking system controlled by the teacher.

Specifically, instead of displaying the descriptors as labels directly associated with the assessment criteria, they would be implemented as a set of checkboxes available during the creation or editing of each criterion.

In this way, the teacher would explicitly determine which operational descriptors are related to each assessment criterion, ensuring greater pedagogical accuracy and avoiding incorrect or unjustified assignments.

mnunezcedec:
Tell your colleagues @pabloamayab  to look into this issue in the iDevice LOMLOE review. Can save work.

erseco:
Hi @mnunezcedec — thanks for the detailed report. Most of the points you raised are now resolved in PR #1827 (commit `2f537ead9`). Status point by point:

### 1. Filter label "Filtrar materias" → "Filter areas/subjects"
[...]

PR #1905, opened 2026-06-05
feat(lomloe): add Extremadura LOMLOE dataset (official siglas + Infantil backfill)

### LINKED ISSUE #1904: feat add LOMLOE Extremadura json dataset
This issue is for the submission of the LOMLOE dataset for Extremadura in JSON format.

pabloamayab:
## Summary

Adds the Extremadura (`ES-EX`) LOMLOE curriculum dataset and aligns it with the iDevice contract already documented in the README.

This revision fixes a regression introduced when the regenerated JSON shipped the generator's **raw, pre-remap output**: Educación Primaria and ESO área codes used the derived siglas (`BIG`, `FQX`, `GEH`, `EPV`, `TYD`…) instead of the official Extremadura ones (`BG`, `FQ`, `GH`, `EPVA`, `TECD`…) documented in the README and used by `ESO_COURSE_SUBJECTS` in `edition/lomloe.js`, and Infantil criterios lost their competencias-clave backfill. As a result the per-course ESO filter hid valid materias in 1º–3º ESO.

The previous revision adapted the **tests** to the broken data; this one fixes the **data** so the JSON, tests, README and runtime describe the same contract.

## Changes

- **Primaria/ESO subject codes → official siglas.** Remapped the 12 derived códigos to the official Extremadura siglas (`BIG→BG`, `FQX→FQ`, `GEH→GH`, `EPV→EPVA`, `TYD→TECD`, `EVC→EVCE`, `LEX→LE`, `EFI→EF`, `EEX→EyE`, `EAR→EA`, `FOP→FOPP`, `CMN→CMNS`) across área keys and every embedded competencia/criterio/saber code. Infantil and Bachillerato keep their derived codes, per the README.
- **Infantil competencias-clave backfill restored** — byte-identical to the `ES` / `ES-MD` datasets, matched by criterio código.
- **Tests realigned to the intended contract** (not the broken output): restored the official-codes and Infantil-backfill assertions, and added a real-dataset ESO render test (1º–3º show the Decreto 110/2022 Anexo V distribution, 4º left unfiltered), a filter↔dataset alignment check, and a stronger no-stale-derived-codes assertion covering embedded criterio/saber codes.
- **Bachillerato 1º/2º duplication** documented as a known, pre-existing, dataset-family-wide limitation (the state dataset and every regional concretion duplicate most Bachillerato subjects into both years), deferred to a follow-up so it can be fixed consistently across all LOMLOE datasets.

Regional Extremadura content (regional criterios + saberes) is preserved; only the área codes and the Infantil links change.

## Verification

- `make fix` clean.
- All `lomloe` frontend unit tests green (158), including the new regression tests; full `make test-frontend` suite green.

Refs #1904.

---

## Bachillerato 1.º/2.º distribution (follow-up to @mnunezcedec's review)

@mnunezcedec reported that Bachillerato mixed 1.º and 2.º subjects. A deep audit confirmed it and that it was **not** specific to Extremadura.

**The issue.** RD 243/2022 (arts. 9–13) assigns each Bachillerato subject to a specific course, but the state-derived datasets duplicated most subjects into **both** years (the `#1827`/`#1828` extraction copied the per-*modalidad* curriculum into each year instead of placing it per course). So single-year subjects — *Física*, *Química*, *Historia de España*, *Historia del Arte* (2.º) or *Física y Química*, *Historia del Mundo Contemporáneo*, *Filosofía* (1.º) — showed up in the wrong year.

**Was it only ES-EX?** No. It affected `ES`, `ES-EX`, `ES-MD`, `ES-EFP`, `ES-NC` and `ES-VC`. Two datasets were already correct: **`ES-CN`** (Canarias) and **`ES-GA`** (Galicia) encode disjoint year sets — they were used as in-repo oracles.

**Normative sources checked.** RD 243/2022 (BOE-A-2022-5521) arts. 9–13; Extremadura **Decreto 109/2022** (DOE) arts. 15–19 — which adopts the state per-course distribution verbatim (no subject reassigned); cross-validated against `ES-CN` and `ES-GA`. The modalidad *General* trap was verified (*Matemáticas Generales* = 1.º, *Ciencias Generales* = 2.º).

**What changed.** A **JSON-level** fix (matching how `ES-CN`/`ES-GA` already encode it — **no runtime `BACH_COURSE_SUBJECTS` filter**): for each affected dataset the wrong-year copy of every single-year subject (identical content, only the `BAC1`/`BAC2` tag differed) was removed; the correct-year copy is kept unchanged. I/II families (*Matemáticas*, *Latín*, *Griego*, *Dibujo Técnico/Artístico*…) keep both years.

| Dataset | Bachillerato status |
|---|---|
| `ES`, `ES-EX`, `ES-MD` | **Fixed** — 29 wrong-year duplicates removed (RD 243/2022 + Decreto 109/2022 / Decreto 64/2022). |
| `ES-EFP` | **Fixed** — same 29 removed (Orden EFP/755/2022). 4 MEFPD-specific optativas left (need the optativa annex). |
| `ES-NC` | **Fixed** — same 29 removed (Navarra adopts the state distribution). |
| `ES-CN`, `ES-GA` | **Already correct** — left unchanged, used as oracles. |
| `ES-VC` | **Unchanged** — Valencian subjects; needs Decret 108/2022 verified first (deferred to avoid a cross-language mismatch). |

**Why it is normatively correct.** Every fixed subject's course matches RD 243/2022 / Decreto 109/2022 **and** the independently-extracted `ES-CN`/`ES-GA` datasets. A final adversarial pass ran 70 counterexamples (14 bellwether subjects × 5 datasets) against the `ES-CN` oracle with **0 mismatches**, and the fix is provably removal-only (29 entries removed per dataset, **0 kept-entry content changed**).

**Tests added.** Per-dataset Bachillerato year-separation tests (ES, ES-EX, ES-MD, ES-EFP, ES-NC), an ES-EX regression test (*Física y Química* → 1.º, *Historia de España* → 2.º, the General-modality trap, I/II families preserved) and a real-dataset Bachillerato **render** test.

**Remaining (documented in the README, needs curriculum validation):**
- `ES-EX` is **missing the 2.º entries** of two I/II subjects — *Artes Escénicas II* and *Matemáticas Aplicadas a las CC.SS. II* (Decreto 109/2022 arts. 18.5 / 17.2). This is an *under-representation*, not a duplication; fixing it needs the 2.º criterios/saberes content from the DOE (a regeneration, not a hand edit), so it is left as a follow-up.
- `ES-EFP` optativas and `ES-VC` (Valencian) — see the table.


erseco (dismissed):
Hi @pabloamayab, thanks for getting the Extremadura dataset in and for the clear write-up of what changed. I have gone through the JSON, the test changes and `lomloe.js`, and I would hold off on approving for now. There are three things I think we should sort out first.

### 1. Infantil competencias clave — possible data regression (blocking)

The previous test (`assertInfantilLinkedToCompetenciasClave`) encoded a decision from #1832: in Infantil, criterios are backfilled with their competencias clave. `ES-MD` and the state dataset still assert this behaviour, and the README documents it explicitly for `ES`, `ES-EX` and `ES-MD`.

In this PR, the new `ES-EX` JSON leaves Infantil criterios with `competencias_clave: []`, and the test was changed to assert that empty array instead of keeping the backfill check. The same helper stays green for Madrid asserting the contrary, so the net effect is that a previous guarantee for `ES-EX` gets removed.

Could we fix the dataset so Infantil keeps the backfilled links, rather than flipping the assertion? If there is a reason Extremadura should now differ from Madrid/State here, it should be documented in the README and covered by a specific test.

### 2. ESO/Primaria subject codes, which source is canonical? (needs confirmation)

The new JSON ships codes such as `BIG`, `FQX`, `GEH`, `EPV`, `CMN`, etc. for Primaria/ESO. The README's “official Extremadura siglas” section currently says these stages should instead use the official codes from the *Orden de 9 dic. 2022* / DOE annexes, with mappings such as `BIG→BG`, `FQX→FQ`, `GEH→GH`, `EPV→EPVA`, etc. The previous test also asserted the documented siglas and explicitly rejected the derived ones, while this PR inverts those checks.

Two possibilities, and I am not assuming which one is correct:

- the regenerated JSON may have skipped the final remap step to the documented official siglas, in which case the dataset should be regenerated or post-processed;
- or the new codes may be the correct official ones from the source used for this extraction, in which case the README is out of date and should be updated with the exact source/code convention.

Either way, the JSON, README and tests currently disagree, so we should confirm the canonical source and align all three.

### 3. ESO per-course filter must be aligned with the chosen codes (blocking)

`ESO_COURSE_SUBJECTS['ES-EX']` in `lomloe.js` currently filters subjects per course using the documented siglas (`BG`, `EF`, `EPVA`, `GH`, `LE`, `FQ`, `TECD`, etc.). If the JSON now ships `BIG`, `FQX`, `GEH`, `EPV`, `TYD`, etc., the per-course filter may hide valid subjects in the editor.

This PR does not touch `lomloe.js`, so the dataset and the runtime filtering logic are currently out of sync unless the JSON keeps the same code convention as the filter. If we keep the new codes, `ESO_COURSE_SUBJECTS['ES-EX']` needs to be updated too, and we should add a regression test proving that the filtered subject list still works for Extremadura.

### Minor

The README's `ES-EX` section should be updated once the code convention and Infantil behaviour are decided.

The main point is to keep `ES-EX` consistent with the iDevice runtime behaviour and with the conventions documented after #1832. Once the Infantil backfill is clarified/restored and the canonical code set is aligned across JSON, tests, README and `lomloe.js`, I am happy to take another look. Thanks again!

mnunezcedec:
Hi @pabloamayab, @erseco.

In a first review, where the objective is to determine whether all the areas and/or subjects from the different curricula are included, it appears that in Early Childhood Education and Primary Education they are present and correctly displayed. In Compulsory Secondary Education (ESO), only some subjects are visible in 1st, 2nd, and 3rd year, while they all seem to be present in 4th year. In Baccalaureate, subjects from 1st and 2nd year are mixed together.

In short, I consider that we should start with this as a first step and then, once all subjects are visible, we could carry out a more in-depth analysis of assessment criteria, competence descriptors, and basic knowledge.

erseco:
Hi @mnunezcedec, @pabloamayab — I've pushed a fix that resolves my own three change requests above and addresses Martin's first-pass review. (Self-responding here, since I made the changes directly on the branch.)

**Root cause:** the "data correction" commit shipped the generator's raw, pre-remap output — Primaria/ESO used the derived siglas (`BIG`, `FQX`, `GEH`…) instead of the official Extremadura ones (`BG`, `FQ`, `GH`…), and Infantil lost its competencias-clave backfill. The README already documented the correct contract; the data and tests had drifted from it. I fixed the **data** to match the contract rather than the tests to match the broken data.

**My three review points**
1. **Infantil backfill (blocking):** restored. `competencias_clave` are backfilled again, byte-identical to `ES`/`ES-MD` (matched by criterio código); `assertInfantilLinkedToCompetenciasClave` + the ÁCA spot-check are back.
2. **Code convention:** kept the documented official Extremadura siglas. Remapped all 12 Primaria/ESO códigos derived→official (`BIG→BG`, `FQX→FQ`, `GEH→GH`, `EPV→EPVA`, `TYD→TECD`, `EVC→EVCE`, `LEX→LE`, `EFI→EF`, `EEX→EyE`, `EAR→EA`, `FOP→FOPP`, `CMN→CMNS`) across área keys and every embedded competencia/criterio/saber code. README, tests and data now agree.
3. **ESO per-course filter:** `ESO_COURSE_SUBJECTS['ES-EX']` already used the official siglas, so it needed no change — the data now matches it. Added a real-dataset render test (1º–3º show the Anexo V distribution, 4º unfiltered) plus a filter↔dataset alignment test.

**@mnunezcedec's points**
- **ESO 1º–3º only some subjects visible:** fixed by the remap — they now show the full Decreto 110/2022 (Anexo V) distribution (8/9/9 materias); 4º stays unfiltered.
- **Bachillerato 1º/2º mixed:** confirmed, but it's *not* specific to Extremadura — the state dataset and every regional concretion duplicate most Bachillerato subjects into both years (Física/Química/Historia de España and Física y Química/Hª del Mundo Contemporáneo all appear twice). Aligning the Bachillerato year distribution needs the authoritative DOE Decreto 109/2022 per-year list and should be applied across **all** datasets, so I've documented it as a known limitation and propose handling it as the deeper follow-up you suggested — once this visibility fix lands.

Regional Extremadura content (regional criterios + saberes) is preserved; only códigos and Infantil links changed. `make fix` clean and all frontend unit tests green.

erseco:
## Bachillerato 1.º/2.º distribution — fixed 

@mnunezcedec, following up on your Bachillerato observation — I went deep on this. You were right, and it turned out **not** to be Extremadura-specific.

**Root cause.** RD 243/2022 (arts. 9–13) assigns every Bachillerato subject to a specific course, but the state-derived datasets duplicated most subjects into **both** years (the `#1827`/`#1828` extraction copied the per-*modalidad* curriculum into each year instead of placing it per course). Single-year subjects therefore appeared in the wrong year.

**Datasets audited**

| Dataset | Before | Action |
|---|---|---|
| `ES`, `ES-EX`, `ES-MD` | 29 subjects duplicated into both years | **Fixed** (29 wrong-year copies removed) |
| `ES-EFP` | 33 duplicated | **Fixed** (29 removed; 4 MEFPD optativas left, pending the optativa annex) |
| `ES-NC` | 29 duplicated | **Fixed** (29 removed) |
| `ES-CN` (Canarias) | already per-course | unchanged — used as **oracle** |
| `ES-GA` (Galicia) | already per-course | unchanged  |
| `ES-VC` (Valencià) | 21 duplicated (Valencian) | **unchanged** — needs Decret 108/2022 verified first |

**Bachillerato changes (per fixed dataset): the same 29 subjects**

- Moved to **1.º only**: Biología Geología y CC.AA., Cultura Audiovisual, Economía, Economía Emprendimiento y Act. Empresarial, Educación Física, Filosofía, Física y Química, Historia del Mundo Contemporáneo, Lenguaje y Práctica Musical, Literatura Universal, **Matemáticas Generales**, Proyectos Artísticos, Volumen.
- Moved to **2.º only**: Biología, **Ciencias Generales**, Diseño, Empresa y Diseño de Modelos de Negocio, Fundamentos Artísticos, Física, Geología y CC.AA., Geografía, Historia del Arte, Historia de España, Historia de la Filosofía, Historia de la Música y de la Danza, Literatura Dramática, Movimientos Culturales y Artísticos, Química, Técnicas de Expresión Gráfico-plástica.
- **Kept in both years** (I/II families, distinct content): Matemáticas, Latín, Griego, Dibujo Técnico, Dibujo Artístico, Dibujo Técnico Aplicado, Análisis Musical, Coro y Técnica Vocal, Lengua Castellana y Literatura, Lengua Extranjera.

**Normative sources.** RD 243/2022 (BOE-A-2022-5521) arts. 9–13; Extremadura **Decreto 109/2022** (DOE) arts. 15–19 (adopts the state distribution verbatim — no subject reassigned). Cross-validated against `ES-CN` and `ES-GA`. The modalidad *General* trap is handled (*Matemáticas Generales* = 1.º, *Ciencias Generales* = 2.º).

**Mechanism.** JSON-level fix (matching how `ES-CN`/`ES-GA` already encode it) — **no runtime filter**. Removal-only: the wrong-year copy (identical content, only the `BAC1`/`BAC2` tag differed) is deleted; the correct-year copy is untouched. Verified: 29 removed per dataset, **0 kept-entry content changed**; a final adversarial pass ran 70 counterexamples vs the `ES-CN` oracle with **0 mismatches**.

**Tests.** Per-dataset year-separation tests, an ES-EX regression (Física y Química → 1.º, Historia de España → 2.º…) and a real-dataset Bachillerato render test. All green; `make fix` clean.

### Where I'd value your curriculum validation 🙏

@mnunezcedec could you please review the Bachillerato 1.º/2.º subject distribution, especially for Extremadura? Two points need a curriculum eye:

1. **ES-EX is missing two 2.º subjects**: *Artes Escénicas II* and *Matemáticas Aplicadas a las CC.SS. II*. Decreto 109/2022 (arts. 18.5 / 17.2) lists them in 2.º, and `ES-CN`/`ES-GA` carry both years, but the ES-EX dataset only has the 1.º entry. This is an *under-representation* — fixing it means adding the 2.º criterios/saberes from the DOE (a regeneration, not a hand edit), so I left it as a follow-up rather than fabricate content. Does that match your reading of the decree?
2. Please sanity-check the 1.º vs 2.º lists above against the Extremadura *Anexo horario* — particularly the optativas and any subject your centres offer differently.

`ES-VC` (Valencian) is intentionally left for a separate pass once Decret 108/2022 is verified. Thanks!

mnunezcedec (commented):
Thanks @erseco 

The subjects are now appearing, but… from a quick review I can see that there are subjects in 1st and 2nd year of Bachillerato that share the same name, for example Music Analysis, while the DOE refers to Music Analysis I (for first year) and Music Analysis II (for second year). I also notice that Applied Mathematics for Social Sciences is only offered in the first year.

Additionally, regarding the way the learning outcomes are ordered— for example in Music Analysis— while the DOE lists them as 1.1.1, 1.1.2, 1.1.3, etc., the iDevice changes this order, which makes it more difficult to select.

I emphasize that this is just a quick review.

Image

Image

erseco:
@mnunezcedec thanks, this is exactly the kind of detailed review we needed. Three points, handled as follows:

### 1. I/II naming (Análisis Musical → I / II) — fixed ✅

You're right. The Castilian datasets showed the bare base name in both years; RD 243/2022 names these *Análisis Musical I* (1.º) / *Análisis Musical II* (2.º), exactly as `ES-CN` already does. I've appended the official course suffix to every genuine I/II family — *Matemáticas*, *Latín*, *Griego*, *Análisis Musical*, *Coro y Técnica Vocal*, *Dibujo Artístico / Técnico / Técnico Aplicado*, *Lengua Castellana y Literatura*, *Lengua Extranjera*, *Matemáticas Aplicadas a las CC.SS.*, *Artes Escénicas* — across `ES`, `ES-EX`, `ES-MD`, `ES-EFP` and `ES-NC`. Single-year subjects (*Física*, *Física y Química*, *Matemáticas Generales*…) are left untouched. Commit `cdd96f11d`.

### 2. Matemáticas Aplicadas a las CC.SS. only in 1.º — confirmed, needs regeneration ⚠️

Correct, and it's the same class of issue as *Artes Escénicas*: Decreto 109/2022 (arts. 17.2 / 18.5) lists **MACS II** and **Artes Escénicas II** in 2.º, but the `ES-EX` dataset only has the 1.º entry. This is a *missing* subject (under-representation), not a duplication — so I can't fix it by hand without fabricating the 2.º criterios/saberes; it has to come from the DOE in a regeneration. After the naming fix above, the 1.º entry now reads "… I", which makes the missing "… II" in 2.º more obvious.

### 3. Saberes order (1.1.1, 1.1.2…) — the committed extraction lost the codes ⚠️

This one matters: in the **committed** dataset the Bachillerato *saberes* don't carry the DOE code (`A.1.1.1`, `A.2.1.2`…) at all — I checked, **0 of 2215** Bachillerato saberes have it — and within a block they're out of order with some entries merged/dropped vs the DOE. So I can't re-sort them by hand: there's no code to sort on.

Your screenshot, though, shows the saberes **with** the codes ("AM A.2.1.2. El estudio de partituras…"), one per entry — i.e. a **better extraction than what's in the repo**. @pabloamayab is that the version you have locally? If so, committing that code-preserving Bachillerato extraction would fix both the ordering **and** the completeness (points 2 and 3) in one go, and keep the `A.b.c.d` codes visible/sortable. I've documented this as the regeneration follow-up in the iDevice README.

**Summary:** point 1 is fixed and pushed now; points 2 & 3 are pinned down and waiting on the improved (code-preserving) extraction. Thanks again for the careful look 🙏

mnunezcedec (commented):
Pending a resolution of the issue with the codes so that the order and content of the subjects can be displayed in accordance with the regulations, I tested another Year 1 Baccalaureate subject, specifically Physics and Chemistry. The system also generates several errors in the ordering of the codes, as well as omissions and the inclusion of learning content that is not defined in the regulations. (@erseco @pabloamayab )

I hope these issues can be resolved across all subjects, as reviewing each one individually could become an endless task.

erseco:
Hi @mnunezcedec — thanks again, your Física y Química read was exactly right, and digging in found the real cause. I've pushed a fix for the **ordering** half; the **completeness** half needs your curriculum eye, and I've turned it into a concrete checklist instead of a "review every subject" slog.

**Root cause.** The official DOE saber codes (`FYQ E.1.1`, `E.2.1`…) were only ever present as free text at the start of each saber's description — never stored as structured data. The iDevice renders saberes in the order they sit in the JSON, so wherever that hand-entered order didn't match the decree it showed wrong (e.g. FYQ 1.º block *E* was stored as `E.2.1, E.2.2, E.1.1`). On top of that the overlay is partial: some saberes carry a DOE code, others are still the state-level fallback text with no code.

**Fixed in this push:**
- The DOE code is now lifted into a structured `codigo` field for **3399 of 4601** ES‑EX saberes, and the prose no longer repeats it.
- Every block where **all** saberes carry a code is now sorted by it with a natural comparator (so `A.1.10` follows `A.1.2`). **220 blocks** were reordered. FYQ 1.º block *E* now reads **`E.1.1 → E.2.1 → E.2.2`**.
- The iDevice now **shows the DOE code** (`E.1.1`) on each saber and in the export, instead of the internal id — so it lines up with the decree when you select.

**Still needs your validation (the completeness half).** This part can't be done safely by hand — it's the DOE re-extraction we agreed to do as a follow-up. I've added `ES-EX-saberes-code-coverage.md`, which lists, per subject and block, every saber with **no DOE code** and every **sequence gap** — a finite checklist (**518 blocks** flagged) so no subject has to be eyeballed blind. Your FYQ 1.º example shows up precisely in block *A*:
- `A.1.1` and `A.2.2` present and ordered;
- **`A.2.1` missing** (a real omission);
- two saberes with **no code** (`SB01-002`, `SB01-003`) that are state-fallback text — likely the "content not in the Extremadura decree" you flagged.

**Where I'd value your curriculum eye:**
1. Does the coverage report match Decreto 109/2022 — i.e. are the un-coded saberes genuinely *not* in the DOE (drop them) vs. mis-transcribed (re-code them)?
2. Same for the FYQ gaps (`A.2.1`, and the un-coded items in `B`/`D`).
3. I'll fold the missing 2.º subjects you noted earlier (*Artes Escénicas II*, *Matemáticas Aplicadas a las CC.SS. II*) into the same DOE re-extraction pass.

The ordering fix + visible codes should make this usable now; the checklist drives the follow-up. 🙏

PR #1906, opened 2026-06-06
fix(idevices): dedupe electrical-circuits TikZJax to a single export copy

erseco:
## Summary

The `electrical-circuits` iDevice shipped the **TikZJax** library duplicated byte-for-byte in both `edition/` and `export/`:

| Asset | Per copy | Copies |
|-------|----------|--------|
| `tikzjax.js` | 6.7 MB | edition + export |
| `fonts/` (140 BaKoMa TeX `.ttf`) | 3.5 MB | edition + export |

That's **10.2 MB of raw duplication inside one iDevice** — and it ships twice in the static bundle (loose under `files/perm/...` **and** inside `bundles/idevices.zip`).

`export/` is the canonical, self-contained copy that gets bundled into exported content (SCORM/HTML5/EPUB). The `edition/` copy is only used by the authoring preview. This PR keeps a single copy in `export/` and points the edition preview at it.

## What changed

- **Load TikZJax in the edition form from `export/`** (single source of truth). `edition/electrical-circuits.js` now loads `../export/tikzjax.js` explicitly, guarded so it never double-loads when the export runtime already injected it.
- **Removed `edition/tikzjax.js` and `edition/fonts/`.**
  - `edition/fonts/` was *never referenced* — the edition CSS has no `@font-face`; runtime tracing confirmed zero `.ttf` requests from the editor.
  - `edition/tikzjax.js` was only auto-included by the folder scan in `idevice-parser` (`getValidFilenames` lists every `.js` in the folder when `config.xml` declares none). It's now loaded explicitly from `export/`.
- **Updated third-party license paths** in `public/libs/README.md` to point only at `export/` (the BaKoMa font licence file remains in `export/fonts/`).

`export/` is untouched, so **exported content is byte-identical** — this only affects what the authoring app loads.

## Impact

Static editor `.zip` (`exelearning-static-*.zip`):

| | Size |
|---|---|
| before | 66.9 MB |
| after | **59.4 MB** |
| **saved** | **−7.2 MiB (−11%)** |

(The full duplication is 10.2 MB raw; in the compressed zip it nets ~7.2 MiB once across the loose files and `idevices.zip`.)

## Testing

- Added `test/e2e/playwright/specs/idevices/electrical-circuits.spec.ts`: adds the iDevice, enters a `circuitikz` diagram, triggers the preview, asserts the SVG renders **and** that `tikzjax.js` loads from `/export/` and **never** from `/edition/`. ✅
- `make fix` clean; frontend + backend unit suites green (7112 pass / 0 fail).

> Note: `make test-unit`'s global-coverage gate flags `src/websocket/asset-coordinator.ts` (88.2%), a file this PR does not touch (pre-existing; it's also covered by the integration suite).

## Follow-up (out of scope)

The `export/` BaKoMa fonts are still 140 raw `.ttf` (~3.5 MB). Converting them to WOFF2 would save another ~1.5–2 MB, but it changes the shipped export output and needs font tooling + rewriting 140 `@font-face` rules — better as its own PR.


erseco:
Follow-up commit `harden electrical-circuits TikZJax load guard`:

The first attempt loaded fine locally but the circuit preview stayed blank on the Cloudflare Pages **static** preview. Root cause: the edition guard was `document.querySelector('script[src*="tikzjax"]')`. In static builds every asset URL carries the cache-busting `?v=<build-version>` query, and this branch's version string contains "tikzjax" (`v0.0.0-...-tikzjax-...`), so the selector matched unrelated scripts (e.g. `jquery.min.js?v=...-tikzjax-...`). The guard then short-circuited and the real `tikzjax.js` was never injected. `bun x serve` didn't reproduce it because the local build version had no "tikzjax" substring.

Fix: match the actual `tikzjax.js` filename (last path segment), not any src containing the substring. Verified by rebuilding the static bundle with a `...-tikzjax-...` version and running the E2E against `--project=static` — preview renders and TikZJax loads from `export/`.

erseco:
Related: #1907 re-encodes this iDevice's TeX fonts to WOFF2 (export/ only, independent of this PR — they merge in any order).

erseco:
Closed as superseded by #1912

PR #1907, opened 2026-06-06
perf(idevices): serve electrical-circuits TeX fonts as WOFF2

erseco:
## Summary

The `electrical-circuits` iDevice ships the BaKoMa TeX fonts that **TikZJax** uses to render circuit labels — it emits `<text font-family="cmXXX">` and the iDevice's export CSS declares 140 `@font-face` rules for them. They shipped as **140 TTF files (~3.4 MiB)**.

This converts them to **WOFF2** (lossless — identical glyph outlines, just better compression; supported by every browser since 2016) and points the `@font-face` rules at the `.woff2` files.

| | Fonts (raw) |
|---|---|
| before (TTF) | 3.44 MiB |
| after (WOFF2) | **1.99 MiB** (−42%) |

Static editor zip: **−~0.9 MiB** (the fonts ship both loose under `files/perm/...` and inside `bundles/idevices.zip`; in the zip TTF is already deflated, so the net is smaller than the raw saving).

## Scope / independence

This touches **only** `export/fonts/*` and `export/electrical-circuits.css`, so it is fully independent of the edition/ dedup PR #1906 (which touches `edition/*`) — the two merge in any order without conflict.

> Note: the third-party license path in `public/libs/README.md` (`export/fonts/*.ttf`) is owned by #1906; once both land it should read `export/fonts/*.woff2`. Left untouched here to avoid a cross-PR conflict.

## Testing

- New E2E `test/e2e/playwright/specs/idevices/electrical-circuits-fonts.spec.ts`: loads the real export stylesheet and asserts the browser loads the faces (`cmr10`, `cmmi10`, `cmsy10`, `cmex10`, `msbm10`) as valid WOFF2, and that **no `.ttf`** is ever requested. Green in both `--project=chromium` and `--project=static`.
- `make fix` clean.

## Conversion provenance

Generated once with fonttools (lossless container reflavor, no glyph changes):

```python
from fontTools.ttLib import TTFont
f = TTFont('cmXXX.ttf'); f.flavor = 'woff2'; f.save('cmXXX.woff2')
```

Related: #1906 (edition/ TikZJax dedup).


erseco:
@mnarvaezm a question for you as the author of this iDevice 🙏

This PR only **re-encodes** all 140 BaKoMa TeX fonts to WOFF2 (lossless, keeps every glyph). But while measuring, I found that a representative set of circuits only ever pulls in **8 of the 140 fonts**: `cmr10`, `cmr7`, `cmmi10`, `cmmi7`, `cmsy10`, `cmex10`, `msbm10`, `eufm10` (roman, math italic, math symbols, large operators, AMS blackboard, fraktur).

So there's a much bigger optimization on the table — **dropping the never-used fonts**:

| Keep | Size | Extra saving vs all 140 |
|---|---|---|
| all 140 (this PR) | ~2.0 MiB | — |
| only the 8 observed | ~0.12 MiB | **~1.87 MiB** |
| all sizes of the 6 used families (cmr/cmmi/cmsy/cmex/msbm/eufm) | ~0.58 MiB | **~1.41 MiB** |

The reason I **didn't** just delete them: the question labels are free LaTeX (`l=$...$`), so a teacher could pull in a font that isn't in the trimmed set and it would render with a wrong fallback — e.g. `\textsc{}` → `cmcsc*`, `\texttt{}` → `cmtt*`, or an uncommon symbol from another family. The declared packages are `circuitikz, amsmath, amssymb`.

**Could you confirm** whether it's safe to drop the unused fonts, and which families circuitikz + your typical label usage actually needs? If you're comfortable bounding the set, we can ship the trimmed list (≈1.4–1.9 MiB more saved) instead of all 140. Otherwise we keep all 140 as WOFF2 (this PR), which is risk-free.

Related: #1906 (edition/ TikZJax dedup).

erseco:
Closed as superseded by #1912 @ignaciogros @mnarvaezm  feel free of apply the same solution to edition/ folder.

PR #1910, opened 2026-06-06
perf(static): assemble resource bundles from loose files, drop shipped zips (−9.8 MB / −19.7% vs main)

erseco:
## Summary

The static PWA build (`make build-static`) shipped **two parallel copies of every resource**:

| Source | Consumed by |
|---|---|
| **Loose tree** (`files/perm/**`, `app/common/**`, `libs/**`, `style/workarea/**`) | The **live workarea** — `theme.js`/`idevice.js` inject `<link>`/`<script>`/`` at loose URLs in static mode; plus fonts, tinymce theme CSS, rubric `html2canvas.js`. |
| **Bundle zips** (`idevices.zip`, `themes/*.zip`, `common.zip`, `libs.zip`, `content-css.zip`, ~18 MB) | **`ResourceFetcher` only** — fetched + unzipped into `Map<path, Blob>` to build the downloadable HTML5/SCORM/EPUB export, and for the "download theme" button. |

The loose tree is **mandatory** (the editor renders from it) and is already shipped, so the zips were ~18 MB of **incompressible, redundant duplication** in the download.

This PR makes static mode **assemble each bundle in-memory from the loose files on first use**, driven by per-bundle file lists in `bundles/manifest.json`, then **persist the result to IndexedDB** so later sessions are a single cache read. The zips are no longer shipped in `dist/static` (only the manifest is).

**Result: static `.zip` 47.15 → 37.85 MB (−9.76 MB, −19.7%), re-measured 2026-07-03 against current `main` (`c3c3273c8`).** Manifest grows ~230 KB raw (~16 KB gzipped).

_(Originally measured −18.6 MB / −29% against the `main` baseline at the time this PR was opened; `main` has since dropped in size independently through unrelated work, so the absolute delta shrank, but the mechanism and relative savings this PR provides are unchanged.)_

## Scope guard

**Server mode is untouched.** `build-resource-bundles.js` still builds `public/bundles/*.zip`, and `src/routes/resources.ts` still serves them via `/api/resources/bundle/*` for the web-server/Docker deployment. Only the **static distribution** and **static-mode runtime** change.

This covers the web PWA, LMS iframe embedding, and the **Electron desktop app** (also static mode, served over the custom `app://` protocol). Assembly is plain `fetch` on the main thread — **no service worker is involved**, so the custom `app://` protocol keeps working (a SW-based approach would not).

## Changes

- **`scripts/build-resource-bundles.js`** — emit a `staticFiles` map of `{ s: sourceUrl, t: targetPath }` entries per bundle (themes per-name, iDevices per-name, libs flat, common per-lib, contentCss flat) alongside the zips. The same enumeration that builds the zips emits the list → **single source of truth**, no path logic duplicated on the client.
- **`scripts/build-static-bundle.ts`** — copy `bundles/manifest.json` only, not the zips. Loose trees are already copied.
- **`public/app/yjs/ResourceFetcher.js`** — `loadStaticManifest`, `getStaticFileList`, `assembleBundleFromLoose`, shared MIME helper, `fetchThemeZipBlob`. The five static methods now assemble from loose files; **lazy per-iDevice** assembly replaces loading the whole 11 MB `idevices.zip` up front. Adds IndexedDB persistence for iDevices and common libraries (themes/libs/content-css already persisted via their wrappers).
- **`navbarStyles.js` / `modalStyleManager.js`** — in static mode, assemble the theme zip client-side via `fetchThemeZipBlob` (fflate) instead of fetching the shipped zip.

## Why not the alternatives

- **Service worker synthesizes the zip on request** — rejected: SWs don't run on Electron's `app://` protocol, so desktop would break.
- **Drop the loose tree, keep zips + a SW virtual filesystem** — rejected: larger refactor touching workarea rendering, offline, and export; would force re-bundling `edition/` (19 MB), fonts, `data/`, `config.xml`, clawing back most of the saving.

## Tests

- **`build-resource-bundles.spec.ts`** — new `staticFiles` assertions, including **assembled set === zipped set** for `idevices.zip` / `content-css.zip` / `libs.zip` (byte-fidelity guarantee) and that every source URL exists on disk.
- **`ResourceFetcher.test.js`** — rewritten static-mode suite: assembly fetches loose URLs (not `/bundles/*.zip`), correct `Map<targetPath, Blob>`, content-css `content/css/` prefix, lazy per-iDevice, IndexedDB read-through + write, updated loading chain, `fetchThemeZipBlob`, `assembleBundleFromLoose`/`getStaticFileList` helpers.
- **UI specs** — `downloadThemeFromBundle` static path (assemble + zip client-side).

### Verification

- `make build-static` → `dist/static/bundles/` contains **only `manifest.json`** (0 zips); static `.zip` 47.15 → 37.85 MB (re-measured 2026-07-03 against current `main`).
- Static e2e (`--project=static`): `save-detection-static`, `url-param-static`, and `component-export-import` all green — the last roundtrips export→import and **renders images in preview**, proving assembly produces working exports/previews in a real browser.
- `make fix` clean; frontend + script unit suites green.

CI runs the full static e2e matrix on this PR.

## Fix — test sources listed in the static manifest caused 404 noise

`scanDirectory()` only skipped dotfiles, so colocated `*.test.js` sources ended up in `staticFiles.idevices`/`.common` and the zips — but `dist/static` excludes them, so first-load assembly 404s and logs a `console.warn` per file, and static manifests diverged from the server-mode zips. `scanDirectory()` now filters `.test.js`/`.spec.js`, matching the copy exclusion. Commit `f34966d3`; test added.

## Architecture record

This PR adds **[`ADR-1910-01`](doc/architecture/adr/ADR-1910-01-assemble-static-resource-bundles-from-loose-files.md) — "Assemble static-mode resource bundles from loose files instead of shipping zips"** (`status: Proposed`), recording the decision, the four options considered (status quo · client-side assembly · service-worker zip synthesis · dropping the loose tree) with pros/cons and a valuation each, the consequences, and a risk table with likelihood/severity.

The size figures in the ADR were **re-measured at the current branch head** (`067e7eabc`) rather than carried over from the July measurement, comparing two `dist/static` builds — the shipped one, and a reconstruction of the pre-change variant (same tree plus `public/bundles/*.zip` and the manifest with `staticFiles` stripped):

| Variant | Files | Tree bytes | `zip -9` bytes |
|---|---:|---:|---:|
| Ships zips (pre-change equivalent) | 3,498 | 94,088,041 | 51,099,669 |
| Assembles from loose (this PR) | 3,487 | 84,361,742 | 41,388,557 |
| **Delta** | −11 | **−9,726,299 (−10.3%)** | **−9,711,112 (−19.0%)** |

Manifest cost: +216,973 B raw, ≈ +13.7 KB gzipped. Measured fan-out from the manifest: 52 iDevices / 976 files (median 14, max 158 = `map`), themes 56–97 files, `exe_math` 85.

The ADR records the trade-offs honestly, including the ones this PR does not solve: per-file request fan-out on first assembly, the loss of a per-bundle integrity hash (a missing file is skipped with a `console.warn` instead of failing the bundle), and the new obligation that `manifest.json` and the files `dist/static` actually copies never diverge — citing the `.test.js` incident fixed in `f34966d3` as evidence that the class of bug is real. Two follow-ups are recorded: make an incomplete assembly fail loudly, and add a drift check against the **built** `dist/static` (the current spec checks `public/`, which is exactly why that divergence was invisible to it).


ignaciogros:
Hi @erseco,

Could you please check the merge conflicts?

Thanks.

erseco:

- **Minor (confirmed):** `scripts/build-resource-bundles.js` `scanDirectory()` skips only dotfiles, so `manifest.staticFiles.idevices` (and `.common`) list the colocated `*.test.js` sources — 38 under iDevice `export/` dirs (e.g. `checklist/export/checklist.test.js`) plus 4 common-lib test files. But `dist/static` copies those trees via `copyDirRecursive` (`scripts/build-static-bundle.ts:1141`, default `excludePatterns=['.test.js','.spec.js']`), so the files are absent. On first assembly `assembleBundleFromLoose` (`public/app/yjs/ResourceFetcher.js:458`) fetches each `s`, 404s, and emits one `console.warn` per file before skipping it gracefully. No functional break (test files are never runtime assets and the assembled Map/IndexedDB copy just omits them), but it produces avoidable 404/console noise per iDevice on first load and makes static-mode exports diverge from server-mode zips (which still bundle the test files). Cleanest fix: filter `.test.js`/`.spec.js` in `scanDirectory` for both the manifest lists and the zips.
- **Test gap:** the guard `build-resource-bundles.spec.ts:193` asserts each `s` exists under `public/` (source tree, where test files live), not under the built `dist/static`, so it cannot catch this; `static-bundle-assembly-static.spec.ts` only asserts no `/bundles/*.zip` request and that one image renders — it does not assert absence of 404s/console warnings.

Verified: read at head 7c0490e; server-mode path (`src/routes/resources.ts` zips) is untouched, and the assembly null-skip logic is correct. CI: all checks green on 7c0490e ✅.
Residual risk: no local `make build-static` was run; the 404 behavior is inferred from the copy-exclusion vs. manifest-inclusion asymmetry in code, not observed at runtime. The issue is cosmetic/consistency-level and does not block merge.

---
**Update — resolved (`f34966d3`):** `scanDirectory()` filters `.test.js`/`.spec.js`, so those files appear in neither the manifest nor the zips (no more first-load 404s); test added.

ignaciogros (changes requested):
Hi @erseco,

Thanks for this work. I think it can be approved.

One thing to consider before merging: changing how resources are shipped in the static distribution looks like a durable architecture decision under `AGENTS.md`, so it would be worth adding a short ADR (or linking an existing one) to record it.

That’s the only reason I’m requesting these changes. It can also be merged now if you prefer.

erseco:
Hi @ignaciogros — you're right, this is a durable decision, so I've added an ADR rather than merging without one.

**[`ADR-1910-01`](https://github.com/exelearning/exelearning/blob/feat/static-assemble-bundles-from-loose/doc/architecture/adr/ADR-1910-01-assemble-static-resource-bundles-from-loose-files.md) — "Assemble static-mode resource bundles from loose files instead of shipping zips"** (`status: Proposed`, you're listed as reviewer). Commit `7da1b5fbc`; `make architecture-check` passes.

It records four options with pros, cons and a valuation each — status quo, client-side assembly (chosen), a service worker that synthesizes the zips, and dropping the loose tree in favour of a zip-backed virtual filesystem. The last two are rejected on the same blocker: service workers don't run on Electron's `app://` protocol, so either one would break exports (or all rendering) in the desktop app.

Rather than reuse the July figures — `main` has moved since — I re-measured at the current head (`067e7eabc`), comparing the shipped `dist/static` against a reconstruction of the pre-change variant:

| Variant | Files | Tree bytes | `zip -9` bytes |
|---|---:|---:|---:|
| Ships zips (pre-change equivalent) | 3,498 | 94,088,041 | 51,099,669 |
| Assembles from loose (this PR) | 3,487 | 84,361,742 | 41,388,557 |
| **Delta** | −11 | **−9,726,299 (−10.3%)** | **−9,711,112 (−19.0%)** |

The manifest costs +216,973 B raw (≈ +13.7 KB gzipped) for the 9.94 MB of zips removed.

The negatives are written up with the same weight as the benefits: per-file request fan-out on first assembly (median 14, worst case 158 for the `map` iDevice), the loss of a per-bundle integrity hash — a missing file is skipped with a `console.warn` rather than failing the bundle — and the new obligation that the manifest and what `dist/static` actually copies never diverge. That last one isn't hypothetical: it's the `.test.js` bug fixed in `f34966d3`, and the ADR cites it as evidence of the failure mode. Two follow-ups are recorded for it: make an incomplete assembly fail loudly, and add a drift check against the **built** `dist/static` (the current spec checks `public/`, which is precisely why it couldn't catch that one).

Server mode is untouched throughout — the zips are still built and still served by `/api/resources/bundle/*`.

Ready for another look whenever suits you.

ignaciogros:
Thanks, @erseco.

I'll change the ADR status to "Accepted". It's ready to be merged.

PR #1914, opened 2026-06-07
Fixes #1911 - "Creado con eXeLearning" always in Spanish

### LINKED ISSUE #1911: "Creado con eXeLearning" link does not change with package language
1) Create a new file
2) Both languages in Basque (interface and content)
3) Choose "Created with eXeLearning link" option
4) Preview and exports show the message in Spanish

I think package language should be the language used in the link

ignaciogros:
The “Made with eXeLearning” link was always in Spanish ("Creado con eXeLearning") because it was hardcoded in `src/shared/export/renderers/PageRenderer.ts`. 

It is now displayed in the content language (not necessarily the application language).

### DIFF
--- a/src/cli/commands/translations-format.spec.ts
+++ b/src/cli/commands/translations-format.spec.ts
@@ -34,6 +34,66 @@ function makeXlf(targetLang: string, units: { id: string; resname: string; sourc
 }
 
 // ---------------------------------------------------------------------------
+// Unit tests — fixAttributeQuotes
+// ---------------------------------------------------------------------------
+
+describe('fixAttributeQuotes', () => {
+    let fixAttributeQuotes: (content: string) => string;
+
+    beforeEach(async () => {
+        ({ fixAttributeQuotes } = await import('./translations-format'));
+    });
+
+    it('replaces left/right curly double-quote attribute delimiters with straight quotes', () => {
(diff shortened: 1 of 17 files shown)


erseco (changes requested):
Thanks, the implementation looks aligned with #1911: the hardcoded Spanish “Made with eXeLearning” text is now resolved from the package/content language, and the fallback path through `trans()` makes sense.

I would request one blocking change before merging:

`translations/messages.ro.xlf` now contains typographic quotes in XML attributes:

```xml
<trans-unit id=”MwExe01” resname=”Made with eXeLearning”>
...
<trans-unit id=”AK7eXCi” resname=”Page counter”>
````

These should use regular XML quotes:

```xml
<trans-unit id="MwExe01" resname="Made with eXeLearning">
...
<trans-unit id="AK7eXCi" resname="Page counter">
```

This also changes the existing `Page counter` entry, so Romanian translations may fail to parse or fall back incorrectly.

Optional follow-up: it would be useful to add an XML/XLF validation step in CI, because the current checks are green but did not catch this.

ignaciogros:
Thanks for the feedback! 

I had noticed some “odd” quotation marks, but only in the translations, not in the strings themselves. I'll fix it right away and update the `make translations-format` command so that it handles those cases too. There is already some cleanup in place, but it's incomplete.

erseco (approved):
LGTM!

PR #1915, opened 2026-06-07
fix(form): prevent question and answer loss or corruption

### LINKED ISSUE #1887: True-false questions changes from true to false
1) Create a form iDevice
2) Create a True-false question
3) Select True
4) Save question, iDevice and file
5) Edit the question: it has changed and now is false

mnarvaezm:
This PR groups several fixes for the **Form** iDevice aimed at preventing
questions and answers from being lost or corrupted.

- **Correctly persist True/False answers (#1887)** — the answer no longer changes
  from *True* to *False* when reopening or changing the language.
- **Preserve special characters in selection questions** — texts such as `A<B`
  are no longer truncated in the editor, export, or legacy import.
- **Preserve questions when saving an edited question** — editing a question no
  longer causes neighboring questions to be lost or misplaced.
- **Preserve the question order when adding them with the up/down buttons.**

## Tests

Includes unit tests (Vitest), backend tests (bun test), and E2E tests (Playwright)
for each fix.

Closes #1887


erseco (approved):
The changes look good to me.

The True/False fix is consistent: answers are normalized to stable `1`/`0` values, the preview renderer only serializes the value into the generated HTML, and the exported runtime converts the value back with `Number(...)` / `parseInt(...)` when checking or showing answers. So string values are handled correctly.

The other changes also make sense: option texts are escaped as plain text, the radio lookup is scoped to the current container, and question order is synchronized from the DOM with guards against malformed state.

Approved.

PR #1917, opened 2026-06-07
fix(export): render LaTeX in JSON-based interactive iDevices (preview and exports)

### LINKED ISSUE #1916: LaTeX is not rendered in JSON-based interactive iDevices (preview and exports)
LaTeX formulas written inside **JSON-based interactive iDevices** are not rendered: they appear as raw source code (`\( ... \)`, `\[ ... \]`, `$$ ... $$`) both in the **preview** and in **exports**.

Affected iDevices:

- **Adaptative Quiz** (`adaptative-quiz`)
- **Form** (`form`)
- **True/False** (`trueorfalse`)
- **Scrambled List** (`scrambled-list`)

The rest of the iDevices, such as text iDevices, render LaTeX correctly, so the issue is specific to this family of activities.

## Steps to reproduce

1. Create a page with an **Adaptative Quiz** iDevice, or with Form / True-False /   Scrambled List.
2. Add a question or answer containing LaTeX, for example:    `\( x^2 + y^2 = z^2 \)`.
3. Open the **preview** and/or export to **HTML5 / single page / SCORM / IMS / ELPX**.
4. Check the result.

**Expected result:** the formula is rendered correctly with MathJax.  
**Current result:** the raw LaTeX code is displayed without being rendered.

## Origin of the problem

The problem is that these four iDevices use **`componentType: 'json'`**: they do notstore the content of questions and answers as static HTML, but as **serialized JSON properties** of the component. That content is only converted into DOM at runtime,when the iDevice JavaScript renders the activity: starting the game, changing round, showing a message, displaying the report, and so on.

As a result, the LaTeX in these activities falls between the two paths:

1. **The pre-rendering step does not reach it**, because at export time the formula  does not exist yet in the DOM; it is still inside the JSON data.
2. **MathJax is not packaged**, because the decision only depends on the `addMathJax`    option, which is disabled by default.

mnarvaezm:
LaTeX written inside the JSON-based interactive iDevices — **Adaptative Quiz**, **Form**, **True/False** and **Scrambled List** — was shown as raw source (`\( ... \)`, `\[ ... \]`, `$$ ... $$`) in the preview and in every export.
This PR makes those formulas render correctly with MathJax, including the LaTeX placed in the **instructions** / "after" text of the activities.
Closes #1916 
Closes #1899 

## Root cause

These four iDevices are `componentType: 'json'`: questions, answers and instructions are not stored as static HTML but as serialized JSON properties, and are only turned into DOM **at runtime** by the iDevice JavaScript. As a result the LaTeX fell between the two existing paths:

- **Pre-rendering never reached it** — at export time the formula is not in the  DOM yet, it still lives inside the JSON.
- **MathJax was not bundled** — the decision depended only on the `addMathJax`  option, which is off by default.

## Changes

**Export pipeline — detect runtime JSON LaTeX and bundle MathJax**
- `BaseExporter` now detects LaTeX stored in the JSON properties of
  `adaptative-quiz`, `form`, `trueorfalse` and `scrambled-list`
  (`pagesHaveRuntimeJsonLatex` / `componentHasRuntimeJsonLatex`).
- When detected, MathJax is bundled and referenced and pre-rendering is skipped
  for those pages, in **preview, HTML5, single page, SCORM 1.2/2004, IMS and
  ELPX**, regardless of the `addMathJax` setting.
- **EPUB3 is intentionally excluded**: EPUB readers do not run JavaScript, so
  these interactive iDevices never render at runtime there — EPUB keeps
  pre-rendering visible LaTeX to SVG+MathML.

**Runtime — typeset the dynamically rendered content**
- `adaptative-quiz` calls `updateLatex(...)` after each dynamic render (initial
  template, messages, round changes, report) so freshly injected questions are
  typeset.
- `trueorfalse` now typesets the whole `.exe-trueorfalse-container` (the wrapper
  that holds the **instructions**, the questions and the after-text) instead of
  only `#tofPMainContainer`, which left the instructions untouched.

## Tests

- **Unit:** export-detection logic for the four iDevice types
  (`Html5Exporter`, `Page`, `Ims`, `Scorm12`, `Scorm2004`, `Elpx`),
  `adaptative-quiz` runtime typesetting, and `trueorfalse` instructions
  typesetting.
- **Integration:** LaTeX-in-JSON export (`latex-text-idevice-export.spec.ts`).
- **E2E (Playwright):** preview and browser exports include MathJax when JSON
  properties contain LaTeX (`runtime-json-latex-export.spec.ts`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


erseco (changes requested):
I think we should revisit this approach.

The exporter already has a LaTeX pre-rendering path intended to avoid bundling MathJax when `addMathJax` is false. It also already handles some JSON iDevices through `data-idevice-json-data`, and DataGame content is processed recursively during export.

The current PR detects runtime JSON LaTeX and forces MathJax instead of extending the existing SVG+MathML pre-rendering path. That seems less compatible and inconsistent with the current export strategy.

A better fix may be to make JSON-property pre-rendering recursive for `data-idevice-json-data`, similar to `preRenderLatexInGameData()`, so nested fields such as questions, answers and feedback are converted to SVG+MathML during export. MathJax should remain the fallback only when pre-rendering fails or when `addMathJax` is explicitly enabled.

mnarvaezm:
LaTeX in JSON iDevices — Adaptative Quiz, True/False, Scrambled List and Form — is now pre-rendered to SVG+MathML in preview/export, so MathJax no longer needs to be loaded at runtime.  The necessary modifications have been made to the affected iDevices to ensure compatibility with the pre-rendering of their elements. Questions, options and feedback are processed, including nested fields.

 The now-unnecessary RUNTIME_JSON_LATEX_IDEVICES machinery has been removed.

Covered with unit, integration and e2e tests. 

Please take a look when you can.
 @erseco  @ignaciogros

erseco (approved):
Approved. I reviewed the updated approach and verified locally that the export path now pre-renders LaTeX in the affected JSON iDevices to SVG+MathML instead of bundling/runtime-loading MathJax when addMathJax is disabled.

I also ran the targeted unit/integration tests and the new Playwright spec against localhost:8082. The browser export checks pass and confirm that MathJax is neither bundled nor referenced while the math is rendered via exe-math-rendered output.

This addresses the previous review concern about reusing the existing pre-rendering strategy.

PR #1929, opened 2026-06-10
fix(export): align pre-rendered LaTeX with the text baseline (#1919)

### LINKED ISSUE #1919: Pre-rendered LaTeX (span.exe-math-rendered) is not vertically aligned with the paragraph baseline
### Affected area
LaTeX-to-SVG+MathML pre-rendering in exports with `addMathJax = false`.
Wrapper generators: `createRenderedWrapperHtml()` in `public/app/common/LatexPreRenderer.js` (browser) and `src/shared/export/prerender/ServerLatexPreRenderer.ts` (CLI/API).

### Observed behavior
For **inline** math, the SVG-rendered formula is vertically offset relative to the surrounding paragraph text (it sits too high/low on the baseline) instead of being centered on the baseline the way MathJax does at runtime. The shift is most visible with fractions, sub/superscripts, and radicals (glyphs with depth below the baseline).

### Expected behavior
The SVG should align its baseline with the surrounding text, matching the live MathJax render (`addMathJax = true`).

### Exported DOM structure
​```html
<p>… text
  <span class="exe-math-rendered" data-latex="\frac{1}{2}">
    <svg ... style="vertical-align: -1.005ex;" ...>…</svg>
    <math>…</math>
  </span>
… text</p>
​```

### Technical analysis (root-cause hypotheses)
1. **No CSS for `.exe-math-rendered`.** The project defines no rule for the wrapper (no matches under `assets/styles/**`). The span stays `display: inline` by default, so alignment relies **solely** on the inline `style="vertical-align: -X.XXXex"` that MathJax sets on the `<svg>`. The paragraph `line-height`, or the SVG being a replaced element inside an uncontrolled inline span, can break that alignment.
2. **Assistive MathML without its hiding container.** The wrapper concatenates `inner = svg + mathml`, where `mathml` is the `<math>` extracted from `mjx-assistive-mml math` — but **without** the surrounding `<mjx-assistive-mml>` element or its CSS (`position:absolute; clip; …`) that normally takes it out of flow. Exports omit that CSS, so in MathML Core browsers (Chrome/Firefox) the `<math>` **occupies layout box and/or renders**, adding height/width and shifting the baseline (possible double render).
3. **Possible loss of the inline `vertical-align`.** Verify that the export chain (HTML sanitization, serialization into the `data-idevice-json-data` attribute, `DOMParser`/`innerHTML` round-trips) does not strip or normalize the `<svg>` inline `style`. If lost, the SVG falls back to `vertical-align: baseline` and misaligns.

### Reproduction
1. iDevice with inline math, e.g. `Let \(\frac{a}{b}\) be the quotient`.
2. Export to HTML5/SCORM/IMS/Page/ELPX with **MathJax disabled**.
3. Open the page: compare the vertical position of the SVG against adjacent text (and contrast with the same content using `addMathJax = true`).

### Proposed fix direction
- Add a minimal stylesheet for `.exe-math-rendered` that normalizes the wrapper's box model and respects the SVG's `vertical-align` (e.g. wrapper `inline`, SVG inheriting `vertical-align`; `data-display="block"` → `display:block; margin:auto`), shipped with exports alongside the pre-render.
- Ensure the assistive MathML **does not affect layout**: either re-wrap it in a reusable hidden container (replicating `mjx-assistive-mml` with `position:absolute; clip-path:inset(50%)`), or apply `.exe-math-rendered math { display:none }` in MathJax-free exports.
- Confirm the `<svg>`'s `style="vertical-align:…"` survives the full export pipeline intact.

### Notes
- Affects only the pre-render path (`addMathJax = false`); with runtime MathJax the alignment is correct because its own CSS manages the container and assistive MathML.
- Applies to all iDevices with pre-rendered math (text, and the recent JSON ones: `trueorfalse`, `adaptative-quiz`).

mnarvaezm:
In exports without MathJax (`addMathJax = false`), pre-rendered inline formulas rendered as SVG were vertically offset relative to the paragraph text.

This affected formulas with fractions, sub/superscripts and radicals, which did not sit on the text baseline as they do when MathJax renders them at runtime.

## Root cause

The `.exe-math-rendered` CSS forced `vertical-align: middle` on the wrapper and on the SVG.

That overrode the inline `vertical-align: -X.XXXex` style that MathJax sets on the `<svg>` itself to place the glyph depth correctly below the baseline.

## Changes

* Drop `vertical-align: middle`.
* The wrapper now uses `display: inline-block; line-height: 0`.
* The SVG is allowed to drive its own alignment through the inline `vertical-align` generated by MathJax.
* Replace deprecated `clip` with `clip-path: inset(50%)` on the hidden MathML.
* Extract the CSS into a shared `PRERENDERED_LATEX_CSS` constant in `constants.ts`.
* Use `PRERENDERED_LATEX_CSS` as the single source of truth in:

  * `Html5Exporter`
  * SCORM 1.2 / 2004 exporters
  * IMS exporter
  * Page exporter
  * ELPX exporter
  * `Epub3Exporter`
* Remove the duplicated literal CSS.

## How to test

1. Create a new project.
2. Add a Text iDevice.
3. Add some text.
4. Insert an inline LaTeX formula in the text, for example: `\( \frac{x^2 + 1}{\sqrt{x + 2}} \)`.
5. Save the changes.
6. Open the preview.
7. Check that the formula is rendered correctly and aligned with the surrounding text.

## Tests

* Add regression coverage in `constants.spec.ts`.
* Add regression coverage in `Html5Exporter.spec.ts`.
* Add regression coverage in `Epub3Exporter.spec.ts`.
* Check that the wrapper does not force `vertical-align: middle`.
* Add an E2E test in `latex-rendering.spec.ts`.
* Check that the exported `<svg>` keeps its inline `vertical-align`.
* Check that the export includes the corrected wrapper rule.


erseco (approved):
Approved. I tested this locally against localhost:8082 and the fix works as expected.

I also generated a screenshot from the exported HTML5 output and the inline fraction/subscript/radical now sit correctly on the surrounding text baseline.

Image

PR #1930, opened 2026-06-10
fix(export): keep exe-node: internal links in content.xml across round trips (#1927)

erseco:
Fixes #1927

## Problem

`BaseExporter.preprocessPagesForExport()` rewrote every `href="exe-node:<pageId>[#anchor]"` into a **static export path** (`index.html`, `html/about.html`, `../index.html#x`) **in the source HTML**, and that same mutated source then fed `generateOdeXml()` → `content.xml`.

Since the importer (`ElpxImporter.remapInternalPageLinks`) and the editor (`a[href^='exe-node:']`) only understand the `exe-node:` form — and there is **no reverse mapping** from `html/*.html` back to `exe-node:` anywhere in the import pipeline — internal page links were **silently degraded on every export → re-import round trip**. They keep resolving until the target page is renamed/moved/reordered, then break in every subsequent export.

This is the same class of issue already fixed for `exe-package:elp` (moved to render time). The asset half of preprocessing (`{{context_path}}/content/resources/…`) **is** reversed on import, confirming the asymmetry — only the internal-link rewrite leaked into the persisted XML. SCORM 1.2/2004 + EPUB3 dodged it because they build `content.xml` from raw `getContentXml()`.

## Fix

Move the internal-link rewrite out of preprocessing and into **render time**, so it runs on the rendered HTML output and never mutates the source that feeds `content.xml`.

- **`PageRenderer`** — new `replaceInternalLinks()` (multi-page) reuses the existing `getPageLink()` helper, so internal links and navigation links share one source of truth. New `namespaceSinglePageAnchors()` + `replaceSinglePageInternalLinks()` handle single-page export (in-page anchors). Applied in `renderPageContent()` (multi-page) and `renderSinglePage()` (single-page).
- **`BaseExporter.preprocessPagesForExport()`** — no longer rewrites `exe-node:`; keeps only the XML-safe asset-URL rewrite. The dead `buildPageUrlMap` / `replaceInternalLinks` helpers are removed (single source of truth).
- **`PageExporter`** — drops its source-mutating overrides; the single-page rewrite now happens at render time too.

Net effect: `content.xml` retains `exe-node:` links (and anchors); exported HTML pages still resolve to the correct static paths / in-page anchors. ELPX, HTML5 and IMS are fixed; SCORM/EPUB are unaffected for `content.xml` and keep correct HTML links via the shared render path.

## How to reproduce / test

- [ ] Create a project with two pages, **Home** (index) and **About**.
[...]

PR #1934, opened 2026-06-10
fix(assets): recover MIME/extension for extension-less assets to stop PDF auto-download

### LINKED ISSUE #1933: PDFs embedded in iframes auto-download repeatedly in the static editor (assets exported without file extension)
## Summary

In the **static PWA editor**, a PDF embedded in a page is downloaded to the user's Downloads folder **repeatedly / on every render** instead of being displayed, when the asset was stored/exported **without a file extension**.

Reported by @greynoa.

> **Scope (from a real affected project):** the package carried **48** extension-less `asset-<uuid>` resources — PNG, JPEG, MP3, ZIP, an embedded HTML site, and the PDF. All are recoverable from their content, so the fix sniffs magic bytes rather than special-casing PDF.

## Repro

1. Open (or import a `.elpx` of) a project that embeds a file via an `<iframe>` whose `src` points to an extension-less asset, e.g.:
   ```html
   <iframe src="{{context_path}}/content/resources/asset-e9e79be2-7b98-3e8c-0143-91e790c196f8"></iframe>
   ```
   The referenced resource is a real PDF (`%PDF-1.4`) but the file in `content/resources/` has **no extension** (`asset-<uuid>`).
2. Open the project in the static editor.
3. The PDF is downloaded automatically, again and again as the page re-renders.

Assets that kept their original filename render correctly; only the extension-less ones break.

## Root cause

The viewer decides a resource's type **solely from its file extension**:

- The asset's stored `filename` was lost and fell back to `asset-<uuid>` (no extension), so its derived MIME is `application/octet-stream`.
- In the editing view, the `asset://<id>` iframe resolves to a blob whose `type` is `octet-stream`; an `<iframe>` loading `octet-stream` is **downloaded** by the browser instead of rendered.
- In the Preview panel, the service worker maps unknown extensions to `application/octet-stream`, and its PDF handling / embed detection only act on `.pdf` / `application/pdf`.

Because the embed reloads on every render, the download repeats continuously.

## Proposed fix

1. **Detect type by content, not extension** — a shared magic-byte sniffer (`%PDF-`, image/zip/media/html signatures).
2. **Normalize on ingest** — when an asset's MIME is unknown/`octet-stream`, sniff the bytes, set the correct MIME, re-type the blob, and append the correct extension to the filename.
3. **Self-heal on open** — repair extension-less / `octet-stream` assets when opening a project (and after `.elpx` import), so existing broken projects are fixed automatically.
4. **Harden the Preview SW** — correct `Content-Type` by sniffing the in-memory bytes.
5. Stop persisting/exporting extension-less filenames (`asset-<uuid>` → `asset-<uuid>.<ext>`).

## Note

This is **not** related to PR #1930 (which only moves the rewrite of internal `exe-node:` page links to render time).

erseco:
Fixes #1933

## Problem

In the **static PWA editor**, a PDF embedded in a page is downloaded to the user's Downloads folder **repeatedly / on every render** instead of being displayed, when the asset was stored/exported **without a file extension**. Reported by @greynoa.

Confirmed with a real `.elpx`: the PDF is embedded via `<iframe src="{{context_path}}/content/resources/asset-<uuid>">` where the resource (a real `%PDF-1.4`) has **no extension**. The viewer decides a resource's type **solely from its extension**, so an extension-less PDF resolves to `application/octet-stream`; an `<iframe>`/`<object>` loading octet-stream is **downloaded** by the browser instead of rendered, and re-fires on every render. The same project carried **48** extension-less resources (PNG, JPEG, MP3, ZIP, an embedded HTML site and the PDF) — all recoverable by content.

Not related to #1930 (which only moves the rewrite of internal `exe-node:` page links to render time).

## Fix

Detect type by **content (magic bytes)**, not extension, and recover the MIME + filename extension + blob type so embeds render inline:

- **`public/app/common/mime-sniff.js`** (new) — `sniffMimeFromBytes` / `extensionForMime` for PDF, PNG, JPEG, GIF, WEBP, ZIP, MP3, MP4, OGG, WAV, SVG and HTML. IIFE that exposes `window.eXeMimeSniff` in the browser and `module.exports` under Vitest (no top-level globals leak into the shared classic-script scope).
- **`AssetManager`** — normalizes MIME/extension on ingest (`putAsset`/`putBlob`): when the MIME is missing/`octet-stream`, sniff the bytes, fix the MIME, re-type the blob (so embeds render inline) and append the extension to the filename. New **`repairAssetsWithoutType()`** self-heals already-broken projects; it runs on project open and after import.
- **`preview-sw.js`** — corrects `Content-Type` by sniffing the in-memory bytes when the extension is unknown.
- **`BaseExporter`** — appends the MIME extension to extension-less export filenames so future exports are not lossy.
- **`yjs-loader.js`** — loads `mime-sniff.js` (Group 0) before `AssetManager.js` so `window.eXeMimeSniff` is ready for the classic-script load.

## Tests

- **Unit (RED→GREEN):** `mime-sniff.test.js`, `AssetManager.test.js` (putAsset/putBlob recovery + `repairAssetsWithoutType` incl. idempotency), `preview-sw.test.js` (sniffer + `resolveServedMime`), `BaseExporter.spec.ts` (extension-less filename handling).
- **E2E:** `test/e2e/playwright/specs/pdf-noext-iframe-no-download.spec.ts` imports an extension-less-PDF `.elpx` fixture and asserts the asset is recovered to `application/pdf` (with `.pdf`) and that nothing auto-downloads. Passes in `chromium` and `static`.
- Fixture generated by `test/fixtures/scripts/build-pdf-noext-fixture.mjs`.

## Verification

- `make fix` clean (one pre-existing `noExplicitAny` warning in `src/routes/export.ts`, untouched).
- Frontend `vitest` — 13103 pass. Backend `bun test` — 6917 pass (the only failures are pre-existing local-env DB-path errors in API-v1/Config route suites, unrelated to this change).
- New E2E spec passes (chromium + static).


ignaciogros (approved):

Issue #1932, opened 2026-06-10
Collaborative editing: project stalls on simultaneous open, and edits can silently disappear across sessions


erseco:
> Reported by @greynoa and @montesdeocaleocadia-sys, based on feedback from teams using the
> real‑time collaborative editor on a hosted cloud environment. Filed after reproducing both issues.

## Summary

Teams working on the same project report two recurring problems:

- **A — "It never finishes loading."** When more than one person opens the **same** project at the
  same time, the editor often gets stuck on the loading screen / takes very long to open. It used to
  tolerate concurrent users; lately it frequently does not.
- **B — "Content disappeared from one day to the next."** Despite always saving (and periodically
  downloading backups), parts of a project come back missing in a later session, forcing re‑import of
  sections from older copies.

Both reproduce. Details, root cause, and an assessment of whether **#1896** fixes them below.

## Environment

- Cloud: a hosted cloud environment, an internal alpha build (built 2026‑06‑09). This build
  **predates #1896** (still open).
- Two accounts, two isolated browser sessions (separate IndexedDB, i.e. two real machines).
- Baseline: a single user opens the project cleanly in ~6 s.

## Symptom A — stalled / very slow load on simultaneous open

**Steps**
1. User 1 creates a project, adds a page + a text iDevice, saves, and shares it (public).
2. User 1 and User 2 open the **same** project URL **at the same instant**, in separate sessions.
3. Observe each client's time to a usable editor (loading screen gone + document synced).

**Observed (cloud, 3 runs, 30 s cap; baseline ≈ 6 s):**

| Run | User 1 | User 2 |
|---|---|---|
| 1 | 7.9 s | 12.6 s |
| 2 | 21.7 s | **> 30 s** (loading screen still up; WebSocket not yet connected) |
| 3 | 6.0 s | 9.9 s |

Load time degrades severely and unpredictably under concurrent open (8–22 s typical, one client
≳30 s vs ~6 s baseline). The slow client had not reached a connected, synced WebSocket — the stall is
in the **client‑side initialization / sync handshake**, before any asset transfer. (In run 2 the
≳30 s client did eventually render, so what is confirmed here is *severe intermittent slowdown*, which
users experience as "it never loads"; a permanent hang is plausible from the same code path but was
not captured in 3 runs.)

**Likely root cause** (`public/app/yjs/YjsDocumentManager.js`):
- The initial full‑sync wait resolves *silently* after a 3 s timeout, and `wsProvider.once('sync')`
[...]

PR #1936, opened 2026-06-10
fix(yjs): collaborative editing — cross-day data loss + full-sync wait hardening (#1932)

erseco:
## Summary

Addresses #1932 (collaborative editing reliability). Two parts:

- **Part 1 — cross-day data loss (fix):** cherry-picks the self-contained persistence fix from
  #1896 (commit `3d37fa0b`, "fix(yjs): correct document mutex and unify persistence version scheme").
- **Part 2 — stall on simultaneous open (hardening):** makes the collaborative full-sync wait robust
  against a missed-sync-event race.

## Part 1 — data loss (cherry-picked from #1896)

Two deterministic backend mechanisms caused committed content to disappear across sessions:

1. **`saveFullState` stamped every write `version:'1'`.** On load, `loadDocumentWithUpdates` keeps only
   updates with `version > snapshot_version`, and a snapshot carries a `Date.now()` millisecond
   version, so `1 > 1.7e12` is false and the write was **silently discarded** on the next load.
2. **Broken per-project document mutex.** A single `release()` resolved one shared promise that *all*
   waiters were `Promise.race`-ing, so multiple concurrent writers each believed they held the lock —
   allowing interleaved writes to the same project.

Verified by A/B before cherry-picking: on `main` these reproduce deterministically (5 concurrent lock
acquirers → up to **4** simultaneous holders; a `saveFullState` write returns **0 rows** on reload),
and are fixed after (1 holder; write retained). The commit ships its own backend tests
(`doc-lock.spec.ts`, `db/queries/yjs.spec.ts`, `routes/yjs.spec.ts`).

> Note: this part overlaps #1896 (same author/commit). If #1896 merges first, drop this commit on
> rebase; kept here so #1932's data-loss half is self-contained.

## Part 2 — full-sync wait hardening

`YjsDocumentManager._waitForFullSync` used `wsProvider.once('sync', cb)` and only resolved on
`sync(true)`. y-websocket also emits `sync(false)` on a reconnect/handshake blip; `once` consumed that
first event and removed the listener, so a joining client then **waited out the entire
`fullSyncTimeout`** even though a real `sync(true)` followed. It could also miss a sync that completed
between the early `synced` check and attaching the listener.

Replaced with a **persistent** `on('sync')` listener that ignores `sync(false)`, resolves on the first
`sync(true)` or the timeout, **detaches itself in every path**, and re-checks `synced` after attaching
to close the race window.

## Investigation note (scope honesty)

With the correct collaborative flow (make the project public in Share, then open via the **share
link**), a concurrent joiner **does** load. For a large project (13 MB → 9 pages; 30 MB → 50 pages) it
takes ~20 s — dominated by app bootstrap (~12 s) and asset processing, **not** a Yjs sync hang. Some
earlier "never loads" symptoms trace to an access artifact (opening before the public/share state had
propagated → "access denied"), not a product hang.

So this PR ships the definitive **data-loss fix** plus a real **latent-race hardening**; it does not
add an invasive load-path change, because no hard hang reproduced via the correct flow. Deeper
concurrent-load latency for large projects (app bootstrap + asset processing) is left as a follow-up.

## Tests

- Backend (Part 1): document-mutex mutual-exclusion (FIFO) + persistence version-scheme specs — green.
- Frontend (Part 2): `YjsDocumentManager._waitForFullSync` — resolve on `sync(true)` with listener
  cleanup; **ignore a `sync(false)` blip then resolve on `sync(true)`** (regression); resolve on
  timeout with cleanup; already-synced / no-provider fast paths. 205 tests green in the suite.
- `make fix` clean (one pre-existing lint warning in an unrelated admin-route file).
- `make test-unit`: **7136 unit tests pass, 0 fail.** Its strict *global* coverage gate flags a
  pre-existing gap in `src/websocket/asset-coordinator.ts` (88.20%) — a file **not touched by this
  PR** (restored separately by #1896's test-only commit `8a0f6a03`). Patch coverage on this PR's
  changed lines is met.


ignaciogros (approved):

PR #1942, opened 2026-06-12
fix(assets): context-independent asset id hash to stop collaborative image loss (#1941)

### LINKED ISSUE #1941: Collaborative save loses all images (corrupted asset UUIDs `--` + silent export degradation) — reported by @USSEFXBEN
## Summary

In a **desktop → online collaborative** flow, **all images disappear** from a project after saving. In the editor and in exported packages the images are replaced by dangling URLs like `{{context_path}}/content/resources/<uuid>` (the user sees literal `context/path…` paths), and the `content/resources/` binaries are **missing entirely** from the export.

Reported by **@USSEFXBEN**: created a project in **local/desktop** eXeLearning, opened it in an **online** install to collaborate, added **image descriptions** and **iDevices** (a presentation + a video iframe), saved (v5 → v6) and every image vanished — both online and locally.

- **Environment:** v4.0.0-beta2, **Windows**, desktop app, collaborating against an online instance.
- **Severity:** data loss, silent. Should be **bug / high priority**.

## Evidence (user's two ELPX: v5 good, v7 broken)

| | v5 (good) | v7 (broken) |
|---|---|---|
| image `src` | `{{context_path}}/content/resources/2.png` | `{{context_path}}/content/resources/5947b408-…` (raw UUID, **no extension**) |
| binaries in ZIP | 9 images under `content/resources/` | **directory absent** (0 image files) |
| some ids | resolved to real names | **3 of 8 malformed**: `4e47ddc2-cbf7-7ede--7a4-…` (double dash) |

`{{context_path}}` is a normal placeholder; "context/path" is just that placeholder with no image behind it.

### 1. Signed-XOR corrupts asset UUIDs — `public/app/yjs/AssetManager.js:686`

Asset ids are content-derived (`hashToUUID(calculateHash(blob))`). In a **non-secure context** (HTTP-on-IP / some desktop contexts) `crypto.subtle` is unavailable and `calculateHash()` uses a JS FNV fallback. One term was not forced unsigned:

```js
(hash ^ sizeHash).toString(16).padStart(8, '0'),   // ← missing >>> 0
```

JS `^` returns a **signed Int32**; when bit 31 is set the value is **negative**, so `.toString(16)` emits a leading `-`. That `-` lands in the **4th UUID group** (`hashToUUID` uses `substring(16,20)`), producing ids like `…-7ede--7a4-…`. Those ids no longer match the stored blobs, so resolution fails and the images are dropped.

Reproduced deterministically: bytes for the string `"exe-0"` (`[101,120,101,45,48]`) force the negative branch and yield `f266d980-1715-6075--1a8-c460bf266d9e`. **Present in the `v4.0.0-beta2` tag (`AssetManager.js:372`) and still in `main` (`:686`).** Fix: `((hash ^ sizeHash) >>> 0)`.

### 2. Context-dependent hash scheme — `AssetManager.js:659-694`

Secure context → SHA-256; non-secure → FNV. **The same image gets a different id depending on the client/context**, so when a desktop-created project (SHA ids) is opened/re-hashed online (FNV ids), the content references and the stored asset ids diverge and stop resolving.

### 3. Silent export degradation — `src/shared/export/exporters/BaseExporter.ts:751-754` and `:508-541`

When an `asset://<uuid>` cannot be resolved, the exporter leaves the **raw UUID (no extension)** in `src` (`:751-754`) and **skips writing the binary** (`:508-541`) — **with no error surfaced to the user**. That is why the loss is silent and gets persisted to both copies.

## Why existing PRs do not fix it

- **#1936** (collab data loss): persistence/version-scheme + mutex only; explicitly does **not** touch assets/blobs or the hash.
- **#1934** (extension-less assets): recovers by magic bytes **only if the blob is present**; here the blobs are gone.
- **#1896**: same persistence fix as #1936.

None touch `AssetManager.js:686`, the dual hash scheme, or the silent export fallback. Related prior report: #711 ("Images lost in preview", closed).

## Fix in progress (branch `fix/asset-uuid-fnv-signed-xor`)

TDD, RED→GREEN:

- ✅ **UUID corruption:** `((hash ^ sizeHash) >>> 0)` + regression tests in `AssetManager.test.js` (asserts the FNV-fallback hash is 64 lowercase hex and the derived UUID is well-formed, no `--`).
- ✅ **Export integrity (detection):** `BaseExporter` now records unresolved UUID asset refs (`getUnresolvedAssetRefs()`) so the save/export flow can surface dangling references instead of losing data silently; tests in `BaseExporter.spec.ts`.
- ⏳ **Context-independent hashing** (so secure/non-secure produce the same id), surfacing unresolved assets in the save UI + server blob recovery, malformed-id detection/repair on open, and an E2E asset-integrity spec.

## Recovery for affected users

The lost blobs survive **only in the older good package (v5)** — the v7 has no `content/resources/`. Practical recovery: **re-import the images from the v5 file** (or, if the online instance still has them, from `FILES_DIR/assets/{uuid}/`). The fix prevents recurrence; it cannot magically rebuild the already-stripped v7.

erseco:
Fixes #1941

## Problem

A user (reported by @USSEFXBEN, v4.0.0-beta2, Windows desktop) created a project locally, opened it in an **online collaborative cloud reached over a bare IP** (`http://<IP>`, no TLS), added image descriptions + iDevices, saved — and **all images disappeared**, both online and locally. In the package the images became `{{context_path}}/content/resources/<uuid>` (no extension, the user sees `context/path…`) with the `content/resources/` binaries **missing entirely**; several asset ids were malformed like `4e47ddc2-cbf7-7ede--7a4-…` (double dash).

## Root cause

The asset id is content-derived: `hashToUUID(calculateHash(blob))`. `calculateHash()` used `crypto.subtle` SHA-256 in **secure** contexts and an **FNV fallback** when `crypto.subtle` is unavailable — which only happens in a **non-secure context** (`http://<IP>` without TLS; desktop `app://`, `localhost` and HTTPS clouds are all secure). The fallback had two defects:

1. **Divergence:** FNV ≠ SHA-256, so the *same image* gets a *different id* depending on the context. A project created on desktop/HTTPS (SHA-256 ids) and edited on an IP-served cloud (FNV ids) ends up with content references that no longer match the stored assets.
2. **Corruption:** the term `(hash ^ sizeHash).toString(16)` was not forced unsigned (`>>> 0`). JS `^` returns a signed Int32; with bit 31 set the value is negative and `.toString(16)` emits a leading `-` that lands in the 4th UUID group (`hashToUUID` uses `substring(16,20)`), producing ids like `…-7ede--7a4-…`. Reproduced deterministically with the bytes for `"exe-0"`.

When an `asset://` reference cannot be resolved, `BaseExporter` silently left the raw uuid in `src` and skipped the binary — so the loss happened with **no warning** and was persisted.

Verified present in the `v4.0.0-beta2` tag and in `main`; not fixed by #1936 (persistence only), #1934 (recovers extension-less only when the blob is present) or #1896.

## Fix

- **Keep `crypto.subtle` SHA-256 (native path unchanged)** and replace **only the non-secure FNV fallback** with a **pure-JS SHA-256** (`AssetManager._sha256Hex`) that returns the *identical* digest. An asset's content-derived id is now the **same in every context** → no divergence, **no duplicates**, **zero id churn** for existing projects. The signed-XOR corruption disappears with the FNV block.
- **Defense in depth:** `BaseExporter.getUnresolvedAssetRefs()` records `asset://` references that don't resolve to a bundled file, so a dangling `content/resources/<uuid>` URL is **detectable** instead of being dropped silently.

No new dependency; the server never re-hashes/compares the client hash (`folder-manager.ts` uses `crypto.randomUUID`; `findAssetByHash` is never called), so SHA-256 client-side is sufficient and consistent.

## Tests (TDD, RED→GREEN)

- `public/app/yjs/AssetManager.test.js`:
  - the non-secure fallback now produces a real SHA-256 (known vector `"abc"`);
  - the **same hash/id with and without `crypto.subtle`** — i.e. **no duplicate** when a project moves between desktop/HTTPS and a non-secure HTTP-on-IP cloud;
  - the fallback never emits `-` and derives a well-formed UUID (regression for the old `--` corruption; trigger bytes `"exe-0"`).
- `src/shared/export/exporters/BaseExporter.spec.ts`: unresolved asset refs are recorded; resolved refs are not.

## Verification

- `make fix` clean (one pre-existing `noExplicitAny` warning in `src/routes/export.ts`, untouched).
- `make test-unit`: **frontend 13072 pass / backend 7134 pass**, 0 fail.
- `make test-integration`: **709 pass**, 0 fail.
- E2E runs in CI (the change affects only the non-secure fallback, which Playwright on `localhost` cannot reach, and an internal export-tracking method with no output change in secure contexts).

## Ops note

eXeLearning should be served over HTTPS (or `localhost`); a bare `http://<IP>` deployment is a non-secure context. A follow-up could surface an explicit in-app warning when running without a secure context.

## Recovery for already-affected projects

The lost binaries survive only in the older good package (e.g. the reporter's v5) — re-import the images from it (or from the server's `FILES_DIR/assets/{uuid}/` if still present). This fix prevents recurrence; it cannot rebuild an already-stripped package.


ignaciogros (approved):

PR #1954, opened 2026-06-16
Fix duplicate asset removal for self-contained HTML bundles

### LINKED ISSUE #1951: Duplicate asset removal breaks self-contained HTML bundles (e.g. Hype exports)
When adding/unzipping multiple HTML bundles (e.g. Tumult Hype exports), eXeLearning's asset system removes identical files (like HYPE runtime JS files).
This causes missing files in the final output and breaks embedded content (iframes fail with 404 errors).
These bundles rely on:

- their own folder structure
- duplicate runtime files per instance
- relative paths staying intact

Because of this removal, only one copy is kept and others are not correctly available where expected.

**Steps to reproduce**

- Export multiple Hype projects (HTML + .hyperesources)
- add into eXeLearning file manager
- Embed via iframe 'text' idevice
- Preview/export

**Result**

- Missing runtime files (HYPE-*.js)
- Console 404 errors
- Only one instance works, others fail

**Expected**
Each imported HTML bundle should work independently with its own required files.

**Note**
This isn’t about storage efficiency — it breaks valid self-contained web exports.
There needs to be a way to preserve folder structure / treat these imports as isolated bundles instead of removing their assets. Perhaps in the file manager allow an option on a choosen folder to not remove duplicate files.

erseco:
## Summary

Fixes #1951.

Self-contained HTML bundles such as Tumult Hype exports rely on their own relative folder structure: each `index.html` resolves its runtime and resources through **path-relative** URLs from its own folder. Two byte-identical files (e.g. `HYPE-runtime.js`) living in different bundle folders must therefore be preserved as **separate** assets — collapsing them by content hash breaks one bundle's links.

ZIP extraction (`ModalFilemanager.extractZipAsset`) inserted every entry through `assetManager.insertImage(file, { folderPath })` **without** `forceNewId`. So the second identical file resolved to the same content-addressed asset ID, found the existing asset, and **overwrote its `folderPath`** — collapsing both bundles into one asset at a single path and breaking the other bundle.

The fix passes `forceNewId: true` during ZIP extraction (the same approach already used by `duplicateSelectedAsset`), so each extracted entry becomes its own asset. Ordinary image/file uploads keep content-hash deduplication unchanged.

The export side already handles this correctly: `BaseExporter.buildAssetExportPathMap()` keys by asset id, builds paths from `folderPath/filename`, and resolves same-path collisions with a counter — so two distinct IDs with distinct folderPaths are both written to `content/resources/<folderPath>/<filename>`. No exporter change was needed.

## Tests

- `AssetManager.insertImage`: byte-identical files at different paths with `forceNewId` produce two distinct asset IDs with the same content hash but distinct `folderPath`; ordinary inserts without `forceNewId` still dedupe by hash.
- `ModalFilemanager.extractZipAsset`: a synthetic ZIP with two bundles containing identical `index.hyperesources/HYPE-runtime.js` files asserts every entry is inserted with `forceNewId: true` and that the two runtimes keep their distinct bundle folder paths.
- No copyrighted Hype exports added — fixtures are small synthetic byte arrays built in-test.
- `make fix` clean for changed files; full `modalFileManager` + `AssetManager` vitest suites green (1300 tests).

## How to test

Two ready-made sample ZIPs are attached to this PR (synthetic, no copyright — the "runtime" is a tiny script that prints a green line when it loads):

- [hype-two-bundles.zip](https://github.com/user-attachments/files/28994959/hype-two-bundles.zip) - `bundle-a` and `bundle-b`, each with its own `index.html` + a **byte-identical** `index.hyperesources/HYPE-runtime.js`.
-  [hype-two-bundles-with-image.zip](https://github.com/user-attachments/files/28994957/hype-two-bundles-with-image.zip) - same, plus a byte-identical `poster.png` in each bundle (shows it also covers images, not just JS).

Steps:

1. Download one of the attached ZIPs from this PR.
2. Open a project → toolbar **File manager** → **Upload** → pick the ZIP.
3. Select the uploaded ZIP in the grid → **More → Extract ZIP** → confirm (any target folder, or root).
4. **Check both copies survived** (this is the fix): open `bundle-a/index.hyperesources/` and `bundle-b/index.hyperesources/` in the file manager — **each** must contain its own `HYPE-runtime.js` (and `poster.png` for the second ZIP).
   - On `main` (before this PR), only one folder keeps the file; the other is empty, because both collapse into a single content-addressed asset.
5. **Optional end-to-end:** export the project as **HTML5**, unzip the result, and open `content/resources/bundle-a/index.html` and `.../bundle-b/index.html` in a browser. Each should show the green "✅ runtime loaded…" line, proving its relative `<script src="index.hyperesources/HYPE-runtime.js">` resolved. Both `content/resources/bundle-{a,b}/index.hyperesources/HYPE-runtime.js` must be present in the export.


erseco:
Resolved the 6 `github-code-quality` (CodeQL) inline comments on `modalFileManager.test.js` (lines 1455–1470) as **false positives**.

They all report the same thing — *"Property access on null or undefined — the base expression is always undefined"* on `modal` / `modal.assetManager`. The analyzer doesn't model Vitest's `beforeEach` lifecycle, so it assumes `modal` is never initialized. In practice `modal` is created in `beforeEach` (`modal = new ModalFilemanager(...)`, `modal.assetManager = ...`), the new test passes, and it uses the **exact same pattern** as the pre-existing `should extract files from ZIP and reload assets` test in the same file (which isn't flagged only because it's outside this PR's diff).

I deliberately did **not** add the suggested `modal.assetManager = modal.assetManager || {}` guard: it would be inconsistent with every other test in the file and could mask a genuine setup regression, so it's noise rather than a real defect.

erseco:
## Review — correct for the primary path, one important round-trip gap

> **Update:** the two gaps below (`.elpx` round-trip + drag-and-drop) are now **resolved in this PR** — see the "Resolved in follow-up commits" section at the bottom.

Validated against the real `hypes.zip` from #1951 (two bundles sharing 5 byte-identical files: `HYPE-772.full.min.js`, `HYPE-772.thin.min.js`, `PIE.htc`, `blank.gif`, and the shared `.jpg`).

**What works**
- Root cause is right: `extractZipAsset` inserted each entry with content-hash dedup, so the 2nd identical file resolved to the existing asset and **overwrote its `folderPath`**, collapsing both bundles into one path.
- `forceNewId: true` is the correct minimal fix and mirrors `duplicateSelectedAsset`. `__MACOSX`/dotfiles are still skipped.
- Export side is unaffected: `BaseExporter.buildAssetExportPathMap()` keys by asset id, so two distinct ids with distinct `folderPath`s yield two distinct `content/resources/.../` paths (the collision counter only fires on identical path+name).

**Gaps found in review — now resolved (see below)**

1. ~~**`.elpx` round-trip re-collapses the bundles.**~~ `AssetManager.extractAssetsFromZip()` still used `hashToUUID(hash)` + skip-if-exists, so a fixed project exported to `.elpx` and re-opened lost the duplicate again — the source of the reporter's actual 404. **Fixed.**
2. ~~**Drag-and-drop / upload into a folder still dedups.**~~ `uploadFiles()` had no `forceNewId`, so a byte-identical file dropped into a different folder overwrote the first one's `folderPath` ("files disappeared after preview/save"). **Fixed.**

---

## Resolved in follow-up commits

**Fix A — ELPX import round-trip** (`AssetManager.extractAssetsFromZip`)
Track the content hashes seen within a single import. The **first** occurrence of a hash keeps the content-addressed id (preserving same-project skip and cross-project blob reuse); any **subsequent** occurrence of the same hash in the same import gets a fresh `crypto.randomUUID()` and its own `folderPath`. Safe because import reconciles references **by path** (`assetMap` is `Map<path,id>`), and bundle resources are referenced only by relative links inside the stored `.html` asset — never by `asset://` in `content.xml`. A project fixed by this PR now survives a `.elpx` export → reopen.

**Fix B — drag-and-drop / upload** (`ModalFilemanager.uploadFiles`)
Opt out of content-hash dedup **only for sub-folder uploads** (`forceNewId: Boolean(currentPath)`). The damaging `folderPath` overwrite only happens for non-root uploads (at root the `folderPath` is empty, so `insertImage` never rewrites it), so root uploads keep dedup — preserving **cross-project content reuse** — while files dropped into named folders stay distinct. `insertImage`'s default behavior is unchanged.

> The first revision used a blanket `forceNewId` on every upload, which regressed the cross-project dedup e2e (`file-manager.spec.ts:975`). Caught by CI and narrowed to subfolder uploads in `ce832ece6`.

**Tests**
- `AssetManager.test.js`: 3 new import-dedup tests (2 fail on pre-fix code — verified meaningful).
- `modalFileManager.test.js`: updated `uploadFiles` tests (root keeps dedup, subfolder forces new) + a dual-folder regression test.
- **E2E** `test/e2e/playwright/specs/hype-bundle-dedup.spec.ts`: File manager → upload synthetic 2-bundle ZIP → Extract, then asserts both bundles keep their own `HYPE-runtime.js` (distinct ids, distinct `folderPaths`, identical bytes). Green on chromium + firefox.

**Verification:** `make fix` clean; full frontend unit suite green (13118 tests); `file-manager.spec.ts` (incl. cross-project dedup) + `hype-bundle-dedup.spec.ts` green on chromium; cross-project test re-confirmed green after the Fix B narrowing.

*(Resolved in commits `3a487071d` and `ce832ece6`, pushed to this branch.)*

pabloamayab (approved):
LGTM. 
I followed the five steps described for testing, and everything is fine.

Image

Image

PR #1955, opened 2026-06-16
fix(magnifier): preserve exported image path so HTML export shows the image (#1953)

### LINKED ISSUE #1953: Magnifier iDevice fails to load image after HTML export ("Image not available")
Description
Environment
eXeLearning version: 0.0.0-alpha-build202606091131
Export format: Website (HTML5)
Browser: Google Chrome (also tested in Firefox)
OS: macOS Sequoia (same behaviour reported on Windows)
Problem

The Magnifier (Lupa exploradora) iDevice works correctly inside the eXeLearning editor, but after exporting the project as HTML5 the image is not displayed.

Instead, the iDevice shows the message:

Image not available.

The audio and the rest of the iDevice load correctly, only the image fails.

Steps to reproduce
Create a new Magnifier iDevice.
Select an image from the resource manager.
Save the project.
Export as Website (HTML5).
Open index.html.
Expected behaviour

The image should be displayed and the magnifying glass effect should work normally.

Actual behaviour

The exported page displays:

Image not available.

The image is not loaded.

Possible cause

After inspecting the exported files, it appears that the JavaScript (magnifier.js) rewrites the image path by inserting the iDevice ID as an intermediate directory.

For example, the exported JSON contains:

"imageResource":"../content/resources/img_animales.jpg"

but changeDirectory() transforms it into something like:

../content/resources/idevice-1780487482081-mnjl9i39r/img_animales.jpg

However, the exported image is actually stored as:

content/resources/img_animales.jpg

so the generated path does not exist and the image fails to load.

This seems to be a path rewriting issue inside the Magnifier iDevice.

Additional information

I have attached screenshots showing:

The iDevice working correctly inside the editor.
The exported HTML showing "Image not available".

If needed, I can also provide the generated ZIP package for testing.

Image
Image

mnarvaezm:
The Magnifier iDevice rendered fine in the editor but showed "Image not available" in HTML5/SCORM/IMS/EPUB exports (Fixes #1953).

The exporter already rewrites the iDevice's `imageResource` to its final exported location (`content/resources/<filename>` from `index.html`, `../content/resources/<filename>` from subpages). At runtime, `changeDirectory()` discarded that correct path, kept only the basename, and rebuilt it as `content/resources/<ideviceId>/<filename>`. The exporter never creates that per-iDevice folder, so the image 404'd.

## Fix

Align `changeDirectory()` with the image-gallery path handling pattern:

- keep `blob:`, `data:`, and `asset://` URLs untouched
- preserve already-rewritten `content/resources/...` paths
- apply the correct page-relative prefix: `content/resources/...` from `index.html`, `../content/resources/...` from subpages

This removes the invalid `ideviceId` subfolder from exported image paths.

### Manual reproduction (issue #1953 flow)
1. Create a new **Magnifier** iDevice.
2. Select an image from the resource manager.
3. Save the project.
4. Export as **Website (HTML5)**.
5. Unzip the package and open `index.html` (and a subpage under `html/`).
6. ✅ The image is shown and the magnifier works (before: "Image not available").

### DIFF
--- a/public/files/perm/idevices/base/magnifier/export/magnifier.js
+++ b/public/files/perm/idevices/base/magnifier/export/magnifier.js
@@ -82,25 +82,41 @@ var $magnifier = {
 
     changeDirectory(data) {
         const $node = $('#' + data.ideviceId),
-            isInExe = eXe.app.isInExe();
-        if (isInExe || $node.length == 0 || !data.imageResource)
-            return data.imageResource;
+            isInExe = eXe.app.isInExe(),
+            file = data.imageResource;
 
-        // Don't transform blob:, data:, or asset:// URLs - they're already valid or will be resolved
-        if (data.imageResource.startsWith('blob:') ||
-            data.imageResource.startsWith('data:') ||
-            data.imageResource.startsWith('asset://')) {
-            return data.imageResource;
+        if (isInExe || $node.length == 0 || !file) return file;
+
+        // Don't transform blob:, data:, or asset:// URLs - they're already
+        // valid display URLs or will be resolved by the asset resolver.
+        if (
+            file.startsWith('blob:') ||
+            file.startsWith('data:') ||
+            file.startsWith('asset://')
+        ) {
+            return file;
         }
 
-        const pathMedia = $('html').is('#exe-index')
-            ? 'content/resources/' + data.ideviceId + '/'
-            : '../content/resources/' + data.ideviceId + '/';
+        // The exporter already rewrites the stored asset URL to its final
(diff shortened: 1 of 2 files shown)


erseco (approved):
The exporter writes the **correct** path into the iDevice JSON (`content/resources/<file>`) and the file is placed there. The failure was purely at **runtime**: the old `changeDirectory()` rebuilt the path and injected a
phantom `<ideviceId>/` folder that the export never creates → 404 → "Image not available".

Image
Image

PR #1956, opened 2026-06-16
fix(3dmol): support molecule files in the file manager and viewer

### LINKED ISSUE #1946: Regarding the new iDevice 3DMol in new eXeLearning 4.0.1
Regarding iDevice 3DMol, I just noticed this bug in the recently released version 4.0.1. I've seen the bug in both the offline version for GNU/Linux and the online demo (https://exeleaning.online.net), as well as in an installation of the online version on my server, and you can see it in the attached videocaptures:

Part one
https://ismailaligago.es/nextcloud/index.php/s/2nYKRHt2on6mJEr

Part two:
https://ismailaligago.es/nextcloud/index.php/s/Qm68bykLoARr4eS

So long!

mnarvaezm:
The 3Dmol iDevice could not use molecule files reliably. This PR fixes the upload → select → render → switch flow and aligns the file manager with the formats the iDevice actually supports.

Closes #1946

## Fixed issues

- **Uploaded molecules didn't appear in the file manager** (`.sdf`, `.pdb`,  `.mol2`, `.xyz`, `.cif`, `.mmcif`): the picker inherited the mesh-only  `accept='3d'` filter because its input id contains "model". The model input  now opts into its own `data-filemanager-accept="molecule"` filter, and the  picker gives priority to the explicit `data-filemanager-accept` value over  the id-based heuristic.
- **SDF files with an empty title line** are no longer trimmed before  rendering/export; empty checks still use `trim()`.
- **Switching to a second molecule loaded the previous one**: the change  handler read the stale jQuery `.data('blobUrl')` cache after the shared  picker had written the native `dataset.blobUrl`. The `blobUrl` is now read  and cleared through the native dataset.
- **Blob URL persisted as the model reference**: it is now sanitized to recover  the canonical `asset://` reference, or discarded, as in `three-d-viewer`.
- **`.tar.gz` files are now rejected** while `.gz`/`.tgz`/`.zip` containers  remain supported, keeping the file manager and the iDevice parser aligned.

## Improvements
- The type filter separates **3D Models** (meshes: `three-d-viewer`) from  **3D Molecules** (3Dmol formats).
- Editing renders the model by resolving its asset reference, with a fallback  to legacy inline data; the bundled default model is persisted as a project  asset to follow the same `asset://` discipline.
- The file manager "Type" dropdown no longer shows unusable options (e.g.  "Images") when opened from 3Dmol — it lists only types selectable under the  active accept filter.

## Tests

- Unit tests for 3dmol editing, `modalFileManager` and `ideviceNode`.
- Playwright spec covering upload, render, SDF with an empty title, and model  switching.
- New molecule fixtures (`sample-molecule.sdf`, `sample-molecule-empty-title.sdf`).


public/app/workarea/modals/modals/pages/modalFileManager.js
```diff
@@ -1558,25 +1552,95 @@ export default class ModalFilemanager extends Modal {
...
+            'chemical/x-mmcif',
+        ]);
+        if (supportedMimes.has(m)) {
+            return true;
+        }
+        const name = filename || '';
+        if (/\.tar\.gz$/i.test(name)) return false;
+        return (
+            /\.(pdb|sdf|mol2|xyz|cif|mmcif)$/i.test(name) ||
+            /\.(zip|tgz|gz)$/i.test(name)
```
erseco: `isMoleculeAsset()` also feeds `getAssetTypeCategory()` and `getFileTypeLabel()`, which run across the **whole** media library, not only under `accept='molecule'`. Since any `.zip`, `.gz` or `.tgz` returns `true` here, a generic `backup.zip` or `archive.gz` now shows up labeled **"3D Molecule"** and grouped under the **"3D Molecules"** dropdown in the general file-manager view (they used to be `other` / `File`).

Inside the 3Dmol picker it's fine to offer compressed containers; outside that context it's a user-visible misclassification.

Suggestion: treat `.zip` / `.gz` / `.tgz` as a molecule **only** when the active filter is `molecule` (i.e. in `passesAcceptFilter`), and keep `getAssetTypeCategory` / `getFileTypeLabel` on pure chemical-MIME / molecular-extension detection. It would also be worth a test pinning the expected result for a generic `.zip` (`'other'` / `File`).

erseco (changes requested):
I reviewed the full diff, the affected sources, and ran the unit tests for the 4 changed files: **822 tests passing** (the `stderr` lines are error-path assertions, not failures).

Verified and correct:

- **File manager filter:** `data-filemanager-accept="molecule"` takes priority over the `id` heuristic — it leaves the `three-d-viewer` mesh filter untouched. ✔
- **Stale `blobUrl` cache:** reading `el.dataset.blobUrl` (native attribute) instead of `jQuery .data()` is the correct fix when switching molecules. ✔
- **SDF with an empty title line:** keeping the model data raw and applying `.trim()` only in the emptiness checks, consistently across edition and export. ✔
- **`blob:` sanitization:** `sanitizeModelPath` recovers the canonical `asset://` reference or drops it, mirroring `three-d-viewer`. I confirmed `AssetManager.insertImage()` does not validate that the file is an image, so reusing it for molecules is legitimate. ✔
- Solid unit + E2E coverage (including the molecule switch with a real atom count).

Just **one change** requested before merging (inline comment on `isMoleculeAsset`). Everything else 👍.

mnarvaezm:
Thanks, @erseco — fixed!

isMoleculeAsset() no longer treats .zip, .gz, or .tgz as molecules.

erseco (approved):
Thanks for the thorough work on this, @mnarvaezm! 🙏

Reviewed the full diff and ran the changed unit suites locally — all green (3dmol edition/export + modalFileManager + ideviceNode). The fixes are clean and well-structured:

- Dedicated `molecule` accept filter via `data-filemanager-accept`, with `passesAcceptFilter` centralizing the rule so the grid and the type dropdown stay consistent.
- Raw MDL molfile data preserved (emptiness checks still use `trim()`) so empty-title SDFs render correctly.
- Native `dataset.blobUrl` read/clear fixes the stale-model-on-switch bug.
- `blob:` URLs sanitized back to `asset://` on save, mirroring `three-d-viewer`.
- Consistent `.tar.gz` rejection across the parser and the file manager.

Great test coverage too — the unit regressions plus the WebGL-optional E2E spec are a nice touch. 👍

ismagago:
Thanks for the Improvement.

Now the 3DMol iDevice works properly here:

https://221191cf.exelearning.pages.dev/

here:

https://1946-regarding-the-new-idevi.exelearning.pages.dev/

and here:

https://online.exelearning.dev

But it doesn't work properly neither here:

https://online.exelearning.net 

nor in the offline version 4.0.1

Sure you will correct this soon.

Thanks a lot

Ismail Ali Gago

PR #1959, opened 2026-06-16
fix(three-d-viewer): hide empty-state overlay over a loaded model in export (#1957)

### LINKED ISSUE #1957: 3D Viewer displays “Select a 3D model to display” over an already loaded model in export
In the 3D Viewer iDevice (three-d-viewer), when the content is exported or viewed in the preview panel (/viewer/...), the 3D model loads correctly and is rendered, but the “empty state” layer (.viewer-empty) is not hidden.

That layer contains the text “Select a 3D model to display” and uses a semi-transparent gradient background. As a result, the model appears faded/washed out with an incorrect message over it, giving the false impression that no model has been selected.

This only happens outside edition mode. In the editor, the layer is hidden correctly because it uses a different runtime logic.

**Expected result**

The 3D model should be displayed clearly inside the viewer, without the empty-state layer and without the text “Select a 3D model to display”.

**Actual result**

The model is rendered, but it remains covered by the .viewer-empty layer, which adds a semi-transparent gradient and the text “Select a 3D model to display” on top of it.

Image

erseco:
## Summary

Fixes #1957. In exported content (and the preview panel), the **3D Viewer** iDevice kept the `.viewer-empty` layer — the semi-transparent gradient with the text *"Select a 3D model to display"* — on top of an already rendered model, so the model looked washed out behind a misleading message. This only happened outside edition mode; the editor was unaffected because it uses different runtime logic.

## Root cause

The export runtime's `ThreeDViewerRuntime.toggleEmpty()` decided overlay visibility from a narrow scheme allowlist:

```js
const hasValidSrc = src && (src.startsWith('blob:') || src.startsWith('http'));
const hasConfigSrc = this.config.src && this.config.src.startsWith('asset://');
```

- **Live editor / preview iframe:** the preview reaches the parent window's `AssetManager`, so the configured source stays an `asset://` handle, `hasConfigSrc` is true and the overlay is hidden.
- **Static HTML5 export:** there is no `AssetManager`, and `IdeviceRenderer.fixAssetUrls` rewrites the source to a **relative** `content/resources/<file>` path. That value starts with neither `blob:`/`http` nor `asset://`, so `hasModel` was `false` and the overlay stayed on top of the model. This is the bug.

STL was unaffected (the shared runtime hides the overlay after parsing) and edition was unaffected (it owns `[data-empty-state]` and keys off `state.src`).

## Fix

Decide overlay visibility from whether a model source is configured **at all** — the same single-source-of-truth rule edition already uses — instead of an allowlist of URL schemes. The decision is extracted into a pure, edge-case-tested helper:

```js
function computeEmptyStateDisplay(configSrc, viewerSrc) {
    const hasModel = !!(
        (configSrc && String(configSrc).trim()) ||
        (viewerSrc && String(viewerSrc).trim())
    );
    return hasModel ? 'none' : 'grid';
}
```

`toggleEmpty()` now delegates to it, so a relative `content/resources/...` path, an `asset://` handle, a `blob:` URL or an absolute URL all correctly mean "model selected". The overlay is also hidden right after the shared STL runtime boots, so STL matches the GLB/GLTF path deterministically.

Only the export runtime is touched — no edition, CSS or markup changes, and no existing behaviour is removed.

## Tests

- **Unit (Vitest)** — `three-d-viewer.test.js`: pure-function edge cases (including the relative `content/resources/...glb` regression, `asset://`, `blob:`, `http`, empty/whitespace/null), the `toggleEmpty()` method, and the STL boot path. Verified red before the fix and green after.
- **E2E (Playwright)** — `three-d-viewer.spec.ts`: generates a **real HTML5 export**, serves the unzipped bytes byte-for-byte from a clean origin (no service worker / dev-server cache) and loads the offline page, asserting the `.viewer-empty` overlay is hidden over the rendered model. Verified failing before the fix and passing after. A preview-panel overlay assertion was also added.

## Verification

- `make fix` clean
- `npx vitest run .../three-d-viewer` → 200 passed
- `bun x playwright test --project=chromium .../three-d-viewer.spec.ts` → 20 passed
- Every changed source line is covered by a unit test.

### DIFF
--- a/public/files/perm/idevices/base/three-d-viewer/export/three-d-viewer.js
+++ b/public/files/perm/idevices/base/three-d-viewer/export/three-d-viewer.js
@@ -769,6 +769,31 @@
     }
 
     /**
+     * Decide whether the empty-state overlay ("Select a 3D model to display")
+     * should be shown.
+     *
+     * A model is present when the iDevice has a configured source (single
+     * source of truth, mirroring edition's `[data-empty-state]` logic) OR the
+     * `<model-viewer>` already resolved a `src`. The configured source may be
+     * a relative `content/resources/...` path (static export + the preview
+     * URL-rewrite pipeline), an `asset://` handle (live editor / preview), a
+     * `blob:` URL, or an absolute http(s) URL — all mean "model selected".
+     * The previous blob:/http:/asset:// allowlist missed the relative export
+     * path and left the overlay covering a rendered model (issue #1957).
+     *
+     * @param {string} configSrc - the configured model source
+     * @param {string} viewerSrc - the current `<model-viewer>` src, if any
+     * @returns {'none'|'grid'} CSS display value for the overlay
+     */
+    function computeEmptyStateDisplay(configSrc, viewerSrc) {
+        const hasModel = !!(
+            (configSrc && String(configSrc).trim()) ||
+            (viewerSrc && String(viewerSrc).trim())
+        );
+        return hasModel ? 'none' : 'grid';
+    }
+
+    /**
      * Runtime controller for each wrapper.
      */
     class ThreeDViewerRuntime {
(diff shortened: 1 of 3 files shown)


mnarvaezm (approved):
Works correctly, as shown in the example:

https://quext.educarex.es/00/07/

Approved.

PR #1965, opened 2026-06-17
# fix(electrical-circuits): fix AI prompt identifier and correctly render Ω and Greek/math symbols in circuits

### LINKED ISSUE #1964: The AI generator for the Electrical Circuits iDevice creates questions but they are not loaded into the editor
The prompt for the **Electrical Circuits** iDevice seems to generate the questions correctly using AI, but the result is not loaded afterwards into the iDevice editor.

The AI returns questions in the expected format, but when trying to add them to the iDevice, they do not appear or are not imported correctly. It seems to be an issue with validation, parsing, or loading the generated content, rather than with the text generation itself.

## Steps to reproduce

1. Open the **Electrical Circuits** iDevice.
2. Use the AI generation option.
3. Enter a prompt to generate electrical circuit questions.
4. Wait for the AI to generate the questions.
5. Try to load/import the generated questions into the iDevice.

## Actual result

The AI generates the questions, but they are not loaded correctly into the iDevice.

## Expected result

The questions generated by the AI should be loaded automatically into the iDevice editor, as long as they follow the expected format.

mnarvaezm:
Closes #1964

This PR fixes two issues in the **Electrical Circuits** iDevice:

* **The iDevice AI prompt now generates questions compatible with the activity.
* **Greek and mathematical glyph rendering**: TikZJax emits SVG `<text>` elements without registering `@font-face`, causing browsers to fall back to system fonts. As a result, symbols such as Ω, Σ, and ω were rendered incorrectly (`Ω → ¬`, `Σ → §`, `ω → !`). Additionally, an off-by-one issue affected omega rendering because TikZJax emits U+00AC while the Computer Modern font stores the glyph at U+00AD.

### 1. Insert AI-generated questions

1. Open a project and add the **Electrical Circuits** iDevice (Science category).
2. Use the **AI Generator** to create several questions.
3. Copy the generated questions into the questions dialog.
4. Verify that the questions are inserted into the activity, each with its corresponding circuit code and statement.
5. Navigate through the questions and confirm that each circuit preview is displayed correctly.

### 2. Circuit code containing the Ω symbol

1. In a question, enter a circuit containing a literal Ω symbol, for example:

   ```latex
   \begin{circuitikz}
   \draw (0,0) to[battery1,l=6 V] (0,2) -- (2,2) to[R=1 kΩ] (4,2) to[R=2 kΩ] (4,0) -- (0,0);
   \end{circuitikz}
   ```

2. Click **Preview** and verify that Ω is rendered correctly (not as `¬`). The same should apply to symbols such as Σ and ω.

3. Save the iDevice and open the activity preview.

4. Verify that the circuit still renders correctly, using the self-contained SVG without requiring a re-render.

## Tests

* Added unit tests for Unicode-to-LaTeX normalization.
* Added unit tests for SVG text-to-path conversion.

### DIFF
--- a/public/files/perm/idevices/base/electrical-circuits/edition/electrical-circuits.js
+++ b/public/files/perm/idevices/base/electrical-circuits/edition/electrical-circuits.js
@@ -27,6 +27,7 @@ var $exeDevice = {
     version: 3.1,
     renderedTikzPreview: null,
     tikzFinishedHandler: null,
+    tikzCapturePromise: null,
 
     init: function (element, previousData, path) {
         this.ideviceBody = element;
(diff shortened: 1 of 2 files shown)


erseco (approved):
Approved. Nicely done — this correctly nails the root cause of #1964 and goes the extra mile on rendering.

Strengths:
- **Correct, minimal root-cause fix.** The `addEvents(10 → 11)` change aligns question validation with the `getTabIA(11)` prompt format, so electric-circuits lines are no longer silently rejected by the adaptative-quiz regex.
- **Single source of truth.** All sanitization funnels through `normalizeTikzCode`, keeping the rendered LaTeX, the SVG cache key, and the saved code in sync.
- **Self-contained SVG output.** Converting TikZJax `<text>` to real glyph `<path>`s (with the U+00AC→U+00AD Omega fixup) makes circuits render correctly without the unregistered Computer Modern web fonts, and the parser fails safe (falls back to leaving text untouched).
- **Idempotent helpers** and strong unit coverage (29 tests), including a reproduction of the exact Ω label from the bug report.

Pre-merge nits (non-blocking): confirm patch coverage ≥90% on the new font-parser branches via `make test-coverage`, and double-check `data-show-console="true"` on the edition preview is intentional rather than leftover debug. An E2E spec for the generate → insert → preview flow would be a nice follow-up.

mnunezcedec (commented):
Hi @mnarvaezm thanks for your work.

Indeed, the content generated by the AI (from the prompt) is now accepted by the IDevice and the questions are loaded correctly, or at least the first one. However, when moving to the next question, the circuit preview is not generated, and it is therefore not possible to continue or save.

Greetings and thanks

PR #1971, opened 2026-06-18
fix(export): keep the license footer off the global content font (#1939)

### LINKED ISSUE #1939: [Bug]: License text (CC) is clipped in Flux theme when the "Playwrite ES" content font is selected
In the **Flux** theme, selecting **Playwrite ES** as the content font applies it to the *entire* page, including the Creative Commons license block in the footer (`#packageLicense`). Because Playwrite ES is a tall handwriting typeface, the license text no longer fits its line box and is **rendered cut off / partially hidden**, so the license can't be read in full.

The license is a legal/attribution element and should remain fully legible regardless of the author's content-font choice.

#### Steps to Reproduce

1. Create or open a project using the **Flux** theme.
2. Set the project/content font to **Playwrite ES**.
3. Assign a Creative Commons license to the package (e.g., *CC BY-SA*).
4. View the page (editor/preview/export) and look at the license block in the footer.

#### Expected Behavior

The CC license text is fully visible and legible, unaffected by the selected content font.

#### Actual Behavior

The license inherits Playwrite ES and is clipped, so the text is not shown in full.

#### Technical Notes / Root Cause

- The license container `#packageLicense` does **not** declare its own `font-family`, so it inherits the author-selected content font (Playwrite ES) instead of staying on the theme's neutral stack (Fredoka / sans-serif).
- The clipping is amplified by the link styling in `theme/style.css` (~lines 845–865): `#packageLicense { line-height: 1em; }` together with `#packageLicense a { white-space: nowrap; overflow: hidden; text-overflow: ellipsis; line-height: 15px; }`. Playwrite ES glyphs exceed that fixed line box, so the text is clipped.

#### Proposed Solution

Pin the license block to a neutral font regardless of the selected content font, as suggested — the field should not switch to Playwrite ES:

```css
#packageLicense,
#packageLicense p,
#packageLicense a {
    font-family: "Fredoka", Arial, Verdana, Helvetica, sans-serif !important;
}
```

Optionally, make the block robust to any tall/handwriting font by letting the license text breathe instead of clipping it:

```css
#packageLicense a {
    white-space: normal;      /* allow wrapping */
    overflow: visible;        /* don't clip */
    line-height: 1.3;         /* room for ascenders/descenders */
}
```

#### Acceptance Criteria

- [ ] With **Flux** + **Playwrite ES**, the CC license text in the footer is shown **in full** (not clipped) in editor, preview, and exported package (HTML/SCORM).
- [ ] The license block keeps a neutral, legible font even when the content font is a tall/handwriting typeface.
- [ ] All CC variants (`cc-by-sa`, `cc-by-nc`, `cc-0`, etc.) and the license icon remain correctly aligned.
- [ ] No regression with the default font (Fredoka) or other content fonts.

#### Environment

- Theme: **Flux** (`config.xml`: version `2025`, default font *Fredoka*)
- Content font triggering the bug: **Playwrite ES**
- eXeLearning version: 

Image

[prueba.elpx.zip](https://github.com/user-attachments/files/28831168/prueba.elpx.zip)

erseco:
Closes #1939

## Problem
Selecting a tall handwriting content font (Playwrite ES) applied it to the whole page via the global-font CSS, including the Creative Commons license footer (`#packageLicense`). The legal text no longer fit its line box and was clipped, so the license couldn't be read.

## Fix
One change in the shared `GlobalFontGenerator`, which is used by the preview and every exporter, so it applies to **all themes** at once:

```css
#packageLicense,
#packageLicense * {
    font-family: var(--exe-license-font, Arial, Verdana, Helvetica, sans-serif) !important;
}
```

- `!important` so the global content font can't override it.
- The override is only emitted when a global font is selected — with no global font the license keeps the theme's default font.
- A theme can pin the license to its own font by defining `--exe-license-font`; otherwise it falls back to a neutral, legible stack.
- Fixing the font removes the root cause of the clipping, so no per-theme line-height hacks are needed.

## Tests
- Unit: `GlobalFontGenerator.spec.ts` asserts the override is emitted for every global font (and not for `default`).
- E2E: `license-font-global.spec.ts` (flux + nova) selects Playwrite ES and asserts the body uses it while `#packageLicense` does not.

## Verification
- `make fix` clean; `GlobalFontGenerator`, `Html5Exporter` and `teacher-mode-toggle` suites green.
- New E2E passes for both themes; confirmed it fails when the generator change is reverted.

@greynoa could you review?


cristinavaldera (dismissed):
Now the text is being cut off at the bottom, causing the last part of the content to be partially hidden.
Image

erseco:
> Now the text is being cut off at the bottom, causing the last part of the content to be partially hidden. 

This already happens on main, check it in online.exelearning.net

ignaciogros (approved):
I was going to review all the styles to make them compatible with any font-family, but there will be errors in styles already created by users anyway. Your solution is simple and it'll work in all cases. I really prefer the current approach now.

Thank you.

PR #1972, opened 2026-06-18
Hide Teacher Mode content by default in exports

### LINKED ISSUE #1772: Invert Teacher Mode default in exports: hidden by default, opt-in to reveal
## Summary

Currently, content marked as **Teacher Mode** in iDevices/pages is **visible by default** in exported content (HTML5, SCORM, etc.). To hide it from students, authors must rely on an external plugin/extension or post-export tweaks.

This proposal flips the default: **Teacher Mode content should be hidden in exports by default**, and only revealed via an explicit opt-in (a URL parameter, a query flag, a toggle in the viewer, or a documented mechanism — to be decided in discussion).

## Motivation

Teachers consistently ask the same question: *"if I upload my eXe content to a public website, will my answer keys / teacher notes be exposed?"*

The current behavior says yes — unless they install a plugin or manually edit the export. That is the wrong default for the most common publishing scenario:

- **Safer default for the dominant use case.** Most exports end up on a public LMS or static website. Hiding teacher content by default protects authors who don't realize the content is visible.
- **Reveal-on-demand is easier than hide-on-demand.** Asking authors to opt-in to expose teacher content (e.g. for a printed teacher guide, or a private LMS area) is a one-step action. Asking them to install a plugin or strip content after export is brittle and easy to forget.
- **Aligns with author intent.** When a user explicitly marks a block as *Teacher only*, they are signalling \"not for students\". The export should honor that signal by default.

## Proposed solution

1. **Default behavior in all exports**: Teacher Mode content is excluded/hidden from rendered output (CSS `display: none` at minimum, ideally not emitted into the DOM at all when feasible, to prevent View-Source leaks).
2. **Opt-in reveal mechanism**: a single, documented switch — e.g. `?teacher=1` query param, an export-time flag in the export dialog, or a viewer toggle — re-enables visibility. (Open to discussion in this thread.)
3. **Tooltip / help text** next to the *Teacher Mode* checkbox in the iDevice property editor, explaining the new default: *\"Marked content is hidden in the exported package by default. It can be revealed using the teacher view (see documentation).\"*
4. **Migration note** in the changelog so existing authors are aware that re-exporting an old project changes the visibility default.

## What this proposal explicitly rejects

> **Do not generate two parallel exports** (one \"student\", one \"teacher\").

That \"Solomonic\" split breaks the **edit-once, export-anywhere** model that eXeLearning is built on, doubles maintenance cost, and forces users to track two artifacts that drift out of sync. It violates [DRY (Don't Repeat Yourself)](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) — the teacher and student versions share 99% of the content; the difference is presentation, not source. The correct lever is a **single export with a visibility flag**, not two builds.

## Related issues

Open:
- #1242 — *Optional Global Password for \"Teacher Mode\" Access* (complementary: a password gate would compose naturally on top of an opt-in reveal)

Closed (history / context):
- #204 — original *Add \"Teacher Mode\"* feature request
- #607 — *Add teacher mode to pages* (extended teacher mode from iDevices to whole pages)
- #1340 — *Custom visual styling for \"Teacher Mode\" content via Style Manager*
- #1401 — *Teacher mode CSS class missing in iDevice*
- #786 — *The \"Teacher mode\" is not working*
- #462 — *Teacher mode CSS in new themes overrides form-check inputs*
- #1233 — *Visually indicate in the structure panel which pages are not visible in the export*

## Out of scope (for this issue)

- The exact opt-in mechanism (query param vs. dialog flag vs. viewer toggle) — to be decided in discussion.
- Strong access control (password, auth) — tracked in #1242 and would build on top of this change.

erseco:
## Summary

- Hides Teacher Mode content (the `teacher-only` class on blocks and iDevices) by default in exported packages and in the preview.
- The in-page self-serve toggle is now **opt-in**: it appears only when the content is opened with `?exe-teacher=1` (also `true`/`yes`, the generic alias `?teacher-mode=1`, and the legacy `?exe-teacher-toggler=1`), and it is **OFF by default** — the viewer activates it to reveal teacher content (state remembered in `localStorage`). The parameter never reveals content on its own.
- Without the parameter there is no toggle and teacher content stays hidden (student view).
- Removes the need for host plugins (WordPress, Moodle, Omeka-S) to inject CSS/JS to control Teacher Mode — they make the toggle available by changing the content URL only.
- Adds an authoring hint on the "Teacher only" property.

## Context

Fixes #1772.

Previously the only way to reveal teacher content was a self-serve toggle shown to **everyone** (any student could flip it), which is why every host plugin injected CSS to hide it. Now the toggle is hidden by default and appears only when the host opts in via the URL parameter.

Host integrations switch by changing the URL only:
- Student view: `index.html`
- Teacher-layer selector available: `index.html?exe-teacher=1`

## Implementation

- All runtime lives in `public/app/common/exe_export.js` (`teacherMode`), loaded in `<head>` for every export format and the preview:
  - `bootstrap()` runs at parse time: it sets `_showToggler` from a truthy parameter and, when the toggle is available, restores the viewer's stored on/off choice flicker-free (adds `mode-teacher` to `<html>` only if `localStorage` says on). It never reveals on its own.
  - `init()` renders the toggle only when the parameter made it available **and** `.teacher-only` content exists; the toggle's change handler adds/removes `mode-teacher` and persists the choice.
  - The active parameter is propagated onto in-package navigation links so the toggle stays available across pages (same-origin and opaque-origin iframes).
- `base.css` (`html:not(.mode-teacher) .js .teacher-only { display:none !important }`) is unchanged.
- The authoring preview (`previewPanel.js`) loads the viewer with `?exe-teacher=1`, so the toggle is available while editing (still off by default).
- Authoring hint added via a `help` field on both `teacherOnly` definitions in `src/routes/config-params.ts`.

## Notes

- **Presentation mode, not access control.** `?exe-teacher=1` only exposes a toggle; it is not authentication and not a security boundary. Truly sensitive answer keys need a separate authenticated/password-gated feature.
- Does not generate separate student/teacher exports.
- Existing packages must be re-exported to pick up the new default (changelog/migration note added).
- Page-level Teacher Mode does not exist today (block/iDevice only); deferred.

## Testing

- `public/app/common/exe_export.test.js` (Vitest): the parameter sets `_showToggler` (incl. aliases), does **not** reveal on its own, restores the revealed state from `localStorage` only when the toggle is available, and the toggle's render/checked/change behavior.
- `test/integration/teacher-mode-toggle.spec.ts` (Bun): teacher-only class + the base.css hide rule in the exported HTML.
- `test/e2e/playwright/specs/teacher-mode-url-param.spec.ts` (Playwright): the SW-served preview's runtime makes the toggle available (`_showToggler`) with `?exe-teacher=1`, and not without it.
- `src/routes/config-params.spec.ts`: the authoring hint is present and translation-prefixed.

Full suite green: `make test-unit` (7375), full frontend Vitest (13352), `make test-integration` (721), the E2E spec, and `make fix`.

## Why host plugins get simpler (and where this is heading)

Because teacher-mode visibility is now driven purely by the package's own `?exe-teacher=1` URL parameter, the host plugins (WordPress, Moodle ×3, Omeka-S) **drop the CSS/JS they used to inject into the embedded package** to hide the toggle — they only change the iframe URL. Beyond that cleanup, this advances the longer-term goal of **isolating embedded resource content from the host**: with the host no longer reaching into the package DOM, the resource can later be sandboxed / served cross-origin without losing teacher-mode control. The consuming plugin PRs (kept blocked until this one merges): wp-exelearning#58, omeka-s-exelearning#23, mod_exeweb#65, mod_exescorm#81, mod_exelearning#86.


erseco:
Follow-up commit (316c46f1b): eXeLearning's own authoring **preview** now reveals Teacher Mode by default — the preview panel loads the viewer with `?exe-teacher=1` so authors see their teacher-only content while editing (the author is the teacher). Exported packages still stay hidden by default; only the in-app preview opts in. `previewPanel.js` (SW preview + open-in-new-tab), its tests, the E2E spec, and docs/changelog updated.

greynoa:
Thanks @erseco, this makes a lot of sense to me — yes, go ahead! 👍 I like that teacher-only content now stays hidden by default in the exports, so students don't see it unless it's meant for them, and that authors can still see it while editing. That's exactly the behaviour I'd expect. Looks great to me.

pabloamayab (approved):
LGTM!!! Now I'm going to review Moodle extensions

PR #1886, opened 2026-06-02
Add Navarra and Valencian LOMLOE curriculum datasets

erseco:
## Summary

Adds two new autonomous-community LOMLOE curriculum datasets to the LOMLOE iDevice:

- **Navarra** (`ES-NC`) — official Spanish curriculum from the Gobierno de Navarra / BON.
- **Comunitat Valenciana** (`ES-VC`) — official Valencian curriculum from the Generalitat Valenciana / DOGV.

Builds on the dataset work in #1827 and follows the same JSON schema, code conventions and validation approach as the existing LOMLOE datasets (`ES`, `ES-EFP`, `ES-EX`, `ES-MD`, `ES-GA`, `ES-CN`).

Closes #1884.
Closes #1883.
Related to #1827.

## What changed

- Added `public/files/perm/idevices/base/lomloe/data/lomloe-ES-NC.json` (4 etapas, 1243 competencias, 3016 criterios, 5837 saberes).
- Added `public/files/perm/idevices/base/lomloe/data/lomloe-ES-VC.json` (4 etapas, 1003 competencias, 5001 criterios, 4975 saberes).
- Registered both datasets (`available: true`) in the LOMLOE iDevice `DATASETS` registry.
- Documented sources, language policy, corrections reviewed and mapping decisions in the iDevice `README.md`.
- Added regression tests for both datasets and registry guards in `lomloe.test.js`.

Both datasets are kept as raw JSON in the repo and gzipped at static-build time by the existing pipeline (`scripts/build-static-bundle.ts`), exactly like the other LOMLOE datasets.

### Navarra (`ES-NC`, Spanish)

Official BON curriculum annexes (ANEXO II) of the Decretos Forales:

- `Decreto Foral 61/2022` — Educación Infantil (BON 112).
- `Decreto Foral 67/2022` — Educación Primaria (BON 130).
- `Decreto Foral 71/2022` — Educación Secundaria Obligatoria (BON 155).
- `Decreto Foral 72/2022` — Bachillerato (BON 170).

`Decreto Foral 51/2025` (ESO) and the timetable/implementation Órdenes Forales 62/63/64/67-2022 were reviewed; the timetable orders do not change the stable curriculum content the iDevice represents.

### Comunitat Valenciana (`ES-VC`, Valencian)

Official DOGV curriculum annexes of the Decrets:

- `Decret 100/2022` — Educació Infantil.
- `Decret 106/2022` — Educació Primària.
- `Decret 107/2022` — Educació Secundària Obligatòria.
[...]

PR #1975, opened 2026-06-19
feat(lomloe): nest CC_DESCRIPTIONS under optional Config key

### LINKED ISSUE #1974: [Proposal] Explore multi-language support for "descriptores operativos" in curriculum LOMLOE JSON
## Context
Currently, the texts for the "descriptores operativos" are defined by default inside the `lomloe.js` file. The goal of this issue is to evaluate the addition of a new key within the curriculum JSON data files to allow incorporating these descriptor texts directly in the specific language of each curriculum.

---

## Proposed Solution (Phase 1)

This proposal introduces a new optional key named `Config` at the root level of the dataset JSON file. This key will encapsulate global metadata and settings,  with the `CC_DESCRIPTIONS` property.

### Proposed Schema:
```json
{
  "Etapa label": {
    "...": "..."
  },
  "Config": {
    "CC_DESCRIPTIONS": {
      "...": "..."
    }
  }
}
```

### Behavior:
* **If the `Config` key and `CC_DESCRIPTIONS` are present:** The system will load and display the operational descriptor texts specified inside the JSON file.
* **If the key is missing (Fallback):** The system will default to the current behavior, using the standard texts defined in `lomloe.js`.

### Benefits:
* **Backward Compatibility:** Existing datasets without the `Config` key will continue to work seamlessly without breaking changes.
* **Flexibility:** It enables localized and independent translations of descriptors for each specific curriculum.

> **Note:** This initial proposal will serve as the starting point.

---

## Advanced Proposal (Phase 2 / Future Scope)

As a long-term evolution, a deeper restructuring of the dataset is also under consideration to fully encapsulate the core curriculum data under its own semantic key, making the transition much easier since the `Config` object will already be in place:

```json
{ 
  "Curriculum": {
    "Etapa label": {
      "...": "..."
    }
  },
  "Config": {
    "id": "...",
    "isoCode": "...",
    "label": "...",
    "labelEn": "...",
    "framework": "LOMLOE",
    "community": "...",
    "available": true,
    "descriptorsPerCriterion": ...,
    "CC_DESCRIPTIONS": {
      "...": "..."
    }
  }
}
```

---

## Next Steps
1. Discuss the feasibility and design of Phase 1.
2. Proceed with the development and implementation of the first approach (optional `Config` object with `CC_DESCRIPTIONS`).

franmate:
## Description

This Pull Request nests the `CC_DESCRIPTIONS` competencia-clave descriptions under a new optional `Config` object at the root level of the dataset JSON. This keeps configuration/metadata separated from curriculum stage data at the top level, while remaining backward-compatible: datasets without `Config` continue to work using the built-in `CC_DESCRIPTIONS` fallback in `lomloe.js`.

Closes #1974

---

## Key Changes

* `edition/lomloe.js`: destructure `Config` out of the loaded dataset so it's never treated as a curriculum etapa; read descriptor texts from `Config.CC_DESCRIPTIONS`, falling back to the existing `CC_DESCRIPTIONS` constant when `Config` or `CC_DESCRIPTIONS` is missing.
* `data/lomloe-ES-GA.json`: moved `CC_DESCRIPTIONS` under the new `Config` key.
* `lomloe.test.js`: updated `walkAreas()` to skip the `Config` key; added tests asserting `Config.CC_DESCRIPTIONS` is present with the expected keys and is not iterated as an etapa.
* `README.md`: documented `Config` as a reserved, optional top-level key, and proposed a `Curriculum`/`Config` schema split for future dataset versions.

---

## How to Test

1. Load the Galicia (`ES-GA`) dataset and verify operational descriptor tooltips/badges still show the correct text (sourced from `Config.CC_DESCRIPTIONS`).
2. Remove the `Config` block from `lomloe-ES-GA.json` (or test with another dataset that lacks it) and verify the editor falls back to the default `CC_DESCRIPTIONS` texts in `lomloe.js` without errors.


ignaciogros (changes requested):
Thanks, @franmate!

Could you please review the conflicts?

I agree with you that encapsulating the educational stages would result in a cleaner structure. Perhaps we can make that change if @erseco agrees.

I was not expecting the additional level that was introduced: `Config.CC_DESCRIPTIONS` instead of `CC_DESCRIPTIONS`. I like the `{ "curriculum": {}, "descriptors": {} }` approach.

This is something new, so I think it's worth taking the time to structure it as cleanly and correctly as possible.

Thank you.

ignaciogros:
Hi @franmate,

There was a previous PR (#1886) which is already merged into `main` and includes those changes.

Sorry for the misunderstanding in my last message. There's no need to make any changes. I'm closing the PR since it's already covered.

The Galician translation still needs to be reviewed, but I think that should be part of a separate PR.

Thank you very much for your work.

Issue #1875, opened 2026-06-01
Link to file download, opens two windows


3DCube-KNX:
I have a link to a RAR file to download. When I click it, two pop-up windows in File Explorer open:
One that says "Save file".

The other says "blob:app://localhost/a61add6d-a090-4e22-883e-2da07a032e50". Shouldn't only one pop-up window in File Explorer open to save it?

Exporting to SCORM and uploading to the online platform I use works correctly; it's only in the editor and in the preview when testing.

I'm using version 4.0.0 on Windows 11.

Thanks.

erseco:
Thanks for the detailed report, @3DCube-KNX — the second window showing `blob:app://localhost/…` was the key clue. 🙏

You're seeing a **double save dialog** in the Electron desktop app: when you click a content link to a downloadable file (your `.rar`), the editor handed the `blob:` URL to Electron's `will-download` flow, which shows our own "Save file" dialog *and* Electron's default one at the same time.

This is **related to** #1818 / #1819 but not the same bug. #1819 fixed the double dialog only for the *ELPx source-file* download link; arbitrary file-download links like yours went through a different code path that was still affected.

Fix opened in #1876: non-image file downloads are now routed through a single native save dialog. Web export/SCORM were never affected (as you noted), and image/lightbox links are unchanged.

It'll ship in an upcoming release — we'd appreciate it if you could confirm once it's available. Thanks again!

PR #1980, opened 2026-06-22
fix(tools): make Download CSV work in link validation and resource report

### LINKED ISSUE #1947: Validación de Enlace: CSV
Por favor,

En validación de enlace o Informe de recursos, el botón **Descargar CSV** no descarga nada desde la última versión de ExeLearing con Window 11.

Gracias

erseco:
Fixes #1947

## Cause

The **Download CSV** button in *Tools → Link validation* and in the *Resource Report* downloaded nothing and threw in the console:

> `TypeError: Cannot read properties of undefined (reading 'showToast')`

The toast/notification manager is registered as `eXeLearning.app.toasts` (a `ToastsManager`) and its method is `createToast(data)`. But several call sites instead called `eXeLearning.app.alerts.showToast({ type, message })` — an `alerts` manager and a `showToast` method that **do not exist anywhere** in the codebase.

In the CSV flow, when there were **no broken links** (or no resources) to export, the empty-case branch ran `eXeLearning.app.alerts.showToast(...)`. Since `eXeLearning.app.alerts` is `undefined`, this threw, aborted the click handler, and nothing downloaded — exactly the reported error.

The image-optimizer modal had the same wrong call but with optional chaining (`alerts?.showToast`), so it silently never showed its success/error toast.

The existing unit tests **masked** the bug by mocking the non-existent `eXeLearning.app.alerts.showToast`, so they passed while production crashed.

## Fix

- Replace all 5 `eXeLearning.app.alerts.showToast({ type, message })` calls with the real `eXeLearning.app.toasts.createToast({ title, body, icon, modal, remove })` API (mapping `type` → `icon`/`title`, `message` → `body`).
- Update the 3 unit test files to mock and assert the real `app.toasts.createToast`, and add regression tests asserting the empty-export case no longer throws.

Touched files:
- `modalOdeBrokenLinks.js` — Link Validation (validation error + empty CSV)
- `modalOdeUsedFiles.js` — Resource Report (empty CSV)
- `modalImageOptimizer.js` — success/error toasts

## Testing (TDD)

The updated/added tests fail on `main` (7 failures pinpointing the `app.alerts.showToast` lines) and pass with the fix.

- `npx vitest run` on the three modals → **210 tests pass**
- Patch coverage **100%** on changed lines · `make fix` clean · no new translation keys (all strings already registered)

### DIFF
--- a/public/app/workarea/modals/modals/modal.js
+++ b/public/app/workarea/modals/modals/modal.js
@@ -657,7 +657,12 @@ export default class Modal {
         downloadLink.click();
         document.body.removeChild(downloadLink);
 
-        URL.revokeObjectURL(url);
+        // Defer revocation so asynchronous download handlers can still read the
+        // blob. In Electron the global click interceptor in asset_url_resolver.js
+        // routes blob: downloads through saveBufferAs, which fetches the blob URL
+        // on a later microtask; revoking synchronously here would invalidate the
+        // URL first and abort the save with ERR_FILE_NOT_FOUND (issue #1947).
+        setTimeout(() => URL.revokeObjectURL(url), 10000);
     }
 
     /*******************************************************************************
--- a/public/app/workarea/modals/modals/modal.test.js
+++ b/public/app/workarea/modals/modals/modal.test.js
@@ -1160,12 +1160,24 @@ describe('Modal', () => {
       vi.restoreAllMocks();
     });
 
-    it('revokes blob URL after download', () => {
+    it('defers revoking the blob URL so async (Electron) download handlers can read it (issue #1947)', () => {
+      // In Electron the global click interceptor (asset_url_resolver.js) routes
+      // blob: downloads through saveBufferAs, which fetches the blob URL on a
+      // later microtask. Revoking synchronously here would invalidate the URL
+      // before that fetch runs, aborting the save with ERR_FILE_NOT_FOUND.
+      vi.useFakeTimers();
       const modal = new Modal(mockManager, 'test-modal', 'Default Title', false);
 
       modal.downloadCSVFile('content', 'test.csv');
 
+      // Not revoked synchronously while async consumers may still read the blob.
+      expect(URL.revokeObjectURL).not.toHaveBeenCalled();
+
+      // Eventually revoked once pending async work has had a chance to run.
+      vi.runAllTimers();
       expect(URL.revokeObjectURL).toHaveBeenCalledWith('blob:test-url');
+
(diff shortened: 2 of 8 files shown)


ignaciogros (changes requested):
Thanks @erseco,

I tried the Windows Electron version, but it's still failing.

Now I see these JavaScript errors in the console: 

```
asset_url_resolver.js?v=v0.0.0-alpha-build8ee6e5cc:823  GET blob:app://localhost/d328bfef-9bf9-4f9b-84e1-95dfaf64ca1e net::ERR_FILE_NOT_FOUND
asset_url_resolver.js?v=v0.0.0-alpha-build8ee6e5cc:827 [AssetResolver] Electron saveBufferAs failed: TypeError: Failed to fetch
```

I used [broken-links.zip](https://github.com/user-attachments/files/29205344/broken-links.zip) for testing.

erseco:
Thanks @ignaciogros — that console error pointed at a second, deeper bug in the actual download path (different from the toast crash), so I pushed a follow-up fix (3ff90ebd6).

### Cause

`Modal.downloadCSVFile()` triggered the `<a download>` click and then revoked the blob URL **synchronously**:

```js
downloadLink.click();
URL.revokeObjectURL(url); // synchronous
```

In Electron, the global capture-phase click interceptor in `asset_url_resolver.js` routes blob: downloads through `saveBufferAs`, which reads the blob with an **asynchronous** `fetch(blobUrl)` on a later microtask (to avoid the double save dialog from #1875). The synchronous revoke ran first, so by the time that `fetch` executed the URL was already gone — hence `net::ERR_FILE_NOT_FOUND` and `saveBufferAs failed: TypeError: Failed to fetch`. In a browser the `<a download>` reads the blob during click dispatch, so the immediate revoke was harmless — that's why it only failed in Electron.

### Fix

Defer the revocation so the async save path can read the blob first (the same `setTimeout(() => URL.revokeObjectURL(url), 10000)` pattern already used in `WebFileSystem.js` and `previewPanel.js`).

### Verification

Reproduced in real Chromium (same engine as Electron):

```
old_sync_revoke:      { ok: false, error: "TypeError: Failed to fetch" }   <- your error
new_deferred_revoke:  { ok: true,  bytes: 11 }                             <- fix
```

I couldn't fully launch Electron locally (the Electron binary download kept failing in my environment), so a Windows Electron re-test with `broken-links.elpx` would be appreciated to confirm end to end. Unit tests cover the deferred-revoke behavior as a regression.

ignaciogros (approved):

PR #1981, opened 2026-06-22
feat(electron): add "File → Close" desktop menu option

### LINKED ISSUE #1847: Missing File->Close option
Hi. Congratulations for this new version. 

User interface has changed and maybe I´m wrong, but I can´t close a file. I think it would be good to keep this menu option.

Regards :)

erseco:
## Summary

Restores the missing **File → Close** menu entry reported in #1847. It closes the current editor window in the packaged desktop (Electron) app.

## Why it is Electron-only (and how)

The static browser/PWA build and the Electron app both run in the same `static`/offline runtime, so `config.isOfflineInstallation`, `window.__EXE_STATIC_MODE__`, `runtimeConfig.isStaticMode()` and storage capabilities **cannot** distinguish them. The only reliable Electron signal is the presence of `window.electronAPI`.

So the entry is rendered **hidden** in the offline navbar markup (`d-none exe-electron-only`) and is revealed by `navbarFile.js` **only** when `window.electronAPI.closeCurrentWindow` is a function. In the browser/PWA build it stays hidden; the server (online) build never renders the offline block at all.

## Unsaved-changes handling

Close goes through a new `app:closeCurrentWindow` IPC handler that calls `BrowserWindow.close()` on the sender window. The existing `attachEditorWindowCloseGuard(win)` still intercepts the close and prompts for unsaved changes — no confirmation logic is duplicated.

## Changes

- `app/preload.js` — expose `closeCurrentWindow()` on the `electronAPI` bridge.
- `app/main.js` — add the `app:closeCurrentWindow` IPC handler (resolves the sender window via `BrowserWindow.fromWebContents`, returns `true`/`false`).
- `views/workarea/menus/menuNavbar.njk` — add the hidden `File → Close` item in the offline block, after *Save* and before *Export as…*.
- `src/routes/pages.ts` — register the `close` label in the navbar translation map.
- `public/app/workarea/menus/navbar/items/navbarFile.js` — query the button, add `setCloseFileEvent()` (reveal + wire click only under Electron; respects `checkOpenIdevice()`).
- `public/app/workarea/menus/navbar/items/navbarFile.test.js` — tests for the new behavior.

## Testing

- `make fix` — clean (only a pre-existing unrelated `any` warning in `export.ts`).
- Frontend (Vitest) — full suite green; `navbarFile.test.js` covers: button is queried/stored, `setEvents()` calls `setCloseFileEvent()`, entry shown + close invoked when the Electron bridge is present, idevice-open guard blocks the close, and the entry stays hidden with nothing wired when `electronAPI` is absent. Patch coverage on the changed file is 100%.
- Backend (Bun) — `src/routes/pages.spec.ts` green with the new `close` label.

Note: the Electron-only visibility is verified by unit tests. The Playwright E2E harness runs the web/static build where `window.electronAPI` is absent, so it cannot exercise the Electron path.

Closes #1847

### DIFF
--- a/app/main.js
+++ b/app/main.js
@@ -1534,6 +1534,18 @@ ipcMain.handle('app:getMemoryUsage', async (e) => {
     };
 });
 
+// Close the window that initiated the request (File > Close). The unsaved
+// changes confirmation is handled by the close guard attached in
+// attachEditorWindowCloseGuard(), so we only trigger the close here.
+ipcMain.handle('app:closeCurrentWindow', async (e) => {
+    const win = BrowserWindow.fromWebContents(e.sender);
+    if (win && !win.isDestroyed()) {
+        win.close();
+        return true;
+    }
+    return false;
+});
+
 function normalizeBinaryPayload(bufferData, base64Data) {
     if (bufferData instanceof Uint8Array) {
         return Buffer.from(bufferData);
--- a/app/preload.js
+++ b/app/preload.js
@@ -44,6 +44,9 @@ contextBridge.exposeInMainWorld('electronAPI', {
     }
   },
   openElp: () => ipcRenderer.invoke('app:openElp'),
+  // Close the current window (File > Close). The existing close guard in the
+  // main process still prompts for unsaved changes.
+  closeCurrentWindow: () => ipcRenderer.invoke('app:closeCurrentWindow'),
   readFile: (filePath) => ipcRenderer.invoke('app:readFile', { filePath }),
   getMemoryUsage: () => ipcRenderer.invoke('app:getMemoryUsage'),
   notifyRendererReadyForOpenFile: () => ipcRenderer.send('app:renderer-ready-for-open-file'),
--- a/public/app/common/shortcuts.js
+++ b/public/app/common/shortcuts.js
@@ -174,7 +174,7 @@ get isStatic() {
 
 getComboRemap() {
   const isStatic = this.isStatic;
-  return {
+  const map = {
     'mod+alt+n'   : 'navbar-button-new',
     'mod+o'       : isStatic ? 'navbar-button-open-offline'    : 'navbar-button-openuserodefiles',
     'mod+s'       : isStatic ? 'navbar-button-save-offline'    : 'navbar-button-save',
@@ -184,6 +184,18 @@ getComboRemap() {
     'mod+p'       : 'navbar-button-preview',
     'mod+k'       : 'navbar-button-global-search',
   };
+
+  // File -> Close (Cmd/Ctrl+W). This closes the current window, so the
+  // standard accelerator is Cmd+W / Ctrl+W (not Q, which is reserved for
+  // quitting the app). macOS already provides a native Cmd+W via the Window
+  // menu role, so we only wire the in-app binding on Windows/Linux. Gating on
+  // the Electron bridge keeps the browser/PWA build from swallowing Ctrl+W
+  // (which natively closes the browser tab there).
+  if (!this.isMac && typeof window?.electronAPI?.closeCurrentWindow === 'function') {
+    map['mod+w'] = 'navbar-button-close-file';
+  }
+
+  return map;
 }
 
 resolveTarget(combo) {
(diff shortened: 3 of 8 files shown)


ignaciogros (dismissed):
Thank you, @erseco.

I think these changes would improve the presentation and the usability:

* Use "Exit" instead of "Close".
* Move the link to the last position in the list (below "Print"), preceded by a separator: `<li class="dropdown-divider"></li>`.
* If possible, add the "Q" keyboard shortcut for the different operating systems.

Could you please evaluate these changes?

erseco:
Thanks for the review, @ignaciogros 🙏 I've pushed the changes — here's what I did, plus one point I'd like to discuss with some references.

✅ **2) Moved below *Print*, with a divider.** The entry now sits in the last position of the File menu, below *Print*, preceded by a `dropdown-divider`. The divider and the entry are Electron-only and revealed together only when the desktop bridge is present.

✅ **3) Keyboard shortcut — added as ⌘W / Ctrl+W** (rather than *Q*). Since this item closes the **current window** (`BrowserWindow.close()`), the standard accelerator is **⌘W / Ctrl+W**. *Q* (⌘Q / Ctrl+Q) is the universal *Quit the application* accelerator. On macOS ⌘W is already provided natively by the Window-menu role, so the in-app binding is added for Windows/Linux; the menu shows the ⌘W / Ctrl+W hint on all platforms.

🟡 **1) "Exit" vs "Close" — I'd suggest keeping "Close".** This item closes the **current window/document**, not the application (the app keeps running and may have several project windows/tabs open). On every desktop platform, "Close" is the window command and "Exit"/"Quit" is reserved for terminating the app:

- **Apple HIG** — [*Close* closes the current window/document (⌘W); *Quit* quits the app (⌘Q)](https://developer.apple.com/design/human-interface-guidelines/the-menu-bar)
- **GNOME HIG** — [Ctrl+W closes the window/tab; Ctrl+Q quits the app](https://developer.gnome.org/hig/reference/keyboard.html)
- **Microsoft** — [Ctrl+W / Ctrl+F4 close the document "but not the entire app"; Alt+F4 exits the app](https://support.microsoft.com/en-us/windows/keyboard-shortcuts-in-windows-dcc61a57-8ff0-cffe-9796-cb9706c75eec)
- **Electron** — [`role: 'close'` (⌘/Ctrl+W) vs `role: 'quit'` (⌘/Ctrl+Q)](https://www.electronjs.org/docs/latest/tutorial/menus); Electron even labels the *quit* role **"Exit"** on Windows.

This also matches Nielsen Norman Group's usability heuristics:

- [**#2 — Match between the system and the real world**](https://www.nngroup.com/articles/ten-usability-heuristics/): *"Use words, phrases, and concepts familiar to the user… Follow real-world conventions."*
- [**#4 — Consistency and standards**](https://www.nngroup.com/articles/consistency-and-standards/): *"Users should not have to wonder whether different words, situations, or actions mean the same thing. Follow platform and industry conventions."*
- [**Command-name guidance (NN/g)**](https://www.nngroup.com/articles/ui-copy/): *"Commands should always reflect the state the system will move into."*

Labeling a window-close as "Exit" (and binding it to the *Q* / Quit accelerator) would make the same word and shortcut mean something different than in every other desktop app — raising cognitive load and risking data-loss confusion (a user who reads "Exit" as "leave the app" only closes one window, or vice-versa).

That said, your "Exit" request does point at a real gap: there's currently **no app-level Quit/Exit on Windows/Linux** (the native Quit only exists on macOS). If you'd like, I'm happy to add that as a **separate** item — a proper *Quit / Exit* (⌘Q / Ctrl+Q via `app.quit()`) — alongside *Close*, either in this PR or as a follow-up. Your call.

Re-requesting your review on the updated branch — thanks!

ignaciogros (approved):
I misunderstood this feature. Thanks for the clarification. All of my suggestions (menu position, using "Exit", the Q shortcut) were related to closing the application, not the document.

However, after testing the feature, at least on Windows, I can see that it behaves exactly like that: if I press Ctrl+W, the application is closed (the instance of the application I'm currently using). The reason is that there is not (yet) any welcome screen (the application is open, but no document is open).

For this case, I still prefer "Exit" and the current position (last in the list). In any case, it's just a word. It's not important at all. I think we can merge the PR, and we can change that later if necessary.

Thank you!

PR #1983, opened 2026-06-22
The iDevices panel is disabled when the current page is document root.

### LINKED ISSUE #1967: Disable Add iDevice in Project Properties
Disable the ability to add iDevices while in Project Properties.

ignaciogros:


erseco (approved):
Tested in the online deploy and it worked! 

Image

PR #1989, opened 2026-06-23
feat(ai): add configurable providers for game generation

erseco:
## Summary
- Adds an environment- and admin-configurable AI subsystem for game-question generation, so regional/institutional installations can hide public AI options and use only their configured provider.
- New AI service layer (`src/services/ai/`): typed config parsing + provider allow-list, a fail-closed manager with request timeout, and Azure OpenAI / Ollama / OpenAI-compatible providers.
- New `POST /api/ai/generate-text` (authenticated, online-only) and admin-only `POST /api/ai/test-connection` diagnostic.
- Public, secret-free AI config `{ enabled, provider, mode, configured }` added to the parameter response; the `defaultAI` user preference is shown only in external mode.
- Admin **AI** tab backed by `app_settings` (env vars as defaults). Provider API keys are **write-only**: masked on read, and a blank value on save keeps the stored key.
- Editor behaviour matrix preserved/extended in `common_edition.js`.
- Docs: `doc/development/ai-configuration.md` + `.env.dist` entries.

## Requested by
Requested by @ignaciogros.

## Behaviour matrix
| `AI_FEATURES_ENABLED` | `AI_PROVIDER` | Game editor | `defaultAI` |
|---|---|---|---|
| `false` | (any) | No AI tab / buttons | Hidden |
| `true` | `external` | Assistant selector + **Send to AI** (current behaviour) | Shown |
| `true` | `azure` / `ollama` / `openai_compat` | Server-side **Create** button; no external options | Hidden |

When a managed provider is selected but misconfigured, generation fails closed with a clear error — it never falls back to a public AI service.

## Security
- Provider credentials stay server-side. They are never returned by `/api/config`, `/api/parameter-management/parameters/data/list`, `GET /api/admin/settings` (masked → `value: ''` + `isSet`), the static bundle, or any frontend global.
- `AI_PROVIDER` is validated against an allow-list; managed endpoint URLs are validated as `http(s)` on save.
- The provider endpoint is admin-configured (trusted); the browser only ever sends the prompt text — it cannot choose the provider URL.

## Testing (run locally on this branch)
- `make fix` — clean (only a pre-existing `noExplicitAny` warning in `src/routes/export.ts`, untouched here).
- `bun test src` — 7382 pass. The only 3 failures are the pre-existing `Resources Routes` "version prefix" tests (environmental: `getAppVersion()` returns the `v0.0.0-alpha` fallback in an untagged checkout; unrelated to this PR and unchanged by it).
- `bun test test/integration` — 721 pass, 0 fail.
- Frontend `vitest run` (full) — 13437 pass, 0 fail.
- New colocated tests added for every new file (AI service + providers, route, admin masking/validation, config gating, static bundle, `apiCallManager.getGenerateQuestions`, `common_edition` matrix, `admin/ai.js`, `pages` secret-masking).
- E2E: `test/e2e/playwright/specs/ai-configuration.spec.ts` added (disabled / external / managed UI rows). Not executed in the local sandbox (the Chromium headless-shell binary could not be downloaded here); it runs in CI.

## Reviewer verification checklist
- [ ] `make fix && make test-unit && make test-integration` are green.
- [ ] `make test-e2e` green, including `ai-configuration.spec.ts`.
- [ ] **Disabled** (`AI_FEATURES_ENABLED=false`): no AI tab/buttons in a game iDevice; `defaultAI` absent from Preferences; `POST /api/ai/generate-text` returns a safe `AI_DISABLED` error.
- [ ] **External** (`AI_PROVIDER=external`): assistant selector + “Send to AI” present; `defaultAI` shown; opening a public assistant still works.
- [ ] **Managed** (`AI_PROVIDER=ollama`/`azure`/`openai_compat`): external selector hidden; “Create” generates questions via the backend; misconfigured provider shows a clear error and does **not** open a public AI.
- [ ] No secret leak: inspect `/api/parameter-management/parameters/data/list`, `/api/config`, `GET /api/admin/settings`, and `data/bundle.json` — none contain provider keys/endpoints.
- [ ] Admin **AI** tab: provider field groups toggle by provider; saving a blank API key keeps the stored key; “Test connection” reports success/failure.

### DIFF
--- a/.env.dist
+++ b/.env.dist
@@ -261,3 +261,41 @@ CLEANUP_UNSAVED_AGE_HOURS=24
 # Age in days after which guest user projects are deleted (default: 7)
 CLEANUP_GUEST_AGE_DAYS=7
 ###< Project Cleanup Settings ###
+
+###> AI Configuration ###
+# Controls how AI assistance is offered in game-generation activities.
+# These values are DEFAULTS; administrators can override them at runtime from
+# the admin panel "AI" tab (stored in app_settings, which takes precedence).
+# API keys are never exposed to the frontend.
+
+# Master switch. When false, all AI buttons and AI preferences are hidden.
+AI_FEATURES_ENABLED=true
+
+# Provider mode: external | azure | ollama | openai_compat
+#  - external      : keep the public AI assistant links (ChatGPT, Claude, ...).
+#  - azure/ollama/openai_compat : eXeLearning generates questions server-side
+#    through the configured provider; the public AI options are hidden.
+# Invalid values fall back to "external" (fail-safe).
+AI_PROVIDER=external
+
+# Azure OpenAI (used when AI_PROVIDER=azure)
+AI_AZURE_ENDPOINT=
+AI_AZURE_API_KEY=
+AI_AZURE_DEPLOYMENT=
+AI_AZURE_API_VERSION=
+
+# Ollama (used when AI_PROVIDER=ollama). Endpoint defaults to a local instance.
+AI_OLLAMA_ENDPOINT=http://localhost:11434
+AI_OLLAMA_MODEL=
+
+# OpenAI-compatible endpoint (used when AI_PROVIDER=openai_compat).
+# AI_COMPAT_API_KEY is optional (some self-hosted gateways are unauthenticated).
+AI_COMPAT_ENDPOINT=
+AI_COMPAT_API_KEY=
+AI_COMPAT_MODEL=
+
+# Generation options (managed providers only)
+AI_REQUEST_TIMEOUT_MS=60000
+AI_MAX_OUTPUT_TOKENS=4096
+AI_TEMPERATURE=0.2
+###< AI Configuration ###
--- a/.env.dist
+++ b/.env.dist
@@ -268,7 +268,9 @@ CLEANUP_GUEST_AGE_DAYS=7
 # the admin panel "AI" tab (stored in app_settings, which takes precedence).
 # API keys are never exposed to the frontend.
 
-# Master switch. When false, all AI buttons and AI preferences are hidden.
+# Master switch. When false, the AI tab stays visible in copy-only form (prompt +
+# Copy) but the provider selector, "Send to AI", the managed Generate surface and
+# the defaultAI preference are all hidden, and the server endpoint refuses to run.
 AI_FEATURES_ENABLED=true
 
 # Provider mode: external | azure | ollama | openai_compat
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -327,6 +327,7 @@ Domain-specific guidance lives in `.agents/skills/*/SKILL.md`.
 | Real-time/Yjs | [doc/development/real-time.md](doc/development/real-time.md) |
 | REST API v1 | [doc/development/rest-api.md](doc/development/rest-api.md) |
 | Embedding in LMS | [doc/development/embedding.md](doc/development/embedding.md) |
+| AI configuration | [doc/development/ai-configuration.md](doc/development/ai-configuration.md) |
 | Profiling | [doc/development/profiling.md](doc/development/profiling.md) |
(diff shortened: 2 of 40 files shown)


ignaciogros (dismissed):
Thanks, @erseco.

I wasn't expecting `AI_FEATURES_ENABLED=false` to have that behavior. I expected the AI tab to remain visible, but without the provider selector and the "Send to AI" button. That way, users can still copy and paste the code without any issues. Otherwise, I think we're removing a feature that is heavily used for game creation.

I've tested with both Azure and Ollama as `AI_PROVIDER`, with and without a custom `BASE_PATH`, and the result has always been the same (on `localhost:8080` and without HTTPS):

```
XHR POST
http://localhost:8080/api/ai/generate-text
[HTTP/1.1 502 Bad Gateway 11348ms]
```

The UI shows the following message:

> The questions could not be generated: The AI provider returned an error (HTTP 500).

I get the same results with both `make up-local` and `make up`. The container logs do not show any related information.

erseco:
Thanks for the careful review, @ignaciogros — both points are addressed in `66692fc20`.

### 1. `AI_FEATURES_ENABLED=false` keeps the AI tab (copy-only)

You're right: fully removing the tab dropped a workflow people rely on. Disabled mode now **keeps the AI tab visible** with the prompt and the **Copy** button (and the **Questions** → **Save** flow), but **without** the provider selector, the **Send to AI** button and the managed **Generate** surface. So teachers can still copy the auto-built prompt, take it to any AI tool of their choice, and paste the questions back — eXeLearning just never calls a provider itself.

Updated behaviour matrix:

| `AI_FEATURES_ENABLED` | `AI_PROVIDER` | Game editor |
|---|---|---|
| `false` | (any) | Prompt + **Copy** (+ Questions/Save); no selector / Send to AI / Generate |
| `true` | `external` | Prompt + assistant selector + **Send to AI** |
| `true` | `azure` / `ollama` / `openai_compat` | Server-side **Generate** / **Create** |

### 2. The opaque `HTTP 500` and empty logs

I reproduced your symptom. The `502` + *"The AI provider returned an error (HTTP 500)"* means the **upstream provider** (your Ollama/Azure) actually replied `500` — for Ollama that is almost always *the model has not been pulled* — but we were **discarding the upstream body and logging nothing**, which is exactly why the container logs were empty. That opacity was the real bug.

Fixed by adding server-side observability:

- Every managed-provider failure is now logged with provider, endpoint host, HTTP status and a truncated response-body snippet, e.g.

  ```
  [ai] provider "ollama" at localhost:11434 returned HTTP 500: {"error":"model 'llama3' not found, try pulling it first"}
  ```

- The admin-only **Test connection** diagnostic (`POST /api/ai/test-connection`) now returns that snippet directly, so an admin sees the exact provider error without reading logs.
- The public `POST /api/ai/generate-text` response is unchanged and still returns only a safe `{ error, message }` (no upstream details) — nothing leaks to ordinary users.

So your next run will show the real cause in the logs. For a local Ollama the usual fix is `ollama pull <AI_OLLAMA_MODEL>`; note that a `localhost` Ollama is **not** reachable from inside the `make up` container, so use a host-reachable address there (the docs now mention this).

### Tests / docs

- New disabled-mode and logging/diagnostic assertions across vitest, `bun test` and the Playwright `ai-configuration.spec.ts`.
- `doc/development/ai-configuration.md` (matrix + a new *Troubleshooting a managed provider* section) and the `.env.dist` comment updated to match.
- `make fix` clean; AI unit suites green.

erseco:
Hi @ignaciogros 👋

I'm closing this one in favour of #1998, which carries the exact same work but lives on a branch in `exelearning/exelearning` directly (`feature/configurable-ai-mode`) instead of my fork — that should make it easier for you to push to it.

I've assigned #1998 to you and attached your review there for continuity, so you can pick it up and keep refining it. Thanks a lot for the thorough testing and feedback! 🙏

ignaciogros:
OK. 

Thank you, @erseco!

PR #1994, opened 2026-06-24
Defer tooltips execution to ensure $exeFX.init() has completed first.

### LINKED ISSUE #1966: Tooltip does not work in FX
When inserting a Tooltip in a text inside a FX effect, it doesn´t show in Preview:

**Edition**:

Image

**Preview**:

Image

ignaciogros:
**How to test:**

Add a tooltip inside an accordion or other effect. It should work in all cases now.


mnunezcedec (approved):
Works as expected!!! 
Thanks

PR #1995, opened 2026-06-24
iDevice icon selector contrast

### LINKED ISSUE #1761: iDevice icon selector contrast
In some cases there's not enough contrast between icons and background. Maybe that CSS opacity should no be used.

ignaciogros:
**How to test:**

Just add a box and select different icons. It'd be good if you test it with different styles.


mnunezcedec (approved):
It is working as expected

PR #1998, opened 2026-06-25
feat(ai): add configurable providers for game generation

erseco:
> **Note:** This PR supersedes #1989. It is pushed to a branch on `exelearning/exelearning` directly (`feature/configurable-ai-mode`) instead of my fork, so that @ignaciogros can take over and continue refining the work. The previous review is attached at the bottom for continuity.

## Summary
- Adds an environment- and admin-configurable AI subsystem for game-question generation, so regional/institutional installations can hide public AI options and use only their configured provider.
- New AI service layer (`src/services/ai/`): typed config parsing + provider allow-list, a fail-closed manager with request timeout, and Azure OpenAI / Ollama / OpenAI-compatible providers.
- New `POST /api/ai/generate-text` (authenticated, online-only) and admin-only `POST /api/ai/test-connection` diagnostic.
- Public, secret-free AI config `{ enabled, provider, mode, configured }` added to the parameter response; the `defaultAI` user preference is shown only in external mode.
- Admin **AI** tab backed by `app_settings` (env vars as defaults). Provider API keys are **write-only**: masked on read, and a blank value on save keeps the stored key.
- Editor behaviour matrix preserved/extended in `common_edition.js`.
- Docs: `doc/development/ai-configuration.md` + `.env.dist` entries.

## Requested by
Requested by @ignaciogros.

## Behaviour matrix
| `AI_FEATURES_ENABLED` | `AI_PROVIDER` | Game editor | `defaultAI` |
|---|---|---|---|
| `false` | (any) | Prompt + **Copy** (+ Questions/Save); no selector / Send to AI / Generate | Hidden |
| `true` | `external` | Assistant selector + **Send to AI** (current behaviour) | Shown |
| `true` | `azure` / `ollama` / `openai_compat` | Server-side **Create** button; no external options | Hidden |

When a managed provider is selected but misconfigured, generation fails closed with a clear error — it never falls back to a public AI service.

## Security
- Provider credentials stay server-side. They are never returned by `/api/config`, `/api/parameter-management/parameters/data/list`, `GET /api/admin/settings` (masked → `value: ''` + `isSet`), the static bundle, or any frontend global.
- `AI_PROVIDER` is validated against an allow-list; managed endpoint URLs are validated as `http(s)` on save.
- The provider endpoint is admin-configured (trusted); the browser only ever sends the prompt text — it cannot choose the provider URL.

## Testing (run locally on this branch)
- `make fix` — clean (only a pre-existing `noExplicitAny` warning in `src/routes/export.ts`, untouched here).
- `bun test src` — 7382 pass. The only 3 failures are the pre-existing `Resources Routes` "version prefix" tests (environmental: `getAppVersion()` returns the `v0.0.0-alpha` fallback in an untagged checkout; unrelated to this PR and unchanged by it).
- `bun test test/integration` — 721 pass, 0 fail.
- Frontend `vitest run` (full) — 13437 pass, 0 fail.
- New colocated tests added for every new file (AI service + providers, route, admin masking/validation, config gating, static bundle, `apiCallManager.getGenerateQuestions`, `common_edition` matrix, `admin/ai.js`, `pages` secret-masking).
- E2E: `test/e2e/playwright/specs/ai-configuration.spec.ts` added (disabled / external / managed UI rows). Not executed in the local sandbox (the Chromium headless-shell binary could not be downloaded here); it runs in CI.

## Reviewer verification checklist
- [ ] `make fix && make test-unit && make test-integration` are green.
- [ ] `make test-e2e` green, including `ai-configuration.spec.ts`.
- [ ] **Disabled** (`AI_FEATURES_ENABLED=false`): AI tab stays visible with the prompt + **Copy** (and Questions → Save); no selector / Send to AI / Generate; `defaultAI` absent from Preferences; `POST /api/ai/generate-text` returns a safe `AI_DISABLED` error.
- [ ] **External** (`AI_PROVIDER=external`): assistant selector + "Send to AI" present; `defaultAI` shown; opening a public assistant still works.
- [ ] **Managed** (`AI_PROVIDER=ollama`/`azure`/`openai_compat`): external selector hidden; "Create" generates questions via the backend; misconfigured provider shows a clear error and does **not** open a public AI.
- [ ] No secret leak: inspect `/api/parameter-management/parameters/data/list`, `/api/config`, `GET /api/admin/settings`, and `data/bundle.json` — none contain provider keys/endpoints.
- [ ] Admin **AI** tab: provider field groups toggle by provider; saving a blank API key keeps the stored key; "Test connection" reports success/failure.

---

### Review — @ignaciogros

> Thanks, @erseco.
>
> I wasn't expecting `AI_FEATURES_ENABLED=false` to have that behavior. I expected the AI tab to remain visible, but without the provider selector and the "Send to AI" button. That way, users can still copy and paste the code without any issues. Otherwise, I think we're removing a feature that is heavily used for game creation.
>
> I've tested with both Azure and Ollama as `AI_PROVIDER`, with and without a custom `BASE_PATH`, and the result has always been the same (on `localhost:8080` and without HTTPS):
>
> ```
> XHR POST
> http://localhost:8080/api/ai/generate-text
> [HTTP/1.1 502 Bad Gateway 11348ms]
> ```
>
> The UI shows the following message:
>
> > The questions could not be generated: The AI provider returned an error (HTTP 500).
>
> I get the same results with both `make up-local` and `make up`. The container logs do not show any related information.

### Response — @erseco (addressed in `66692fc20`)

Thanks for the careful review, @ignaciogros — both points are addressed in `66692fc20`.

**1. `AI_FEATURES_ENABLED=false` keeps the AI tab (copy-only)**

You're right: fully removing the tab dropped a workflow people rely on. Disabled mode now **keeps the AI tab visible** with the prompt and the **Copy** button (and the **Questions** → **Save** flow), but **without** the provider selector, the **Send to AI** button and the managed **Generate** surface. So teachers can still copy the auto-built prompt, take it to any AI tool of their choice, and paste the questions back — eXeLearning just never calls a provider itself.

| `AI_FEATURES_ENABLED` | `AI_PROVIDER` | Game editor |
|---|---|---|
| `false` | (any) | Prompt + **Copy** (+ Questions/Save); no selector / Send to AI / Generate |
| `true` | `external` | Prompt + assistant selector + **Send to AI** |
| `true` | `azure` / `ollama` / `openai_compat` | Server-side **Generate** / **Create** |

**2. The opaque `HTTP 500` and empty logs**

The `502` + *"The AI provider returned an error (HTTP 500)"* means the **upstream provider** (Ollama/Azure) actually replied `500` — for Ollama that is almost always *the model has not been pulled* — but we were **discarding the upstream body and logging nothing**, which is why the container logs were empty. That opacity was the real bug. Fixed by adding server-side observability:

- Every managed-provider failure is now logged with provider, endpoint host, HTTP status and a truncated response-body snippet, e.g.

  ```
  [ai] provider "ollama" at localhost:11434 returned HTTP 500: {"error":"model 'llama3' not found, try pulling it first"}
  ```

- The admin-only **Test connection** diagnostic (`POST /api/ai/test-connection`) now returns that snippet directly, so an admin sees the exact provider error without reading logs.
- The public `POST /api/ai/generate-text` response is unchanged and still returns only a safe `{ error, message }` (no upstream details) — nothing leaks to ordinary users.

For a local Ollama the usual fix is `ollama pull <AI_OLLAMA_MODEL>`; note that a `localhost` Ollama is **not** reachable from inside the `make up` container, so use a host-reachable address there (the docs now mention this).

## Fix — managed-mode Create inserted generated questions twice

In managed provider mode, the "Create" flow auto-inserted the generated questions AND then re-staged them into the Save textarea and switched to the Save tab, so clicking Save inserted every question a second time (N → 2N). Create now auto-inserts once and no longer primes the Save surface. Commit `2f0cd912`; reproduction test added.

## Fix (docs) — ai-configuration.md described behavior the code lacks

The troubleshooting section claimed managed-provider failures are logged server-side as `[ai] …` lines and that admin "Test connection" returns the upstream `details` snippet. Neither exists — `http.ts` discards the upstream body and the endpoint returns only a generic typed error. The doc now describes the actual no-leak behavior. Commit `bd7c3fc2`.


erseco:

- **Important:** `public/app/common/common_edition.js:1194-1196` — in managed mode the **Create** flow both inserts the questions (`saveQuestions(correctsQuestions)` → `insertAIContent`, which appends to `questionsGame`, e.g. `adaptative-quiz.js:1772`) *and* then repopulates `#eXeEQuestionsArea` and switches to the Questions tab, which shows the primed **Save** button (`#eXeESaveButton`, handler at `common_edition.js:1055`). Clicking Save — the surface the code navigates the user to — re-runs `validateAndSave` + `saveQuestions(validLines,…)` → `insertAIContent` on the same lines with no dedup, so every generated question is inserted twice (N → 2N). Pick one behaviour: stage-only (drop the auto-insert) or insert-only (don't repopulate/switch to the Save surface). This flow is only reachable in the new managed mode, so it's a newly introduced footgun.
- **Test gap:** the managed generate→Save path is untested. `common_edition.test.js:1120-1133` drives `genarateIAQuestons` with a `vi.fn()` mock and never fires a subsequent `#eXeESaveButton` click; the Playwright spec only asserts which controls render. Green CI cannot observe the double insert.
- **Minor (doc/claim mismatch):** `doc/development/ai-configuration.md:122-134` and the PR description state managed-provider failures are logged server-side as `[ai] …` lines and that Test connection returns the upstream `details` snippet. Neither exists at head: `src/services/ai/providers/http.ts:49-50` discards the upstream body on non-2xx, there is no logging anywhere under `src/services/ai/`, and `src/routes/ai.ts` test-connection returns only the generic `AiError.message` (no `details`). The prior reviewer's "opaque HTTP 500 / empty logs" complaint is not actually addressed by the code.

Verified: static reading of the diff/files at `dc5d00b917`. `CI: all checks green on dc5d00b ✅` — but CI does not exercise the managed generate→Save flow. Residual risk: I did not run the app; the double-insert should be confirmed manually (configure a managed provider, click Create, then Save, and count the inserted questions — expect N, observe 2N). Findings on the 14 serial settings reads (`config.ts loadAiConfig`) and the absence of a rate limit on `/api/ai/generate-text` are hardening suggestions, not defects, and are not blocking.

---
**Update — resolved (`2f0cd912`):** managed Create auto-inserts once and no longer repopulates `#eXeEQuestionsArea` / switches to the Save tab, so Save cannot re-insert (N→2N fixed); reproduction test added.

---
**Update — doc finding resolved (`bd7c3fc2`):** `ai-configuration.md` corrected to match the code (no `[ai]` server logs, no upstream `details` snippet; failures return a generic error).

mnarvaezm:
Hi everyone,
I have taken another look at the PR and introduced several changes to the AI question generator:

**Generate form interface.** The form container was clipped to the height of a single field, causing its content to overflow outside the box. In addition, the fields were displayed inline instead of being aligned below their labels. I have rebuilt the form using the Bootstrap grid: each field now appears below its label and uses the full available width, while the Topic field occupies the remaining space and the Create button is positioned to its right.

**Content of the prompt sent from the Generate tab.** The prompt was missing the specific instructions for each game and contained a duplicated code block. These issues have now been fixed.

**Prompt unification.** The Prompt tab displayed different text from the prompt that was actually sent and ignored the Specialty, Course and Topic fields. I have unified both prompts through a single generator. When the Prompt tab is available, it now displays exactly the same prompt that will be sent, including the context provided in the form. If an AI provider is configured, this tab is hidden and the Generate tab is selected by default when entering AI mode.

**Number of questions.** The number appeared twice: within the instructions for each game and in an additional sentence, sometimes with contradictory values. It is now specified only once. When the prompt is editable, the game’s default value is used; when a configured provider is used, the number entered in the form is applied.

**Review before insertion.** Generated questions are no longer added directly to the game. Instead, they appear in the Questions tab so that users can review and correct them before inserting them. Questions with an invalid format are also displayed, whereas they were previously discarded silently. When saving, invalid questions are identified and remain visible so that users can make minor corrections and reuse them.

On the other hand, this PR does not address the possibility of each user using their own AI account within eXeLearning. The configuration it introduces is installation-wide only: `app_settings` is a global key-value table with no per-user dimension, and it can only be modified by an administrator. The only individual preference is `defaultAI`, which does not contain credentials; it simply determines which public assistant is opened by the “Send to AI” button.

It would be useful to add the connection settings to the user preferences and store them on the user’s own computer rather than on the server. The order of preference when using AI could be:

* The user’s local configuration, if it exists and is complete.
* Otherwise, if the installation has a server, the installation-wide configuration.
* If neither is available, external mode using public assistants.

However, it would probably be better to leave this enhancement for a separate PR.

Best regards.

@erseco  @ignaciogros

PR #1968, opened 2026-06-18
Opaque editor preview v2: HTTP/app isolation, trusted static Service Worker, incremental revisions

erseco:
## Summary

This pull request establishes the final eXeLearning editor-preview architecture:

- **Opaque HTTP preview v2** for server/cloud, embedded hosts, and Electron through `app://`.
- **Trusted same-origin Service Worker preview** only for standalone static/PWA.
- **No authored-content `srcdoc` transport** and no silent fallback chain.
- **Incremental protocol v2** separating fixed resources, session assets, and generated-document revisions.
- **External-media relay** for opaque contexts.
- **Normalized `previewHttp` embedding contract** for Moodle, WordPress, Omeka S, Nextcloud, and Procomún.

ADR-0015 records the transport decision and its security consequences.

## Final transport matrix

| Runtime | Provider | Posture |
|---|---|---|
| Server/cloud | `HttpPreviewProvider` | opaque HTTP capability URL |
| Electron | `HttpPreviewProvider` over `app://` | opaque/cross-origin from renderer |
| Embedded host with valid `previewHttp` | `HttpPreviewProvider` | opaque; fails closed if missing or invalid |
| Standalone static/PWA | `StaticServiceWorkerPreviewProvider` | same-origin trusted-content mode, not a security sandbox |
| php-wasm playground | disabled by default; explicit development-only unsafe opt-in | never silently secure |

The provider set is intentionally limited to:

```text
HttpPreviewProvider
StaticServiceWorkerPreviewProvider
```

`SrcdocPreviewProvider`, `srcdocInliner`, Base64 resource budgets, runtime URL interception, custom srcdoc navigation, and popup handoff machinery are removed.

## Transport-selection security gates

Selection is deterministic and cannot be bypassed through a loose override:

- Server and Electron always use HTTP/app and reject a Service Worker downgrade.
- Embedded HTTP requires a valid protocol-v2 `previewHttp` block even when `previewTransport: "http"` is explicitly set.
- A standalone runtime forcing HTTP must provide valid endpoint configuration.
- Embedded same-origin Service Worker preview requires both fields:

```jsonc
{
  "previewTransport": "static-service-worker",
  "allowUnsafeEmbeddedPreview": true
}
```

The transport name alone is rejected. Unknown and removed override values fail closed.

## Static/PWA warning

Static/PWA remains a deliberate compatibility mode because a Service Worker cannot serve an opaque-origin iframe.

[...]

PR #1999, opened 2026-06-25
Add unified compact activity authoring format parser

### LINKED ISSUE #1228: A unified format using # separators and optional @parameters
(proposal from an user)

A unified format using # separators and optional @parameters:

Correct_answers]#[Question]#[Option_1]#[Option_2]#[...]@[OptionalL_Params]

Examples

- Multiple choice with explanation:
012#Which are the three main axes?#Poverty blame#Dehumanization#Dissent criminalization#Democratic participation@explain=The three axes are poverty as guilt, dehumanization, and dissent criminalization.

- True/False with hint:
1#Is the Earth round?#True#False@hint=Think about the horizon@explain=The Earth is a geoid, not flat.

- Fill-in-the-blank:
eXeLearning is a @@free@@ and open source editor to create @@educational@@ resources.@explain=eXeLearning creates interactive learning content.

- Flashcards:
Photosynthesis#Process by which plants convert light into energy@type=flashcard

So we can have a single parser for most iDevices, there is a backward compatibility - old formats auto-convertible - Less code duplication - Easy to extend (@hint, @points, @feedback, etc.)
Simpler maintenance. If you wish, it is possible a gradual migration: new format alongside existing ones, with automatic conversion tools for legacy content

erseco:
## Summary

Fixes #1228.

This PR adds a compact, line-based authoring parser for activity-like iDevices using `#` field separators and optional `@key=value` parameters.

The new format is implemented as an **authoring/import normalization layer**. It does **not** replace eXeLearning's canonical iDevice storage formats — iDevices keep their existing, per-type storage models. The parser turns one compact line into a normalized, iDevice-agnostic model; separate adapters convert that model into the concrete data structure of a specific iDevice.

## What changed

- **Parser** `src/shared/parsers/unified-activity-format.ts` — a pure (no DOM/fs) parser producing a normalized model for:
  - selection (single/multiple choice), true/false, fill-in-the-blank, and flashcard aliases.
- **Adapters** `src/shared/parsers/unified-activity-adapters.ts` — convert the normalized model into **confirmed** iDevice shapes:
  - selection → `form` `questionsData` (`activityType: 'selection'`)
  - true/false → `trueorfalse` game question `{question, feedback, suggestion, solution: 1|0}` (and, alternatively, a `form` `true-false` question)
  - fill-in-the-blank → `form` `questionsData` (`activityType: 'fill'`, blanks as `<u>answer</u>`)
  - flashcard → `flipcards` (Memory cards) card (`eText`/`eTextBk`, URI-encoded with the same `encodeURIComponentSafe` rule the editor uses)
- **Diagnostics** — every malformed/ambiguous input yields a stable, line-accurate diagnostic instead of a silent misparse; the parser never throws for user input.
- **Docs** — new `doc/elpx-format/idevices/unified-authoring-format.md` (grammar, escaping, parameter registry, iDevice mapping, diagnostics, limitations, migration). Linked from `llms.txt` and `doc/elpx-format.md`.
- **Catalog fix** — `doc/elpx-format/idevices/catalog.md` listed legacy `MultichoiceIdevice`/`EleccionmultiplefpdIdevice` under `quick-questions-multiple-choice`, but `MultichoiceHandler` and `LegacyXmlParser` both convert them to `form`. Moved them to the `form` row to match the code and the legacy-alias table.

## Design decisions

- `#` separates payload fields only when not escaped as `\#`; `@key=value` starts a parameter only when not escaped as `\@`. `@@…@@` (fill blanks) is never mistaken for a parameter because it is never followed by `=`.
- Escapes: `\#`, `\@`, `\\`, `\n`, `\t`. Unknown escapes are preserved literally with a warning (error in `strict` mode).
- `@type=` disambiguates; otherwise the kind is inferred (blanks → fill; 4 fields with a boolean + `true`/`false` labels → true/false; ≥3 fields with a digit-leading first field → selection; otherwise an `AMBIGUOUS_TYPE` error). **Flashcards always require `@type=flashcard`.**
- **True/false `1` means the statement is true** (not option index 1), matching the issue example. Tokens `1/true/t/yes/y` and `0/false/f/no/n` normalize to a boolean, then `solution: 1|0` / `answer: '1'|'0'`.
- Selection answers are 0-based option indexes; `012` is compact (0,1,2), `0,2`/`0|2` are also accepted.
- `@hint` maps to the iDevice `suggestion` field — **not** `hint`, which the `form` iDevice never reads (the legacy `MultichoiceHandler` writes a dead `hint` key; noted as a latent issue, left out of scope here).
- Adapters omit the question `id` (the editors generate ids), keeping output deterministic and testable, and never invent storage fields. Recognized parameters without a target field survive in the normalized model rather than being dropped.
- **Scope:** parser + adapters + docs + tests. A textual bulk-question UI already exists (`getTabIA`/`insertQuestions` in `public/app/common/common_edition.js`); wiring this format into it is a documented follow-up. No user-visible behaviour changes, so no E2E spec is needed and existing import/export is untouched.

## Adversarial review

The parser was hardened by an adversarial review pass that executed crafted
inputs against it. It surfaced (and this PR fixes, with regression tests) bugs
that the happy-path tests missed:

- `@@blank@@` answers containing `=` (e.g. `@@x=5@@`) were silently destroyed
  because the `@` of `@@` + `key=` matched the parameter boundary. The scanners
  now skip `@@…@@` regions when locating parameters.
- The parameter-boundary regex accepted keys that validation later rejected
  (e.g. `@8am=`), consuming payload then erroring. The boundary now matches only
  valid letter-leading keys; other `@token=` text stays literal.
- `@type=single` / `@type=multiple` did not constrain the selection mode (a
  `@type=single` line with two correct answers was silently accepted). They now
  seed the selection mode, with `@selection=` taking precedence.
- `@points=` (empty) silently became `0`; `@@@@` emitted a spurious
  `EMPTY_ALTERNATIVE` warning. Both fixed.

### Automated checks
- [x] `make fix` — clean (2 pre-existing `any` warnings in an unrelated file).
- [x] `make test-unit` — 7532 pass, 0 fail; all 222 files meet the 90% threshold.
- [x] `make test-coverage` — new parser and adapter files are at **100%** line/function coverage.
- [x] Frontend `vitest` — 13600 pass; the single failure (`focusedEditMode.test.js`) is a pre-existing flake that passes in isolation and is unrelated (no `public/` files changed).
- [x] `make test-e2e` — N/A (no user-visible editor/import/export behaviour changed).

### Parser checks
- [x] Single- and multi-answer selection (compact/comma/pipe indexes).
- [x] True/false using boolean semantics (the issue example).
- [x] Fill-in-the-blank with `@@answer@@` (and `@@a|b@@` alternatives).
- [x] Flashcard alias lines.
- [x] Escaped `#`, `@`, `\`, `\n`, `\t`; Unicode/emoji; CRLF; long input; literal `<script>` kept as text.
- [x] Diagnostics for out-of-range/duplicate indexes, empty fields, duplicate/unknown/invalid params, malformed blanks, ambiguous types; batch line numbers.

### Adapter checks
- [x] Selection → `form` questionsData (`[isCorrect, text]` tuples, `selectionType`).
- [x] True/false → `trueorfalse` `{question, feedback, suggestion, solution}`.
- [x] Fill → `form` `fill` with `<u>` gaps; flashcard → `flipcards` card.
- [x] `@hint` → `suggestion` (never `hint`); unsupported params not silently lost.

### Documentation checks
- [x] Grammar, escaping, parameter behaviour, iDevice mapping and limitations documented; examples match the tests; linked from the docs indexes.

## Notes for reviewers

Please pay special attention to:
- The distinction between this authoring format and canonical ELPX/iDevice storage.
- The true/false `1#Question#True#False` boolean semantics.
- The `MultichoiceIdevice` → `form` mapping (code + catalog now agree).
- Whether true/false should primarily target `trueorfalse` (current) or `form` (both adapters provided).

## Architecture records

- **SDD:** `doc/architecture/sdd/SDD-0008-unified-compact-activity-authoring-format.md`
- **ADRs:**
  - `doc/architecture/adr/ADR-0032-compact-activity-authoring-format-dsl.md`
  - `doc/architecture/adr/ADR-0033-authoring-parser-adapter-normalization-layer.md`
  - `doc/architecture/adr/ADR-0034-diagnostic-driven-non-throwing-parser.md`

## Fix — fill-blank alternatives encoded with extra pipes

Compact blanks with alternatives were encoded as `<u>|wet|moist|</u>` instead of `<u>wet|moist</u>`. The wrapping pipes produced empty-string alternatives (`['','wet','moist','']`), which the `form` runtime and editor accepted as a valid answer. `renderFillBaseText` now emits bare pipe-separated alternatives. Commit `2a6a4ad1`; spec corrected.


erseco:
cc @mnarvaezm — you're our foremost expert on the iDevices, so I'd really value your eyes here. This PR maps the compact authoring syntax onto the existing iDevice models as follows:

- selection / multiple-choice → `form` `questionsData` (`activityType: 'selection'`, `[isCorrect, text]` answers)
- true/false → `trueorfalse` game question `{question, feedback, suggestion, solution: 1|0}` (and, alternatively, a `form` `true-false` question)
- fill-in-the-blank → `form` `questionsData` (`activityType: 'fill'`, blanks as `<u>answer</u>`)
- flashcard → `flipcards` card (`eText`/`eTextBk`)

A couple of points I'd especially like you to sanity-check: that hints should map to `suggestion` (not the `hint` key, which the `form` iDevice never seems to read), and that legacy `MultichoiceIdevice` truly converts to `form` (I aligned `catalog.md` with `MultichoiceHandler`/`LegacyXmlParser` on this).

Could you let me know whether these mappings are correct, or whether there are actually flaws I've missed? Please feel free to discard this entirely if it's nonsense — no hard feelings. Thanks a lot!

erseco:
@ignaciogros, could you give us your opinion on whether this PR is worth keeping?

Right now we do not have a clear stakeholder for #1228: @cristinavaldera does not remember who originally requested it, and @mnarvaezm has already raised doubts about its practical usefulness, especially if the current workflow is based on the AI question format rather than this proposed compact/XML-related format.

My understanding is that #1999 is technically reasonable, but it adds a parser/adapters layer that is not currently exposed in the UI and that nobody has clearly confirmed they need.

So the decision seems more product-related than technical:

* If you see a real use case for this compact authoring/import format, we could keep it, probably changing `Fixes #1228` to `Refs #1228` or `Partially addresses #1228`.
* If not, I think it may be better to close it to avoid adding code and maintenance surface for something we are not sure anyone will use.

Since you know the product direction better, I would leave the final decision to you.

ignaciogros:
I haven't reviewed the details of the proposal yet, but I think standardizing/normalizing the data input and output formats is essential. In that sense, I think this proposal is solid, and we should take it into account as part of the iDevices review process and the definition of their development API.

What I'm not sure about is which format we should use, because development is already quite advanced and there are many existing examples based on the current syntax. I think we need to read @mnarvaezm's opinion before making a decision.

Earlier, I suggested using GIFT (General Import Format Technology), partly because it would make integration with Moodle question banks really easy. I believe this discussion should be part of the API definition process.

In my opinion, it's worth evaluating this carefully before making any changes, to avoid having to revise the process more than once.

erseco:

- **Important:** `src/shared/parsers/unified-activity-adapters.ts` (`renderFillBaseText`) wraps multi-alternative fill blanks in pipes — `@@wet|moist@@` → `<u>|wet|moist|</u>`. The `form` runtime populates `fillAnswer` from the trimmed `<u>` inner text (`export/form.js:1192-1203`) and grades via `.text().split('|')` with no empty-string filter (`export/form.js:1706-1708`), so `|wet|moist|` splits to `['','wet','moist','']`. The empty answer is accepted: in strict mode an empty submission grades correct; in the default levenshtein-≤1 mode any single-character submission also matches `''`. The canonical encoding for alternatives is unwrapped `a|b` (the runtime and editor both split on bare `|`). Fix: emit `<u>${answers.join('|')}</u>` and update the spec assertion (`unified-activity-adapters.spec.ts:124-126` currently enshrines the wrapped output). Note this is latent today — no production code imports these adapters (the PR is a normalization layer with wiring deferred) — but the follow-up that wires it in inherits an always-correct grade for empty/1-char input on every multi-alternative blank.
- The second proposed finding (unescaped `@key=` inside payload text truncating the payload) is working-as-designed: the parser is non-throwing, emits a line-accurate diagnostic, and the `\@` escape is documented. Not a defect.

Verified: code paths re-inspected at head; grading mechanics traced through `form.js` export runtime. CI: all checks green on e89b34e ✅.

Residual risk: no local build/test/server was run by this reviewer — the empty-answer grade path was confirmed by reading `form.js`, not executed. Because the adapter is unwired, no integration/E2E currently exercises it, so this cannot surface until the documented wiring lands.

---
**Update — resolved (`2a6a4ad1`):** `renderFillBaseText` emits `<u>a|b</u>` (no wrapping pipes), matching how the runtime/editor split `<u>` inner text; spec corrected.

erseco:
Thanks everyone for the feedback, and especially @ignaciogros for raising the broader API question.

We have now recovered the original request that led to #1228. The proposal came from F. Petrossi, one of the Italian translators, and its main motivation was to reduce the number of different parsers/formats used by similar interactive activities, with the expectation of reducing duplication and maintenance costs.

After revisiting that requirement, I don't think this PR is the right solution to merge at this point.

#1999 implements the proposed `#` / `@parameter` syntax as an additional authoring/import normalization layer, while deliberately keeping the existing per-iDevice storage formats and parsers. This means that it does not currently remove the duplication that motivated the original request; instead, it introduces another format, parser and adapter layer that we would also need to maintain.

The normalization work itself may still be useful, but I agree that we should first define the iDevice API and decide which interchange/authoring format we actually want to support. That discussion should also consider existing formats such as GIFT and potential Moodle interoperability before committing to a custom DSL.

Since there is no user-facing integration depending on this code yet, I think the best option is to close this PR rather than merge an API/format that we may later need to replace.

The work and discussion here remain useful input for the future iDevice API/normalization design.

Closing this PR as not planned.

PR #2001, opened 2026-06-25
Prepare v4.0.2 release

### LINKED ISSUE #1992: Prepare v4.0.2 release
- Update changelog.
- Update third-party packages list.
- Do a full translation maintenance cycle.
- Complete Spanish translation.
- Add placeholder translations for new strings in incomplete translations.
- Update styles documentation (no `downloadable` config option is required now).
- Remove de `downloadable` config option from all the default styles.
- Add new agents rule to avoid string extraction and translation.

ignaciogros:
- Updated changelog.
- Updated third-party packages list.
- Full translation maintenance cycle.
- Complete Spanish translation.
- Placeholder translations for new strings in incomplete translations.
- Update styles documentation (no `downloadable` config option is required now).
- Remove de `downloadable` config option from all the default styles.
- Add new agents rule to avoid string extraction and translation.

### DIFF
--- a/.agents/skills/i18n/SKILL.md
+++ b/.agents/skills/i18n/SKILL.md
@@ -45,24 +45,23 @@ The distinction matters because the GUI language and the content language can di
 ## Commands
 
 ```bash
-make translations                        # Extract new translation keys from source
-make translations-cleanup LOCALE=es      # Remove obsolete keys for a locale
-make fix                                 # Lint
+make fix                                 # Lint (the only command this skill ever runs)
 ```
+
+> `make translations` and `make translations-cleanup` are **never run by agents**. Key extraction and all XLF changes are managed by a dedicated separate process. See the prohibition below.
 
 ## Gotchas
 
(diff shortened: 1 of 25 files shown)


erseco (approved):
Approving for the 4.0.2 release. ✅

One addition folded into this release PR: I've **temporarily hidden the Galicia (`ES-GA`) LOMLOE dataset** by setting `available: false` on its `DATASETS` entry. The dataset JSON stays in the repo untouched — only the availability flag is off, so it's no longer selectable and re-enabling it later is a one-line change.

The reason is the open discussion in #1900: the Galicia concretion still needs **formal decisions on its official curriculum requirements** before it can faithfully match the regulation — specifically (1) a hybrid "mandatory + optional" descriptor mode that the iDevice doesn't support yet, and (2) the official Galician descriptor labels. Rather than ship a half-supported behaviour that wouldn't align with the official curriculum, we prefer to keep it hidden until those requirements are defined.

Big thanks to @franmate for the work and the patience throughout #1900 — this isn't a rejection of the contribution, quite the opposite: we want to land it properly once the spec is settled. The issue/PR stays open as the reference to pick it back up.

Changes in this commit:
- `lomloe.js`: `ES-GA` → `available: false` (with a comment pointing to #1900 / #1898)
- `lomloe.test.js`: regression guard updated to assert the on-hold flag; the per-course-filter test now exercises `ES-NC` instead of `ES-GA`
- `CHANGELOG.md`: noted under v4.0.2 → Changed

PR #2002, opened 2026-06-25
Allow renaming page titles from the workspace

### LINKED ISSUE #1600: Rename nodes in the editing area
Suggested by @jcayetano-spain in #1518:

> For better usability, it would be highly intuitive to allow renaming pages by double-clicking their title directly in the main editing workspace (at the top), rather than forcing the user to do it exclusively from the left-side structure tree.

erseco:
## Summary

Closes #1600.

This PR lets users rename the current page directly from the main editing workspace by editing the page title shown at the top of the page.

The interaction is aligned with the existing box/iDevice title editing UX — a **single click** on the title (or on a **pencil "Edit title" button**) enters inline edit mode — and the change is written through the **same canonical page/node rename path used by the structure tree**, so the workspace heading, the structure tree, and the underlying project data stay in sync. This is not a cosmetic DOM mutation: the source of truth is the same one used when renaming from the tree.

### Handling the visible-vs-node title distinction

As noted on the issue, the workspace `<h1>` does not always show the node name. The implementation follows an **"edit what you see"** rule:

- **Default page** (heading shows the node name) → editing renames the node (`titleNode`) via `structureEngine.renameNodeAndReload(...)`, so the structure tree updates too.
- **Page with a separate in-page title** (`editableInPage`, heading shows `titlePage`) → editing updates only that `titlePage` via the page-properties path (`node.apiSaveProperties({ titlePage })`), leaving the tree node name untouched.
- **Hidden title** (`hidePageTitle`) → not editable from the workspace.

## What changed

- **New shared helper** `public/app/workarea/project/idevices/inlineTitleEditor.js` — encapsulates the contenteditable edit lifecycle (Enter/blur save, Escape cancel, trim, reject empty/whitespace-only/unchanged). It is reused by **both** the workspace title and the structure-tree inline rename, so there is no duplicated title-editing logic.
- **`idevicesEngine.js`** — wraps the page title in a `.content-editable-title` container with a pencil "Edit title" button (same markup/classes as box titles). A single click on the title or the pencil enters edit mode; the pencil is hidden while editing.
- **`menuStructureBehaviour.js`** — the tree's `startInlinePageRename` now delegates its lifecycle to the shared helper (behaviour preserved, covered by its existing tests).
- **i18n** — the only new visible string is the editable control's accessible label, wrapped in `_('Page title')`; the pencil reuses the existing `_('Edit title')` key. No `translations/` files are touched — registering the new key in the catalog is left to the translation team.
- **Styles** — `assets/styles/layout/_nodecontainer.scss`: inline layout for the title + pencil and a focus outline while editing.
- **Tests** — colocated Vitest specs and a Playwright E2E spec (below).

## How it was fixed

The workspace page title is no longer a static-only heading. Clicking it (or the pencil) enters inline edit mode seeded with the current raw title. On confirmation the value is normalized (trimmed) and rejected if empty/whitespace-only/unchanged; otherwise it is routed to the matching canonical path above, after which the heading, the structure tree, and the Yjs model all reflect the new title. Escape cancels and restores the previous value.

## Tester checklist

- [x] Open or create a project with at least one page.
- [x] Go to the editing workspace.
- [x] Click the page title at the top of the page (or the pencil button next to it).
- [x] Rename the page to a new valid title and confirm with Enter (or by clicking away).
- [x] Check that the workspace title displays the new title.
- [x] Check that the left structure tree displays the same new title.
- [x] Switch to another page and return to confirm the title remains changed.
- [x] Save and reload/reopen the project, and confirm the title persists.
- [x] Try entering an empty title and verify it is rejected (previous title restored).
- [x] Try entering only spaces and verify it is rejected.
- [x] Press Escape while editing and verify the previous title is preserved.
- [x] Confirm that renaming from the structure tree still works.
- [x] Confirm that box/iDevice title editing still works.
- [x] Confirm there are no browser console errors.

## Tests

Commands run:

```bash
make fix
npx vitest run public/app/workarea/project/idevices/inlineTitleEditor.test.js
npx vitest run public/app/workarea/project/idevices/idevicesEngine.test.js
npx vitest run public/app/workarea/menus/structure/menuStructureBehaviour.test.js
make test-frontend
bun x playwright test --project=chromium test/e2e/playwright/specs/page-title-workspace-rename.spec.ts
```

Results:

- `make fix` — clean.
- Vitest (helper + engine + tree): **390 passed**. Patch coverage on changed lines: `inlineTitleEditor.js` 100% lines; new `idevicesEngine.js` methods 98.3%; the tree refactor is fully covered by its existing suite.
- `make test-frontend` — green (one pre-existing, unrelated flake in `focusedEditMode.test.js` that passes when run standalone; it does not touch any file changed here).
- E2E — the new spec covers: click/pencil rename → workspace heading + structure tree + Yjs model sync; persistence across page switch and save/reload; Escape cancel; empty/whitespace rejection. **3/3 passed.**

> Backend integration tests were not re-run: this change is entirely frontend (`public/app/**`, `assets/styles/**`, `translations/**`, `test/e2e/**`); no `src/**` files are touched.

## Notes

- **E2E test added: yes** (`test/e2e/playwright/specs/page-title-workspace-rename.spec.ts`), and it runs on the standard `chromium` project in CI.
- The bundled Playwright chromium download stalled in this local environment (a CDN/network issue, not a code issue), so the committed spec was executed locally against the installed Google Chrome (`channel: 'chrome'`) via a temporary config that was removed afterwards. The spec itself is unchanged.

### DIFF
--- a/assets/styles/components/_buttons.scss
+++ b/assets/styles/components/_buttons.scss
@@ -6,6 +6,40 @@
     padding: 0;
 }
 
+// Edit-title button: hidden accessibly by default, revealed on hover/focus-within
+.content-editable-title {
+    .btn-edit-title {
+        position: absolute;
+        width: 1px;
+        height: 1px;
+        padding: 0;
+        margin: -1px;
+        overflow: hidden;
+        clip: rect(0, 0, 0, 0);
+        white-space: nowrap;
+        border: 0;
+        opacity: 0;
+        transition: opacity 0.2s ease;
+    }
+
+    &:hover,
+    &:focus-within {
+        .btn-edit-title {
+            position: static;
+            width: auto;
+            height: auto;
+            padding: 0;
+            margin: 0 0 0 .5em !important;
+            margin-inline-start: 0.25rem;
+            overflow: visible;
+            clip: auto;
+            white-space: normal;
+            border: 0;
+            opacity: 1;
+        }
+    }
+}
+
 // Ensure translated text spans inside buttons inherit the button's own color
 // (e.g., white for .btn-primary, white for .btn-danger) instead of var(--text).
 // Static mode wraps {{ 'Text' | trans }} in <span data-i18n="Text"> for runtime translation.
--- a/assets/styles/layout/_nodecontainer.scss
+++ b/assets/styles/layout/_nodecontainer.scss
@@ -71,8 +71,33 @@
     }
     #node-content {
         height: auto;
+        // Page title with inline rename affordance (single click on the title
+        // or the pencil button), mirroring box/iDevice title editing.
+        .content-editable-title {
+            display: flex;
+            flex-flow: row nowrap;
+            align-items: center;
+            gap: 4px;
+            .btn-edit-title {
+                cursor: pointer;
+                .small-icon {
+                    width: 16px;
+                    height: 16px;
+                }
+            }
+            // Hide the pencil when the page title itself is hidden/empty.
+            .page-title.hidden ~ .btn-edit-title {
+                display: none;
+            }
+        }
         .page-title {
             margin-top: 15px;
+            cursor: text;
+            &[contenteditable="true"] {
+                outline: 2px solid var(--dark-exe-color);
+                outline-offset: 4px;
+                border-radius: 2px;
+            }
         }
         &[mode="edition"] {
             #eXeAddContentBtnWrapper {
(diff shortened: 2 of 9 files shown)


mnunezcedec (approved):
One more step forward, thank you.

ignaciogros (approved):
Thanks, @erseco!

I suggest the button to be hidden accessibly by default, and revealed on hover/focus-within.

My last commit does that. Please feel free to revert it if you find any problem.

ignaciogros:
> I suggest the button to be hidden accessibly by default, and revealed on hover/focus-within.

Maybe we should do that with the iDevice title button too. What do you think?

erseco:
@ignaciogros i like it, go ahead with the others!

ignaciogros:
OK. Thanks. I'll do the iDevice title part in a separate issue: #2005

PR #2008, opened 2026-06-28
`exe_media` is no longer required for `rel="lightbox"`.

### LINKED ISSUE #1548: exe_media is no longer required for rel="lightbox"
When exporting content that includes links with `rel="lightbox"` pointing to a video or audio file, the `exe_media` library (MediaElement.js) is also exported. This is no longer necessary, because most of the browsers in use natively support those formats, and the Flash fallback is no longer used.

ignaciogros:
How to test:

Create three links with `rel="lightbox"` or `rel="lightbox[galleryID]"`: one linking to an image, one to an MP3 file and one to an MP4 file. Verify that all three are displayed correctly, without using the old JavaScript player.

### DIFF
--- a/doc/elpx-format/libraries.md
+++ b/doc/elpx-format/libraries.md
@@ -99,7 +99,6 @@ const { files: allRequiredFiles, patterns } = this.getRequiredLibraryFilesForPag
 | `exe_tooltips` | Tooltips iDevice | `exe_tooltips/jquery.qtip.min.js`, `exe_tooltips/jquery.qtip.min.css` |
 | `exe_magnify` | Image magnifier iDevice | `exe_magnify/exe_magnify.js`, CSS |
 | `exe_media` | Embedded media (audio/video) iDevice | `exe_media/exe_media.js`, CSS |
-| `exe_media_link` | Linked media iDevice | `exe_media_link/exe_media_link.js` |
 | `abcjs` | ABC music notation iDevice | `abcjs/abcjs-basic.js` |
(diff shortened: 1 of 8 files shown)


erseco (approved):
Reviewed exhaustively and verified functionally against the real export runtime on `make up-local` (image + MP3 + MP4 lightbox links): all three open correctly in prettyPhoto with native HTML5 controls, the legacy `mediaelementplayer` is never invoked, and the video container is sized correctly. The removal of `exe_media_link` is clean — no dangling references and `exe_media` stays available for the embedded-media iDevice.

Multi-angle review surfaced one actionable, confirmed (minor) issue, which I have fixed and pushed to this branch:

- The download-link label derived the extension with `src.split('.')[1]` (second dot-segment, not the last). The new `<source>`-based extraction makes this reachable for videos too, so `lesson.part1.mp4` was labelled `part1` and `clip.mp4?token=abc` leaked the query string. Fixed to take the last dot-segment and strip query/fragment.
- Added frontend tests for the new branches (extraction from `<source>`, video-height recalc for both >0 and zero-height, and the corrected multi-dot/query label). This also resolves the failing `codecov/patch` check.

After the changes: `common.test.js` 361/361, full frontend suite 13714/13714, backend export specs 144/144, `make fix` clean.

The other observations are intentional (loss of the MediaElement.js skin is the PR's goal; native controls remain) or negligible (`.flv` only ever played via dead Flash). LGTM.

[PR-2008-review-EN.pdf](https://github.com/user-attachments/files/29440165/PR-2008-review-EN.pdf)

PR #2011, opened 2026-06-30
feat(idevice): restore file attachment iDevice

### LINKED ISSUE #1858: [Feature Request / Regression] Restore the "File Attachment" iDevice removed in eXeLearning 4.0
**Summary**
The **File Attachment** iDevice, which was available in eXeLearning 2.9, is no longer present in version 4.0. This is a significant regression for educators who use it regularly to provide downloadable resources directly within their learning packages.

**Expected behavior (v2.9)**
In eXeLearning 2.9, there was a dedicated **File Attachment** iDevice that allowed authors to:
- Attach one or more files directly to a page node.
- Display a clear, visual download link for the learner (with a file icon and filename).
- Embed the file inside the `.elp`/exported package with no extra steps.

**Current behavior (v4.0)**
The File Attachment iDevice does not appear in the iDevice panel. According to the official eXeLearning 4.0 manual: https://descargas.intef.es/cedec/exe_learning/Manuales/manual_exe40/

The only known workaround is:
1. Add a **link** iDevice or a text block with a hyperlink.
2. Go to **Sources** and upload the file there.
3. Manually copy the generated URL and paste it into the link.

This multi-step process is cumbersome and, more importantly, **the result is a plain text hyperlink**, which is not visually intuitive for young learners. It lacks the file-type icon and the clear "download" affordance that the old iDevice provided.

**Impact**
This affects teachers and instructional designers who build content for **primary and secondary school students**. A simple, visual file download element is essential for usability in educational contexts.

**Request**
Please restore the **File Attachment** iDevice in eXeLearning 4.0 (or add an equivalent component), so that authors can:
- Attach files with a single, dedicated iDevice.
- Have the exported page display a visually clear download link (icon + filename).
- Keep the workflow simple for non-technical educators.

**Environment**
- eXeLearning version affected: **4.0**
- Last version where the feature worked: **2.9**
- Reference manual (v4.0, no File Attachment iDevice listed): https://descargas.intef.es/cedec/exe_learning/Manuales/manual_exe40/

**Screenshots**
- *(Attached: screenshot from v2.9 showing the File Attachment iDevice in the panel)*

Image

*Thank you for the great work on eXeLearning. This iDevice is very useful in everyday classroom content creation and its absence is noticed immediately.*

erseco:
## Summary

Closes #1858.

Restores the **File Attachment iDevice** as a modern, Media Library-backed iDevice. Authors can attach one or more existing/uploaded files to a page and render a clear learner-facing download area with file-type icons, filenames (or author titles), optional descriptions, and file sizes.

Built against the eXeLearning 4.x architecture (JSON api-v3 iDevice + AssetManager); the eXeLearning 2.9 class was used only as a functional reference.

cc @greynoa for review and sign-off, and @ignaciogros to help us give it a nicer visual style.

## What changed

- New `file-attachment` iDevice under `public/files/perm/idevices/base/file-attachment/`, auto-discovered via `config.xml` (`component-type: json`, `api-version: 3.0`).
- **Edition UI:** add/select/upload files through the existing Media Library (with a native-input fallback that still routes through the AssetManager). Multiple attachments, reorder, remove, optional per-attachment title/description in a collapsible section, and a collapsible rich-text **Instructions** panel (shown to learners only when it has content).
- **Export:** one accessible download link per attachment ("Download <name>"), self-contained inline SVG file-type icons, a safe `download` filename, missing-asset fallback, and long-filename wrapping. All user-controlled text is escaped.
- **Asset model:** attachments store a stable `asset://<uuid>.<ext>` reference in `jsonProperties` — the AssetManager binary is the source of truth, packaged under `content/resources/` on export. Editor/preview resolves references live; static/server exports use packaged paths.
- **Media Library sync:** the editor reconciles attachments with the live asset metadata. Renaming a referenced file updates the attachment's name/icon/URL; deleting it flags the attachment as missing (keeping the reference so the author can re-add or remove it).
- Registered `file-attachment` in the `jsonOnlyIdevices` allow-list in `exe_export.js`, and updated the iDevice catalog/patterns/snippets docs.

## Legacy import

Legacy aliases now import as `file-attachment` (previously `text`): `FileAttachIdevice`, `FileAttachIdeviceInc`, `AttachmentIdevice`. The converter preserves the instructions, the per-file descriptions (used as the link label) and `showDesc`; missing/duplicate/empty legacy files are handled without crashing.

## How to test

A guide for testers — tick each step as you go.

- [ ] **Add the iDevice:** on a page, open the iDevices panel → *Information and presentation* → **File attachment**.
- [ ] **Add files:** click **Add file** and upload or pick one or more files from the Media Library. Each appears as a row with a file-type icon and the filename.
- [ ] **Title/description:** expand **Title and description** on a row and set a title and a description.
- [ ] **Instructions:** expand the **Instructions** panel and add some rich text.
- [ ] **Reorder/remove:** use the up/down arrows to reorder and the ✕ to remove an attachment.
- [ ] **Save the iDevice** and confirm the rendered view shows one download link per attachment ("Download …") with the icon, filename and size; the description shows when present; the instructions show only when you added text.
- [ ] **Preview:** open the preview panel and confirm the download links are present and work.
- [ ] **Media Library — rename:** rename one of the referenced files in the Media Library and confirm the attachment's name/icon update in the iDevice.
- [ ] **Media Library — delete:** delete a referenced file and confirm the attachment is flagged as missing in the editor and degrades gracefully in preview/export (no broken link or crash).
- [ ] **Export:** export HTML5 (and SCORM/IMS/EPUB if convenient) and confirm the files are packaged under `content/resources/` and the links download them.
- [ ] **Legacy import:** open an eXeLearning 2.x `.elp` containing a File Attachments iDevice and confirm it imports as a **File attachment** iDevice (not Text), preserving the instructions, file descriptions and the attached files.
- [ ] **Accessibility:** the controls are keyboard-operable and links read as "Download <name>".

## Architecture records

- **SDD:** `doc/architecture/sdd/SDD-0009-file-attachment-idevice-restoration.md`
- **ADRs:**
  - `doc/architecture/adr/ADR-0035-file-attachment-json-idevice-media-library.md`
  - `doc/architecture/adr/ADR-0036-file-attachment-asset-reference-model.md`
  - `doc/architecture/adr/ADR-0037-legacy-file-attachment-import-remap.md`
  - `doc/architecture/adr/ADR-0038-file-attachment-resilience-accessibility.md`

## Fix — `file-attachment` was treated as HTML on the browser export path

The browser export path now marks `file-attachment` as a JSON iDevice, matching its `config.xml` (`<component-type>json</component-type>`) and the server loader. Previously `getIdeviceConfig('file-attachment')` returned `componentType: 'html'`, so `IdeviceRenderer` never emitted `data-idevice-component-type="json"` and browser preview/export skipped `renderView`/`renderBehaviour` (broken preview download links / empty `db-no-data` render). Commit `d0cb0fba`; regression test added.


greynoa:
This is brilliant — exactly the fix I've been hoping for!  Tested it and it works like a charm. Huge thanks for tackling this, @erseco — great job as always. Green light from me, let's ship it! 🚀

franmate:
Tested this with an eXeLearning 2.x .elp file containing a File Attachments iDevice. I can confirm that it imports correctly as a File Attachment iDevice. The instructions, file descriptions, and all attached files were perfectly preserved.

This is a great solution for maintaining content compatibility with eXeLearning version 2. In my opinion, it should be merged into main with high priority once this issue is resolved:
To see the files, it is necessary to edit the iDevice. If an ELP is opened without editing the iDevice, the files are not shown in the export, even though they do appear in the editor.
attachment

erseco:
@franmate, thank you for the thorough review. This is now fixed: imported legacy File Attachment iDevices generate their initial file list immediately, so the files are visible in preview/export without opening and saving the iDevice first. Regression and integration tests were added, and CI is green. Could you please test it again?

franmate:
I can confirm that the changes work perfectly now. When opening the ELP file, the attached files are correctly displayed in the export without needing to edit the iDevice first. Thank you so much for the fix!

cristinavaldera (approved):

cristinavaldera:
Maybe @ignaciogros can take a look to the interface

franmate:
Feature Request: Additional File Property Option

I believe it is important to add a new option to the properties section of each file within the iDevice to improve link behavior.

### Detailed Requirements

1. Toggle/Switch for `target` attribute:
  Description: Allow the user to choose whether the file should open in a new tab/window.
   UI/UX: An on/off switch or checkbox.
     ON  (Default): Applies `target="_blank"` (Opens in a new tab/window).
     OFF: No target attribute or `target="_self"` (Opens in the same tab/window).

(Particularly useful for PDF documents)

erseco:
@franmate, this has been implemented.

### Changes

- Added a per-file **Open in a new tab or window** checkbox under **File properties**.
- The option is enabled by default, including for existing and imported attachments that do not yet contain the new property.
- When enabled, the exported link includes `target="_blank"` and `rel="noopener noreferrer"`.
- When disabled, the link has no `target` attribute and opens in the current browsing context.
- Added frontend tests for the default, enabled, disabled, persistence, Media Library, and missing-file cases.

Both **Build & Test Docker Images** and the complete **E2E Tests** workflow are passing on commit `f707ee7`.

One browser-level detail: the links still retain the existing `download` attribute, so a browser may download the file directly rather than navigate to a new tab, depending on the file origin and browser behavior.

erseco:

- **Important:** `src/shared/export/browser/idevice-config-browser.ts:53` — the hardcoded `jsonIdevices` allow-list (lines 53-76) omits `file-attachment`, while `config.xml` declares `<component-type>json</component-type>`. In the browser bundle `idevice-config` is aliased to this file, so `getIdeviceConfig('file-attachment')` returns `componentType:'html'`. `IdeviceRenderer.ts:143` therefore never emits `data-idevice-component-type="json"`, and `exe_export.js:287` (which only collects `json` nodes) never invokes `renderView`/`renderBehaviour` for this type — making the PR's own `jsonOnlyIdevices` entry (`exe_export.js:371`) inert on the browser path. Concrete impact: in the **preview panel**, `fixAssetUrls` intentionally keeps `asset://` hrefs (IdeviceRenderer.ts:139), and `renderBehaviour` is the only code that resolves them to blob URLs — so preview download links stay `href="asset://…"` and don't work; a file-attachment with an empty stored htmlView (`db-no-data`, e.g. legacy import with intro but zero resolvable files) renders nothing on the browser path while server/CLI export shows it. Add `'file-attachment'` to the browser `jsonIdevices` list to match `config.xml` and the server loader.
- **Test gap:** the colocated `src/shared/export/browser/idevice-config-browser.spec.ts` was not updated to assert `getIdeviceConfig('file-attachment').componentType === 'json'`, and the e2e preview spec only asserts the link is visible and has a `download` attribute — both present in the stored htmlView regardless of whether the JSON re-render runs — so green CI does not exercise the broken browser render path.

Verified: claims traced through `idevice-config-browser.ts`, `IdeviceRenderer.ts`, `exe_export.js`, `fixAssetUrls`, and the export/edition `file-attachment.js` at the reviewed head. `CI: all checks green on f707ee7 ✅` (the browser HTML5 *download* export still works because `fixAssetUrls` rewrites `asset://`→`content/resources/` in export mode; the defect is on the preview runtime and the db-no-data path).

Residual risk: the preview href-resolution impact is derived from the code (no local build/preview was run). The two other proposed findings were reviewed and are not blockers: the legacy escaped-href vs. rewriter mismatch is real but the JSON `attachments[].url` (source of truth) is correctly remapped and the effect is edge-only; the Yjs asset observer's lazy unobserve self-heals on any subsequent asset-map change.

---
**Update — resolved (`d0cb0fba`):** `'file-attachment'` added to the browser `jsonIdevices` list so it reports `componentType: 'json'`; regression test asserts parity with `config.xml`.

greynoa:
Overall it still works really well. I inserted the iDevice (Information and presentation → File attachment), added files from the Media Library, and confirmed each attachment shows as a row with a file-type icon and filename. Title/description editing works cleanly per file, and the Instructions panel saves and renders correctly (and only shows up when there's actual content). Reordering with the up/down arrows and removing attachments both work as expected. The rendered view shows one download link per file with icon, filename and size, and the description shows right below when present. I also like that each download link has hidden "Download" text for screen readers, in addition to the visible title and filename. The Preview panel mirrors all of this correctly.

I also tested Media Library integration: renaming a referenced file updates it correctly in the iDevice once you reopen/save it, and deleting a referenced file gets flagged clearly ("The attached file is missing. Re-add it from the Media Library.") without crashing — nice touch. That said, while doing the rename/delete I saw this in the browser console both times:

`[MediaLibrary] Failed to rename asset: TypeError: Cannot read properties of undefined (reading 'refreshAttachmentsFromAssets')` (same message for delete).

The iDevice still ends up looking correct after reopening the edit panel, but this error shouldn't be happening silently in the background — as far as I can tell it's still there on the latest commit.

More importantly, when exporting to HTML5 ("Sitio web") I got a real, repeatable error:

`ReferenceError: c_ is not defined`, logged as `[exe_export] Could not load template: c_ is not defined`.

This happens every single time I export, so I'd consider this blocking. It looks unrelated to the other export fix that was already merged for this iDevice — this seems to be its own separate issue, still present.

One more thing, on accessibility: "Open in a new tab or window" is checked by default for every new file, and when it's on it just opens the link in a new tab with no warning of any kind — no label, no extra text, nothing that tells someone (especially people using a screen reader) that it's about to open in a new tab. This can be disorienting: they lose their place and the back button doesn't take them back to where they were. I'd suggest making this off by default instead of on, and/or adding a small visual/accessible note like "(opens in a new tab)" next to the link when it's enabled.

I saw that @franmate asked for a way to turn this off per file, and that's now been added — thank you for that! But it's still on by default, and there's still no indicator telling people the link opens in a new tab, so I think this part is still open.

One thing I didn't test myself: importing a 2.x `.elp` file with a File Attachments iDevice. I saw @franmate already tested this twice and confirmed it works correctly, so that one's covered.
```

franmate:
Hi, @erseco 
An error is thrown because the c_() function is not recognized. 
The translation function has not been defined.

erseco:
@greynoa @franmate thank you both for the report — you were right, there was a real bug. All three points you raised are now fixed.

**`c_ is not defined`.** The iDevice's export renderer (`export/file-attachment.js`) called `c_()`, the *editor-only* translation helper. It does not exist in a generated site, so `renderView()` threw and `exe_export.js` swallowed it as `[exe_export] Could not load template`, leaving the iDevice unrendered. Export strings now resolve from the `$exe_i18n` bundle that every export already ships (`libs/common_i18n.js`), with an English fallback. Our unit tests missed it because the shared Vitest setup defines a `c_` mock — the new regression test runs with `c_` removed from the global scope, and the preview E2E spec now fails on that console output.

**Open in a new tab.** Agreed, and it is now **off by default** (opt-in per file). Opening a new tab with no warning is an unrequested change of context — WCAG 2.2 [SC 3.2.5 Change on Request](https://www.w3.org/WAI/WCAG22/Understanding/change-on-request.html) and technique [G200](https://www.w3.org/WAI/WCAG22/Techniques/general/G200). `target="_blank" rel="noopener noreferrer"` is emitted only when the author ticks the box; existing and imported attachments stay in the current tab.

**`refreshAttachmentsFromAssets` on rename/delete.** Also fixed. The Yjs asset observer called the global `$exeDevice`, which the core resets when edition ends while the iDevice node stays on the page — so the callback threw, and because Yjs runs observers synchronously inside the transaction, the error surfaced in the Media Library caller even though the rename itself had succeeded. The observer is now bound to the iDevice instance.

Commits `ea53b37` and `863a1e3`, CI green on the branch.

franmate:
In eXeLearning 2, the Attachment iDevice always opens links in a new window (`target="_blank"`).

To preserve the appearance and behavior of educational resources created with previous versions of eXeLearning, I propose the following approach:

When opening a legacy ELP file, the **&quot;Open in a new tab or window&quot;** option should be checked by default. This default behavior should apply exclusively to imported legacy ELPs.

greynoa:
Thanks @erseco — the three fixes look right to me, and I appreciate that the "off by default" change wasn't treated as a matter of taste but grounded in WCAG 2.2 SC 3.2.5 (Change on Request) and technique G200. That's the correct basis for this decision.

On @franmate's latest proposal — defaulting `target="_blank"` to on for legacy imports — I don't think we should do it, and I'd rather explain why than just vote against it:

1. **eXeLearning 2's hardcoded `target="_blank"` was not a design decision we need to honour.** It was a known accessibility defect. Reproducing it on import doesn't preserve fidelity; it carries a WCAG failure forward into a version that just fixed it. "v2 did it this way" isn't a criterion — SC 3.2.5 is.

2. **A legacy `.elp` carries no per-file author intent about link target.** There is nothing in the source to preserve. When the intent is unknown, the default should be the accessible one, and the author can opt in per file after importing — which the new checkbox already allows.

3. **Opening in a new tab is not part of the "appearance" of the resource.** Nothing visual changes. And as @erseco noted, the `download` attribute means many of these links will be downloaded by the browser regardless, so the target frequently has no observable effect at all.

4. **If we do want new-tab links anywhere — legacy or not — the prerequisite is the accessible indicator I raised earlier**: a visible and announced "(opens in a new tab)" on the link. That's still missing. Re-enabling the option by default before it exists would reintroduce exactly the problem this thread just closed.

So my position is: keep it off by default everywhere, legacy imports included. I'm open to being persuaded, but it would need an argument stronger than parity with a version that got this wrong.

Remaining from my side: the "(opens in a new tab)" indicator for when the option *is* enabled. I'll re-test the `c_` crash and the `refreshAttachmentsFromAssets` console errors on 863a1e3 and confirm here.

franmate:
In fact, my proposal stems directly from teachers' concerns regarding MED accessibility. There are hundreds of OERs created with previous versions of eXeLearning that adhere to the UNE 71362 standard by providing explicit descriptions of both the link's purpose and its behavior.

I believe CEDEC needs to make a decision regarding backward compatibility with existing resource catalogs created in eXeLearning 2, as they would be directly affected if editing them in a newer version changes their original behavior. This is not a trivial detail.

greynoa:
Thanks @franmate — I hear the concern about the v2 catalogues, and it's a fair one to raise. But I don't think defaulting target="_blank" on for legacy imports is the right way to address it, and I'd argue UNE 71362 actually points the other way.

A few things:

What UNE 71362 asks for is the announcement, not the new tab. Its accessibility criteria are about link behaviour being predictable and communicated to the user — which is exactly the "(opens in a new tab)" indicator I flagged as still missing. If those v2 OERs "explicitly describe the link's purpose and behaviour," then that description is content, and it's preserved verbatim on import — nothing about the target default strips it. So the part the standard actually cares about is already carried over. What's left is the raw target attribute, and defaulting that to _blank with no announcement is precisely the SC 3.2.5 failure we'd be reintroducing.

A legacy .elp carries no per-file author intent about the target. v2 hardcoded target="_blank" on every attachment — that's a global default in the old tool, not a per-link decision the author made. There's no intent to "preserve." When the intent is unknown, the accessible default is the correct one, and the author can opt in per file (the checkbox already allows that).

In practice the target is frequently inert. As @erseco noted, the download attribute means many of these links get downloaded by the browser rather than navigated, so _blank often has no observable effect at all. The "appearance and behaviour" of the resource doesn't really change here.

On governance: I agree that backward-compatibility with the existing v2 catalogues is a legitimate decision for CEDEC to make explicitly, and I'm glad to see it raised rather than settled implicitly. But the default shouldn't regress accessibility while that's pending — and the prerequisite for enabling new-tab links anywhere, legacy or not, is still the visible and announced "(opens in a new tab)" indicator. Ship that first and "default on for legacy" becomes a much smaller, safer conversation.

Happy to be persuaded, but it would take a clause in UNE 71362 that specifically calls for the new-tab attribute to be preserved — not the description of it, which import already keeps.

erseco:
Following up on the `target="_blank"` thread (@franmate, @greynoa) — while reviewing the code I found something that I think reframes the whole discussion.

**The checkbox was effectively a no-op.** Every attachment link always carries the `download` attribute, and the checkbox only added `target="_blank"` on top. Per the [HTML spec](https://html.spec.whatwg.org/multipage/links.html#downloading-resources), following a hyperlink with a `download` attribute downloads the resource instead of navigating — `target` is never consulted on that path. Exports reference files same-origin (`content/resources/…`) and the preview uses `blob:` URLs, so `download` is honored everywhere this iDevice runs. Even with the box checked, browsers download the file in the current tab; no new tab opens. The option promised behavior the browser doesn't deliver.

That left three coherent paths: **(1) remove the option — the iDevice always downloads**; (2) make it real — drop `download` when checked, add the visible + announced "(opens in a new tab)" indicator @greynoa asked for ([G201](https://www.w3.org/WAI/WCAG22/Techniques/general/G201), [GOV.UK pattern](https://design-system.service.gov.uk/styles/links/)), plus i18n and the platform work below; or (3) keep a dead control — worst of both.

Why (1):

- **This iDevice's contract is file delivery.** It renders "Download <name>" links with type icon and size. WCAG technique [G200](https://www.w3.org/WAI/WCAG22/Techniques/general/G200) says to open new windows only when necessary and lists the necessary cases (context-sensitive help, preserving a secure session) — handing the learner a file isn't one of them. Same-tab is also the blanket recommendation from [NN/g](https://www.nngroup.com/articles/new-browser-windows-and-tabs/) and [GOV.UK](https://design-system.service.gov.uk/styles/links/).
- **Option (2) has real platform costs.** In sandboxed iframes, `target="_blank"` [silently fails without `allow-popups`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/iframe#sandbox) — directly relevant to the hardened opaque embedding/preview work in progress, where we don't want to grant popups. And in the desktop app, `target="_blank"` asks Electron to [create a new chromeless `BrowserWindow`](https://www.electronjs.org/docs/latest/api/window-open) needing an explicit `setWindowOpenHandler` policy, whereas downloads already flow through the app's existing download handling. A per-file boolean that must behave consistently across web, preview, desktop, embeds and five export formats is exactly the kind of variability that costs more than it gives.
- **It doesn't serve legacy fidelity either.** Since `download` always wins, checking the box did *not* reproduce v2's new-window navigation — the link still downloaded. Truly reproducing v2 would mean removing `download`, i.e. option (2) with all its costs, and would re-import thousands of unannounced new-tab links — the exact SC 3.2.5 issue this thread just fixed. On UNE 71362: its web-accessibility criteria are normatively built on UNE 139803:2012, the Spanish adoption of WCAG (see the [published excerpt](https://www.ucm.es/data/cont/docs/758-2017-07-11-UNE%2071362_extracto.pdf), normative references), so the WCAG-based reasoning above argues from that standard's own basis — and the behavior descriptions those OERs contain are authored content that import preserves verbatim, as @greynoa noted.
- **Anyone importing a 2.x resource reviews it editorially anyway** — themes, iDevices and embedded media all change. A link-target nuance the browser ignores is not what preserves those catalogues.

**The PDF use case** (@franmate's original motivation) is legitimate but is a different feature — "view this document in the browser" rather than "deliver this file". Today it's covered by a normal link in a Text iDevice; if the catalogues show real demand, a dedicated document-viewing option can be designed properly (indicator included) in its own issue.

This is now implemented in 3f4332667: the per-file checkbox is removed, links always render as plain download links (no `target`/`rel`), any stored or imported `openInNewWindow` value is ignored and dropped on save, tests cover the new invariant, and the decision is recorded in ADR-0038. It also closes the remaining "(opens in a new tab)" item by construction, and leaves the door open for CEDEC to reopen in-browser viewing as a scoped feature if the v2 catalogues need it — happy to revisit if there are strong objections.

greynoa:
Thanks @erseco, this analysis changes everything. It makes sense: if download always wins over target, the checkbox was never doing what it promised, so removing it is the right call, not a real functional regression.

I'm especially convinced by the point that "reproducing" v2's behavior would have required dropping download, which would have reintroduced exactly the unannounced new-tab links this thread just fixed — that would have been worse, not better, for legacy fidelity.

I also agree that the PDF use case @franmate raised is legitimate but is a different feature ("view in browser" vs. "download"), and deserves its own issue if there's real demand from the catalogues.

This resolves all three points I reported (c_, refreshAttachmentsFromAssets, and the new-tab indicator, the last one by construction now that the option is gone). Proposal approved on my end — green light to move forward. 🚀

ignaciogros (approved):
I've tested the iDevice functionality, and I agree with @cristinavaldera that it's ready to be merged into `main`. I've pushed a few minor UI changes to simplify the interface. I also changed the title from "File attachment" to "File attachments".

@erseco:

Please confirm that you agree with these changes, and let me know if you think it's necessary to rename the directory and the files themselves. I don't think it's essential.

Thanks.

erseco:
Thanks, @ignaciogros. I agree with the UI changes and with using the plural title.

In eXeLearning 2.x, the multi-file iDevice was displayed as **“File Attachments”** — also translated as the plural **“Ficheros adjuntos”** in Spanish — while its internal implementation remained singular: `fileattachidevice.py` and `FileAttachIdeviceInc`.

There was also a separate legacy `AttachmentIdevice` for a single attachment, but this implementation restores the multi-file iDevice.

I therefore don't think we should rename the directory or the files. `file-attachment` works well as the internal technical identifier, and changing it would create unnecessary churn across the configuration, legacy import mappings, documentation, tests, and export code without providing any user-facing benefit.

The visible title can remain **“File attachments”**, while the internal slug stays `file-attachment`.

Thanks for the changes.

PR #2016, opened 2026-07-01
fix(file-manager): keep selection on collaborative Yjs refresh; deterministic save wait

erseco:
Fixes two flaky E2E tests by addressing the underlying app races.

1. **File Manager selection lost on collaborative refresh (real UX bug).** In single-select mode, `renderCurrentView` cleared `selectedAsset` on every re-render, so an incoming **remote Yjs asset change** wiped the current user's selection and disabled the rename button. Now the selection is preserved when the asset still exists (mirrors the multiselect branch); it is cleared only when the asset is gone. Fixes `collaborative/file-manager.spec.ts` and adds 4 Vitest cases.
2. **Async save race (#1667).** `block-reorder-stability.spec.ts` now waits on the real save lifecycle (save button disabled → edition closed + button re-enabled) instead of racing a bare 10s edition-close wait. No timeout bump.

Vitest: `modalFileManager.test.js` green (360 passed).

### DIFF
--- a/public/app/workarea/modals/modals/pages/modalFileManager.js
+++ b/public/app/workarea/modals/modals/pages/modalFileManager.js
@@ -1044,10 +1044,37 @@ export default class ModalFilemanager extends Modal {
                 });
             }
         } else if (!this.multiSelect) {
-            // Single select mode - reset selection on view change
-            this.selectedAsset = null;
-            this.selectedAssets = [];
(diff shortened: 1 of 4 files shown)


ignaciogros (approved):

ignaciogros (approved):
Manually tested.

Great catch!

Thanks.

Issue #1805, opened 2026-05-15
Turn the “Slide” iDevice into a multi-slide “Presentation” iDevice


jjdeharo:
The current **Slide** iDevice, which can be tested at:

https://f782eaab.exelearning.pages.dev/

only allows the creation of a single slide. The proposal is to evolve it into a more general **Presentation** iDevice, capable of containing any number of slides.

The idea would be to let users add, duplicate, delete, and reorder slides within the same iDevice, in a similar way to a simple presentation editor. Each slide could keep the current options available in the Slide iDevice, but integrated into a navigable sequence.

This would make it possible to create short presentations, step-by-step explanations, visual summaries, learning sequences, or brief interactive content without having to insert several independent iDevices.

Possible improvements:

- Rename the iDevice from **Slide** to **Presentation**.
- Allow an unlimited number of slides.
- Include controls to move forward and backward between slides.
- Allow slides to be reordered in the editor.
- Consider the possibility of exporting the content to presentation-specific formats.

I think this change would significantly increase the usefulness of the iDevice, turning it from a single-purpose resource into a more flexible tool for creating complete presentations within eXeLearning.

erseco:
Thanks @jjdeharo for the suggestion. After reviewing it, and re-thinking about it in the past weeks, I would suggest **not implementing this feature as proposed**.

The current Slide iDevice is intentionally focused on a single editable canvas/slide. eXeLearning already provides a higher-level structure for creating, ordering and navigating content through project pages. Adding multiple internal slides inside this iDevice would introduce a second content hierarchy inside the existing one: pages in eXeLearning, and then pages/slides again inside the Slide iDevice.

That would make the authoring model harder to understand and maintain. It would also add extra complexity to editing, duplication, deletion, reordering, preview, export, accessibility, backwards compatibility and tests. In practice, it would move the Slide iDevice towards a nested presentation editor, almost like embedding a small eXe-like structure inside another eXe project.

From a UX/instructional design point of view, adding another layer of navigation and management may also increase cognitive load unless the need is very clear and cannot be solved with the existing project/page structure.

What I do think could be useful is a smaller, more focused improvement: keep the Slide iDevice as a single slide, but improve the presentation experience. For example, we could explore a simpler “next/previous” flow between consecutive Slide iDevices or between eXeLearning pages, especially in fullscreen/presentation mode.

So my recommendation would be:

* Do not turn the Slide iDevice into a multi-slide Presentation iDevice.
[...]

PR #2020, opened 2026-07-01
Add a reader-activated Presentation mode to web site exports

### LINKED ISSUE #2019: Add keyboard navigation to exported and previewed content
## Summary

It would be useful to add keyboard navigation support to exported and previewed eXeLearning content, inspired by the legacy eXeLearning 2.9 `Presentation` style.

This would let users move through exported packages with the keyboard:

- `ArrowLeft`: previous page
- `ArrowRight`: next page
- `ArrowUp`: first page
- `ArrowDown`: last page
- `m` / `Alt+M`: toggle the navigation menu
- a safe search shortcut when the search box is available

## Why

This improves usability for presentation-like use cases and keyboard-only navigation. It also brings back part of the practical behavior that existed in the old `Presentation` style, without requiring authors to use a specific theme.

This is also related to #1805, which proposes evolving the `Slide` iDevice into a multi-slide `Presentation` iDevice. This issue does not implement that full proposal, but it would be a useful partial step toward presentation-friendly navigation in exported content.

## Suggested approach

- Implement the behavior in shared export/preview runtime code, not per-theme, if possible.
- Use existing exported navigation elements such as `.nav-buttons`, `#siteNav`, `#siteNavToggler`, `#searchBarTogger`, and `#exe-client-search-text`.
- Follow WCAG keyboard accessibility guidance:
  - do not intercept shortcuts while users are typing;
  - avoid browser-reserved shortcuts;
  - avoid keyboard traps;
  - provide a safe handling strategy for single-character shortcuts like `m`.
- Add frontend/unit tests for the keyboard handler.
- Add an E2E test that verifies the shortcuts in preview or exported content.

Partially addresses #1805.

erseco:
## Summary

Adds a **reader-activated Presentation mode** to web site exports, so an eXeLearning site can be presented with the keyboard or a presenter remote (issue #2019, inspired by the legacy eXeLearning 2.9 `Presentation` style).

The first iteration of this PR shipped the behaviour as an export option. Review (@ignaciogros, @cristinavaldera) rejected storing it in the project: an option leaves a trace in the `.elpx` that changes navigation for whoever imports it, decides for the reader what only the reader knows, and adds one more option to an already large export configuration. This revision follows the design closed in the thread on 2026-09-15, with one deliberate deviation explained below: **no export option, no export type, nothing stored in the project. The reader switches the mode on with the URL parameter and can leave it from a visible control.**

## How it works

- `?exe-presentation=1` (aliases `true` / `yes`) **enters the mode** and shows a visible **"Exit presentation mode"** control; `?exe-presentation=0` shows the control with the mode off; without the parameter nothing is injected and no key is captured.
- **The parameter is the state.** Entering or leaving rewrites it (`=1` / `=0`) in the current URL (`history.replaceState`) and in the menu, previous/next and search-result links, so the choice survives page changes and reloads with **no storage at all**. The control is a plain `<button>` whose label names the action.
- **Why the parameter activates, unlike Teacher Mode:** `?exe-teacher=1` must not reveal content by itself because revealing answers is a content change. Here the only reason to put the parameter in a link is to present, so an extra click would be pure friction. The concern that motivated "the parameter changes nothing" (no invisible behaviour, no invisible exit) is met by the always-visible exit control.
- While presenting: the navigation menu **collapses** (through the style's own `#siteNavToggler`, so it can still be opened normally), `←` / `PageUp` go to the previous page and `→` / `PageDown` to the next one, **`M`** shows/hides the menu and **`T`** toggles Teacher Mode where its toggle exists on the page (plain keys only, so `Ctrl`/`Cmd+T` is never shadowed). Pages remain the navigation unit. The control's tooltip lists the keys and points to `F11` for full screen.
- **Not part of the mode:** fullscreen (left to the browser, `F11`), `Esc` (it belongs to lightboxes, dialogs, videos and fullscreen), `↑` / `↓` (long pages keep scrolling), any traversal of feedback buttons, FX effects, galleries or video controls.
- **Safeguards kept:** no key with a modifier, none while typing in a form field, none while an overlay is open (prettyPhoto, SimpleLightbox, the Games-* fullscreen image overlay, MediaElement fullscreen video) and, new, none while `document.fullscreenElement` is set, so an iDevice that goes fullscreen on its own keeps its keys.
- **Scope:** web site exports opened as the top-level document. Never SCORM/IMS (the LMS owns navigation), EPUB, or content embedded in an iframe. That includes the editor's preview panel, which is an iframe: to try it, export the web site and open `index.html?exe-presentation=1`.
- **Placement:** the control is appended to `<body>` next to `#made-with-eXe` and styled in `base.css` with the same fixed-badge recipe (bottom-right, neutral look, hidden in print, shifted left when the badge is present, stacked below the badge so its hover expansion covers the control). It lives outside `.exe-content` and the footer, so third-party styles cannot break it and may restyle it by id.

## Implementation

- `public/app/common/exe_export.js`: new `presentationMode` module mirroring `teacherMode` (`bootstrap()` in `<head>`, `init()` with the style's togglers ready, `enter()` / `leave()` / `toggle()`, `handleKeydown()`, `overlaySignals`). The Teacher Mode link-propagation code becomes the shared `$exeExport.withNavParam()` / `propagateNavParam()`, used by both modes and by search hits.
- `public/style/workarea/base.css`: `#exe-presentation-toggler`.
- `public/app/common/common_i18n.js`: `presentation_mode`, `exit_presentation_mode` and `presentation_mode_keys` (tooltip) labels (translations are managed separately; the runtime falls back to English if a bundle predates the keys).
- **Removed:** `pp_addKeyboardNavigation` and all its plumbing (properties UI, `metadata-properties`, Yjs mapping, `YjsDocumentAdapter`, `Html5Exporter`, `PageRenderer` flag, `ElpxImporter`, their tests), the `?keyboard-navigation=false` opt-out, and the `Alt+/` / `ArrowUp` / `ArrowDown` shortcuts.

## Testing

- Unit: `public/app/common/exe_export.test.js`, `describe('presentationMode')` (bootstrap, scope guard, control injection, URL and link rewriting, enter/leave/toggle, arrow/PageUp/PageDown/M/T keys, typing guard, overlay signals). Every line of the module is covered.
- E2E: `test/e2e/playwright/specs/presentation-mode.spec.ts` exports the project in the browser, serves it from its own origin through `page.route()` (new `serveWebSiteExport()` helper) and opens it **top-level**: no parameter → nothing; `=1` → mode on, keys (including `M`), propagation; leave → `=0` in URL and links, survives reload and navigation; re-enter. The SimpleLightbox case in `idevices/image-gallery.spec.ts` runs on that served export with the mode active.
- Locally green: `make fix`, `make test-unit` (8807), `make test-frontend` (16161), `make test-integration` (770), and both specs under `--project=chromium` and `--project=static`.

## Architecture records

- New `ADR-2019-04` — reader activation, scope, control placement, and why the export option, a new export type and fullscreen-driven activation were rejected.
- `ADR-2019-02` (export option) set to **Rejected**, kept as the record of the first iteration.
- `ADR-2019-01` and `ADR-2019-03` updated; `doc/architecture/changes/2019-keyboard-navigation-export-preview/design.md` rewritten.

## Follow-ups (not in this PR)

- User-guide entry for `?exe-presentation=1`, alongside Teacher Mode, `pp_addExeLink` and the accessibility toolbar.
- Optional pointer / presentation cursor, and a `pp_addPresentationControl` option if a project wants the control without the parameter.

Closes #2019. Related: #1805.


ignaciogros (commented):
Thanks for the proposal, @erseco.

In my opinion, it's better to keep this behavior for presentation styles or (an alternative that hasn't been considered yet) introduce a new export option that allows users to enable this type of keyboard navigation. I think I prefer the first option, to avoid exporting content with this behavior without the user's knowledge.

**Reasons:**

* It changes the standard navigation behavior of a website (it's a website, not necessarily a presentation).
* It can introduce navigation inconsistencies. For example, a component such as a gallery using `rel="lightbox[galleryID]"` may rely on the arrow keys to navigate between its items. That's impossible to control in an editor that allows users to insert almost any kind of content. When that happens, if our JavaScript doesn't handle it, using those navigation keys will cause unexpected behavior.

In any case, I think we should wait for the feedback from other team members.

erseco:
@ignaciogros good points, thank you — you're right, this shouldn't ship silently on every export, and the gallery should fully prevail over our shortcuts. I've addressed both parts of your review:

**1. Made it an opt-in export option.** Added a new **"Keyboard navigation" export option** in Project Properties → Export options, modeled exactly on the existing "Include MathJax" / "Accessibility toolbar" checkboxes. It's off by default, so:

- No existing export or preview changes behavior.
- Authors get an explicit, informed choice, with help text calling out that it changes standard navigation and may conflict with other keyboard-driven content (like a lightbox gallery using arrow keys, per your example).
- The runtime opt-out (`?keyboard-navigation=false` / localStorage) still exists as a secondary safety net for viewers, on top of the author-level opt-in.

**2. Made it defer to any open lightbox/gallery/overlay.** I looked at how other libraries handle this exact class of conflict (a global keyboard handler vs. an embedded widget that also wants the keyboard):

- **reveal.js** (the most popular JS presentation framework — same problem, global arrow-key slide nav vs. embedded interactive content) exposes a `keyboardCondition` hook: an optional function that blocks *all* keyboard handling when it returns `false`. ([docs](https://revealjs.com/config/))
- **Bootstrap** modals add a `.modal-open` class to `<body>` as a queryable "something modal is active" signal — this app already loads Bootstrap unconditionally, for what it's worth.
- **PhotoSwipe** (a modern popular lightbox) adds `.pswp--open` to its root element while open, same convention.
- The **WAI-ARIA APG Dialog (Modal) Pattern** states the general accessibility principle: while a modal/overlay is active, background content should not receive keyboard interaction — the active surface owns the keyboard. ([docs](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/))

So I added the same kind of mechanism: `isOverlayActive()` checks a small, documented list of "is some overlay currently open" signals (`overlaySignals`), and `handleKeydown` bails out entirely — not just for arrows, for every shortcut — when any of them is true. Concretely it now covers:

- **exe_lightbox** (prettyPhoto, your `rel="lightbox[galleryID]"` example)
- **SimpleLightbox** (the Image Gallery iDevice) — this one especially matters, since its own arrow-key nav runs on `keyup` while ours is on `keydown`; without suppression our page-navigation would fire *before* the gallery's own handler ever got a chance
- The fullscreen image overlay used by Magnifier and several other Games-* iDevices
- MediaElement.js fullscreen video (Interactive Video)

It's a flat array, so covering a future conflicting widget is a one-line addition, not a rewrite. I verified it against a real, actually-opened SimpleLightbox gallery in a new E2E test (not just a synthetic check), since that's the case you raised.

Two things I found along the way that I deliberately left out of scope (documented in code comments, not swept aside):
- MediaElement's *non-fullscreen* arrow-key seeking (e.g. clicking into an inline video without fullscreening it) — that's a different, focus-scoped problem, not an overlay-open one.
- The third-party `mediateca`/`jwplayer()` video embed — no stable selector to key off of from this codebase.

I don't have a general fix for every possible keyboard-driven widget an author could insert — that would need per-widget coordination and is a much larger problem. Making it opt-in, plus deferring to the known overlays above, means an author who inserts something not on that list can simply not enable this option, or verify compatibility before publishing. Happy to extend the list further as a follow-up if more conflicts turn up.

Does this combination — opt-in export option + overlay suppression — address your concerns?

erseco:

- **Important:** `src/shared/import/ElpxImporter.ts:1196` (`extractMetadata`) and `:1242` (`setMetadata`) — the new `addKeyboardNavigation` option does not survive an ELP/ELPX round-trip. Export writes `pp_addKeyboardNavigation` (OdeXmlGenerator serializes all `ExportMetadata` keys, which the PR extended), but the import path was left untouched: `extractMetadata` reads a hardcoded `pp_*` list that omits it, `setMetadata` never writes it to Yjs, and `OdeMetadata` (`interfaces.ts:268`) lacks the field. An author who enables the option, exports, and reopens the file silently loses the setting — every sibling export option (`addMathJax`, `addSearchBox`, `addAccessibilityToolbar`, …) persists; this one alone reverts to `false`. Fix: add `addKeyboardNavigation` to `OdeMetadata`, read `pp_addKeyboardNavigation` in `extractMetadata`, and `metadata.set(...)` it in `setMetadata`. This also contradicts `metadata-properties.ts`'s docstring, which claims `ElpxImporter` is driven by `METADATA_PROPERTIES` — it references none of it and keeps a parallel list that has drifted.
- **Test gap:** no round-trip test feeds a `content.xml` containing `pp_addKeyboardNavigation` back through `ElpxImporter` (`ElpxImporter.spec.ts` covers `addMathJax` but not the new key), so green CI does not catch the loss. Add a colocated round-trip assertion alongside the fix.
- The reviewed `ArrowUp`/`ArrowDown` first/last-page behavior is an opt-in, documented design decision (ADR-0040/SDD-0010) — noted, not a defect.

Verified: static read of the full export→import path (OdeXmlGenerator, YjsDocumentAdapter, ElpxImporter + its spec) at head; `CI: all checks green on ac6ace9 ✅`. No local build/test was run — the round-trip conclusion is a code-path fact, not a runtime claim.

---
**Update — resolved (`5d73956c`):** `ElpxImporter` now reads `pp_addKeyboardNavigation` and restores it to Yjs metadata (mirroring `addMathJax`); round-trip test added.

cristinavaldera (dismissed):
Thanks, Ernesto.

I think we should think about how right/left buttoms should behave when the project include:
- feedback buttoms
- FX effects
- big content with scroll
- ligthbox effect
- embed videos 

Should right/left buttoms travel through all these elements and not only go from one page to another one?

I think this is really helpful, more than having a specific style for presentations, but I would like to understand better why @ignaciogros  prefers the other option.

erseco:
Thanks, @ignaciogros. I think the main goal here is to make eXeLearning usable as a presentation with a standard presenter remote, without turning this PR into a generic keyboard-navigation system for every possible interactive element.

Perhaps the export option should be called **“Presentation mode”** rather than “Keyboard navigation”. When enabled, it could:

* Start the exported content with the navigation menu collapsed.
* Enable previous/next navigation with the keys commonly sent by presenter remotes (`ArrowLeft` / `ArrowRight`, and possibly `PageUp` / `PageDown`).
* Keep the current safeguards: no shortcuts while typing, and give priority to open lightboxes or fullscreen overlays.

I would keep the scope small for now:

* Pages remain the navigation unit.
* We do not try to automatically traverse feedback buttons, FX effects, galleries, or video controls.
* We do not change the normal scrolling behavior of long pages.
* Interactive elements keep their own keyboard behavior when they have focus or are open.

A future presentation-oriented feature could let authors define internal “steps” within a page, but I think that needs an explicit authoring model and should be addressed separately.

@ignaciogros, @cristinavaldera, do you think this focused **Presentation mode** would cover the immediate use case? Would you prefer it as an export option, as it is now, or as an always-available toggle in the exported content?

ignaciogros:
I prefer the idea of making it an export preference, but if we add it, I wouldn't refer only to keyboard navigation, but also to the presentation cursor. Perhaps something like "Export as presentation", with an explanation of what will happen when you use it: keyboard and pointer controls.

**Two concerns:**

* Someone could share a link that includes an option to download the content, and another person could import it and generate a website with that option enabled without being aware of it. This significantly affects the final user experience and should be done deliberately.
* There could be a style with these options enabled (and perhaps managed differently), leading to inconsistencies.

If I had to choose between the proposals mentioned, I would go with the preference, but I think there's another possible solution: a new export type that, for now, would be identical to the website export but with these features enabled. This would leave no trace in the `.elpx` file, which could be an advantage in this case.

The disadvantage is that it would make the current menu more complex, and it already contains more options than most users use.

erseco:
After thinking about the different alternatives, I think there may be an even simpler solution that gives us the best of both worlds.

Rather than storing **Presentation mode** as a project/export preference, or adding another export type, what if **entering fullscreen from eXe itself was the action that enables presentation mode**?

We could add a single fullscreen/presentation control to exported and previewed content. When the user activates it:

* the document enters fullscreen;
* (optionally) the navigation menu is collapsed, although it can still be opened normally if needed; 
* presentation keyboard navigation is enabled (`ArrowLeft` / `ArrowRight`, and possibly `PageUp` / `PageDown`);
* the existing protections for inputs, galleries, lightboxes, videos and other interactive elements remain in place.

When leaving fullscreen with `Esc`, presentation mode is disabled and the content returns to normal website behaviour.

I would keep `ArrowUp` / `ArrowDown` for normal scrolling, so long pages still behave like normal web content.

There is one important implementation detail: **presentation mode should only be activated by our global fullscreen/presentation control, not by any generic `fullscreenchange` event**. Some iDevices already enter fullscreen themselves, and in those cases their keyboard controls must continue to take precedence.

I think this would be simpler conceptually:

* no additional project property;
* no presentation flag stored in the `.elpx`;
* no extra export type;
* no need for authors to anticipate whether the content will later be presented;
* the user decides to present precisely at the moment they enter presentation/fullscreen mode.

It also seems quite intuitive from a UX perspective: when I put an eXe resource fullscreen, most of the time I am trying to present or consume it without the surrounding navigation. We can make that mode presentation-friendly automatically, while still allowing the menu to be shown when needed.

This seems to me simpler than the current export preference and avoids adding another option to an already fairly large export configuration. What do you think, @ignaciogros @cristinavaldera?

ignaciogros:
I like the latest proposal (presentation mode activated by the end user, web site export only). A few points to close the design:

**Activation, like Teacher Mode.** `?exe-presentation=1` shows a "Presentation" control; the user enters and leaves the mode with it. The parameter itself changes nothing, and without it nothing is injected and no key is captured. The state travels between pages as Teacher Mode does (parameter in the navigation links + `localStorage`).

**Fullscreen should not be part of the mode.** The mode collapses the menu, enables `←`/`→` (and `PageUp`/`PageDown` for presenter remotes) and, later (or from the first moment), the pointer. Fullscreen is left to the browser (`F11`), which is what presenters already do: standard, survives page navigation, and the URL stays reachable. The Fullscreen API needs a user gesture, is lost on navigation in Firefox/Safari and `file://`, is unavailable on iPhone, and would require a fallback and an exit control of its own. Caveats (tablets, kiosk/Electron, macOS shortcut) belong in the user guide. An iDevice that goes fullscreen on its own (`document.fullscreenElement`) counts as an open overlay and the keys stay inactive.

**Scope.** Web site export only, opened as the top-level document. Not for SCORM/IMS (navigation belongs to the LMS) nor for content embedded in an iframe, where presenting makes no sense.

**A visible control is required** to keep the current accessibility level: the user must be able to see the mode, enter it and leave it from the interface. A mode enabled only by a parameter, with no visible way out, is the "invisible behaviour" we rejected earlier. `Esc` should not be the exit key: it already closes lightboxes, dialogs, videos and fullscreen, and would conflict with the very overlays the mode has to respect. `Esc` makes sense for a fullscreen mode, not for this one.

**Placement.** Third-party styles are the concern: injecting into `.page-header` can break their layout, and each style customises Teacher Mode differently. `#siteFooter` is not a safe anchor either: it is hidden when it has no license and no footer content (`siteFooter-empty`, #2145), so the control would disappear with it. The safest default is to append the control to `<body>`, right before `</body>`, as a sibling of `#made-with-eXe`: that element already lives outside `.exe-content` and outside the footer, exists in every web site export regardless of style, and is not affected by anything a style does to the page layout. Do not anchor to `#made-with-eXe` itself (optional via `pp_addExeLink`), but follow its recipe: `base.css` already renders it as a fixed badge in the bottom-right corner, hidden in print, and styles only touch its dark-mode colour. A control styled the same way is always visible, never depends on page layout, and any style can override it if it wants to. I suggest a neutral look, detached from the custom presentation of the styles, as the accessibility toolbar does.

No export preference for now; a `pp_addPresentationControl` option (like the accessibility toolbar) can be added later if needed.

**Documentation.** All of these options (`?exe-presentation=1`, Teacher Mode, `pp_addExeLink`, the accessibility toolbar and the like) should be properly documented in the user guide: most users will never use them simply because they do not know they exist.

erseco:
Updated the PR to the design closed above (@ignaciogros's points of 2026-09-15 on top of the reader-activated proposal). Summary of what changed:

- **Activation like Teacher Mode.** `?exe-presentation=1` (also `true` / `yes`) shows a "Presentation mode" control; the reader enters and leaves the mode with it. The parameter alone changes nothing, and without it nothing is injected and no key is captured. The state travels between pages through the navigation links plus `localStorage`, reusing the Teacher Mode propagation code (now shared by both modes).
- **No fullscreen in the mode, no `Esc`.** Entering collapses the menu (through the style's own toggler, so it can still be opened) and enables `←` / `PageUp` and `→` / `PageDown`. `↑` / `↓` keep scrolling. Fullscreen is left to `F11`. An iDevice that goes fullscreen on its own (`document.fullscreenElement`) now counts as an open overlay and the keys stay inactive, as do open lightboxes and form fields.
- **Scope.** Web site exports opened as the top-level document only: never SCORM/IMS/EPUB nor content in an iframe (which includes the editor preview, so it has to be tried on an exported site).
- **Visible control, safe placement.** A real `<button>` with `aria-pressed`, appended to `<body>` right next to `#made-with-eXe` and styled in `base.css` with the same fixed-badge recipe (neutral look, hidden in print, offset when the badge exists). Not anchored to the header or the footer.
- **The export option is gone.** `pp_addKeyboardNavigation` and all its plumbing (properties UI, metadata, Yjs mapping, exporter/renderer flag, ELP import) are removed, together with the `?keyboard-navigation=false` opt-out and the `m` / `t` / `Alt+/` shortcuts. Nothing is stored in the `.elpx`. `ADR-2019-02` is marked Rejected and the new `ADR-2019-04` records this decision; the design doc is rewritten.
- **Tests.** The E2E spec now exports the site in the browser, serves it from its own origin and opens it top-level, with and without the parameter; the real SimpleLightbox case runs on that export with the mode active.

Left for follow-ups, as suggested: the user-guide entry for `?exe-presentation=1` (with Teacher Mode, `pp_addExeLink` and the accessibility toolbar), the pointer, and an optional `pp_addPresentationControl`.

I am dismissing the previous "changes requested" review since it referred to the export-option design, and requesting a fresh review from @ignaciogros and @cristinavaldera.

erseco:
One follow-up change after trying it on a real export, in favour of the reader:

- **`?exe-presentation=1` now enters the mode directly** and the control reads "Exit presentation mode". This deviates from "the parameter itself changes nothing" on purpose: that rule protects Teacher Mode because revealing answers is a content change that must stay behind a click. Here the only reason to put the parameter in a link is to present, so the extra click was pure friction. The concern behind the rule (no invisible behaviour, no invisible way out) is met by the exit control, which is always visible.
- **No `localStorage`.** The parameter is the state: leaving rewrites it to `?exe-presentation=0` in the URL and in the navigation links, so the choice survives page changes and reloads, and `=0` keeps the control available to come back in. Simpler, and nothing is remembered outside the link itself.
- The `#made-with-eXe` badge is stacked above the control, so its hover expansion covers the control instead of the other way round.

Unit, chromium and static E2E updated; ADR-2019-04 and the design document record the reasoning. If you would rather keep the parameter inert, it is a two-line change back.

erseco:
Two keys from the first iteration are back, only while presenting and only as plain keys (never with `Ctrl`/`Cmd`/`Alt`, never while typing or while an overlay is open):

- **`M`** shows/hides the menu through the style's own `#siteNavToggler` (matched by physical key, so it works on any layout).
- **`T`** toggles Teacher Mode through its own toggle, which only exists on the page when `?exe-teacher=1` made it available; otherwise the key does nothing.

The control's tooltip now lists the keys and points to `F11` for full screen. Full screen itself stays out of the mode, as agreed: the Fullscreen API is per document and is lost on page changes outside Chrome, while `F11` is per window and survives navigation.

PR #2031, opened 2026-07-02
Disable File Manager Insert button outside iDevice edit mode

### LINKED ISSUE #2030: Disable the "Insert" button in File Manager when no iDevice is being edited
When opening the File Manager from Utilities, it is possible to select an image and click Insert. However, if no iDevice is currently being edited, the Insert action has no effect. 
This can be confusing, as the button appears to be available even though it cannot perform any action.

Expected behavior
The Insert button should be disabled unless an iDevice is currently open in edit mode.

erseco:
### Summary

Fixes #2030.

The File Manager's `Insert` button was enabled purely based on file selection, even when no iDevice was currently open in edit mode (e.g. opened from Utilities on a freshly imported project). Clicking it in that state either silently copied the asset URL to the clipboard or inserted into whatever editor happened to still be active, with no indication to the user that nothing useful would happen — confusing, since the button looked available.

`Insert` is now only enabled while an iDevice is actually in edit mode (`eXeLearning.app.idevices.getIdeviceActive()`), regardless of how the File Manager was opened. Selecting a file no longer enables `Insert` on its own; an editable iDevice context is also required. Existing insertion behavior (via TinyMCE's image dialog, or the file/image pickers inside an iDevice's edit form) is unaffected, since all of those paths only exist while an iDevice is being edited.

### Quick testing checklist

- [x] Open the File Manager from Utilities without editing any iDevice.
- [x] Select an image or another file.
- [x] Confirm that the `Insert` button remains disabled.
- [x] Open an iDevice in edit mode.
- [x] Open the File Manager from that editing context.
- [x] Select an image or another file.
- [x] Confirm that the `Insert` button is enabled.
- [x] Click `Insert` and confirm the selected file is inserted as before.
- [x] Confirm there are no visual regressions in the File Manager.

### Automated tests

- [x] Added a failing unit test first (`modalFileManager.test.js`, `updateButtonStates` — "should disable insert button when file is selected but no iDevice is being edited"), confirmed it failed against the pre-fix code.
- [x] Implemented the minimal fix (`hasEditableIdeviceContext()` gating `canInsert`) and verified the new test passes, along with the full `modalFileManager.test.js` suite (361 tests) and the full frontend Vitest suite.
- [x] Added an E2E Playwright test (`test/e2e/playwright/specs/file-manager.spec.ts`, "Insert Button Disabled Outside iDevice Edit Mode (#2030)") covering the full flow: open File Manager from Utilities → select a file → Insert disabled → start editing an iDevice → reopen File Manager → Insert enabled → click Insert → file is actually inserted. Verified it fails against the pre-fix code and passes after the fix.
- [x] Ran the full `file-manager.spec.ts` suite (24 tests) plus every other spec that exercises File Manager insertion (`text-multi-image`, `file-manager-special-chars`, `collaborative/file-manager`, `project-clone-duplicate`, `3dmol-molecule-filemanager`) — all green, no regressions.


cristinavaldera (approved):
Thanks!!

ignaciogros (approved):

PR #2033, opened 2026-07-03
Fix GeoGebra activity display sizing

### LINKED ISSUE #2029: Suggestion: Improve the GeoGebra iDevice Display in eXeLearning
I would like to suggest an improvement to the GeoGebra iDevice available in eXeLearning.

Currently, the display area where the GeoGebra construction is shown is too small, especially the coordinate axes and the interactive workspace. This makes it difficult for learners to properly view and interact with the content, reducing its educational effectiveness.

It would be very helpful if the GeoGebra iDevice allowed a larger display area or provided an option to adjust its size more easily. This would significantly improve usability and the overall learning experience.

A screenshot illustrating the current issue is attached for reference.

Image

Thank you for considering this suggestion.

erseco:
## Summary

Fixes #2029.

The GeoGebra Activity iDevice editor already had Width/Height inputs that correctly round-tripped through save/load into `auto-geogebra-width-N`/`auto-geogebra-height-N` CSS classes, and the export runtime already read those classes correctly when instantiating `GGBApplet`. The only real gap was that the controls were hidden behind a Bootstrap `d-none` class *and* a matching CSS `display: none` override in the editor, so authors had no way to make the embedded construction bigger — leaving every activity stuck at the 565×363 fallback.

## Changes

- Reveals the Width/Height controls in "General Settings" (`d-flex` instead of `d-none`, CSS override removed), with `800`/`600` placeholder hints. Uses `placeholder`, not `value`, so legacy content without explicit dimensions keeps its existing fallback behavior even if just re-saved.
- Fixes a stray `maxlength="1500"` on the size inputs down to `4`, matching the existing digit-only input restriction.
- Hardens `indicator.getSize()` in the export runtime to validate width/height the same way `addActivity()` already did, falling back to the 565×363 default on invalid/zero values.
- Adds `max-width: 100%; overflow-x: auto` to `.auto-geogebra-wrapper` so a configured width wider than the content column scrolls instead of clipping or distorting the applet.
- Adds Vitest coverage for the edition (visibility, save/load round-trip, legacy compatibility) and export (`addActivity`/`getSize` sizing behavior) code.
- Adds a Playwright E2E spec that mocks the GeoGebra CDN script deterministically and verifies the controls are visible and a configured 800×600 size reaches the preview applet.

## Testing

- `npx vitest run public/files/perm/idevices/base/geogebra-activity/edition/geogebra-activity.test.js`
- `npx vitest run public/files/perm/idevices/base/geogebra-activity/export/geogebra-activity.test.js`
- `bun x playwright test --project=chromium test/e2e/playwright/specs/idevices/geogebra-activity.spec.ts`
- `make fix && make test-unit && make test-frontend && make test-integration && make test-e2e` (full suite: 485 passed, 6 pre-existing skips, 0 failed)

## Manual test plan

- [x] Add a "GeoGebra activity" iDevice (Science category) to a page in the workarea
- [x] Expand "General Settings" and confirm the Width/Height fields are now visible (with `800`/`600` placeholders), next to the URL/Title/Authorship fields
- [x] Enter a GeoGebra material ID or URL (e.g. `VgHhQXCC`), set Width = `800` and Height = `600`, and save the iDevice
- [x] Open Preview and confirm the GeoGebra applet renders at the larger configured size (not the old 565×363 default) and is not clipped
- [x] Re-open the iDevice for editing and confirm Width/Height still show `800`/`600` (round-trip)
- [x] Open/import an existing GeoGebra activity that has no explicit width/height saved and confirm it still displays at the default size (backward compatibility — no dimensions get silently injected)
- [x] Set a width larger than the content column (e.g. `1400`) and confirm the applet scrolls horizontally instead of being clipped or visually distorted
- [x] Export as HTML5/SCORM and confirm the exported page also renders the GeoGebra applet at the configured size

## Fix (refactor) — duplicated size-class parsing

The GeoGebra width/height class parsing was duplicated between `getSize()` and `addActivity()`. Extracted a single shared `parseSizeClasses()` helper in the export file that both delegate to; behavior preserved (same 565×363 defaults and `parseInt`/`!isNaN`/`>0` guards). Commit `91705ffa`; colocated tests cover valid/garbage/missing classes.


erseco:

The fix is sound: revealing the pre-existing Width/Height controls (`d-none`→`d-flex`, dropping the CSS `display:none`) with `placeholder` rather than `value` preserves legacy backward-compat, and the hardened `getSize()` parse (`parseInt` + `!isNaN` + `>0`) falls back to the 565×363 default on garbage classes. `overflow-x:auto` on `.auto-geogebra-wrapper` handles oversized applets. No correctness, data-loss, or back-compat regression found.

- **Minor (optional):** `export/geogebra-activity.js:195-204` (`getSize`) now duplicates the width/height class-parsing already in `addActivity` at `:256-268`; the same parse also recurs in `edition/geogebra-activity.js:409-414`. Since these run in separate iDevice contexts, a cross-file shared helper is impractical, but a small `parseSizeClasses()` inside the export file would collapse the two near-identical loops and their two unit tests into one. Purely a maintainability nit, not a defect.

Verified: reviewed the diff and full file at head; parsing guards, default fallback, and placeholder-vs-value round-trip confirmed by reading. No local build/test run — CI: all checks green on 18bb305 ✅. Residual risk: the runtime `GGBApplet` injection and actual applet render size are only exercised by the Playwright spec (which mocks the GeoGebra CDN); real geogebra.org rendering was not exercised, but that is out of scope for this sizing fix.

---
**Update — resolved (`91705ffa`):** `getSize()` and `addActivity` now delegate to a shared `parseSizeClasses()` (single source of truth); tests added.

ignaciogros (commented):
Everything works correctly, but I'm not sure if we should include 800×600 as the default or recommended size. The iDevice used to retrieve the dimensions automatically by making an AJAX request to `https://www.geogebra.org/api/json.php` (see lines 260 and 261 of `edition/geogebra-activity.js`).

If you load the sample activity (`https://www.geogebra.org/m/VgHhQXCC`), it returns dimensions of 365×363, which do not have the same aspect ratio as 800×600. If users choose 800×600 because the application recommends it, some content may not be displayed correctly.

It may be enough to simply remove the "Recommended: 800x600 px" text. If you agree, we can remove it and merge the change into `main`.

Thanks.

erseco:
Yes, I agree. Please remove the **“Recommended: 800x600 px”** text.

Thanks, @ignaciogros.

ignaciogros (approved):

PR #2035, opened 2026-07-03
Fix Text iDevice video subtitles upload and export

### LINKED ISSUE #2034: Video subtitles (.srt) can't be added and never show in the Text iDevice
## Summary

When you add a video inside the **Text iDevice** and try to attach an `.srt` subtitle file, the subtitles never appear in the preview or in the exported content. While testing this I ran into what looks like three different problems, one after another.

What makes me think this is a real bug and not just me: the subtitles option clearly says it supports this format — the field is labelled **"Subtitles (srt / vtt)"** and there's even a checkbox that says **"Player compatible with .srt subtitles. Check only if using .srt subtitles."** So `.srt` is meant to work, but I couldn't get it to.

## Environment

- Version: 4.0.0
- Tested in Chrome and Firefox
- iDevice: **Text**, with a video inserted in it

## Steps to reproduce

1. Add a Text iDevice and insert a video (mp4).
2. Go to the **Subtitles** tab of the video window.
3. Click **"Choose a file"** to add an `.srt` subtitle → the File Manager opens.
4. Click the upload button and try to pick the `.srt` file.
5. Attach the subtitle, save, export (as Web site and as SCORM) and open the result.

## What I expected

- Being able to upload the `.srt` file from the File Manager.
- The subtitles showing on the video, both in the preview and in the export.

## What actually happens

1. **The file chooser won't let me pick the `.srt` file at all** (`.vtt` is greyed out too). The only way I found to get it in was to drag the file directly onto the File Manager window — but that's not obvious, and if you drop it slightly off, the app thinks you're trying to open a project and shows a message saying only `.elpx`, `.elp` or `.zip` files can be opened that way.

2. **Once the subtitle is attached, it never shows.** I tried it with the "compatible player" box unchecked and checked, and in both exports (Web and SCORM). I even clicked the **CC** button on the player and nothing happened.

## What seems to be going on (from my testing)

I compared the two exported files and noticed a few things that might point the team in the right direction:

- **`.srt` vs `.vtt`:** as far as I know, videos on the web normally need subtitles in the **`.vtt`** format, not `.srt`. In the export the subtitle is left as `.srt`, so the normal player just ignores it. It might be that eXeLearning needs to convert the `.srt` into `.vtt` when you add it or when you export.

- **The "compatible player" checkbox does change the output**, but even then the subtitles didn't appear for me. Two things I noticed that could be the cause: the subtitles only seem able to load when the content is opened from a real web address (like inside an LMS or a server), **not** when you just double-click the exported file on your computer — so testing locally always looked broken. And the subtitle track doesn't seem to be set as the default one, so even when the CC button is there, nothing turns on automatically.

## Suggestions

1. Let people actually **select `.srt` (and `.vtt`) files** in the File Manager, instead of having them greyed out in the file chooser.
2. **Convert `.srt` to `.vtt`** when the subtitle is added or exported, so it works with the normal video player without extra steps.
3. Make the subtitle turn on by default (or make it clearer how to enable it), and make sure it works when previewing, not only inside an LMS.

## Attached test files

- `srt-native-player-no-subtitles.elpx` — project with the "compatible player" box **unchecked**.
- `srt-native-player-web-export.zip` — its Web export; subtitles never show.
- `srt-compatible-player.elpx` — project with the box **checked**.
- `srt-compatible-player-scorm-export.zip` — its SCORM export; subtitles still don't show.
- `test-subtitle.srt` — the subtitle file used in the tests.
- `test-video.mp4` — the video used in the tests.

[subtitles.zip](https://github.com/user-attachments/files/29644496/subtitles.zip)

erseco:
## Summary

Video subtitles attached to a video inside the Text iDevice never worked, for two separate reasons:

1. The File Manager's upload window would not let you pick a `.srt` or `.vtt` file, even though the subtitles field is labeled "Subtitles (srt / vtt)".
2. Even when a `.srt` file was attached anyway, browsers only understand the WebVTT subtitle format for native video playback — not SubRip (`.srt`). The app was inserting the raw `.srt` file straight into the video without ever converting it, so no subtitles ever appeared, whether in the Preview panel or in a downloaded Web Site / SCORM package.

This PR fixes both: `.srt` and `.vtt` can now be selected in the File Manager, and `.srt` files are automatically converted to a valid `.vtt` file wherever the video is shown or exported (Preview, Web Site export, SCORM export), so the referenced subtitle file and the video track always stay in sync.

**Scope:** this PR deliberately stays on the native `<video><track>` path and does **not** modify the legacy MediaElement.js JavaScript player (the "Player compatible with .srt subtitles" checkbox). With the SRT → WebVTT conversion in place, native captions work everywhere without that player. Removing the player itself is tracked separately in #1763.

Fixes #2034

## How to test this manually

- [x] Open the app and create (or open) a project.
- [x] Add a Text iDevice to a page.
- [x] Insert a local video into the Text iDevice (Insert/Edit Media dialog).
- [x] Switch to the "Subtitles" tab in that same dialog and click "Choose a file" to browse for a subtitle.
- [x] Confirm you can now actually select a `.srt` file (and a `.vtt` file) — they used to be grayed out / not selectable.
- [x] Pick a `.srt` file, save the media dialog, then save the Text iDevice.
- [x] Open the Preview panel: the video should play and show the subtitle text at the right time (enable captions/subtitles on the player if they're not shown by default).
- [x] Export the project as a Web Site, unzip it, and check that the subtitle file included is a `.vtt` file (not `.srt`) and that it actually contains readable subtitle text.
- [x] Export the project as SCORM and check the same thing.

## Fix — Windows-1252 `.srt` showed mojibake in the editor

The workarea editor decoded subtitle blobs with a plain UTF-8 `Blob.text()`, so a Windows-1252/Latin-1 `.srt` with accented cues showed mojibake in the live editor while the export/preview (which does strict-UTF-8-then-Windows-1252) stayed correct. The editor display path now shares the same fallback decode, so accented cues (e.g. "Canción") render identically in the editor and the export. Commit `471d83c8`; test added.


ignaciogros (dismissed):
Tested in a local Windows environment.

Although the solution works, I think a few details should be reviewed:

* The player (MediaElement.js) is not loaded inside eXe. This was already happening. Related: `/public/app/yjs/AssetManager.js:5444` replaces videos with others that never include the `mediaelement` class.
* `/public/app/common/common.js` checks for `mediaelement` elements (there are none inside eXe) by looking for the `mediaelement` class, and if it finds any that have subtitles, it initializes `mediaelementplayer`.

If these two issues are addressed, and the player is only loaded when the video has SRT subtitles, the JavaScript player would work correctly inside eXe.

However, a conversion to VTT is currently being performed. That works, and it may be the better approach, but in that case MediaElement.js is no longer needed, either inside eXe or outside it (when exporting).

It may be better to postpone merging this change into `main` until the release of eXeLearning v4.0.3, and use that time to remove the player from the application entirely, including the configuration options in the TinyMCE plugin.

Please let me know what you think.

erseco:
Agreed on the redundancy: with the SRT → WebVTT conversion in this PR, native `<video><track>` captions work in the Preview panel and in the Web Site / SCORM exports without MediaElement.js, so the player is no longer needed for subtitles.

I've kept this PR scoped to that native flow and backed the JS player out of it (commit 24118bd):

- Reverted the follow-up commit that patched the player (the default `srclang` in `common.js` and the `.vtt` gate). `common.js` and `common.test.js` are now identical to `main` again, so this PR no longer touches the old player at all.
- Dropped the corresponding "Player compatible with .srt subtitles" (jsplayer) E2E case. The remaining specs cover the native Preview cues and the Web Site / SCORM `.vtt` exports.

I'd rather not remove MediaElement.js, the TinyMCE checkbox, the vendored `exe_media` library or the translation strings in this PR: that also involves the separate `interactive-video` iDevice (which uses the same player independently) and the XLF catalogs, so it's a larger change that fits your "postpone until v4.0.3" suggestion better as its own PR. It's already tracked in #1763.

Could you take another look?

ignaciogros (dismissed):
You’re right: it’s better not to remove the player yet, both because of the interactive video feature and the legacy content that uses `.srt` files (unless an automatic conversion to `.vtt` is performed when opening them).

**With the current solution:**

* VTT works in all cases.
* SRT does not work inside eXe in edition mode (the `src` attribute of the `track` element is a `blob:` URL with no file extension).
* It does work in preview, web, and SCORM exports, but in all cases it is treated as `.vtt`, and the player is not used, even it's included in the exports: `libs/exe_media`.

I suggest waiting until these issues are fixed before merging into `main`, and, as you mentioned, keeping MediaElement.js for now to avoid compatibility issues or functionality loss.

erseco:
@ignaciogros thanks for the detailed review. Both points are addressed in bc82207a:

- **SRT in edition mode**: the workarea live view now serves the `.srt` track as WebVTT on the fly (`AssetManager._displayBlobURLForAsset`), so the native `<video><track>` shows cues in edition mode too — no more extension-less `blob:` SRT. The stored asset stays `.srt`; only the display blob is converted.
- **`exe_media` player**: agreed it's dead weight for this native flow, but it's still a live dependency of the `interactive-video` iDevice, so full removal stays scoped to #1763.

While reviewing this I also found and fixed a related bug: the SRT→VTT conversion was only wired into the Web/SCORM/Preview writers, so EPUB3, component and per-page/branch exports shipped `.vtt` files containing raw SubRip bytes (zero cues). All exporters now share one converter. Added a Windows-1252 decode fallback, graceful degradation, and an edition-view E2E test.

Could you take another look and let me know if you'd like any further changes?

erseco:

- **Minor:** `public/app/yjs/AssetManager.js:864` (`_displayBlobURLForAsset`) decodes the subtitle blob with `asset.blob.text()`, which per the Blob spec always decodes as UTF-8 (invalid bytes → U+FFFD). The export/preview path `BaseExporter.readAssetDataAsText` (`src/shared/export/exporters/BaseExporter.ts:633`) deliberately does strict UTF-8 then a Windows-1252 fallback for exactly the "very frequent Latin-1 `.srt`" case its own comment calls out. Net effect: a Windows-1252 `.srt` with accented cues shows mojibake in the **live workarea editor** while the downloaded Web/SCORM/EPUB export stays correct — display and export diverge. Note the Preview panel is *not* affected: it routes through `Html5Exporter.addAssetsToPreviewFiles` → `resolveAssetExportData` → `readAssetDataAsText` and inherits the fallback. Low impact (editor-only, cosmetic, stored asset untouched), but reusing the strict-then-1252 decode via `blob.arrayBuffer()` would restore single-source consistency.
- **Test gap:** the only fixture (`test/fixtures/subtitles/test-subtitle.srt`) is UTF-8, and `AssetManager.test.js` has no Windows-1252 case, so this display branch is unexercised. A Latin-1 fixture asserting accented cue text survives in the display blob would close it.

Verified: read the changed paths at head; the SRT→WebVTT conversion, File Manager `.srt/.vtt` accept change, and export-side fallback are otherwise sound. CI: all checks green on bc82207 ✅. Residual risk: no local build/test/server was run; the editor-display divergence above is inferred from the code, not reproduced at runtime.

---
**Update — resolved (`471d83c8`):** editor subtitle display and export now share one decode (strict UTF-8 → Windows-1252 fallback), so Latin-1 `.srt` cues no longer diverge; test added.

ignaciogros (approved):
Thanks, @erseco.

It works with both `.srt` and `.vtt` files in edit mode, as well as when previewing or exporting.

I'm only adding this comment as a reminder to improve the usability and avoid confusing users. The application still displays the file as `.srt` when an `.vtt` file is exported, and it also allows users to select a player that, although it is included in the export, is never used.

PR #2036, opened 2026-07-04
perf(static): drop source maps, .d.ts, and unused slide/src from static build

erseco:
## Summary

Follow-up to the static build size audit related to #1910. `dist/static` was shipping a handful of dev-only files that are never fetched at runtime:

- Third-party **source maps** (`.map`) — Bootstrap, Interact, DOMPurify, Showdown
- **TypeScript type declarations** (`.d.ts`) — TinyMCE, Pixo
- The `slide` iDevice's unbundled **TS source tree** (`files/perm/idevices/base/slide/src/`) — the editor already ships its built JS; the `.ts` sources next to it were never referenced by any `<script>` tag

`copyDirRecursive()` in `scripts/build-static-bundle.ts` now excludes `.map`/`.d.ts` by default (alongside the existing `.test.js`/`.spec.js` filter), and the `files/perm` copy also excludes any directory named `src`.

**Out of scope on purpose:** `.md`/`README*`/`CHANGELOG*`/`LICENSE*` are **not** touched by this PR — keeping this change small and focused. Those can be a separate follow-up if we decide to pursue them.

## Measured impact

Rebuilt `dist/static` from a clean `make build-static` on `main` (`c3c3273c8`):

| | Before | After | Savings |
|---|---|---|---|
| `dist/static` (uncompressed) | 89,844,141 B | 88,020,746 B | **−1.74 MB** (25 fewer files) |
| Release ZIP (`dist/static.zip`) | 49,436,162 B | 49,024,254 B | **−0.39 MB** |

## Files removed from the bundle

- `libs/bootstrap/bootstrap.min.css.map`
- `libs/bootstrap/bootstrap.bundle.min.js.map`
- `libs/interact/interact.min.js.map`
- `app/common/edicuatex/js/DOMPurify/purify.min.js.map`
- `libs/showdown/showdown.min.js.map`
- `libs/tinymce_5/js/tinymce/tinymce.d.ts`
- `libs/pixo/pixo.d.ts`
- `files/perm/idevices/base/slide/src/**` (18 `.ts` files)

## Tests

- `scripts/build-static-bundle.spec.ts` — 7 new cases for `copyDirRecursive` covering the new exclusions, plus regression cases for the existing `.test.js`/`.spec.js` filter and non-existent source dirs.
- `make fix` clean, `make test-unit` green (7448 tests, 90%+ coverage maintained).

## QA checklist

- [ ] `make build-static` completes without errors
- [ ] `du -sh dist/static` is smaller than on `main` (expect ~1.7 MB less)
- [ ] `find dist/static -name "*.map"` returns nothing
- [ ] `find dist/static -name "*.d.ts"` returns nothing
- [ ] `find dist/static -path "*/slide/src*"` returns nothing
- [ ] `dist/static/files/perm/idevices/base/slide/edition/slide.js` and `slide-editor.bundle.js` still exist (the iDevice's actual runtime files are untouched)
- [ ] `dist/static/libs/tinymce_5/CHANGELOG.md`, `libs/README.md`, and third-party `LICENSE`/`license.txt` files are still present (confirming we did *not* remove docs/licenses)
- [ ] Workarea loads and the `slide` iDevice can still be added/edited/saved
- [ ] At least one export (HTML5 or SCORM) still generates correctly
- [ ] `make test-e2e-static` passes


erseco:

- Verified the change at head: `copyDirRecursive` default `excludePatterns` now includes `.map`/`.d.ts` (build-static-bundle.ts:1138), and the `files/perm` copy passes `exclude=['src']` (build-static-bundle.ts:1265). Confirmed the only `src` dir under `public/files/perm` is `idevices/base/slide/src`, and every excluded `.map`/`.d.ts` file is a dev-only third-party artifact (bootstrap, interact, showdown, DOMPurify sourcemaps; tinymce/pixo declarations) — no runtime asset is dropped.
- The proposed "over-exclusion footgun" (any future dir named `src`, or any future runtime file ending in `.map`) is a hypothetical maintainability nit, not a defect: the intent is documented in a code comment and the PR body, the behavior is exact-suffix/name matching, and new colocated tests in `build-static-bundle.spec.ts` cover the exclusions and the empty-source no-op. It does not warrant a change here.
- Scope is tight and matches the stated intent; `exclude=['src']` is applied only to the `files/perm` copy, not to `public/app` (which still uses `['test','spec']`).

Verified: static reading of the diff and full function at head; `find` confirmed the only affected `src`/`.map`/`.d.ts` paths. CI: all checks green on 44b7a33 ✅.
Residual risk: no local `make build-static` was run by this reviewer, so the measured −1.74 MB size delta and the QA checklist items (workarea/slide iDevice load, export, `test-e2e-static`) rely on CI and the author's manual QA rather than independent execution.

ignaciogros (approved):
Thanks, @erseco.

Two minor, non-blocking notes:

1. The `.min.js`/`.min.css` files retain their `//# sourceMappingURL=` comment, so DevTools will return a 404 when opening them. This has no runtime impact, but the comment could be removed in a follow-up.
2. The exclusions are global and based on filename: a future `*.map` data file or a runtime `src/` directory under `files/perm` would silently be excluded from the bundle. This is acceptable now, but it’s worth mentioning.

erseco:
Thanks @ignaciogros — both notes are valid, and the first one turned out to be bigger than this PR. Both are fixed here.

### 1. Dangling `sourceMappingURL`

You're right, and while measuring it I found this PR is not the origin of the problem. On a clean checkout 13 files under `public/` announce a source map; this PR orphans 5 of them, the other 8 have been dangling on `main` all along:

| Already dangling on `main` (8) | Orphaned by this PR (5) |
|---|---|
| `libs/fabric/fabric.min.js` | `libs/bootstrap/bootstrap.bundle.min.js` |
| `libs/chartjs/chart.umd.min.js` | `libs/bootstrap/bootstrap.min.css` |
| `libs/dompurify/purify.min.js` | `libs/interact/interact.min.js` |
| `app/common/exe_tooltips/imagesloaded.pkg.min.js` | `libs/showdown/showdown.min.js` |
| `idevices/base/{checklist,rubric,progress-report}/export/html2canvas.js` | `app/common/edicuatex/js/DOMPurify/purify.min.js` |
| `idevices/base/three-d-viewer/export/model-viewer.min.js` | |

Four of those eight sit in `export/` trees, so they already 404 inside exported packages that use those iDevices. A fix scoped to "the map we just excluded" would have covered 5 of 13 and left the pre-existing ones, so this is now a single pass over the output: `scripts/static-bundle/strip-source-map-refs.ts`, run after the copies. `public/` is untouched — only the dist copy is rewritten, same "act on dist, never on the sources" rule as the rest of the build. The pass throws if a reference survives, so an unforeseen comment shape fails the build loudly instead of shipping a silent 404.

The matchers are `convert-source-map` v2.0.0's (MIT — the pair Babel and istanbul use), copied with attribution rather than added as a dependency. Worth recording *why*, because the obvious hand-rolled version is subtly wrong: in JavaScript `$` **without** the `m` flag anchors only to the absolute end of the string, never before a trailing newline, so `/…sourceMappingURL=[^\r\n]*$/` silently misses every file that ends with a line break — 6 of the 11 files I tested (`interact`, `showdown`, `chart.umd`, `html2canvas`, `model-viewer`, edicuatex's `purify`). Tests pin that case plus the legacy `//@` spelling and the inline `data:` URI form.

Measured on a real `make build-static`:

```
  Stripped 13 sourceMappingURL reference(s) (625 bytes)
```

| Check on `dist/static` | Result |
|---|---|
| `grep -rl sourceMappingURL dist/static` | 0 |
| `find dist/static -name '*.map'` | 0 |
| `find dist/static -name '*.d.ts'` | 0 |

625 bytes, so this is not a size optimisation and I'm not presenting it as one — and only DevTools ever requests a `.map`, so there was never a runtime cost for end users. The value is that the distribution stops promising files it does not ship.

### 2. Filename-based exclusions

Fixed rather than just acknowledged:

- `.map` → `.js.map` + `.css.map`. Covers all five real files; a future `*.map` data file no longer disappears (there is a test for exactly that).
- `['src']` → exclusion by path relative to the copy root, so it reads `idevices/base/slide/src` instead of "any directory named `src` at any depth". `copyDirRecursive` still accepts bare names, so `public/app`'s `['test','spec']` is unchanged, and a test asserts a same-named directory elsewhere in the tree survives.

### One caveat while this lands on its own

`dist/static/bundles/libs.zip` still contains both Bootstrap source maps (183,303 B of a 318,229 B archive): `BASE_LIBS` in `build-resource-bundles.js` packs from `public/` before the dist exists, so neither the exclusion nor the strip pass reaches them. I left them deliberately — `libs.zip` is also what `src/routes/resources.ts:469` serves to assemble every exported package, so dropping the `.map` entries without dropping the announcement would just move the 404 into published content. The tripwire in `constants.spec.ts` encodes that: a library in `BASE_LIBRARIES` that still announces a source map must ship its `.map`.

**Update** — #2260 has now landed the other half (`e0245e3b9`): the `//# sourceMappingURL=` line is deleted from the two vendored Bootstrap dist files (48 and 45 bytes), so exports, the resource bundle, the static distribution and server mode all agree with no per-provider transform, and the `.map` files stay in git to attach by hand. Two consequences for this PR, both already pushed: the tripwire's plausibility guard now counts entries that resolve on disk rather than entries that announce a map, because zero announcements is the correct end state and used to be indistinguishable from a broken lookup; and once both branches are in, the pass here reports 11 stripped references instead of 13, since Bootstrap's two are fixed at the source.

### Tests

- `make fix` — clean, no new warnings.
- `bun test ./src/shared/export ./scripts` — 2,378 pass / 0 fail.
- `make build-static` — exit 0, output verified above.
- `make test-e2e-static` — 432 passed, 0 failed, 134 skipped (13.3 min).

Prior art, in case it's useful: jQuery removed the comment from its distributed minified build in 1.11/2.1 for exactly this reason (bugs.jquery.com ticket 14415), and jsDelivr and Drupal reached the same conclusion — if you don't ship the map, don't advertise it.

ignaciogros:
Hi @erseco,

Apart from that, I think that server mode announces missing maps too, but maybe that can be fixed in a different issue/PR.

Thank you.

erseco:
@ignaciogros conflicts are gone — just merged `main` into this branch.

On server mode: serving the `.map` files there is not a problem, and this PR does not touch that. The point of dropping maps here is the **static distributable** (`dist/static` / the PWA zip): `copyDirRecursive` skips `.js.map`/`.css.map`, then `strip-source-map-refs.ts` deletes the leftover `sourceMappingURL` from that copy so the zip does not advertise files it no longer contains. Server mode keeps serving `public/` as-is, maps included, so DevTools still resolves them.

#2260 is the Bootstrap exception, and for a different reason: it stops putting the two `.map` files into **exported packages** (`BASE_LIBRARIES` / `libs.zip`, ~0.9 MB). Those packages copy the same two vendored files that server mode serves, so the announcement has to come out of the files themselves (`e0245e3b9`, 48 + 45 bytes) or every SCORM/HTML5 zip 404s in DevTools. The `.map` files stay in git — this is not a “don’t serve maps on the server” policy.

`interact`, `showdown` and edicuatex’s `purify` still have their `.map` next to them in `public/`, so server mode is fine; only the static copy loses the file, and the strip pass here covers that.

The other `sourceMappingURL`s already dangling on `main` (fabric, chartjs, `libs/dompurify`, html2canvas ×3, model-viewer, imagesloaded’s broken CDN URL) are a pre-existing DevTools 404, unrelated to this size work. Leaving them.

ignaciogros:
You're right about the server mode. I should've seen that.

Ready to be merged.

Thank you!

PR #2037, opened 2026-07-04
perf(static): switch LOMLOE/DigCompEdu datasets from gzip to zstd

erseco:
## Summary

Related to #2036 (strip source maps/.d.ts/slide-src from the static build) and #1910 (assemble resource bundles from loose files, drop shipped zips) — both part of the same static-build size reduction effort.

The LOMLOE/DigCompEdu curricular JSON datasets were the single largest remaining chunk of the static build after those two: **9.57 MB gzipped from 50.56 MB raw**. Zstandard (level 19, native `zlib.zstdCompressSync` at build time) compresses the same 12 files to **3.01 MB** — a 94% reduction, vs gzip's 81%.

The browser doesn't support zstd in `CompressionStream`/`DecompressionStream` (verified empirically against a real Chrome 150 — no format support, no roadmap; Firefox/Safari don't implement it either), so decompression uses **`fzstd`** (101arrowz, MIT, ~3.9 KB gzipped) — vendored the same way as the existing `fflate` UMD build. `lomloe.js`/`digcompedu.js` each gained a `.zst`-via-`fzstd` tier ahead of their existing plain-JSON/XHR fallback chain; the `.gz` tier is removed since the build no longer produces it.

`fzstd`'s only known issues (an int32 overflow bug on >256MiB compressed frames, and a reported streaming-mode corruption bug >128KB) don't apply to this use case: single-shot `decompress()`, largest real frame here is 0.49 MiB.

## Motivation

A prior compression-analysis spike measured gzip vs brotli vs zstd on the real datasets and found zstd matches brotli's ratio almost exactly while needing a ~45x smaller browser decoder (`fzstd` 3.9 KB vs the best available brotli WASM fallback, `tiny-brotli-dec-wasm`, ~173 KB) — brotli also isn't natively supported in Chrome's Compression Streams API (confirmed with a Chromium engineer's own explanation: a 190 KB binary-size cost for the built-in dictionary that was never judged worth shipping). zstd itself isn't in any browser's native Compression Streams support either, so a JS decoder is required either way; `fzstd` is the smallest and best-fit option for this payload shape (small/medium JSON, no bundler available in the iDevice loading context, no streaming needed).

## Measurements

Rebuilt `dist/static` from a clean `make build-static` on this branch:

| Scenario | Before (gzip) | After (zstd) | Delta |
| --- | ---: | ---: | ---: |
| `dist/static` (uncompressed) | 89,844,141 B | 82,977,196 B | **−6.55 MB** |
| Release ZIP (`dist/static.zip`) | 49,436,162 B | 42,592,408 B | **−6.53 MB (−13.8%)** |
| LOMLOE + DigCompEdu datasets only | 9.57 MB | 3.01 MB | −6.56 MB (−94% vs raw) |

Byte-for-byte integrity was verified for all 12 real datasets (compress with native zstd, decompress with `fzstd`, compare SHA-256 against the original) in both Bun and a real Chrome 150 browser (smallest file 101 KB, largest 8.3 MB) before wiring this into the loaders.

## Compatibility

- Datasets remain plain JSON after decompression — no format/schema change.
- Existing plain-JSON and XHR fallback tiers are untouched; sessions without `fzstd` loaded (or where the `.zst` fetch fails) fall through exactly as before.
- No change to `.elp`/`.elpx` project files, exports, or any ZIP container format.
- `fzstd` is now a direct `dependencies` entry in `package.json` (previously only ever resolved as a transitive dep for a different package in an unrelated investigation — not applicable here, this is a fresh addition).

## Tests

- [x] `bun test scripts/build-static-bundle.spec.ts`
- [x] `npx vitest run public/files/perm/idevices/base/lomloe/edition/lomloe.test.js` (111 tests, incl. 3 new for the `.zst` tier)
- [x] `npx vitest run public/files/perm/idevices/base/digcompedu/edition/digcompedu.test.js` (new file, 5 tests)
- [x] `make fix`
- [x] `make test-unit` (backend, full suite)
- [x] `make test-frontend` (frontend, full suite)
- [x] `make build-static` and manually confirm `dist/static/files/perm/idevices/base/{lomloe,digcompedu}/data/*.zst` exist with no `.gz` left behind
- [x] Manually open the LOMLOE and DigCompEdu iDevices in the workarea and confirm the curriculum/framework data still loads and renders
- [x] `make test-e2e-static`

## Risk

Low. The decompression tier is purely additive-with-fallback: if `fzstd` fails to load, throws, or the `.zst` fetch 404s, the code falls through to the pre-existing plain-JSON fetch (and then XHR) exactly as it did before this change — the worst case is a slower load, not a broken iDevice. `fzstd` itself hasn't been updated since March 2024 and has a couple of open, unfixed upstream issues, but neither applies to the payload sizes/usage pattern here (verified with the real datasets, not assumed). Rollback is a straight revert — no data migration involved, since the source `.json` files and build pipeline are the only things that changed.


erseco:

The gzip→zstd swap is clean: encoder (`zstdCompressBuffer`, level 19) and the `.zst`-via-`window.fzstd` decode tier in both loaders are consistent, `fzstd` is vendored + registered in `static-index.html`, and the build spec round-trips real datasets. One test gap is worth closing:

- **Test gap:** `public/files/perm/idevices/base/lomloe/edition/lomloe.js:429` (`fetchJsonMaybeGzipped`) has no automated coverage of the real native-zstd→`fzstd` decode path. Its unit tests mock `window.fzstd.decompress` to return the JSON bytes, and no static-scoped Playwright spec opens the LOMLOE iDevice — DigCompEdu is covered by `test/e2e/playwright/specs/idevices/digcompedu.spec.ts` under the `static` project, but that exercises the separate `digcompedu.js` loader, not `lomloe.js`. This matters because `compressJsonInDir` deletes the raw `.json` (`fs.unlinkSync`), so in the static/Electron build only `.zst` ships: if a real LOMLOE `.zst`→`fzstd` decode ever regresses, the "plain-JSON fallback" fetches a deleted file (404) then XHR-404s and the iDevice fails to load — the graceful degradation only holds in dev. Mirror `digcompedu.spec.ts` for LOMLOE. (Current datasets were byte-for-byte verified manually per the PR body; this is regression protection, not a live defect.)

Verified: read the loader diffs, `build-static-bundle.ts` (`zstdCompressBuffer`/`compressJsonInDir` deleting raw `.json`), both spec files, and `playwright.config.ts` (the `static` project runs all specs against `dist/static`). No local build/tests run. CI: all checks green on cebf686 ✅.

Residual risk: I did not run a browser against a fresh `dist/static`, so real-browser `fzstd` decode of the LOMLOE `.zst` payloads is unverified in CI — exactly the gap above. The naming of `fetchJsonMaybeGzipped` (no longer gzip) is stale but harmless.

ignaciogros:
Hi, @erseco.

Thank you.

erseco:
@ignaciogros done — merged `main` into this branch, conflicts are resolved.

ignaciogros:
Another PR's just been merged, and there are new conflicts. 

Could you please take a look at them?

Thanks.

erseco:
Hi @ignaciogros I fixed the conflicts!

ignaciogros (approved):
Manually tested in both the online and Electron versions.

A Playwright spec for LOMLOE under the static project, mirroring digcompedu.spec.ts, could be added as a follow-up.

Thanks, @erseco!

PR #2147, opened 2026-07-08
[proposal] Refactor interactive-video iDevice

erseco:
Full rebuild of the **Interactive Video** iDevice: inline tabbed editor (replacing the detached fullscreen popup), versioned JSON schema with automatic legacy migration, provider adapters driven over native postMessage (no provider SDKs), honest watched-progress tracking with SCORM/xAPI reporting, and support for multiple instances per page.

Interactive Video inline editor: source field, proportional timeline with per-kind markers, and in-place row editing

## Highlights

- **Video sources**: one auto-detecting source field — local file, YouTube (nocookie), **Vimeo (new)**, EducaMadrid Mediateca, **direct video URL (new)**.
- **Editor**: inline in the page, proportional timeline with colour-coded markers, in-place row editing with "Saved ✓" feedback and per-kind save validation.
- **Interactions**: the 8 legacy question kinds plus **cover**, **pause** and **jump**; unknown kinds are preserved losslessly.
- **Compatibility**: legacy `.elp`/`.elpx`/`.idevice` migrate automatically on open/import (including the video URL); the old "content before/after" boxes become sibling Text iDevices in the same block.
- **Tracking**: completion computed over actually-watched ranges (seeking to the end no longer completes); rewinding re-arms interactions; reporting via the shared SCORM/xAPI flow.
- **Accessibility & privacy**: full keyboard operation, aria-live feedback, colour never the only cue, reduced motion honoured; privacy-variant embeds controlled via postMessage with strict origin validation.
- **Implementation**: TypeScript source under `src/` compiled to two classic IIFEs by the new centralized iDevice build (`scripts/build-idevices.ts`, ADR-0006); 760 colocated unit tests (~97% statements), bundle-contract smokes and Playwright E2E.

## Before / after comparison (with captures)

- 📄 [English report (PDF)](https://github.com/user-attachments/files/30530243/interactive-video-whats-new-en.pdf)
- 📄 [Informe en español (PDF)](https://github.com/user-attachments/files/30530240/video-interactivo-novedades-es.pdf)

## Docs

ADR-0002…ADR-0006 (schema, runtime, providers, build convention), SDD-0001, and `doc/development/idevices-typescript.md`.

## Pending (external to this PR)

- ES translation: "+ Responder" → "+ Respuesta" (handled by the translations flow).
- Visual pass by @ignaciogros.


cristinavaldera (dismissed):
I really think this idevice needed a change! However, the structure of this proposal is different from the rest of idevices. The fields in tabs "Video" ,"Interactions" and "Behaviour and scoring" are always in a first tab called "General settings" (in fact, in order to be consistent with the other idevices, I think that the order should be: fields from "beahaviiour and scoring", "Video", "interactions".  

Other issues:

- the "Use current time" does not work. I have just seen the warning that it only works with uploaded videos. Would it be possible to work also with external videos, as it works now? 
- multiple errors in questions:
- in single choice you can select 2 answers as correct (you should be able to mark just one)
- in true-false you have the same fields as in multiple answers (you should have a button for marking true or false)
-  when you mark a word as blank, html code appears ("el caballo <span style="text-decoration: line-through;">blanco</span>"....

- each question appears below the video, and afterwards the list of all the questions. Something cleaner??

Image

- it doesn´t work in Preview, only in edition (the video does not stop and teh questions are not shown)

cristinavaldera (dismissed):
Thnak you for the changes. I like this first option better, it is more similar to the other idevices.

Change proposals:

1) To be more uniform, Video and Interactions should be in the same section.

2) translation:  SingleChoice: To add a new answer it says " + Responder" instead of "+ respuesta"

3) Dropdown.  May be change color of the button so that it is more visible? (same with Cloze)

Image

4) Warning icon appears although you have filled all teh fields. It only disappears when you close teh question and open ir again.

Image

5) Order list. The numbers should be skipped. It may be confusing, especifically if the items you need to order are also numbers.

Image

6) @ignaciogros we need you here to give "a coat of paint" to all of this :)
Image

Thank you, Ernesto!

ignaciogros (commented):
Thank you for this proposal.

I agree that it is necessary to renew this iDevice's UI so that its presentation is more consistent with that of other iDevices, but I think we should first define the presentation we want for all iDevices, establish a design guide (or developer API), and then work on making all iDevices comply with it.

For example, I think the slides in this iDevice should follow a presentation similar to that of the iDevices in the Games category: adding an item using a + icon, with arrows to navigate between and edit questions.

I suggest defining the desired presentation first, then updating simpler iDevices, such as some of the games, and finally renewing this iDevice's UI.

Otherwise, we may end up having to redesign it twice.

cristinavaldera:
It is true that we would need a global approach for all the idevices. 

Martin and I will start working on those presentation definitions this week, so that we can have a first proposal to start with.

PR #2157, opened 2026-07-10
feat(three-d-viewer): add hotspots, guided navigation and questions

### LINKED ISSUE #2153: 3D Viewer Edevice
The "3D Viewer" e-device does not currently allow for interactivity involving tasks performed on the object being viewed. I would like to see the ability to add interactive features to the object itself. For example:
 The ability to include small markers or pins that explain specific parts of the object.
. The ability to include text and visual effects to guide the user while viewing the object.
. An activity that allows for assessment based on the object (such as Q&A, drag-and-drop exercises, etc.).

This would make the e-device much more appealing for use in secondary education. I am specifically thinking of adding a 3D model of the Teide stratovolcano—created by the Government of the Canary Islands—for the 4th-year ESO Biology and Geology curriculum.
Thank you very much.

erseco:
## Summary

Adds an optional **interaction layer** to the 3D Viewer iDevice: informational
**hotspots (markers)** anchored to points on the model, optional **guided
navigation** (previous / next) between markers, and lightweight **single-choice
questions** attached to markers. It works for both render paths the iDevice
already supports — GLB/GLTF via `<model-viewer>` and STL via the shared Three.js
runtime — and across the editor, preview, static HTML and SCORM exports. When
interactions are disabled the iDevice behaves exactly as before.

Closes #2153.

## Motivation

Issue #2153 asked for interactivity on the object being viewed: pins that
explain specific parts, visual guidance while viewing, and object-based Q&A
(the concrete use case being a 3D model of the Teide volcano for 4th-year ESO
Biology & Geology). Until now the iDevice could only display a model.

Design is documented up front in **SDD-0001** and **ADR-0001**
(`doc/architecture/`). Drag-and-drop activities are intentionally out of scope.

## Implementation

- **Versioned state (v2).** A new `interaction` block (`enabled`, `guidedMode`,
  `wrapNavigation`, `showMarkerLabels`, `markers[]`). Each marker has a
  renderer-independent anchor (`position` + `normal` + optional `surface`),
  an optional captured `camera`, and an `action` (`information` / `image` /
  `video` / `link` / `question`). Pure, idempotent normalization helpers are
  mirrored byte-for-byte between edition and export (same convention as the 360
  viewer).
- **Shared runtime controller.** `createInteractionLayer()` in
  `three-d-viewer-runtime.js` (the only file loaded by both editor and export)
  owns marker state, the accessible content dialog, the single-choice question
  renderer, guided navigation and ARIA. Two thin adapters implement one small
  contract: native `<model-viewer>` hotspots for GLB/GLTF, and a per-frame DOM
  overlay (raycast placement + projection/occlusion) for STL, driven off the
  existing render loop via a new `onFrame` hook. See ADR-0001.
- **Editor UX.** A collapsed *Interactions* section: enable, guided/labels/wrap
  toggles, click-to-place marker creation, an editable marker list
  (reorder/edit/delete) and an accessible marker editor with per-action fields
  and camera capture.
- **Export.** `renderView` embeds the interaction state as an escaped JSON
  `<script>` block plus a static, escaped fallback marker list and (in guided
  mode) baked prev/next controls; `renderBehaviour` boots the shared layer.
- **Assessment (SCORM).** Optional: when at least one marker carries a question,
  an *Assessment (SCORM)* section (the standard `gamification.scorm` tab) lets
  the author make it gradable. In a SCORM export the runtime registers the
  activity and reports the fraction of question markers answered correctly
  (`cmi.core.score.raw`) via the shared `registerActivity` / `sendScoreNew`
  helpers — no new scoring engine. Off by default; the whole path is guarded so
  the iDevice never depends on the gamification framework being present.

## State migration

Legacy state has no `version` and no `interaction`. On load it is migrated to
v2 by adding a **disabled, empty** interaction block, so existing projects open
and render identically. Migration branches on the data shape (not the
informational `config.xml` version), is idempotent, tolerates malformed nested
values, and never drops a valid model source. `blob:`/`data:` URLs are never
persisted (marker media included).

## Accessibility

- Each marker is a real `<button>` with a meaningful accessible label;
  decorative icons are `aria-hidden`.
- Content opens in a `role="dialog"` `aria-modal` dialog with focus move-in,
  focus trap, `Escape` to close and focus return to the marker.
- Questions use native `<fieldset>/<legend>/<input type="radio">`; feedback is
  announced via an `aria-live` region and never relies on colour alone.
- Guided nav buttons expose their disabled state and announce the active step.
- A static, escaped fallback list renders every marker's content when WebGL /
  `<model-viewer>` is unavailable.

## Testing

- **Unit (Vitest):** schema/migration/normalization, question grading, JSON
  serialization + `</script>` escaping, no `blob:`, the runtime controller
  (markers, dialog, question feedback, guided nav, fallback, cleanup) and STL
  projection/occlusion/placement math with a THREE stub, and the editor flow.
  261+ tests, all green.
- **E2E (Playwright):** `three-d-viewer-interactions.spec.ts` authors an
  informational marker and a question, saves, reloads, opens the preview, and
  verifies markers render with accessible labels, guided nav is present, the
  dialog opens, and answering the question shows accessible feedback.
- `make fix` clean; `make bundle` run.

## How to test manually

1. Add a **3D Viewer** iDevice and pick a model (GLB/GLTF or STL, e.g.
   `test/fixtures/sample-model.glb` or `test/fixtures/ascii-cube.stl`).
2. Open the **Interactions** section and tick **Enable interactions**
   (optionally **Guided navigation**).
3. Click **Add marker**, then click on the model to place it. In the marker
   editor set a label and content; choose **Question** as the action type to
   attach a single-choice question (add options, mark the correct one).
4. Add a second marker if you like, then **Save** the iDevice and the project.
5. **Reload**, open the **Preview** (or export to HTML5): click the markers to
   open their content, use **Previous/Next** to walk the tour, and answer the
   question to see the feedback.
6. To make it gradable: with a question marker present, open **Assessment
   (SCORM)**, choose *Automatically save the score*, export as **SCORM**, and
   the LMS receives a score based on the fraction of questions answered
   correctly.
7. Verify a legacy 3D Viewer (no interactions) still displays exactly as before.

## Export compatibility

- Marker data reaches the exported HTML; `asset://` media is rewritten to
  `content/resources/...` by the existing exporter (no per-iDevice rewriting);
  no `blob:` URLs are emitted.
- No new export library and no registration changes — all code lives in files
  the iDevice already ships, copied into exports by the existing directory
  recursion.
- Existing 3D Viewer content (no interaction block) exports unchanged.

## Known limitations

- STL marker occlusion uses a surface-normal-facing test (cheap and robust for
  convex-ish models); deeply concave models may show a marker that is partially
  occluded. Frustum culling still hides off-screen markers.
- Animated GLB surface anchoring uses stable position + normal anchors; the
  `<model-viewer>` `surface` reference is best-effort.
- Marker image/video use a URL/asset field in the editor for now (rich text for
  `information` and full media-library picking can be layered on later).

## Screenshots or recordings

The full author → save → reload → preview → navigate → answer flow is exercised
end-to-end by `three-d-viewer-interactions.spec.ts` (Playwright, Chromium). A
short screen recording can be added here.

## Checklist

- [x] `make fix` passes
- [x] Unit tests pass (`make test-unit` scope for this iDevice)
- [x] Playwright E2E added and passing for the new flow
- [x] Patch covered by colocated tests
- [x] No skipped/disabled tests
- [x] SDD-0001 + ADR-0001 added
- [x] No drag-and-drop activity introduced
- [x] No `translations/**` changes


erseco:

- **Minor / Test gap:** `public/files/perm/idevices/base/three-d-viewer/export/three-d-viewer-runtime.js:815` — the question's `attemptsAllowed` cap lives in the `buildQuestion` closure (`let attempts = 0`), which is rebuilt on every `activateMarker` → `openDialog`. Closing and re-clicking a marker resets attempts, so "Attempts allowed = 1" limits per open, not overall. SCORM scoring is unaffected (`correctIds[markerId]` dedupe in export/three-d-viewer.js:317-320), so it is a self-check UX gap only. The test at `three-d-viewer-runtime.interaction.test.js:210` asserts disablement within one dialog but never reopens; persisting attempt/answered state per marker id would close both the gap and the coverage hole. Non-blocking.
- No exploitable or data-loss issue found. The interaction block is off by default and legacy state migrates to a disabled/empty v2 block, so existing projects render unchanged.

Verified: code re-inspected at head `cc345f9`; `git diff` of the touched runtime/export/edition paths. CI: all checks green on `cc345f9` ✅.

Residual risk: the `sanitizeHtmlDom` foreign-content case (uppercase-keyed `banned` map misses lowercase SVG/MathML tagNames) is real but non-exploitable here — output is inserted via `innerHTML` (scripts don't execute), `on*` attrs are stripped globally, and marker `information` HTML is trusted-author content; defense-in-depth only, not a boundary break. The byte-for-byte schema duplication between edition/export follows the documented 360-viewer precedent (SDD-0001) with the real runtime already shared; a drift-detection test would help but is not a defect. No local build/test was run beyond CI.

PR #2158, opened 2026-07-10
fix(ui): compact collaborative autosave status

erseco:
## Summary

The collaborative autosave status introduced in #1985 rendered a persistent, full-sentence pill in the top toolbar (`#exe-collab-save-status`, e.g. *"All collaborative changes are saved."*). Sitting next to Save, Preview, Styles, Settings, Share, the concurrent-user list and the user menu, that sentence consumed too much horizontal space.

This PR replaces the sentence with a **compact status presentation** and keeps the accessibility guarantees intact. The autosave persistence behaviour from #1985 is **unchanged**.

## Context

This is a UI/accessibility follow-up to #1985, which introduced collaborative autosave persistence for online collaborative Yjs sessions. #1985's saving behaviour — retry/backoff, collaboration gating, the shared save path (`SaveManager.save`), persistence and teardown — is **not** altered here, and `CollaborativeAutosaveManager` remains completely DOM-free.

The one intentional behaviour change, requested during review, is the idle-autosave debounce: it is shortened from **15 s to 8 s** so the "saved" state appears sooner after editing stops, at a moderate increase in save frequency. It stays a debounce (it resets while you keep editing) and remains overridable via `window.__EXE_COLLAB_AUTOSAVE_IDLE_MS__`.

The status is now rendered by a small, focused helper, `CollaborativeSaveStatusView`, which the bridge delegates to from `_updateCollaborativeSaveStatus(phase)`. It is not a general-purpose status manager — it only maps the four autosave phases to three surfaces.

## UX behaviour

The state moves onto a single compact **status dot** on the existing **Save** button, reusing the familiar "unsaved" red-dot placement (bottom-left, absolutely positioned so the button never changes width). Colour is never the only cue (WCAG 1.4.1):

| Phase | Dot | Non-colour cue | Toast |
|-------|-----|----------------|-------|
| `pending` | red | the familiar plain "unsaved" dot (baseline) | — |
| `saving` | amber | **pulses** (motion), `prefers-reduced-motion` respected | — |
| `clean` | green | **check** glyph | — |
| `failed` | red | **exclamation** glyph | error toast (see below) |

- To avoid two overlapping markers, the manual red "unsaved" dot is **hidden while the collaborative status dot is showing** (the dot stands in for it); in non-collaborative modes the manual dot behaves exactly as before.
- The **Save** button label stays `Save`; its dirty/saving/saved semantics are untouched, and the button remains clickable in the `failed` state for a manual retry.
- The dot has a `background-color` transition, so a phase change repaints immediately (it never gets stuck on a stale colour until a hover) and the change reads smoothly.
- The status dot uses the **same size and placement as the manual "unsaved" dot** (14 px, bottom-left), so the marker looks identical whichever one is showing.
- The `saving` state is kept on screen for a **minimum (~600 ms)** so a fast save is still perceptible; a `failed` state is never delayed.

**Toasts are intentionally not shown for routine transitions** (`pending`/`saving`/`clean`) — those would be noisy and are already conveyed by the badge and the live region. A single error toast is shown only when autosave reaches the terminal `failed` state. It is:

- **deduplicated per failure episode** (the manager re-emits `failed` across its retry loop; the toast fires once),
- **re-armed only after a genuine recovery** (`clean`), and dismissed automatically on recovery,
- **persistent** (no auto-hide timeout) so the data-loss warning cannot be missed, and dismissable via its close button,
- **worded for the mode**: when the manual Save button is available it says *"…Please click Save before leaving."*; when Save is hidden (platform-integration / `data-exe-hide-save`) it drops the un-followable instruction and says *"…Your latest collaborative changes may not be saved."*

## Accessibility

- **Live region:** `#exe-collab-save-status` is retained but is now **visually hidden** (position/clip, never `display:none`) so it consumes **no** horizontal toolbar space while staying in the accessibility tree. It keeps `role="status"` + `aria-live="polite"` (with `aria-atomic="true"`) and announces every meaningful state transition.
- **Save button semantics:** the accessible **name stays "Save"**. The button gains `aria-describedby="exe-collab-save-status"` so keyboard users hear the current autosave state as the button's description. The badge is `aria-hidden="true"` (decorative), so it never alters the accessible name; its `title` is only a mouse tooltip.
- **Non-colour cues:** each phase differs from the others by a non-colour signal — `saving` pulses (motion), `clean` shows a check glyph, `failed` shows an exclamation glyph, `pending` is the plain baseline dot — satisfying WCAG 1.4.1.
- **No duplicated announcement on failure:** the assertive error toast (`role="alert"`) carries the full actionable sentence, while the polite live region announces only a short status token (*"Autosave failed."*), so a screen reader does not hear the identical sentence twice. The full sentence remains available on the badge tooltip and as the Save button's description.
- **`aria-busy`:** set to `true` on the Save button only during `saving`, removed for every other phase.
- **No layout shift:** the badge is `position: absolute`; the button width is identical across all four states; the live region is out of flow. No inline styles, no `!important`, low-specificity selectors only.

## Testing

Commands run locally and their results:

- `make fix` — Biome clean (only 2 pre-existing, unrelated warnings in `test/e2e/.../focused-edit-mode.spec.ts`).
- `npx vitest run --config vitest.config.mts public/app/yjs/CollaborativeSaveStatusView.test.js` — **24 passed**.
- `npx vitest run --config vitest.config.mts public/app/yjs/` — **3081 passed** (22 files; includes the rewritten `YjsProjectBridge` `_updateCollaborativeSaveStatus` + integration tests).
- Coverage for the new view (`--coverage`): **97.75% lines**, 100% funcs, 95.95% stmts, 90.81% branch. The only uncovered lines are the CommonJS/global dual-export idiom (`else if (typeof window …)`), which is untestable under `require` — same pattern as the sibling `CollaborativeAutosaveManager.js`.
- `bun x sass assets/styles/main.scss … ` — compiles cleanly.
- `bun x playwright test --project=chromium test/e2e/playwright/specs/collaborative-autosave.spec.ts` — **2 passed** (15.1s).

The E2E spec was updated (not weakened): the **persistence proof is preserved** — a collaborative edit is made with no manual save, the second editor sees it live, both editors leave without clicking Save, the project is reopened and the node is still present. The assertions that required the long visible sentence were replaced with:

- the compact Save-button badge reaching the correct `data-collab-phase` (`clean` / `pending`) and being visible,
- the visually-hidden live region carrying the accessible message,
- a narrow-viewport (375px) check that the Save button and badge stay visible.

## Screenshots

The four dot states and the failure toast were **rendered and captured via Playwright** in a real workarea (driving `setPhase()` for each phase and screenshotting the Save button, the toolbar, and the full failure view). Verified visually:

- `pending` → plain red dot (bottom-left of Save); `saving` → amber dot; `clean` → green dot with a check; `failed` → red dot with an exclamation.
- The toolbar shows the status as **just the dot on Save** — no extra horizontal space next to Preview / Styles / Settings / Share.
- The failure toast reads *"Collaborative autosave failed / Autosave failed. Please click Save before leaving."*

(The PNGs are not embedded here because the GitHub API used by the CLI cannot upload PR-body images; they can be attached to the PR via the web UI if desired.)

## Files changed

- `public/app/yjs/CollaborativeSaveStatusView.js` (new) + `CollaborativeSaveStatusView.test.js` (new)
- `public/app/yjs/YjsProjectBridge.js` — `_updateCollaborativeSaveStatus` now delegates to the view; teardown clears it
- `public/app/yjs/YjsProjectBridge.test.js` — delegation + disconnect-teardown tests
- `public/app/yjs/yjs-loader.js` — registers the new module
- `public/app/yjs/CollaborativeAutosaveManager.js` — idle-autosave debounce default 15 s → 8 s (only change)
- `views/workarea/menus/menuHeadTop.njk` — Save-button status dot + repurposed live region + `aria-describedby`
- `assets/styles/layout/_header.scss` — status-dot styles (colour + pulse + transition), reduced-motion, manual-dot takeover, visually-hidden live region
- `test/e2e/playwright/specs/collaborative-autosave.spec.ts` — updated assertions, persistence proof preserved

## Notes

- New user-facing strings are wrapped in `_()` only: `Collaborative autosave failed` (toast title), `Autosave failed. Your latest collaborative changes may not be saved.` (Save-unavailable wording), and `Autosave failed.` (short live-region token). The `clean`/`pending`/`saving` and the *"Autosave failed. Please click Save before leaving."* strings are reused verbatim from #1985. **No file under `translations/` was modified** — extraction is handled by the separate translation process.


erseco:

- **Minor (a11y):** `assets/styles/layout/_header.scss` (`@media (prefers-reduced-motion: reduce)` → `animation: none`) plus `PHASE_BADGES` in `public/app/yjs/CollaborativeSaveStatusView.js:52-57` (`pending.glyph=''`, `saving.glyph=''`). With the pulse suppressed, `pending` (red) and `saving` (amber) are distinguishable only by colour, which contradicts the PR's own WCAG 1.4.1 rationale that names motion as `saving`'s non-colour cue. Impact is low (both are transient "not-yet-saved" states; `clean`/`failed` retain glyphs), but a static shape/glyph for `saving` would make the claim hold under reduced motion. Not a merge blocker.
- Verified the enlarged manual `.unsaved:before` dot (10px→14px) is intentional and disclosed in the PR body — cosmetic, not a defect. The `exe-online` badge is correctly suppressed in static builds via `main.scss`, teardown clears the view, and the `aria-describedby`/visually-hidden live-region wiring is sound.

Verified: static inspection at head only. CI: all checks green on de6807f ✅. Residual risk: no local build/screen-reader run performed, so the reduced-motion colour-only nuance above is reasoned from the CSS/JS, not observed in an AT session.

ignaciogros (approved):
I like this much better than the previous solution. The only issue I see is that when you hover over the button and then over the status icon, the icon's `title` is displayed on top of the button's tooltip, which creates a slightly odd effect:

Image

That said, this may still be the best approach.

Another option would be to include the status information ("All collaborative changes are saved") directly in the button's tooltip, but such a long message might negatively affect the layout.

I like this solution, at least for the moment. Let's see what @cristinavaldera thinks.

Thanks, @erseco.

cristinavaldera (approved):

cristinavaldera:
I agree with Ignacio. We can improve the solution aferwards.

Thanks, Ernesto!

cristinavaldera:
I agree with Ignacio. We can improve the solution aferwards.

Thanks, Ernesto!

PR #2161, opened 2026-07-12
fix: prevent legacy HTML  decoder from corrupting LaTeX sequences

### LINKED ISSUE #2160: Bug: Legacy XML HTML decoder corrupts LaTeX code
## Description
`Decode HTML content from legacy XML format` is incorrectly modifying LaTeX code embedded within the content, resulting in broken or corrupted equations.

## Steps to Reproduce
1. Open legacy ELP file with LaTeX code `\times`.
2. Observe the output where the LaTeX syntax has been altered to `imes`.

## Actual Behavior
The decoder misinterprets the LaTeX control sequence (for example, `\times`), strips the `\t`, and outputs `imes`, which breaks the mathematical rendering.

franmate:
## Description
This PR fixes the bug where open legacy ELP format misinterprets LaTeX control sequences, resulting in corrupted equations.
Fixes #2160

The decoding logic has been updated to safely ignore or correctly escape LaTeX syntax while processing the legacy XML/HTML content.

**Fixes #** ## Changes Made
* Updated `Decode HTML content from legacy XML format` function to handle backslashes followed by specific characters (like `\t`) correctly within LaTeX blocks.
* Added unit tests covering LaTeX expressions containing LaTeX control sequences.

## How to Test
1. Download the attached test file: [latex.zip](https://github.com/user-attachments/files/29946812/latex.zip)
2. Open this legacy ELP (or ZIP)  file.
3. Verify that the output retains `\times` intact instead of converting it to `imes`.

### DIFF
--- a/src/shared/import/LegacyXmlParser.spec.ts
+++ b/src/shared/import/LegacyXmlParser.spec.ts
@@ -139,36 +139,30 @@ describe('LegacyXmlParser', () => {
             const result = parser.preprocessLegacyXml(input);
             expect(result).toBe('textmore text');
         });
-
         it('should remove tabs', () => {
             const input = 'text\tmore text';
             const result = parser.preprocessLegacyXml(input);
             expect(result).toBe('textmore text');
         });
-
         it('should convert Windows line endings', () => {
             const input = 'line1\r\nline2';
             const result = parser.preprocessLegacyXml(input);
             // After preprocessing: \r becomes \n, \n\n becomes \n
             expect(result).not.toContain('\r');
         });
-
         it('should convert hex escape sequences', () => {
             const input = '\\x41\\x42\\x43';
             const result = parser.preprocessLegacyXml(input);
             expect(result).toBe('ABC');
         });
-
         it('should convert literal \\n to entity', () => {
             const input = 'text\\nmore text';
             const result = parser.preprocessLegacyXml(input);
             expect(result).toContain('&#10;');
         });
-
         it('should keep encoded pre content inside unicode value attribute', () => {
             const input = 'x     <unicode value="before\t&lt;pre&gt;\tkeep     spacing&lt;/pre&gt;\tafter     end"/>';
             const result = parser.preprocessLegacyXml(input);
-
             expect(result).toBe('x<unicode value="before&lt;pre&gt;\tkeep     spacing&lt;/pre&gt;afterend"/>');
         });
         it('should preserve \\nabla instead of converting \\n to a line-break entity', () => {
@@ -177,13 +171,11 @@ describe('LegacyXmlParser', () => {
             expect(result).toContain('\\nabla');
             expect(result).not.toContain('&#10;abla');
         });
-
         it('should preserve \\newcommand', () => {
             const input = '\\( \\newcommand{\\vec}[1]{\\mathbf{#1}} \\)';
             const result = parser.preprocessLegacyXml(input);
             expect(result).toContain('\\newcommand');
         });
-
         it('should preserve \\notin, \\neq and \\ngeq', () => {
             const input = '\\( a \\notin B \\quad x \\neq y \\quad x \\ngeq y \\)';
             const result = parser.preprocessLegacyXml(input);
@@ -191,17 +183,32 @@ describe('LegacyXmlParser', () => {
             expect(result).toContain('\\neq');
             expect(result).toContain('\\ngeq');
         });
-
         it('should handle mixed content with real line breaks and LaTeX commands together', () => {
             const input = 'Line1\\nLine2 with formula \\( \\nabla f(x) \\)';
             const result = parser.preprocessLegacyXml(input);
             expect(result).toContain('Line1&#10;Line2');
             expect(result).toContain('\\nabla');
         });
-
         it('should preserve \\nabla inside a real eXe unicode value attribute', () => {
             const input =
                 '<unicode content="true" value="\\( \\nabla \\times \\vec{F} = \\left( \\tfrac{\\partial}{\\partial x} \\right) \\)"/>';
+            const result = parser.preprocessLegacyXml(input);
+            expect(result).toContain('\\nabla');
+            expect(result).not.toMatch(/&#10;abla/);
+        });
+
+        // --- REVIEWER REGRESSION TEST (Currency vs Formula, same fix as decodeHtmlContent) ---
+
+        it('should convert a literal \\n between two currency amounts (not treat it as a LaTeX block)', () => {
+            // Without a currency guard, "$5...$10" could be greedily matched as a single
+            // $...$ LaTeX block, leaving the literal "\n" between them unconverted.
+            const input = 'Price: $5\\nNew price: $10';
+            const result = parser.preprocessLegacyXml(input);
+            expect(result).toBe('Price: $5&#10;New price: $10');
+        });
+
+        it('should still protect a real $...$ formula between two currency amounts', () => {
+            const input = 'Price: $5, formula $x \\nabla y$, total: $10';
             const result = parser.preprocessLegacyXml(input);
             expect(result).toContain('\\nabla');
             expect(result).not.toMatch(/&#10;abla/);
--- a/src/shared/import/LegacyXmlParser.spec.ts
+++ b/src/shared/import/LegacyXmlParser.spec.ts
@@ -171,6 +171,41 @@ describe('LegacyXmlParser', () => {
 
             expect(result).toBe('x<unicode value="before&lt;pre&gt;\tkeep     spacing&lt;/pre&gt;afterend"/>');
         });
+        it('should preserve \\nabla instead of converting \\n to a line-break entity', () => {
+            const input = '\\( \\nabla \\times \\vec{F} \\)';
+            const result = parser.preprocessLegacyXml(input);
+            expect(result).toContain('\\nabla');
+            expect(result).not.toContain('&#10;abla');
+        });
+
+        it('should preserve \\newcommand', () => {
+            const input = '\\( \\newcommand{\\vec}[1]{\\mathbf{#1}} \\)';
+            const result = parser.preprocessLegacyXml(input);
+            expect(result).toContain('\\newcommand');
+        });
+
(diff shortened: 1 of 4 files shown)


erseco (approved):
I generated a .elp and everythinks seems to work ok, the file for testing

[issue-2160-legacy-latex-times.elp.zip](https://github.com/user-attachments/files/29956633/issue-2160-legacy-latex-times.elp.zip)

erseco (changes requested):
I approved this PR after testing only the original `\times` reproduction case. After testing a broader set of valid legacy content, I found several regressions, so I am changing my review to **Request changes**.

The attached file contains isolated test cases for both expected and problematic inputs:

[issue-2160-pr2161-regression-matrix.elp.zip](https://github.com/user-attachments/files/29958286/issue-2160-pr2161-regression-matrix.elp.zip)

The original issue is fixed, but the current implementation can still corrupt valid content in these cases:

* Existing text matching `__LATEX_BLOCK_0__` can be replaced, removed, or duplicated because the placeholder is predictable.
* HTML entities inside LaTeX expressions are no longer decoded. For example, `\( x &lt; y \)` remains encoded instead of becoming `\( x < y \)`.
* The `$...$` detection is too permissive and can interpret escaped dollar signs or ordinary currency text as LaTeX.
* An escaped dollar sign before a real `$...$` expression can leave the real formula unprotected, causing `\times` to be converted into a tab character again.
* LaTeX environments such as `\begin{equation}...\end{equation}` are not protected unless they are also wrapped in one of the recognised delimiters.

The decoder should avoid predictable placeholders and should decode HTML entities independently from the handling of Python-style escape sequences.

Please add regression tests for:

* Existing `__LATEX_BLOCK_0__` content.
* Placeholder collisions when a real formula is also present.
* HTML entities inside LaTeX expressions.
* Escaped dollar signs before `$...$` formulas.
* Ordinary currency text containing multiple dollar signs.
* LaTeX environments without `$`, `\(`, or `\[` delimiters.

franmate:
Hi @erseco , thanks for the detailed review. All requested changes are applied, and I also found a second instance of the same root cause that's worth flagging.

## Changes applied to your requests (`BaseLegacyHandler.decodeHtmlContent`)

- Placeholder is no longer predictable (was `__LATEX_BLOCK_0__`); now uses a randomized token verified against the actual content, avoiding collisions with real text.
- HTML entities are now decoded independently of Python-style escape handling, so `\( x &lt; y \)` correctly becomes `\( x < y \)`.
- Added support for `\begin{...}...\end{...}` as a recognised delimiter, alongside `\( \)`, `\[ \]`, `$ $`, `$$ $$`.
- An escaped `\$` no longer swallows a real `$...$` formula appearing later in the same content.
- Regression tests added for all 6 requested cases (placeholder collision, collision with a real formula present, entities inside LaTeX, escaped `$` before a real formula, currency text with multiple `$` — documented as a known limitation, see below — and `\begin{}...\end{}` without other delimiters).

## Additional finding: same bug in `LegacyXmlParser.preprocessLegacyXml`

While testing with real eXe content, I found the same root cause one step earlier in the pipeline. This function also converted any literal `\n` to `&#10;` without checking whether it was part of a LaTeX command (`\nabla`, `\newcommand`, etc.), so content arrived already corrupted before reaching `decodeHtmlContent`.

Applied the same block-level protection strategy (a lookahead on the next character was tried first but proved ambiguous, since real line breaks are usually followed by a letter too).

Regression tests added in `LegacyXmlParser.spec.ts` covering the same cases, plus the real eXe example with `\nabla \times \vec{F}`.

## Known limitation (documented, non-blocking)

An unescaped currency `$` (e.g. `$5`) followed by a real `$...$` formula in the same text can still be misread as one LaTeX block — no regex can reliably tell currency apart from math without more context. Documented via an explicit test so it isn't rediscovered as a new bug. Recommended mitigation: escape currency dollar signs (`\$5`).

erseco (changes requested):
Thanks for the detailed response and for finding the same issue earlier in `LegacyXmlParser.preprocessLegacyXml`.

Several of the requested changes have been addressed:

* Placeholder collisions are avoided.
* HTML entities are decoded independently from Python-style escapes.
* Escaped dollar signs before a real formula are handled.
* Simple `\begin{...}...\end{...}` environments are protected.
* LaTeX commands such as `\nabla` are now also protected during XML preprocessing.

However, not all requested changes are complete yet.

The ordinary currency case is not only a classification limitation. It can still reproduce the original corruption when currency text appears before a real formula:

```ts
const input = 'Price: $5 and formula $x \\times y$';
```

The current `$...$` pattern can match from the currency `$` to the opening `$` of the formula. This leaves `\times` outside the protected block, where it can be converted into a tab character.

The current test:

```ts
const input = 'Price: $5 and $10 total';
expect(handler.decodeHtmlContent(input)).toBe(input);
```

does not detect that misclassification because the output remains unchanged whether the text was protected as LaTeX or not.

Escaping currency signs is not a sufficient mitigation for legacy imports, since existing ELP content cannot be assumed to use `\$`.

Please add the mixed currency-and-formula case above as a regression test and adjust the detection so the real formula remains protected. Until then, I would keep this as a requested change.

Thanks for your work on this, you've it almost done!

franmate:
Hi @erseco 
Thanks for pointing out the exact reproduction case. You were right: the previous test didn't detect the corruption because the output was identical in both cases.

### The fix

The `$...$` pattern now ignores a `$` when it is followed by a clean currency amount. A digit glued directly to a letter (like `$2x+3$`) doesn't hit that word boundary, so it's still treated as a legitimate formula.

With this:
- `'Price: $5 and formula $x \times y$'`,  the first `$` is ignored (currency), the second `$...$` is correctly protected, and `\times` is never left exposed.
- `'Compute $5x \times 2$'`  , the `$` is accepted because `5x` has no word boundary between the digit and the letter.

### Regression test added

The exact case you flagged was added: `'Price: $5 and formula $x \times y$'`, along with other related cases (multiple currency amounts, digit glued to a variable, etc.).

erseco (changes requested):
Thanks for the careful iterations on this PR. The original `\times` corruption is fixed, and several important issues raised in the previous reviews have been addressed correctly:

- LaTeX placeholders are now collision-resistant.
- HTML entities are decoded independently from Python-style escapes.
- Escaped dollar signs and the mixed currency/formula case are handled better.
- Simple `\begin{...}...\end{...}` environments are protected.
- The added regression coverage for LaTeX commands is useful.

I am sorry for extending the review again, but I found a couple of regressions that should be fixed before merging. A correct solution here, with the surrounding legacy behaviours protected by tests, will be very beneficial for the project because this import path handles old user content where silent changes are difficult to detect later.

## 1. Existing `\n`, `\t`, and `\r` decoding is now broken before letters

The new safety-net expressions:

```ts
.replace(/\\n(?![a-zA-Z])/g, '\n')
.replace(/\\t(?![a-zA-Z])/g, '\t')
.replace(/\\r(?![a-zA-Z])/g, '\r');
```

prevent valid legacy escapes from being decoded whenever the following text begins with a letter. For example:

```ts
handler.decodeHtmlContent('Line1\\nLine2');
handler.decodeHtmlContent('Col1\\tCol2');
```

These cases were covered by the previous tests, but those tests were removed when the LaTeX test block was introduced. Please keep the previous decoding contract and restore regression tests for these cases. Since LaTeX blocks are already protected first, escapes outside those blocks can be decoded without this lookahead.

## 2. `LegacyXmlParser.preprocessLegacyXml()` can still treat currency text as LaTeX

The `$...$` pattern in `preprocessLegacyXml()` has no currency guard. This can prevent a valid literal `\n` from being converted:

```ts
const input = 'Price: $5\\nNew price: $10';
```

The substring between the two currency signs may be protected as a LaTeX block, leaving `\n` unchanged. Please add a regression test for this case and align the LaTeX detection with the safer behaviour used in `decodeHtmlContent()`.

## 3. Please avoid changing `&nbsp;` into a normal space

```ts
.replace(/&nbsp;/g, ' ')
```

A normal space is not semantically equivalent to a non-breaking space and may collapse when rendered. Please preserve the entity, decode it to `\u00A0`, or remove this unrelated change.

It would also be good to restore the TypeScript annotations removed from `preprocessLegacyXml()`, although that is not the main blocker.

Thanks again for the work on this. The approach is close, and once these regressions are covered, the fix should be much safer for legacy imports.

franmate:
@erseco 

Thanks for catching these — all four points are addressed:

1. Removed the `(?![a-zA-Z])` lookahead from `\n` and `\t` in `decodeHtmlContent` (kept only on `\r`, which was pre-existing behavior unrelated to your report). Restored the exact regression tests you listed (`Line1\nLine2`, `Col1\tCol2`), plus one more covering `\n` followed by a letter with no LaTeX nearby.
2. Aligned `preprocessLegacyXml`'s `$...$` detection with the same currency guard used in `decodeHtmlContent`. Added your exact case (`'Price: $5\\nNew price: $10'`) as a regression test, plus one confirming a real formula between two currency amounts stays protected.
3. `&nbsp;` now decodes to `\u00A0` instead of a regular space, with a test asserting it's not collapsed.
4. Restored the TypeScript annotations in `preprocessLegacyXml`.

Let me know if anything else needs another pass!

erseco (approved):
Thank you for addressing all the review feedback and for your patience throughout the several review rounds.

The latest changes resolve the blocking points raised previously:

- Existing `\n` and `\t` decoding behaviour has been restored and covered again by regression tests.
- LaTeX commands remain protected during both HTML decoding and legacy XML preprocessing.
- Currency text no longer interferes with the tested `$...$` formula cases.
- `&nbsp;` is now decoded as a real non-breaking space.
- The TypeScript annotations have been restored.
- The additional tests provide good coverage for placeholders, entities, environments, currency, and mixed legacy content.

The implementation now fixes the original issue while preserving the surrounding legacy behaviour much more safely.

As a non-blocking future improvement, we could consider extending currency detection to less common formats such as `$ 5` or `5 $`. These formats may still be ambiguous when combined with a later formula, but I do not think that edge case should delay this PR.

Thanks again for the changes and the careful follow-up. Approved.

PR #2162, opened 2026-07-13
fix(import): resolve legacy UDL iDevice fields by reference (#2159)

### LINKED ISSUE #2159: Legacy ELP import issue: some nodes fail to display original content and duplicate other nodes
### Bug Description
An issue has been detected when opening an older ELP file. Some nodes do not display their original content and instead duplicate the content of other nodes.

### Expected Behavior
All nodes should correctly load and display their own original content without duplicating any other sections.

### Steps to Reproduce / Attached Files
* Screenshot of the bug.

Image

* Attached is the legacy ELP file causing the issue.

[epvelp.zip](https://github.com/user-attachments/files/29943172/epvelp.zip)

### LINKED ISSUE #2167: Legacy .elp import: game iDevices lose content when their field is a <reference> (GameHandler)
While fixing #2165 I found the same reference-resolution bug as #2159, in a handler that PR #2162 did not cover.

## Problem

`GameHandler.extractHtmlView` (`src/shared/import/legacy-handlers/GameHandler.ts`) collects a game iDevice's fields with:

```ts
const fieldInstances = this.getChildElements(listEl).filter(el => el.tagName === 'instance');
```

Legacy `contentv3.xml` serializes each object once as `<instance reference="N">` and every later mention as a back-pointer `<reference key="N">`. When a game iDevice's content `TextAreaField` is written as a `<reference>` (common — the object was already serialized earlier in the tree), this `instance`-only filter finds nothing → `extractHtmlView` returns `''` → the game (Classify / Crossword / Select media files / Sort / Drag & drop / Hidden image) imports with **empty content** from a legacy `.elp`.

This is the game-iDevice sibling of #2159, which fixed the same reference-blindness in `BaseLegacyHandler` / `DefaultHandler` but not in `GameHandler`. `GameHandler.extractProperties` has the same gap — it calls `extractHtmlView(dict)` without the resolver context.

## Fix

Mirror #2159: use the shared `resolveFieldInstances(listEl, context?.resolveReference)` helper so `<reference>` fields resolve to their instance, and thread the handler context into `extractProperties`.

## Scope

Legacy `.elp` (pre-v3 pickle) imports only. Modern `.elpx` is unaffected (its games round-trip verbatim — that path was the subject of #2165).

erseco:
## Summary

Fixes #2159. Some nodes in older (pre-v3.0) `.elp` files displayed **another node's content** instead of their own.

The legacy `contentv3.xml` is a serialized Python object graph (Twisted Jelly/Marmalade) where an already-serialized object is written as `<reference key="N">`. For a UDL iDevice, the parser *already* resolved `<reference key="33">` in the `fields` list and extracted the correct content — but because `udl-content` has no dedicated handler, `DefaultHandler` then ran and **overwrote** it:

- `BaseLegacyHandler.extractFieldsContent` only read inline `<instance>` fields and **ignored `<reference>`**, so it found nothing for field 33.
- Its fallback then recursively searched *every* descendant `<instance>`, crossing the inlined `_idevice` / `parentNode` / `parent` back-references into a **different** iDevice and returning that foreign field's content (field 36).

This aligns the importer with the behaviour of eXeLearning 2.9 (iteexe `DOMUnjellier`) and 3.0.2 (`OdeXmlUtil.searchTextIdeviceOldElp`): the explicit `fields` list — resolving each reference to exactly one instance — is the authoritative source of an iDevice's content, never a broad recursive descendant search.

## Changes

- **New shared helper `resolveFieldInstances()`** — resolves both inline `<instance>` fields and `<reference key="N">` back-pointers into an ordered list, used by **both** `LegacyXmlParser` and `BaseLegacyHandler` so the two never diverge again (single source of truth).
- **Reference resolver in the handler context** — `IdeviceHandlerContext.resolveReference` (backed by the parser's O(1) instance-by-reference map); `extractFieldsContent` is now reference-aware.
- **Boundary-safe fallback** — `extractAnyTextFieldContent` no longer crosses into a nested iDevice/Node subtree.
- **No clobbering** — the generic fallback handler (`DefaultHandler.isFallback()`) can no longer overwrite an `htmlView` already resolved from an authoritative field reference.

Maps to the four recommendations in the issue.

## Tests

- `resolveFieldInstances.spec.ts` — helper unit tests (instance/reference/mixed/missing/no-resolver).
- `BaseLegacyHandler.spec.ts` — reference resolution + boundary-safe search (no crossing into JsIdevice/Node).
- `handlers.spec.ts` — `DefaultHandler` reference-based extraction and `isFallback`.
- `LegacyXmlParser.spec.ts` — inline regression reproducing the `TextAreaField 33 → JsIdevice 34 → Node 58 → Node 37 → JsIdevice 35 → TextAreaField 36` structure (was red before the fix).
- `ElpxImporter.spec.ts` — full-pipeline integration against a real fixture `test/fixtures/old_epvelp_udl.elp`, asserting the affected node keeps its own content (`Ides cubrir os apartados da Fase 2`) and does not show the duplicated content (`Agora que xa sabedes cal é o reto`).

Patch coverage on changed source is well above 90% (`resolveFieldInstances.ts` 100%, `BaseLegacyHandler.ts` ~99.7%, `IdeviceHandler.ts` 100%). `make fix` is clean; the full `src/shared/import` suite passes (690 tests). The pre-existing failures in `src/routes/**` are unrelated to this change (they also fail on a clean tree).


cristinavaldera (approved):

erseco:
Follow-up: pushed a commit that extends the same reference-resolution fix to `GameHandler` (tracked in #2167). `GameHandler.extractHtmlView` had its own instance-only field filter that this PR did not touch, so legacy `.elp` **game** iDevices (Classify/Crossword/Select media/Sort/etc.) whose content field is a `<reference>` still imported empty. It now uses the shared `resolveFieldInstances(listEl, context?.resolveReference)` helper. Added colocated unit tests.

franmate (approved):
I’ve run several import tests with ELP (legacy eXeLearning) and the results were exactly as expected. The page structure and iDevices content are perfectly preserved.

I highly recommend giving maximum priority to merging this solution, as it addresses a very critical issue: the loss of content between eXeLearning 2.9 and version 4.

Great job!  Approving the PR.

PR #2164, opened 2026-07-13
feat(yjs): add administrator-only version history recovery

### LINKED ISSUE #2163: [Proposal] Keep a configurable Yjs version history for administrator recovery
## Summary

Keep a limited history of previous Yjs document snapshots in the database so administrators can recover a project after accidental overwrites, stale-client saves, corruption, or other data-loss scenarios.

This proposal intentionally starts with a small server-side safety mechanism rather than a full user-facing version-control feature.

## Motivation

The current server save flow replaces the existing row in `yjs_documents`. Once a newer snapshot is stored, the previous state is no longer available through the normal persistence layer.

The database schema and query layer already include `yjs_version_history` and helpers for creating, listing, retrieving, and pruning historical snapshots. Connecting that existing infrastructure to the save flow should provide a relatively low-cost recovery mechanism.

## Proposed behaviour

- Apply only to server/cloud installations.
- Store previous Yjs snapshots in `yjs_version_history`.
- Keep **5 previous versions per project by default**.
- Make the retention limit configurable at installation level.
- Treat `0` as history disabled.
- Create history entries only for explicit/manual saves initially, not collaborative autosaves or unload persistence.
- Store the Yjs binary snapshot in the database rather than generating `content.xml` or complete `.elpx` files.
- Record the project, version, timestamp, user responsible for the save, and an optional description/reason.
- Prune the oldest entries after a successful save so the configured limit is respected.
- Avoid storing duplicate consecutive snapshots where practical.
- Restrict listing, inspection, and restoration to administrators in the initial implementation.
- Before restoring an old version, preserve the current state as a new safety version.
- Delete version history automatically when the project is deleted through the existing cascade relationship.

## Why database storage first

Database storage is recommended for the initial implementation because:

- The schema and query helpers already exist.
- The current Yjs snapshot is already stored as binary data in the database.
- Snapshot creation and replacement can be performed atomically in one transaction.
- It avoids filesystem/database consistency problems, orphan files, shared-volume requirements, and deployment-specific storage paths.
- It works consistently with SQLite, PostgreSQL, and MySQL.

Filesystem or object-storage-backed snapshots could be considered later for very large installations, while keeping metadata in the database.

## Configuration

Suggested setting:

```env
YJS_VERSION_HISTORY_LIMIT=5
```

The exact configuration mechanism may instead use `app_settings` if runtime administration is preferred. The initial implementation should document the chosen source and validate non-negative integer values.

## Initial implementation scope

1. Read and validate the configured history limit, defaulting to `5`.
2. On an explicit save, preserve the current snapshot before replacing it.
3. Save the new canonical snapshot and history entry atomically.
4. Prune history entries beyond the configured limit.
5. Add query/service tests covering disabled history, initial save, subsequent saves, pruning, and transaction rollback.
6. Add administrator-only endpoints for listing versions and restoring a selected version, or clearly separate those endpoints into a follow-up if they make the initial pull request too broad.

## Acceptance criteria

- [ ] The default retention limit is 5 previous snapshots per project.
- [ ] The retention limit is configurable and `0` disables history.
- [ ] The first project save does not create a redundant history entry.
- [ ] A later explicit save preserves the previous canonical snapshot.
- [ ] Autosave/unload persistence does not create history entries in the initial implementation.
- [ ] No project retains more historical snapshots than the configured limit.
- [ ] A failed save does not leave the canonical snapshot and history in an inconsistent state.
- [ ] History access and restoration are restricted to administrators.
- [ ] Restoring a version first preserves the current canonical state.
- [ ] Tests cover SQLite and remain compatible with PostgreSQL and MySQL query paths.

## Out of scope

- End-user access to version history.
- Visual diffs between versions.
- Full `.elpx` backups for every save.
- Versioned copies of unchanged assets.
- Desktop/Electron filesystem backups, which should be handled separately using platform-appropriate local storage.

erseco:
## Summary

Add a limited database-backed history of previous Yjs snapshots for recovery from accidental overwrites or corrupted saves.

The implementation reuses the existing `yjs_version_history` table and keeps the feature restricted to administrators. No database migration is required.

Closes #2163.

## Implementation

- Preserve the previous canonical Yjs snapshot before each explicit save.
- Do not create history entries for unload/automatic persistence.
- Keep 5 previous versions per project by default.
- Read the configurable retention limit from `YJS_VERSION_HISTORY_LIMIT`.
- Treat `0` as automatic history creation disabled.
- Skip consecutive duplicate snapshots.
- Save the history entry, canonical snapshot, and retention pruning in one transaction.
- Preserve the current canonical state as a safety snapshot before restoration.
- Keep restoration transactional and assign a new canonical snapshot version.

## Administrator API

- `GET /api/projects/uuid/:uuid/yjs-history`
  - Administrator only.
  - Returns version metadata and byte size without returning snapshot contents.
- `POST /api/projects/uuid/:uuid/yjs-history/:versionId/restore`
  - Administrator only.
  - Restores the selected version after preserving the current state.

No end-user history UI is included in this initial implementation.

## Tests

Added coverage for:

- Configuration parsing and the default retention of 5.
- First save without a redundant history entry.
- Previous snapshot preservation.
- Duplicate suppression.
- Configurable pruning.
- Disabled automatic history.
- Transaction rollback on canonical snapshot failure.
- Restoration with a safety snapshot.
- Administrator-only listing and restoration routes.
- Explicit save versus automatic persistence behaviour.

The rollback test uses a temporary SQLite trigger to force the canonical update to fail after the history row is inserted. This verifies that the transaction removes the history row and keeps the original canonical snapshot unchanged.

## Architecture

Adds proposed ADR-0001 documenting why the initial implementation stores Yjs history in the database rather than the filesystem or complete `.elpx` packages.

## Validation status

- Biome lint: passed in CI.
- Unit tests and coverage: passed in CI (7,472 tests).
- Integration, frontend, build, and E2E checks are still running for the latest commit.


erseco:

Reviewed against the current draft head. The change is well-scoped: admin-gated routes (`requireAdmin`), project-scoped lookups (`getVersionById` filters by `project_id`), transactional save/restore with duplicate suppression and retention pruning, and a clean `historyLimit <= 0` disable path. No blocker or important defect found.

- Minor / follow-up (not blocking): restore rewrites only the server `yjs_documents` row (`src/db/queries/yjs-history.ts:112`). Because the client is the source of truth and Yjs merges are monotonic, a still-connected client holding the newer post-incident state can re-undo a restore on next sync. This is inherent to the collaboration model and partially noted in ADR-0001; worth calling out the connected-client/IndexedDB re-merge case explicitly when the admin UI lands.
- Cosmetic: the restore response `version` (`src/routes/yjs.ts` restore handler) returns the new canonical `Date.now()` value, not the restored source version — consider adding `restoredFromVersion` for API clarity.

Verified: static inspection of the diff and full files at head; ADR-0001 and route/query DI wiring reviewed. CI: all checks green on cf26f9e ✅.
Residual risk: end-to-end restore effectiveness against a live client + load path was not exercised locally (route tests mock the history dependency; query tests hit the DB layer only). No integration/E2E test covers a full restore round-trip, so the connected-client re-merge limitation above is untested rather than disproven.

Issue #1674, opened 2026-04-14
Incorrect reference count after deleting image in desktop version


mnunezcedec:
**Description**:
In the desktop version, when working with an iDevice Text, deleting an image and saving the changes does not correctly update the reference count in the File Manager.

**Steps to reproduce:**

1. Open a Text Idevice containing an image with the intention of changing it, in the desktop version.
2. Delete a image from the content.
3. Save the resource.
4. Open the File Manager and enable "Show references".

**Expected behavior:**
The deleted image should display a reference count of 0 (in red), indicating it is no longer used.

**Actual behavior:**
The image shows a reference count of 1 (in blue), as if it were still linked somewhere in the resource.

**Additional notes:**

- If the resource is **reopened**, the reference count is corrected and properly shows 0 in red.
- The app (https://app.exelearning.net/) works correctly and does not have this problem.

erseco:
Root-caused and fixed in #1678.

**Why it happens**

`countAssetReferences` in `public/app/workarea/modals/modals/pages/modalFileManager.js` reads `compMap.get('htmlView') || compMap.get('htmlContent')`. The `htmlView` Y.Map entry is a plain string populated **once** during initial ELP import (`ElpxImporter.createComponentYMap` / `YjsStructureBinding.createComponentMapFromApi`) and is **never refreshed** after edits — `updateComponent` writes only to the `htmlContent` Y.Text.

For a Text iDevice (which saves via `jsonProperties` and triggers `apiSaveIdeviceViewHTML` internally), the save cycle updates `jsonProperties` and `htmlContent` but leaves the legacy `htmlView` string pointing at the original imported HTML, still carrying the deleted `asset://` URL. The reference counter reads that stale string first and counts it as a reference.

Reopening the project works around the issue because Y.Doc is rebuilt from the persisted snapshot (where only the authoritative `htmlContent` survives), dropping the stale `htmlView` key. That's also why `app.exelearning.net` doesn't surface it — the web session rebuilds aggressively enough to mask it.

**Fix**

- `countAssetReferences` now prefers `htmlContent` (refreshed on every save) over `htmlView` (legacy import fallback only).
- `YjsStructureBinding.updateComponent` deletes the stale `htmlView` entry once `htmlContent`/`content` holds the authoritative value, so no other reader can hit the same footgun.

Regression tests:
[...]

PR #2166, opened 2026-07-13
fix(yjs): preserve iDevice content on empty content update (#2165)

### LINKED ISSUE #2165: Game iDevices lose all content after editing in 4.0.2 (body disappears, title remains)
Reported by @omartab1312.

## Summary

In eXeLearning 4.0.2 (cloud `.dev` and desktop), some **"game" iDevices lose their entire content** after a project is edited. The iDevice title/heading stays in place, but the body is blank — both in the workarea and in any re-exported `.elpx`. The content is genuinely gone from the stored project, not just hidden: re-exporting the project produces an empty `<htmlView>` for the affected components.

## Affected iDevices

The quext-family game iDevices, whose whole state lives in a single HTML field with an empty `jsonProperties` (so the HTML is the **only** copy of their content):

- **Classify** (`clasifica`)
- **Crossword** (`crucigrama`)
- **Select media files** (`seleccionamedias`)

Other games that store everything in HTML the same way (Sort, Drag & drop, Hidden image, …) are exposed to the same failure.

## When it happens

Observed after heavy/collaborative editing of a project in 4.0.2:

- several people editing the same project,
- images replaced/deleted,
- blocks / iDevices moved or reordered,
- the game iDevices themselves opened for editing and saved.

A validated, ready-to-publish learning object lost the content of several games on pages such as *"El viaje del camello"* and *"El reto del camello"*, while the surrounding text/UDL/other iDevices were untouched.

## Impact

Irreversible content loss for the affected iDevices in the live project. The only recovery is re-importing an older `.elpx` export that still contains the content (import is lossless, so the content comes back).

## Root cause (brief)

`YjsStructureBinding.updateComponent` drops the `htmlView` fallback whenever `htmlContent`/`content` is written (introduced in #1678, issue #1674, to keep File Manager reference counts fresh). When a **transient empty content update** reaches this chokepoint — an interrupted edit, a failed serialization, or a concurrent-edit race — it blanks the `htmlContent` Y.Text **and** deletes the `htmlView` fallback, leaving the iDevice with no content in either field. Game iDevices are worst hit because they keep no copy in `jsonProperties`.

Fix incoming.

erseco:
Fixes #2165. Thanks @omartab1312 for the detailed report and files — they made this reproducible.

## The problem

"Game" iDevices (**Classify**, **Crossword**, **Select media files**, and other quext games) lost **all** their content after a project was edited in 4.0.2: the iDevice title stayed, but the body was blank — both in the workarea and in any re-exported `.elpx`. The content was genuinely gone from the stored project, not just hidden.

These games keep their whole state as an encoded blob in the `htmlView`/`htmlContent` field with an **empty `jsonProperties`**, so that HTML field is the *only* copy of their content.

## Root cause

Every content write goes through `YjsStructureBinding.updateComponent`. When it writes `htmlContent`/`content`, it also deletes the `htmlView` fallback.

If a **transient empty content update** reaches this chokepoint — an interrupted edit, a failed serialization, or a concurrent-edit race — it blanked the `htmlContent` Y.Text **and** deleted the `htmlView` fallback, leaving the component with no content in either field → irreversible loss. Non-game iDevices survive because they keep a copy in `jsonProperties`; games do not.

Confirmed end-to-end: `updateComponent(id, { htmlContent: '' })` on a game with a full `htmlView` (19,386 chars) left **both** fields empty — exactly the reported state.

### Where the regression was introduced

The destructive `compMap.delete('htmlView')` was added in **#1678** (issue **#1674**), commit `01f1c7f9` — *"fix(filemanager): clear stale htmlView so reference count updates after image delete"*. It has existed since then; the reporter's heavy/collaborative editing between two exports triggered the latent hole. 

## The fix

`updateComponent` now treats content as non-destructible:

- it **refuses an empty** `htmlContent`/`content` update when the component already holds content (in `htmlContent`, `content`, or `htmlView`), logging a warning instead; and
- it only drops the `htmlView` fallback when the **new value is non-empty**.

Real edits always emit a non-empty wrapper, so this never blocks legitimate editing, and the original #1674 behaviour (dropping the stale `htmlView` on a real edit) is preserved.

## Tests

- **Unit** (`YjsStructureBinding.test.js`): empty/whitespace content updates preserve `htmlView` and an existing `htmlContent`; the fallback is not dropped for empty values; an empty write is still applied when there is nothing to lose; the #1674 non-empty behaviour still deletes the stale `htmlView`.
- **E2E** (`game-content-empty-write-guard.spec.ts`): imports a project with game iDevices and asserts an empty write can no longer erase one.

## Recovery for already-affected projects

Import is lossless, so re-importing an older `.elpx` export that still contains the content restores it.


ignaciogros (commented):
I ran a test using the [[eXeLearning 2.9 manual](https://descargas.intef.es/cedec/exe_learning/Manuales/manual_exe29/referencias_y_fuentes.html)](https://descargas.intef.es/cedec/exe_learning/Manuales/manual_exe29/referencias_y_fuentes.html).

When I open the content with this branch, go to the **"Classify"** section, and edit the activity, I find that the images have `blob:app://` URLs. I can't save the iDevice without making any changes because I get the following message:

I thought this was the test I was supposed to run to verify the change. Could you please confirm it that's right?

Thanks.

erseco:
Thanks for testing this, @ignaciogros — and no, that is not the test for this PR. Sorry for the confusion. You did find **two real bugs**, but neither is caused by this branch and both reproduce on `main`. I have now reproduced both against the manual's `.elp` and fixed them separately:

- #2186 → **#2188** — the `blob:app://` URLs.
- #2187 → **#2189** — the validation message.

You are on both as reviewer, each with reproduction steps.

## Why it is not a regression from this branch

This PR touches one method, `YjsStructureBinding.updateComponent`, and both hunks only *prevent* writes: they skip an empty content update and keep the `htmlView` fallback when the incoming value is empty.

1. **The import path never calls it.** Importing an `.elp`/`.elpx` builds components through `createComponentMapFromApi` (`YjsStructureBinding.js:3042`, via `ComponentImporter`), not `updateComponent`. Importing the manual and opening the Classify editor does not execute the new code even once.
2. **The guard only fires on an empty content write**, and you never got as far as saving. In your flow the behaviour is identical to `main`.

## 1. The `blob:app://` URLs (#2186 → #2188)

`asset_url_resolver.js` monkey-patches `Element.prototype.getAttribute` so that reading `href` from an anchor holding `asset://` returns the resolved **blob** URL from cache. `classify.js:693` reads its image references exactly that way, so it gets a blob URL even though the DOM holds `asset://<uuid>`. The `app://` part is just Electron's origin; in the web build you would see `blob:http://localhost:8080/...`.

The fix restricts that substitution to anchors **in the live document** — lightbox libraries keep working, while the detached wrappers iDevices use to parse their own stored HTML get the persistent `asset://` back. ~19 iDevices share that pattern.

## 2. The validation message (#2187 → #2189)

Not the blob URLs, and not an import failure — a single bad card plus a validator that ignores `audio`.

**Card 7 (the cow)** is declared `type: 0` (image) with no image, only `COW2.mp3`. It gets no image anchor, so `url` stays `""` and `validateData` rejects it. No validator ever reads `audio`, even though the message promises *"an image, a text **or/and an audio**"* — and the player is built to render an audio-only card as a big audio button (`CQP-LinkAudioBig`). The editor was refusing content the player displays by design.

Two corrections to what I said earlier in this thread, since I got them wrong before checking:

- **The legacy import is faithful** — my first suspicion was an import gap and it was wrong. The anchors are present with the expected class names and `resources/...` hrefs, and the rewriter resolves them to `asset://`. Card 7 is `type: 0` with an empty `url` **in the original 2.9 file**; eXe 2.9 accepted it, 4.x did not.
- **My "normalise into an audio card" suggestion was not viable** — there is no audio card type (only Image / Text / Both), which is exactly why the type must not gate the audio. #2189 instead implements the rule the messages already state, in one shared predicate.

## The test that does exercise this PR

#2165 is a **content wipe after editing**: the iDevice title stays and the body goes blank, in the workarea and in any re-export. So:

1. Open a project containing game iDevices (Classify / Crossword / Select media files).
2. Edit and save — ideally with the messy conditions from the report: images replaced or deleted, blocks/iDevices moved or reordered, concurrent editing.
3. Confirm the iDevice body never goes blank, and that a re-exported `.elpx` still has a populated `<htmlView>`.

The automated equivalent is `test/e2e/playwright/specs/game-content-empty-write-guard.spec.ts`.

erseco:

PR #2171, opened 2026-07-13
Fix FX elements in Form and Magnifier iDevices

### LINKED ISSUE #2154: "FX effect" on the "Magnifying Glass" device.
For the "Magnifying Glass" (Lupa in spanish) e-device, I decided to animate the displayed image using FX effects—specifically, the "tabs" option. It works correctly while I am working on the e-device itself, but the problem arises when I view it in presentation mode; the tabs do not display correctly there.

 In presentation mode; the tabs do not display correctly there.

I am working online, and the version displayed is v4.0.2 – 2026-07-07.

Thank you very much for your work.
Many thanks.

Image

### LINKED ISSUE #2170: Bug: FX elements inside Form iDevice work in the editor but fail to generate HTML upon export
### Title
Bug: FX elements inside Form iDevices work in the editor but fail to generate HTML upon export

### Description
When adding an FX element (visual effect) inside a Form iDevice, there is a discrepancy between the editor environment and the exported output:
* **In the editor:** The HTML structure is correctly created, the nodes are injected, and the visual effect works perfectly.
* **In the exported version (or preview):** The FX HTML is not generated in the exported file.

*Note: Initially, this was suspected to be a conversion/migration bug from legacy ELP, but it has been confirmed that the issue persists even when creating brand-new content from scratch in eXeLearning 4.0.*

### Steps to Reproduce
1. Create a new project in **eXeLearning 4.0.2**.
2. Add a **Form** iDevice.
3. Inside the text field, insert any FX element (e.g., Tabs, Accordion) and add content to it.
4. Verify that the FX element displays and behaves correctly while inside the editor.
5. Save the iDevice and **Preview** or **Export** the project.
6. Verify that the FX element is not displayed.

### Actual Behavior
In the exported output, the HTML block that should contain the FX structure (such as tab containers or collapsibles) is completely missing or stripped out, leaving the content unformatted or invisible.

### Expected Behavior
The FX element's HTML structure generated within the editor should remain intact and be exported properly so that the theme scripts can apply the visual effect in the end-user's browser.

### Related Issues
* May be related to #2154 

[error-fx_web.zip](https://github.com/user-attachments/files/29974940/error-fx_web.zip)

Image

erseco:
## What this fixes

FX elements such as Tabs and Accordions worked in the editor but rendered as plain stacked content in Preview and in exported packages when they were authored inside:

- Form question text (`questionsData[].baseText`)
- Magnifier instructions (`textTextarea`)

Two independent causes:

1. **The libraries were missing.** Library detection only scanned `component.content`, so FX markup living in nested JSON iDevice properties never pulled `exe_effects.js` / `exe_effects.css` into the package.
2. **Nothing ever initialized that markup.** `exe_effects.js` runs `$exeFX.init()` on DOM ready, while the export engine renders JSON iDevices from a `setInterval` (`delayLoadingIdevicesJson: 50`) plus an asynchronous template `fetch()`. Their content therefore always enters the DOM *after* the page-wide initialization has run. For Form there is nothing rendered earlier to preserve: `renderView` leaves `#form-questions-<id>` empty and the questions only exist after `renderBehaviour`.

## Approach

The editor already solves the second problem: `loadLegacyExeFunctionalitiesExport()` (in `ideviceNode.js` and `idevicesEngine.js`) re-runs the shared libraries after an iDevice's export view is loaded. That is why FX inside a Form works while editing and not in the export — **the export runtime had no equivalent**.

This PR adds it rather than teaching each iDevice about `$exeFX`:

- **`exe_export.js`** — new `$exeExport.afterIdeviceRendered(ideviceNode)`, called from the two places where the engine finishes rendering a JSON iDevice (`renderWithTemplate` and the no-template branch). Any future shared enhancement (another kind of TinyMCE-generated content) is registered there, with no iDevice file touched. `form.js` and `magnifier.js` do not mention `$exeFX` at all.
- **`exe_effects.js`** — `init(root)` accepts an optional scope, and the ids each effect generates now come from a monotonic counter instead of the element position. This fixes a latent defect: with effects initialized in more than one pass, an effect injected *before* an already initialized one was handed an id that was already in use (duplicate `exe-tabs-0`), so clicking a tab inside the Form drove the widget of another iDevice.
- **`magnifier.js`** — `renderBehaviour` stops rebuilding the interface `renderView` had already inserted through `magnifier.html`; it only builds it when `.exe-magnifier-container` is missing, which happens when the engine falls back to the bare `{content}` template.
- **Exporters** — library detection now also scans nested JSON iDevice properties, preserving the existing selective MathJax handling. The property traversal and the math exclusion list that `BaseExporter` and `PageRenderer` both need live in a single shared helper (`src/shared/export/utils/jsonPropertyContent.ts`).

The scope argument is kept even though it is called from one place: a page-wide re-init would run through effects initialized elsewhere, and `timeline.init` detaches their handlers and lets `$.timeliner()` re-collapse a timeline the reader had already expanded.

A timer/delay before `$exeFX.init()` was considered and rejected: it races against a `fetch()` whose duration we do not control, it delays the effects of all ordinary page content (flash of raw stacked headings), and it cannot be tested deterministically.

### Form

1. Add a **Form** iDevice.
2. Add a question and insert an FX element, such as **Tabs** or **Accordion**, in the question text.
3. Save and open **Preview**.
4. Export as **HTML5** and open the exported content.
5. Confirm that the FX element is displayed and interactive in both cases.

### Magnifier

1. Add a **Magnifier** iDevice.
2. Insert an FX element in its instructions or text content.
3. Save and open **Preview**.
4. Export as **HTML5** and open the exported content.
5. Confirm that the FX element is displayed and interactive in both cases.

The exported package should include:

- `libs/exe_effects/exe_effects.js`
- `libs/exe_effects/exe_effects.css`

A test project that reproduces the failure is attached in the comments: [form-fx-tabs.zip](https://github.com/user-attachments/files/30227095/form-fx-tabs.zip).

## Tests

- `exe_export.test.js` — the hook runs after `renderBehaviour` in both render paths, and tolerates a missing node or a missing effects library.
- `exe_effects.init.test.js` — scoped initialization, and id uniqueness across passes. That test fails against the previous implementation of this PR (`expected 'exe-tabs-0' to be 'exe-tabs-1'`), so it pins the defect rather than describing it.
- `magnifier.render.test.js` — the three render paths (container present, container missing, editor).
- `jsonPropertyContent.spec.ts`, `BaseExporter.form-fx.spec.ts`, `PageRenderer.form-fx.spec.ts` — nested property detection and the math exclusions.
- `test/e2e/playwright/specs/form-fx-export.spec.ts` — Form and Magnifier end to end: the libraries are bundled and referenced across preview, html5, page, scorm12, scorm2004, ims and elpx, and the tab widget is interactive in the Preview. Both tests fail if the hook is removed and the bundles are rebuilt.

## Validation

- `make fix` clean.
- `make test-unit`: 7691 passing. One unrelated pre-existing failure (`src/redis/client.spec.ts`, a network timeout) reproduces identically on a clean tree.
- `make test-integration`: 721 passing.
- `make test-e2e`: 495 passing, 6 skipped.
- `make test-e2e-static`: 392 passing, 118 skipped.
- Patch coverage: no uncovered statements in the changed ranges.

Fixes #2170
Fixes #2154


erseco:
## Before / after evidence (#2170)

Reproduced with a **Form** iDevice whose question text (`questionsData[].baseText`) contains an FX **Tabs** element (`<div class="exe-fx exe-tabs">…</div>`), rendered through the **Preview** (browser export) pipeline. The only difference between the two captures is the branch — same injected content, same steps.

| Before (`main`) | After (this PR) |
|---|---|
| FX is stripped → plain stacked headings, no tab widget | Interactive One/Two tab widget |
| Before: FX rendered as plain stacked content | After: FX rendered as an interactive tab widget |

With the fix, the Preview/export package includes `libs/exe_effects/exe_effects.js` and `exe_effects.css`, and `$exeFX.init()` runs after the Form renders its dynamic content, so the FX markup becomes an interactive tab widget instead of plain content.

### Test project that reproduces the failure

[form-fx-tabs.zip](https://github.com/user-attachments/files/30227095/form-fx-tabs.zip) — unzip to `form-fx-tabs.elpx`, open it in the editor, then **Preview** or **Export**. On an unfixed build the FX Tabs inside the Form question render as plain stacked content; with this PR they render as an interactive tab widget.

<sub>Captured with the Preview panel on `main` vs this branch, injecting the Form + FX Tabs content used by <code>test/e2e/playwright/specs/form-fx-export.spec.ts</code>.</sub>

erseco:

- **Minor (single source of truth):** the recursive "walk every nested JSON property string" traversal is duplicated as `iteratePropertyStringFragments` (`src/shared/export/exporters/BaseExporter.ts:981`) and `collectPropertyStringValues` (`src/shared/export/renderers/PageRenderer.ts:879`), and the math-exclusion list `['exe_math','exe_math_datagame','exe_math_mathml']` is literal-duplicated at `BaseExporter.ts:1045` and `PageRenderer.ts:51`. Both paths agree today; a future edit to one site (new math lib, key-skipping, depth guard) would silently diverge the two detection paths. AGENTS.md flags duplicated logic explicitly — worth extracting one shared helper + one constant. Non-blocking.
- The other proposed finding (property scanning now over-detecting libraries) is intentional and documented in the PR body, mirrors the existing `component.content` scan, and its only impact is extra bundled JS/CSS — not a defect. Rejected as a blocker.

Verified: diff read at head; duplication and line numbers confirmed by inspection. New code ships colocated tests (BaseExporter/PageRenderer `*.spec.ts`, form/magnifier `*.test.js`, e2e `form-fx-export.spec.ts`). `CI: all checks green on 4ba124f ✅`.

Residual risk: no local build/test/preview was run — runtime rendering of FX inside Form/Magnifier across all export formats is evidenced only by CI and the added e2e spec, not independently reproduced.

ignaciogros (commented):
Thanks, @erseco.

The solution works, and it may be useful for `$exeFX.init` to accept a scope parameter indicating where it should operate. For example: `$exeFX.init($node[0]);` in `public/files/perm/idevices/base/magnifier/export/magnifier.js`.

But for the moment, all calls have been made using `$exeFX.init();`, and changing that could make the application harder to maintain. The issue isn't maintaining the call to that method itself, but rather that if, in the future, we add other common functionality (for example, another type of content generated from TinyMCE), we would have to modify the iDevice's `.js` file to invoke those scripts. It may be better to keep them independent and avoid changing the way `$exeFX.init` is called.

In other words, I'd suggest reverting the current changes in `public/files/perm/idevices/base/magnifier/export/magnifier.js` and `public/app/common/exe_effects/exe_effects.js`, and instead modifying `public/files/perm/idevices/base/magnifier/export/magnifier.js` so that its code doesn't alter content that has already been rendered by the iDevice's JavaScript (`$magnifier.renderView` and `$magnifier.renderBehaviour` are doing that). 

To test that, clear the contents of `$magnifier.renderView` and `$magnifier.renderBehaviour`.

Another option would be to add an additional delay before executing `$exeFX.init();`.

Please let me know what you think. I'm not opposed to merging this solution, but I think it could cause other things to break in the future.

Thanks.

erseco:
Thanks for the careful review, @ignaciogros — your architectural concern is right, and I've reimplemented the fix the way you describe. `form.js` and `magnifier.js` no longer mention `$exeFX` at all.

### Why the root cause isn't (only) overwriting rendered content

The trigger is execution order rather than DOM clobbering:

- `exe_effects.js` runs `$exeFX.init()` on DOM ready (`$(function(){…})`).
- The export engine renders JSON iDevices from a `setInterval` of `delayLoadingIdevicesJson: 50` ms, plus an asynchronous `fetch()` of the iDevice template (`exe_export.js`).

So Form/Magnifier content always enters the DOM *after* `$exeFX.init()` has already run. For **Form** there is nothing previously rendered to preserve: `renderView` leaves `#form-questions-<id>` empty and the questions markup only exists after `renderBehaviour`.

Two consequences for the options you suggested:

1. **Not overwriting already-rendered content** does not fix it on its own — but you were right that magnifier should not be doing it. `renderView` inserts `magnifier.html` (`<div class="exe-magnifier-container">{content}</div>`) and `renderBehaviour` regenerated *the same* HTML and re-inserted it. That's now gone: it only builds the interface when the container is missing, which happens when the engine falls back to the bare `{content}` template because `magnifier.html` could not be loaded.
2. **An additional delay before `$exeFX.init()` is the one option I'd rule out.** It's a race, not a guarantee: the render waits on a `fetch()` whose duration we don't control, so any constant is either too short on a slow machine or wasted time on a fast one, and it would fail exactly on the slow, loaded machines where users hit it. Delaying the page-wide init also means the effects in *all* the ordinary content stop being transformed until the timer fires, so the whole page shows a flash of raw stacked headings. Timing-based fixes also can't be tested deterministically, so nothing would stop it regressing.

### What the fix does instead

The editor already has the mechanism you're asking for: `loadLegacyExeFunctionalitiesExport()` (in `ideviceNode.js` and `idevicesEngine.js`) runs `$exeFX.init()`, `$exeGames.init()`, `$exeHighlighter.init()`, `$exe.init()`… after an iDevice's export view is loaded. That's why FX inside a Form works while editing and not in the export: **the export runtime has no equivalent**.

So this adds it — `$exeExport.afterIdeviceRendered(ideviceNode)`, called from the two places where the engine finishes rendering a JSON iDevice (`renderWithTemplate` and the no-template branch). Any future shared functionality (another kind of TinyMCE-generated content, as you put it) is registered there, in one place, with no iDevice file touched.

### One more thing the review turned up

Deriving the effect index from the element position is unsafe once effects are initialized in more than one pass. Page with a Form (FX inside) placed **above** another FX element: on DOM ready only the second one exists, so it takes `exe-tabs-0`; when the Form renders, its effect comes first in document order and is handed `exe-tabs-0` again. Duplicate ids, and clicking a tab in the Form drives the other widget. `$exeFX.init()` now takes ids from a monotonic counter, which also makes the code shorter than before. No CSS depends on those generated ids.

### Summary of changes

- `exe_export.js`: new `afterIdeviceRendered()`, invoked after `renderBehaviour` + `init`.
- `exe_effects.js`: monotonic counter for generated ids; `init(root)` keeps the optional scope, now called from a single place.
- `form.js`: reverted.
- `magnifier.js`: no longer rebuilds what `renderView` produced.
- `BaseExporter` / `PageRenderer`: the JSON property traversal and the math exclusion list they both used are now a shared helper (`utils/jsonPropertyContent.ts`), as the earlier review comment suggested.

Tests: unit coverage for the hook, for the id uniqueness across passes (it fails against the previous implementation), and for the magnifier render paths; the e2e spec now also covers the magnifier, and both e2e tests fail if the hook is removed.

I kept `init(root)`'s optional scope even though it's now called from one place: a page-wide re-init would run through effects initialized elsewhere, and `timeline.init` detaches their handlers and lets `$.timeliner()` re-collapse a timeline the reader had already expanded. Scoping to the iDevice avoids touching anything else on the page. Happy to drop it if you'd rather keep the call signature untouched.

erseco:
I've also rewritten the PR description: it described the earlier approach (each iDevice calling `$exeFX` itself), which no longer matches the code. It now covers the ordering root cause, the `afterIdeviceRendered()` hook, the id-counter fix, the magnifier change, and why the timer option was rejected.

ignaciogros (approved):
I think that's the best approach.

Thanks, @erseco!

PR #2178, opened 2026-07-14
fix(idevices): tolerate malformed JSON properties

### LINKED ISSUE #2177: Malformed iDevice jsonProperties aborts page loading and hides subsequent blocks
## Description

A user reported that an existing learning object stopped rendering from its second content block onward. The problem is reproducible in the cloud editor on a page containing a True/False iDevice with malformed persisted `jsonProperties`.

The affected project URL and content identifiers are intentionally omitted because they belong to a user project.

## Observed behavior

Selecting the affected page raises an unhandled exception:

```text
Uncaught (in promise) SyntaxError: Expected ',' or '}' after property value in JSON
    at JSON.parse
    at IdeviceNode.setParams
    at new IdeviceNode
    at IdevicesEngine.newIdeviceNode
    at IdevicesEngine.loadComponentsPage
```

The exception aborts the sequential component-loading loop. The malformed iDevice and every later block are therefore absent from the workarea.

The malformed payload was inside an embedded audio fragment. Its relevant shape was:

```text
<audio controls="controls" src=""><a href="">audio.webm</a></audio>
```

Because this HTML was embedded in a JSON string with mismatched escaping, the complete `jsonProperties` value was invalid.

There were also secondary 404 requests for quoted `data:image/svg+xml` and `blob:` URLs. Those requests are noisy, but the unhandled `JSON.parse` exception is what stops the page from loading.

## Steps to reproduce

1. Create a project with one page and two content blocks.
2. Add an iDevice to the first block whose persisted `jsonProperties` contains malformed JSON.
3. Give that iDevice a previously generated `htmlView` / HTML representation.
4. Add a valid iDevice to the second block.
5. Navigate away from the page and back to force component reconstruction.

## Actual result

`IdeviceNode.setParams()` throws while constructing the malformed component. Page loading stops, so later components and blocks are hidden.

## Expected result

A malformed iDevice payload must be isolated. The editor should log a diagnostic, use an empty properties object as a safe fallback, preserve any previously rendered HTML representation, and continue loading all later blocks.

## Scope

This issue covers resilience during initial loading and collaborative updates. It does not claim to repair the already-corrupted activity data automatically; the source of the persisted malformed media markup should be investigated separately if it can be reproduced from a clean project.

## Reported by

Reported by an eXeLearning cloud user and reproduced with Playwright on July 14, 2026.

erseco:
Fixes #2177

## Problem
Malformed persisted iDevice `jsonProperties` aborts page loading and hides all following blocks. Editing a fallback state could also overwrite the original data.

## Fix
Isolate malformed properties so retained HTML and later blocks render, preserve the original payload, and block editing or saving until valid data is available.

Regression coverage includes unit tests and an isolated Playwright scenario.


ignaciogros (commented):
Thanks, @erseco.

If the user who reported the issue provided a sample `.elp` or `.elpx` file to reproduce the problem, could you please attach it?

I tried saving an `.elpx` file and then modifying its `content.xml` to leave malformed JSON in it, but the result was not what I expected: the application hangs, and the browser console reports the following error:

> Uncaught (in promise) TypeError: can't access property "map", data.questionsData is undefined

erseco:
Thanks for digging into this, @ignaciogros — your test found a real bug, and it is one this PR did not fix. It is fixed now, along with the repro file you asked for.

## About the user's file

I cannot attach it, for two reasons. It is a 118 MB real user project (which is why #2177 omits the URL and the content identifiers), and more importantly **it no longer contains any malformed JSON** — all 53 `jsonProperties` blocks in its `content.xml` parse cleanly. It is the aftermath of the bug, not a reproduction of it, so it would not have helped you.

Attached instead is a 2 KB synthetic repro with no user content, also committed here as **`test/fixtures/damaged-trueorfalse-json.elpx`** (zipped, since GitHub does not accept `.elpx` attachments). One page, three blocks:

[damaged-trueorfalse-json.zip](https://github.com/user-attachments/files/30085120/damaged-trueorfalse-json.zip)


| block | iDevice | `jsonProperties` | expected |
|---|---|---|---|
| 1 | `trueorfalse` | malformed | keeps its rendered HTML, cannot be edited |
| 2 | `trueorfalse` | empty | keeps its rendered HTML |
| 3 | `text` | valid | **must still load** |

Block 1 raises the exact error from #2177:

```
SyntaxError: Expected ',' or '}' after property value in JSON
```

## Your error was real: the {} fallback only moved the crash

`public/files/perm/idevices/base/trueorfalse/export/trueorfalse.js:194`

```js
if (typeof data.typeGame === 'undefined') {   // legacy migration branch
    ...
    data.questionsGame = data.questionsData.map((q) => ({ ... }));   // no guard
```

When this PR falls back to `{}`, `data.typeGame` is `undefined`, so the legacy branch runs, `data.questionsData` is `undefined`, and you get your `TypeError`. The PR turned a `SyntaxError` at load time into a `TypeError` at render time — still a broken page. The edition twin at `trueorfalse/edition/trueorfalse.js:52` already guards this with `Array.isArray(data.questionsData) ? data.questionsData : []`; only the export side was missing it.

There is a **second, independent** crash a few lines down at `:210`, `data.numberQuestions = data.questionsGame.length`, reachable with `{"typeGame":"TrueOrFalse"}` and no questions — the shared `getQuestions` helper returns non-array input unchanged, so it offers no protection there. Fixing `:194` alone does not cover it. An audit of the other iDevices found the same defect in `casestudy/export/casestudy.js:92` (`data.activities.map(...)`), with the same export-vs-edition asymmetry. All three are fixed in e498761.

## Why the E2E in this PR did not catch it

`malformed-idevice-json.spec.ts` created the broken component as a **`text`** iDevice. `text` has no legacy migration branch and renders its retained HTML whatever the properties hold — it is the one type that cannot expose this. The test was passing for the wrong reason.

It now uses a `trueorfalse`, imports the fixture above, and asserts `pageErrors` is empty. I checked both directions: reverting each guard makes the new unit tests and both E2E scenarios fail, and restoring it makes them pass.

## One thing worth flagging separately

While checking the user's file I found that **all 17 of its interactive iDevices have lost their configuration** — `trueorfalse`, `classify`, `sort`, `relate`, `flipcards`, `map`, `puzzle`, `select-media-files` are all empty, while all 28 `text` iDevices are intact. In the exported HTML inside that same `.elpx`, every `data-idevice-json-data` is 87–92 characters (`{"ideviceId":"..."}`) against 2.7–6.5 KB in our fixtures, so the activities are dead in the export too.

I found a reachable mechanism that produces exactly that end state. `ElpxImporter.ts:1520` replaces an unparseable payload with `{}` and keeps only a console warning:

```
[ElpxImporter] Invalid JSON for idevice-damaged-malformed, using empty object
```

The original string is not kept anywhere, so once the project is saved the activity's data is gone. Importing well-formed files is unaffected — I checked `todos-los-idevices` and all 36 game iDevices came through intact with no warnings — so this only bites projects that were already damaged. Note the inconsistency with this PR, which on the editor path deliberately keeps `malformedJsonPropertiesRaw` and blocks editing precisely so the payload cannot be overwritten; the importer does the opposite with the same data.

What made those 17 payloads malformed in the first place is still unknown — the pre-damage file no longer exists. Worth noting for #2179 as well: it validates that the payload is *parseable*, but `{"ideviceId":"..."}` and `{}` are valid JSON and pass validation, and here the empty case returns `error: null`, so editing is not blocked either. I would rather track that as its own issue than widen this PR. Happy to open it if you agree.

erseco:

PR #2179, opened 2026-07-14
fix(idevices): validate JSON before Yjs writes

erseco:
Related to #2177 and #2178.

## Motivation
A malformed iDevice `jsonProperties` payload can be persisted and later prevent a page from loading. #2178 recovers malformed data during load; this PR prevents new invalid payloads from reaching Yjs.

## Changes
- serialize and validate `jsonProperties` before component create/update transactions
- validate again after asset URL normalization
- reject failed saves and preserve the last valid component state

## Tests
Covers HTML quotes, line breaks, Unicode, truncated JSON, circular and undefined values, invalid asset transformations, atomic writes, and save/reload persistence.

Patch coverage: 100%.


erseco:
### Review + pipeline fix

**Pipeline fixed.** The only failing check was `e2e (static, 1)` — this PR's new `idevice-json-save-validation.spec.ts` asserted a preserved component survives `page.reload()`, but in the **static** build the app runs offline (`saveToServer()` is a no-op and a reload boots a fresh project), so `getComponent()` returned `null` after reload. Same reason `open-project-clean-state.spec.ts` skips static ("persistence is disabled in static mode").

Fix (commit `e240869`): split the spec — malformed-JSON rejection + last-valid preservation now run in **all** projects (incl. static); the reload-persistence check is server-mode only via `test.skip(project === 'static', …)`. Verified locally:
- `--project=static` → 1 passed, 1 skipped
- `--project=chromium` → 2 passed

**Easy suggestions applied** (commit `5ed27bd`):
- `serializeAndValidateJsonProperties` now uses `new Error(msg, { cause })` instead of assigning `.cause` after construction.
- Documented that `YjsProjectBridge.addComponent`/`updateComponent` can throw `InvalidJsonPropertiesError` (those wrappers aren't wrapped in try/catch by callers).

**Heads-up (semantic change worth a note in the PR body):** passing `jsonProperties: undefined` now throws (the old create path set it explicitly to `undefined`). This is intended and tested.

**Follow-up filed:** #2181 — de-duplicate the triple blob/asset preparation (`convertJsonProperties` → `prepareJsonPropertiesForSync` in apiCallManager → again inside create/updateComponent) down to a single pass. Safe today only because `prepareJsonForSync` is idempotent.

Overall LGTM: validation runs before the Yjs transaction so writes stay atomic, the server path now rejects bad saves instead of silently persisting, and coverage is thorough.

ignaciogros (commented):
Thanks, @erseco.

If the user who reported the issue provided a sample `.elp` or `.elpx` file to reproduce the problem, could you please attach it?

I tried saving an `.elpx` file and then modifying its `content.xml` to leave malformed JSON in it, but the result was not what I expected: the application hangs, and the browser console reports the following error:

> Uncaught (in promise) TypeError: can't access property "map", data.questionsData is undefined

ignaciogros:
Please see https://github.com/exelearning/exelearning/pull/2178#pullrequestreview-4711937148

erseco:

PR #2188, opened 2026-07-16
fix(assets): stop leaking blob URLs into detached anchor href reads (#2186)

### LINKED ISSUE #2186: Game iDevices read ephemeral blob: URLs instead of asset:// when loading editor data from *-Link* anchors
## Summary

Game iDevices carry their media references as hidden anchors in the stored HTML, e.g.:

```html
<a href="asset://<uuid>" class="js-hidden clasifica-LinkImages">0</a>
```

When the editor loads that stored HTML, it reads those references back with `.attr('href')`. The value it gets is **not** the `asset://` URL that is actually in the DOM — it is an **ephemeral `blob:` URL**. In the desktop (Electron) build this surfaces to the user as `blob:app://...`, because `app://localhost` is the Electron origin (`app/main.js:851`).

Surfaced by @ignaciogros while testing an eXeLearning 2.9 manual export in https://github.com/exelearning/exelearning/pull/2166. It is **not** related to that PR: it reproduces on `main` and the code path below is untouched by it.

## Root cause

`public/app/common/asset_url_resolver.js:657-662` monkey-patches `Element.prototype.getAttribute`:

```js
if (name === 'href' && this.tagName === 'A') {
    // If href is still asset://, try to return the resolved blob:// from cache
    // This helps SimpleLightbox and similar libraries get the real URL
    if (value && value.startsWith('asset://') && resolvedCache.has(value)) {
        return resolvedCache.get(value);
    }
```

So any code reading `href` from an anchor whose stored value is `asset://` gets the blob URL instead, as long as that asset is in `resolvedCache` (populated when the iDevice was rendered in the page view before editing).

`public/files/perm/idevices/base/classify/edition/classify.js:693` does exactly that:

```js
dataGame.wordsGame[iq].url = $(this).attr('href');   // jQuery .attr() -> getAttribute -> blob:
```

The patch lives on `Element.prototype`, so it also applies to the **detached** wrapper `loadPreviousValues` builds (`classify.js:681`). The DOM literally holds `asset://<uuid>`; the code reads `blob:app://...`.

Note the comment directly above the patch (`asset_url_resolver.js:630-631`) states the interception exists so that *"code reading these attributes for saving gets the persistent asset:// URL, not the ephemeral blob:// URL"*. The anchor branch does the opposite of that stated intent.

## Why it matters

The blob URL is then written back on save: `classify.js:811` re-emits `<a href="${word.url}">`, and `classify.js:762-763` re-encrypts it into the DataGame JSON. It survives the save-time cleanup because:

- `convertBlobURLsToAssetRefs` strategies 1/2 (`AssetManager.js:3147`) only match `` — **never `<a href=...>`**.
- Strategy 3 (`AssetManager.js:3183-3192`) *is* scheme-agnostic and does recover blobs present in `reverseBlobCache`, so this does not fail on every save.
- The "cannot recover -> clear" fallback (`AssetManager.js:3194`) matches only `blob:https?://`, so an unrecoverable `blob:app://` is **persisted rather than cleared**, leaving a dead reference after reload. In the web build the same fallback instead clears the href to `""`.

## Affected iDevices

`classify` is where it was observed, but the read-`href`-from-`*-Link*`-anchor pattern is shared by ~19 iDevices under `public/files/perm/idevices/base/`, including `identify`, `discover`, `sort`, `az-quiz-game`, `crossword`, `dragdrop`, `map`, `guess`, `relate`, `hidden-image`, `puzzle`, `flipcards`, `select-media-files`, `quick-questions`, `word-search`, `beforeafter`, `trivial`.

## Secondary: `blob:app://` is invisible to two regexes

Both assume a `blob:https?://` shape and therefore never match in Electron (verified by running them):

- `AssetManager.js:3277` — `prepareJsonForSync` `blobUrlPattern`
- `blobPasteGuard.js:63` — carrying the incorrect comment `// Match blob: URLs (always blob:http:// or blob:https://)`

## Not the cause: legacy import

Ruled out. The legacy rewriter is generic with no per-iDevice list: `convertContextPathToAssetRefs` runs unconditionally on `htmlView` (`ElpxImporter.ts:928`) and `properties` (`ElpxImporter.ts:977`), `clasifica-activity` -> `classify` is mapped (`LegacyXmlParser.ts:1324`), and classify already inherits the #2156 fix via `BaseLegacyHandler.extractTextAreaFieldContent`. After import the stored `htmlView` correctly holds `asset://`.

## Resolved: the validation error is a *separate* bug

**Update — reproduced with the source file, so the open question below is now answered.**

The reported *"You must indicate an image, a text or/and an audio for each card"* is **not** caused by the blob substitution described above, and my earlier hypothesis in this issue was **wrong**: legacy 2.9 exports *do* carry `clasifica-LinkImages` anchors under exactly that class name, with `resources/...` hrefs, and the importer resolves them correctly. The import is faithful.

The real cause is a bad card in the manual plus a validator that ignores `audio`. The activity has 8 cards but only 6 `clasifica-LinkImages` anchors; decrypting the `clasifica-DataGame` payload (`unescape` + XOR 146) shows card 7 (the cow) is declared `type: 0` (image) with an empty `url` and only `resources/COW2.mp3`. With no anchor to override it, `url` stays `""` and `validateData` (`classify.js:941-948`) rejects it via `type === 0 && url.length < 4`.

`validateData` never consults `mquestion.audio` — zero references to that field in the block — even though the message it prints promises *"an image, a text **or/and an audio**"*. So an audio-only card is rejected by a validator that claims to accept it. eXe 2.9 accepted that card; 4.x does not.

That is a distinct defect from the `getAttribute` patch this issue is about, and it deserves its own decision (bring `audio` into the condition, or normalise a `type: 0` card with no `url` but with an `audio` into an audio card on load). Tracked for now in https://github.com/exelearning/exelearning/pull/2166#issuecomment-4989832699.

## Suggested direction

Fix the reader rather than the generic rewriter (the pattern established by #2156). Either exempt detached / `js-hidden` game link anchors from the `getAttribute` patch, or have the game iDevices resolve back via `assetResolver.getAssetUrlFromBlob()` (`asset_url_resolver.js:713`), as `image-gallery` already does. Also widen the two `blob:https?://` regexes to be scheme-agnostic.

erseco:
## Problem

`asset_url_resolver.js` patches `Element.prototype.getAttribute` so that code reading `src`/`href` **for saving** gets the persistent `asset://` URL instead of the ephemeral `blob://` one — that is the stated intent in the comment above the patch.

The anchor branch did the exact opposite:

```js
if (name === 'href' && this.tagName === 'A') {
    if (value && value.startsWith('asset://') && resolvedCache.has(value)) {
        return resolvedCache.get(value);   // <- hands out the ephemeral blob URL
    }
```

Game iDevices store their media references as hidden anchors:

```html
<a href="asset://<uuid>" class="js-hidden clasifica-LinkImages">0</a>
```

and read them back by parsing their **stored** HTML (`htmlView`, straight from Yjs) into a **detached** wrapper — `classify.js:681` builds `$('<div></div>').html(originalHTML)` and `classify.js:693` does `$(this).attr('href')`. jQuery `.attr()` calls `getAttribute`, the patch sees the asset is in `resolvedCache` (populated when the iDevice was rendered in page view before editing), and returns `blob:app://...`.

The editor therefore reads a blob URL although the DOM literally holds `asset://<uuid>`, and re-persists it on save (`classify.js:811` re-emits `<a href="${word.url}">`). The blob dies with the page, leaving a dead reference after reload. The read-back pattern is shared by ~19 iDevices (identify, discover, sort, az-quiz-game, crossword, dragdrop, map, guess, relate, hidden-image, puzzle, flipcards, select-media-files, quick-questions, word-search, beforeafter, trivial).

## Fix

Gate the cache lookup on `Element.isConnected`.

The discriminator matches why the branch exists in the first place: SimpleLightbox and friends operate on anchors **in the rendered page** and need a loadable URL in the window before the MutationObserver has rewritten the href. Detached anchors are never rendered — they are throwaway wrappers used to parse stored HTML back into form data, and must see the persistent value.

This does not regress persistence for live anchors. Verified against the MutationObserver (`asset_url_resolver.js:578-604`): for anchors in the document it calls `trackElementAsset()` (storing `data-asset-url` = `asset://...`) **and** rewrites the href to the blob. So a live anchor read for saving hits branch 2 (`data-asset-url` + `blob:` href) and already returns `asset://`. Branch 2 stays ungated, so both detached and connected anchors keep mapping blob back to `asset://`.

No per-iDevice special casing, and no class-name matching — the fix is at the single point where the wrong value is produced.

## How to reproduce

1. Download the eXeLearning 2.9 manual: https://descargas.intef.es/cedec/exe_learning/Manuales/manual_exe29/referencias_y_fuentes.html
2. Import the `.elp` into eXeLearning.
3. Navigate to the **"Classify"** section (`Clasifica`) and click **Edit** on the activity.
4. Inspect the image URLs loaded into the form fields.

**Before:** the URLs read `blob:app://.../<uuid>` in the desktop/Electron build (`blob:http://localhost:8080/...` in the web build) — an ephemeral reference. Saving writes that blob back into the stored HTML, so after a reload the images are gone.

**After:** the URLs read `asset://<uuid>/<filename>` — the persistent reference actually stored in the document. Saving round-trips the same value and the images survive a reload.

Also confirm the lightbox is unaffected: open any iDevice with a linked image in preview and click it — it still opens, because live-DOM anchors keep receiving the resolved blob URL.

## Tests

`public/app/common/asset_url_resolver.test.js`, colocated (Vitest). Added a `describe` block covering the regression, all of which fail on `main`:

- a **detached** `<a href="asset://uuid">` whose asset is in `resolvedCache` returns `asset://uuid`
- a **connected** anchor still returns the cached blob URL (lightbox behaviour preserved)
- a detached anchor whose asset is not cached returns `asset://`
- blob href + `data-asset-url` still maps back to `asset://`, detached **and** connected
- a full classify-style read-back of `clasifica-LinkImages` / `clasifica-LinkAudios` anchors yields no `blob:` URL

The existing test named *"populates cache so getAttribute can return blob URL for asset:// href (SimpleLightbox support)"* asserted the buggy behaviour: it used a **detached** anchor, which is not how SimpleLightbox works. It now appends the anchor to the document, matching the real scenario the branch is there to serve.

While adding these I found the test file never restored `Element.prototype.getAttribute`. Each `vi.resetModules()` + re-import stacked another patch layer with its own `resolvedCache`, and the layers fed each other's return value back through the `asset://` <-> `blob://` mapping, producing values no single layer would return. The `getAttribute interception` block had no import at all and silently relied on a patch **leaked** from earlier blocks. The native method is now restored between tests and that block imports the module itself.

```
npx vitest run public/app/common/asset_url_resolver.test.js
 Test Files  1 passed (1)
      Tests  130 passed (130)

make test-frontend
 Test Files  247 passed (247)
      Tests  13724 passed (13724)
```

`make fix` is clean. Coverage on `asset_url_resolver.js`: 97.58% lines / 84.95% branches; both outcomes of the changed condition are exercised directly.

## Out of scope

The two secondary items in the issue are left for separate PRs, as they are independent defects with their own blast radius:

- the `blob:https?://` regexes that never match `blob:app://` (`AssetManager.js:3277`, `blobPasteGuard.js:63`)
- `validateData` ignoring `mquestion.audio` (the *"You must indicate an image, a text or/and an audio"* error), tracked in #2166

Closes #2186

### DIFF
--- a/public/app/common/asset_url_resolver.js
+++ b/public/app/common/asset_url_resolver.js
@@ -655,9 +655,18 @@
 
         // For anchor elements reading 'href'
         if (name === 'href' && this.tagName === 'A') {
-            // If href is still asset://, try to return the resolved blob:// from cache
-            // This helps SimpleLightbox and similar libraries get the real URL
-            if (value && value.startsWith('asset://') && resolvedCache.has(value)) {
+            // If href is still asset://, return the resolved blob:// from cache, but
+            // only for anchors that live in the document. SimpleLightbox and similar
+            // libraries operate on the rendered page and need a loadable URL before
+            // the MutationObserver has rewritten the href.
+            //
+            // Detached anchors are never rendered: they belong to the throwaway
+            // wrappers iDevices build to parse their own stored HTML back into form
+            // data (e.g. classify's `$('<div></div>').html(previousData)`). Handing
+            // them an ephemeral blob: URL makes them persist a reference that dies
+            // with the page, wiping the media on reload (issue #2186). Those reads
+            // must see the persistent asset:// URL that is literally in the DOM.
+            if (this.isConnected && value && value.startsWith('asset://') && resolvedCache.has(value)) {
                 return resolvedCache.get(value);
             }
             // For persistence: return the original asset:// URL if we have one stored
--- a/public/app/common/asset_url_resolver.test.js
+++ b/public/app/common/asset_url_resolver.test.js
@@ -8,12 +8,21 @@
 
 import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
 
+// The resolver patches Element.prototype.getAttribute on import. Every re-import
+// (vi.resetModules() + await import) installs another layer on top of the previous
+// one, and each layer owns a separate resolvedCache. Stacked layers feed each
+// other's return value back through the asset://<->blob:// mapping, which produces
+// results no single layer would ever return. Capture the pristine method once and
+// restore it between tests so each test exercises exactly one layer.
+const nativeGetAttribute = Element.prototype.getAttribute;
+
 describe('AssetUrlResolver', () => {
   let mockAssetManager;
   let mockAssetWebSocketHandler;
 
   beforeEach(() => {
     vi.clearAllMocks();
+    Element.prototype.getAttribute = nativeGetAttribute;
 
     // Mock console methods
     vi.spyOn(console, 'log').mockImplementation(() => {});
@@ -33,6 +42,7 @@ describe('AssetUrlResolver', () => {
     delete window.eXeLearningAssetResolver;
     delete window.jQuery;
     delete window.eXeLearning;
+    Element.prototype.getAttribute = nativeGetAttribute;
     vi.restoreAllMocks();
   });
 
@@ -253,8 +263,11 @@ describe('AssetUrlResolver', () => {
       });
 
       it('populates cache so getAttribute can return blob URL for asset:// href (SimpleLightbox support)', async () => {
-        // This test verifies the behavior for SimpleLightbox and similar libraries
-        // When they read the href of an anchor, they should get the resolved blob:// URL
+        // This test verifies the behavior for SimpleLightbox and similar libraries.
+        // When they read the href of an anchor *in the live document*, they should
+        // get the resolved blob:// URL. The anchor must be connected: detached
+        // anchors are parsing wrappers reading back stored HTML and must keep the
+        // persistent asset:// URL instead (see issue #2186).
         const assetUrl = 'asset://cached-uuid-test/lightbox-image.jpg';
         const blobUrl = 'blob:http://localhost/cached-blob-test';
         mockAssetManager.resolveAssetURL.mockResolvedValue(blobUrl);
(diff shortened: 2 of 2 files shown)


erseco:

PR #2189, opened 2026-07-16
fix(idevice/classify): accept audio-only cards on save (#2187)

### LINKED ISSUE #2187: Classify: audio-only cards are rejected on save — both validators ignore the audio field
## Summary

Opening the **Classify** iDevice from the [eXeLearning 2.9 manual](https://descargas.intef.es/cedec/exe_learning/Manuales/manual_exe29/referencias_y_fuentes.html) and saving it **without any change** fails with:

The card it rejects has an audio. Three different messages in this iDevice promise that an audio counts, and **no validator ever reads the `audio` field**.

Reported by @ignaciogros on #2166 while testing that PR. It is unrelated to it and reproduces on `main`. Separate from #2186, which covers the `blob:app://` URLs seen in the same test.

## The player supports audio-only cards by design

This is the part that settles it. `public/files/perm/idevices/base/classify/export/classify.js:1491-1496`:

```js
if (audio.length > 0) {
    if (url.trim().length === 0 && text.trim().length === 0) {
        $audio.addClass('CQP-LinkAudioBig');
    }
    $audio.show();
}
```

That branch exists **specifically** for a card with no image and no text. It is not dead code: `CQP-LinkAudioBig` has its own styling (`export/classify.css:520`) and event wiring (`export/classify.js:874, 878, 1005`). The player renders an audio-only card as a big audio button, deliberately.

So the editor refuses to save content the player is built to display.

## Root cause: `audio` is never consulted, in either validator

**Per-card** (`edition/classify.js:852-861`):

```js
if (p.type === 0 && p.url.length < 5) {
    message = msgs.msgCompleteImage;
} else if (p.type === 1 && p.eText.trim().length === 0) {
    message = msgs.msgCompleteText;
} else if (p.type === 2 && (p.eText.trim().length === 0 || p.url.length < 5)) {
    message = msgs.msgCompleteBoth;
}
```

**Whole-form** (`edition/classify.js:941-948`):

```js
if (
    (mquestion.type === 0 && mquestion.url.length < 4) ||
    (mquestion.type === 1 && mquestion.eText.trim().length === 0) ||
    (mquestion.type === 2 && (mquestion.url.length < 4 || mquestion.eText.trim().length === 0))
) {
    $exeDevice.showMessage($exeDevice.msgs.msgCompleteData);
    return false;
}
```

Neither reads `audio`. Yet the messages they print state the intended rule precisely:

| message | text | intended rule |
|---|---|---|
| `msgCompleteImage` (type 0) | "You must link an image **or sound** to this card" | image **or** audio |
| `msgCompleteText` (type 1) | "You must indicate a text **or sound** for this card" | text **or** audio |
| `msgCompleteBoth` (type 2) | "You must select an image **and** text for this card" | image **and** text |
| `msgCompleteData` (form) | "You must indicate an image, a text **or/and an audio** for each card" | any of the three |

The intent was documented all along; the code just never implemented it. The two validators also duplicate the same rule with **different thresholds** (`< 5` vs `< 4`), which AGENTS.md explicitly calls out as a bug waiting to happen.

Note there is no "audio" card type — only Image (0) / Text (1) / Both (2) (`edition/classify.js:395-404`). Audio is an orthogonal field on every card, which is exactly why the type must not gate it.

## Reproduction (verified against the manual's own file)

**Card 7 (the cow) is the trigger**: declared `type: 0` (image) with no image, only `COW2.mp3`. It gets no image anchor, so `url` stays `""` and `type === 0 && url.length < 4` rejects it. Card 6 is the same idea expressed as `type: 1` with the text "Cat", which is why it passes.

Steps:

1. Import the 2.9 manual `.elp`.
2. Go to the "Classify" section and edit the activity.
3. Press save without changing anything.
4. Observe the error, even though every card has content.

## The import is faithful — not an import bug

Worth stating, since it was the first suspicion and it was wrong: the anchors are present in the source with the expected `clasifica-LinkImages` / `clasifica-LinkAudios` class names and `resources/...` hrefs, and the legacy rewriter resolves them to `asset://`. Card 7 is `type: 0` with an empty `url` **in the original 2.9 file**. eXe 2.9 accepted that state; 4.x does not.

## Impact

Legacy content with audio-only cards cannot be re-saved at all without the user mutilating the card, and the error does not say which card is at fault.

## Fix

Implement the rule the messages already state, in **one** shared predicate used by both validators (single source of truth), with a single length threshold:

- type 0 (Image): `url` **or** `audio`
- type 1 (Text): `eText` **or** `audio`
- type 2 (Both): `url` **and** `eText`

erseco:
Closes #2187. Found by @ignaciogros while testing #2166 — thanks. Unrelated to that PR; this reproduces on `main`.

## The bug

The Classify activity of the [eXeLearning 2.9 manual](https://descargas.intef.es/cedec/exe_learning/Manuales/manual_exe29/referencias_y_fuentes.html) cannot be saved **at all**, even without touching anything:

The card it rejects **has** an audio. Neither validator ever read the `audio` field, although three of the messages they print say an audio counts:

| message | text | rule it states |
|---|---|---|
| `msgCompleteImage` (type 0) | "You must link an image **or sound** to this card" | image **or** audio |
| `msgCompleteText` (type 1) | "You must indicate a text **or sound** for this card" | text **or** audio |
| `msgCompleteBoth` (type 2) | "You must select an image **and** text for this card" | image **and** text |

And the player is **built for** audio-only cards — `export/classify.js:1491-1496` gives a card with no image and no text a big audio button (`CQP-LinkAudioBig`), which has its own styling (`export/classify.css:520`) and event wiring. The editor was refusing content the player renders by design.

There is no "audio" card type (only Image / Text / Both), which is exactly why the type must not gate the audio.

## The fix

One shared predicate, `isQuestionComplete`, used by **both** the per-card and the whole-form validation — removing duplicated logic that had drifted to two different thresholds (`< 5` vs `< 4`):

- type 0 (Image): `url` **or** `audio`
- type 1 (Text): `eText` **or** `audio`
- type 2 (Both): `url` **and** `eText`

The whole-form validation now also brings the offending card up before showing the message, so it points at something instead of failing the form anonymously. This needs no new translatable string.

## How to reproduce

1. Import the 2.9 manual `.elp`.
2. Go to the **"Classify"** section and edit the activity.
3. Press save **without changing anything**.

**Before:** rejected with the message above.
**After:** saves cleanly, and the cow card (audio only) is preserved.

To see the offending card: it is **card 7 of 8**. The activity has 8 cards but only 6 `clasifica-LinkImages` anchors; decrypting the `clasifica-DataGame` payload (`unescape` + XOR 146) shows card 7 is `type: 0` (image) with an empty `url` and only `resources/COW2.mp3`. Card 6 is the same idea done as `type: 1` with the text "Cat", which is why it always passed.

This is data from the original 2.9 file — eXe 2.9 accepted that card, 4.x did not. The legacy import is faithful and not at fault.

## Tests

`public/files/perm/idevices/base/classify/edition/classify.test.js` — 16 new cases covering the three types plus edge cases (missing question, non-string fields, sub-threshold values, whitespace-only text), including a regression test for the manual's cow card and one asserting an audio does **not** rescue a "Both" card missing its half.

Verified the tests actually catch the bug: reverting the predicate to the pre-fix semantics fails exactly the two audio-only cases.

```
npx vitest run public/files/perm/idevices/base/classify/edition/classify.test.js
 Test Files  1 passed (1)
      Tests  55 passed (55)

npx vitest run           # full frontend suite
 Test Files  247 passed (247)
      Tests  13734 passed (13734)
```

I also replayed the real payload from the manual through the patched predicate: all 8 cards now pass, card 7 included.

Note: `focusedEditMode.test.js` failed once in a full parallel run and passed in isolation and on re-run. It is unrelated to this change (which only touches `classify`) and looks like pre-existing flakiness — flagging it rather than hiding it.

## Out of scope

- #2186 — the `blob:app://` URLs seen in the same test (fixed in #2188).
- The `useParseIntRadix` lint infos in this file are pre-existing and untouched. Note `make fix` runs `biome lint` over `public/app/` only, so `public/files/perm/idevices/**` is deliberately not formatted; I kept the file's existing style rather than reformatting it.

### DIFF
--- a/public/files/perm/idevices/base/classify/edition/classify.js
+++ b/public/files/perm/idevices/base/classify/edition/classify.js
@@ -36,6 +36,9 @@ var $exeDevice = {
     playerAudio: '',
     isVideoType: false,
     numberCards: 3,
+    // Shortest string accepted as a media URL. Empty media is stored as '', so
+    // this only guards against blank or truncated values.
+    minMediaUrlLength: 4,
     version: 0.8,
     id: false,
     checkAltImage: true,
@@ -849,15 +852,14 @@ var $exeDevice = {
         p.msgHit = $('#clasificaEMessageOK').val();
         p.msgError = $('#clasificaEMessageKO').val();
 
-        if (p.type === 0 && p.url.length < 5) {
-            message = msgs.msgCompleteImage;
-        } else if (p.type === 1 && p.eText.trim().length === 0) {
-            message = msgs.msgCompleteText;
-        } else if (
-            p.type === 2 &&
-            (p.eText.trim().length === 0 || p.url.length < 5)
-        ) {
-            message = msgs.msgCompleteBoth;
+        if (!$exeDevice.isQuestionComplete(p)) {
+            if (p.type === 1) {
+                message = msgs.msgCompleteText;
+            } else if (p.type === 2) {
+                message = msgs.msgCompleteBoth;
+            } else {
+                message = msgs.msgCompleteImage;
+            }
         }
 
         p = $exeDevice.normalizeQuestionByType(p);
@@ -938,13 +940,10 @@ var $exeDevice = {
                 $exeDevice.wordsGame[i]
             );
             $exeDevice.wordsGame[i] = mquestion;
-            if (
-                (mquestion.type === 0 && mquestion.url.length < 4) ||
-                (mquestion.type === 1 && mquestion.eText.trim().length === 0) ||
-                (mquestion.type === 2 &&
-                    (mquestion.url.length < 4 ||
-                        mquestion.eText.trim().length === 0))
-            ) {
+            if (!$exeDevice.isQuestionComplete(mquestion)) {
+                // Bring the offending card up so the message points at something.
+                $exeDevice.active = i;
+                $exeDevice.showQuestion(i);
                 $exeDevice.showMessage($exeDevice.msgs.msgCompleteData);
                 return false;
             }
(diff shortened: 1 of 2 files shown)


erseco:

Issue #2190, opened 2026-07-16
Importing a project with damaged iDevice properties silently discards the activity data


erseco:
## Description

When a project carries an iDevice whose persisted `jsonProperties` cannot be parsed, importing it replaces that payload with an empty object and keeps no copy of the original. The activity's configuration is gone, and once the project is saved the loss is permanent.

The retained `htmlView` still renders, so the page looks intact while the activity underneath is dead. That makes the loss easy to miss until someone opens the activity or exports the project.

## Where

`src/shared/import/ElpxImporter.ts:1520`

```ts
if (!parsed) {
    this.logger.warn(`[ElpxImporter] Invalid JSON for ${componentId}, using empty object`);
    props = {};
}
```

The original string is not stored anywhere, and the only trace is a `console.warn`. There is no user-facing notice naming the affected activities.

## Steps to reproduce

1. Import `test/fixtures/damaged-trueorfalse-json.elpx` (added in #2178).
2. Observe the console:

```
[ElpxImporter] Invalid JSON for idevice-damaged-malformed, using empty object
```

3. Read the stored properties for `idevice-damaged-malformed`.

## Actual result

The stored `jsonProperties` is `{}`. The original payload is unrecoverable.

## Expected result

A payload that cannot be parsed must be preserved, not replaced. The editor already has the machinery for this: #2178 keeps `malformedJsonPropertiesRaw` and blocks editing so the value cannot be overwritten. Import should reach the same end state instead of discarding, and the user should be told which activities are affected rather than only the console.

## Why this is feasible

The importer writes through `compMap.set('jsonProperties', ...)` directly (`ElpxImporter.ts:1767`), so it bypasses the write validation added in #2179. It can therefore carry the raw string into the Y.Doc, after which the existing #2178 logic detects the parse error, blocks editing, and preserves the payload. The E2E in #2178 already covers that half once the raw value is present.

## Scope

Well-formed projects are unaffected. Verified by importing `todos-los-idevices_dos_informes.elpx`, `todos-los-idevices.elp` and `old_el_cid.elp`: all game iDevices came through with their properties intact and zero warnings. This only affects projects that were already damaged.

## Evidence from a reported user project

A user project exported from build `0.0.0-alpha-build202607100302` shows exactly this end state. All 17 of its interactive iDevices lost their configuration, while all 28 `text` iDevices are intact:

| iDevice type | repo fixtures | user project |
|---|---|---|
[...]

PR #2192, opened 2026-07-16
fix(idevices): stop getQuestions handing back non-array data

erseco:
Follow-up hardening split out of #2178. Related to #2177 and #2190.

## The bug

`getQuestions` is shared by **20 gamified iDevices**. It rejected a non-array input and then handed it straight back:

```js
getQuestions: function (questions, percentage, random) {
    if (!Array.isArray(questions)) return questions;   // returned undefined unchanged
```

Every caller assigns the result back and immediately reads `.length` off it:

```js
data.questionsGame = ...getQuestions(data.questionsGame, ...);
data.numberQuestions = data.questionsGame.length;   // TypeError if questions were missing
```

So the guard protected nobody — it just moved the crash to the caller. An activity that lost its questions (the damaged-payload class from #2177 / #2190) took down the whole iDevice with:

```
TypeError: Cannot read properties of undefined (reading 'length')
```

It now returns `[]`, so those activities render empty instead of crashing.

## Changes

- **`public/app/common/common.js`** — `getQuestions` returns `[]` for non-array input.
- **`form/export/form.js`** — one caller depended on the old falsy return: `if (!ldata.questionsData) return;` skipped rendering when there were no questions. `[]` is truthy, so it is now `!ldata.questionsData.length`. Same behaviour as before.
- **`form`, `trueorfalse`, `scrambled-list`** — `JSON.parse(JSON.stringify(odata || {}))`. `adaptative-quiz` already did this; these three now match.
- **`public/vitest.setup.js`** — `mockGamificationHelpers` had no `getQuestions`, so any test reaching it failed with "not a function". Added once, mirroring the real contract.

## How to test

**Automated**

```bash
make fix && make test-unit
```

The guard is verified in both directions: revert `return []` to `return questions` in `common.js` and the new `trueorfalse` / `common` tests fail with the exact `TypeError` above.

**Manually, with the attached example** 

[pr2192-actividades-sin-preguntas.elpx.zip](https://github.com/user-attachments/files/30104193/pr2192-actividades-sin-preguntas.elpx.zip)

It is a hand-made 4 KB package whose payloads are real eXeLearning ones with only the questions array removed. It has four blocks:

| # | Block | On `main` | On this branch |
|---|-------|-----------|----------------|
| 1 | True/False, no `questionsGame` | `TypeError … reading 'length'`, activity dead | renders empty, no error |
| 2 | True/False, `questionsGame` not an array | same crash | renders empty, no error |
| 3 | Form, no `questionsData` | renders nothing | renders nothing (unchanged — proves no regression) |
| 4 | Text | renders | renders |

1. Import the `.elpx` (Project → Open, or drag it in).
2. Open the page and the browser console.
3. On `main`, blocks 1 and 2 throw. On this branch the console is clean and the page loads whole.

## Tests

- `common.test.js` — the old `returns non-array inputs as-is` test encoded the old contract; replaced by `returns an empty array for non-array inputs` (`undefined`, `null`, `false`, plain object).
- `trueorfalse/export/trueorfalse.test.js` — `updateConfig` normalises missing and non-array questions to `[]`. This is the path the fix actually repairs; it is what fails on the old contract.
- `form/export/form.test.js` — `updateConfig` normalises missing questions and survives no data at all; `renderBehaviour` renders nothing when there are no questions (this is what fails if the `.length` guard is missed).
- `scrambled-list/export/scrambled-list.test.js` — `updateConfig` survives undefined and empty data.

Local: `make fix` clean, backend 7496 tests, integration 721 pass / 0 fail, frontend 13725 pass, plus `form.spec.ts` + `relate.spec.ts` green.

**I could not complete a full local E2E run** — the suite needs ~10-19 min here and kept being cut short, so I am relying on CI. The nine E2E tests that looked suspicious in a partial run fail **identically on `main` with these changes stashed** (`text:2434`, `home-is-where-art-is:231`, `download-source-file:590`, `latex:908`, `latex:969`), and the failing set varied run to run (4, then 6, then 5 on `main`), so they are pre-existing and timing-sensitive rather than a regression. `focusedEditMode.test.js` is flaky on `main` the same way.

## Notes

- **`trueorfalse`'s `|| {}` does nothing on its own here.** Its legacy migration branch still dereferences `questionsData` unguarded, so `updateConfig(undefined)` still throws — the guard only changes `SyntaxError` into `TypeError` until #2178 lands. It is included so the three iDevices match `adaptative-quiz`, and the tests here only cover the `typeGame`-present path, which this PR does fix. Branched off `main`; the two changes touch different hunks and merge cleanly in either order.
- **iDevice export tests resolve `getQuestions` through the mock in `vitest.setup.js`, not `common.js`.** The mock's fidelity is load-bearing, so `common.js` carries a comment to keep the two in sync.
- `public/files/perm/**` is excluded from vitest coverage (`vitest.config.mts:103`) because those sources are `eval`-loaded in tests, so the changed iDevice lines are not measured. They do ship with tests, as above.


erseco:

PR #2194, opened 2026-07-17
fix(idevices): iDevice data integrity — root cause of #2177 plus the reunified fix family

### LINKED ISSUE #2165: Game iDevices lose all content after editing in 4.0.2 (body disappears, title remains)

### LINKED ISSUE #2177: Malformed iDevice jsonProperties aborts page loading and hides subsequent blocks

### LINKED ISSUE #2186: Game iDevices read ephemeral blob: URLs instead of asset:// when loading editor data from *-Link* anchors

### LINKED ISSUE #2187: Classify: audio-only cards are rejected on save — both validators ignore the audio field

erseco:
Consolidates the iDevice data-integrity family — the root-cause fix for #2177 plus six sibling PRs — into one reviewable branch, each fix kept as its own commit. Reunifies #2166, #2178, #2179, #2188, #2189, #2192.

## What it fixes

- **Root cause (#2177):** `prepareJsonForSync` corrupted serialized `jsonProperties` when a blob URL sat inside an HTML attribute (the regex ate the escaping backslash). It now walks the parsed JSON — recoverable assets recover, nested values are covered, Electron `blob:app://` included.
- **Write validation (#2179):** invalid payloads are rejected before touching the Y.Doc — validated once, no silent fallback, no orphan blocks.
- **Load tolerance (#2178):** a damaged activity no longer hides the blocks after it; its raw payload is preserved and edit-blocked with a visible notice.
- **Renderer hardening (#2192):** gamified iDevices survive missing or non-array question data.
- **Content loss (#2165 / #2166):** an empty write can no longer wipe an existing game body.
- **Asset leak (#2186 / #2188):** detached anchor reads no longer leak blob URLs into saved content.
- **Classify (#2187 / #2189):** audio-only cards validate again.

## How to test

Download **`exelearning-data-integrity-test-pack.zip`** and unzip it — six `.elpx` files, one per fix. Each row of the README shows the steps and the expected main-vs-fixed behaviour. 

[exelearning-data-integrity-test-pack.zip](https://github.com/user-attachments/files/30662010/exelearning-data-integrity-test-pack.zip)

_Pack refreshed 2026-08-03: `classify-content-2166.elpx` shipped without the eight media files it references, so its cards showed raw `{{context_path}}/…` placeholders when edited (spotted by @ignaciogros). The media is restored; no other fixture changed._

The three quickest:

1. **`repro-2177-blob-en-html-anidado.elpx`** — import, edit the Form, insert an image into the first question, save, reload.
   - On `main`: the save corrupts the activity → `SyntaxError` on reload, the page stops rendering.
   - Here: it saves and survives.
2. **`pr2192-actividades-sin-preguntas.elpx`** — import and open the page with the browser console open.
   - On `main`: `TypeError … reading 'length'`, dead activities.
   - Here: clean console, activities render empty.
3. **`classify-content-2166.elpx`** — import a game with content, then edit heavily (ideally two collaborators).
   - On `main`: the game body can be wiped (title stays, body empty).
   - Here: the body always survives.

No `blob:` URL survives in an `.elpx`, so `repro-2177` stages the project and you fire the bug by inserting media. For a one-liner, run in any open project's console:

```js
const am = window.eXeLearning.app.project._yjsBridge.assetManager;
const out = am.prepareJsonForSync(JSON.stringify({ textTextarea: '<img src="blob:https://x/unknown" alt="x">' }));
try { JSON.parse(out); console.log('OK →', out); } catch (e) { console.error('CORRUPTED →', e.message); }
```
On `main` it prints `CORRUPTED`; here it prints `OK`.

## Validation report

A cross-environment verification (localhost fix, Cloudflare Pages fix-static, `online.exelearning.dev` = main, `online.exelearning.net` = 4.0.2) plus real two-client collaboration and a real Electron `blob:app://` run:

[data-integrity-report-EN.pdf](https://github.com/user-attachments/files/30143156/data-integrity-report-EN.pdf)
same report in spanish:
[informe-integridad-datos-ES.pdf](https://github.com/user-attachments/files/30143158/informe-integridad-datos-ES.pdf)

## Tests

Every fix ships regression tests, verified in both directions (reverting a fix fails its tests). Frontend suite green; patch coverage complete on the modified lines. The six original PRs are merged into this branch and show as **Merged**; review happens here.

Closes #2165. Closes #2177. Closes #2186. Closes #2187.
(#2190 stays open on purpose: it is being re-scoped to import-side resilience and is not implemented here.)


erseco:

- I could not substantiate any confirmed blocker or important defect. The one proposed finding (`public/files/perm/idevices/base/form/export/form.js:206`, `!ldata.questionsData` → `!ldata.questionsData.length`) is not a live bug: `renderBehaviour` derives `ldata` from `updateConfig`, which unconditionally ends with `data.questionsData = getQuestions(...)`, and `getQuestions` (`public/app/common/common.js:2323`) returns `[]` for non-array input and an array in every other branch — so `.length` is always safe. `updateConfig` also parses `odata || {}`, so `undefined`/`{}` inputs are covered, and `form.test.js:206-230` asserts exactly those cases do not throw. The cross-file contract is already documented in a `getQuestions` comment. This is a defensive-style nit (`?.length`), not a defect.
- Spot-checked the core fix (`AssetManager.prepareJsonForSync` now parses then walks the tree instead of regex-rewriting serialized JSON, returning the payload untouched on parse failure) — this correctly avoids the backslash-eating corruption of #2177 without introducing data loss.

Verified: static read of the diff and full files at head 3145e11; no local build/test/server run. CI: all checks green on 3145e11 ✅.
Residual risk: low. Optional (non-blocking) hardening — `ldata.questionsData?.length` at form.js:206 would remove the implicit dependency on the `getQuestions` array contract, but the code is provably safe and tested at this head.

ignaciogros (dismissed):
Hi @erseco,

`classify-content-2166.elpx` doesn't have any images or other files associated with the activity. I don't think it's a bug, but it did seem unusual to me. When I edit the activity, I see values such as `{{context_path}}/rabbit.svg` in the form fields. 

I didn't find any issues when editing a similar complete activity (eXeLearning Manual → Classify).

I did find issues in `pr2192-actividades-sin-preguntas.elpx`. In that case, on Windows (with the "Cloud" version):

* I open the content.
* I go to the "Activities without questions" section.
* I click **Edit**, and the browser console shows the following error: `Uncaught TypeError: can't access property "length", $exeDevice.questionsGame is undefined` [/files/perm/idevices/base/trueorfalse/edition/trueorfalse.js]. From that point on, even if I enter a statement and mark it as true or false, I can't save the iDevice: `Uncaught (in promise) TypeError: can't access property 0, $exeDevice.questionsGame is undefined`

Could you please take a look?

Thanks.

erseco:
Hi @ignaciogros, thanks — both observations were worth raising, and the second one is a real bug that this PR should have caught. Fixed here.

## 1. `classify-content-2166.elpx` and `{{context_path}}/rabbit.svg`

Your instinct is right: not a product bug, but the fixture is at fault, not you.

That package contains only `content.xml` and `content.dtd` — no resources at all — while its stored properties reference `{{context_path}}/rabbit.svg`, `leon.svg`, `ardilla.svg`, `tigre.svg` and the `*_name.mp3` files. `{{context_path}}/…` is the placeholder the exporter writes for an asset reference; on import, `AssetManager.convertContextPathToAssetRefs` resolves each one against the assets actually extracted from the package and rewrites it to an `asset://` URL. With no asset to match, it leaves the reference verbatim — deliberately, since rewriting it would destroy the only record of what the card pointed at. That surviving token is what you saw in the form fields.

So the fixture references files that were never bundled with it. Tracing where it came from makes that obvious: it was hand-derived from the Manual's own **Clasifica** activity — same `ardilla.svg`, `leon.svg`, `rabbit.svg`, `tigre.svg`, `chipmunk_name.mp3`, `lion_name.mp3`, `rabbit_name.mp3`, `tiger_name.mp3` — and whoever cut it copied the activity's properties but not its `content/resources/` files. Your control test resolving correctly is exactly the expected contrast: you were comparing the same activity with and without its media.

**Fixed fixture attached:** [classify-content-2166.elpx.zip](https://github.com/user-attachments/files/30661880/classify-content-2166.elpx.zip) — rename it to `.elpx` (GitHub will not accept the bare extension). Same package, with the eight files restored from the Manual and nothing else changed.

I checked it against the real importer rather than by eye, in both directions:

| | `{{context_path}}` tokens surviving import | references rewritten to `asset://` | assets stored |
|---|---|---|---|
| fixture you tested | 8 | 0 | 0 |
| attached fixture | 0 | 8 | 8 |

Separately, showing a raw `{{context_path}}` token to an author is poor feedback for what is really a missing file. That is a UX gap rather than a defect, and it is outside this PR, so I opened #2223 for a proper "this activity references a file that is not in the package" notice. Your reaction — "I don't think it's a bug, but it did seem unusual" — is precisely the gap it describes, so I quoted it there.

## 2. `pr2192-actividades-sin-preguntas.elpx` — confirmed bug, fixed

You found a genuine hole, and it is in this PR's own scope. Thank you.

**Root cause.** In `public/files/perm/idevices/base/trueorfalse/edition/trueorfalse.js`, `loadPreviousValues()` runs the stored payload through `transformObject()`, which returns an already-migrated activity **untouched**:

```js
transformObject: function (data) {
    if (data.typeGame && data.typeGame === 'TrueOrFalse') {
        return data;      // no normalization on this branch
    }
    …
```

The `Array.isArray(...)` normalization that does exist there only guards the legacy `questionsData` migration branch, which is skipped for a payload already saved in the modern shape. So `$exeDevice.questionsGame = dataGame.questionsGame` assigned `undefined`, and the next read — `addEvents()` doing `$exeDevice.questionsGame.length` — threw exactly what your console showed. Saving then failed the same way in `validateData()` and in the `questionsGame[$exeDevice.active]` write, which is your second error.

Both damaged shapes in that fixture hit it: `tof-missing-questions` (no `questionsGame` key) and `tof-nonarray-questions` (`questionsGame` holding an object).

**Why the PR missed it.** The #2192 hardening covered only the export/runtime side — `export/trueorfalse.js`, `form.js`, `casestudy.js`, `scrambled-list.js` and `getQuestions` in `common.js`. The edition side was never touched. The result was a half-fix that is easy to mistake for a whole one: the activity renders cleanly, so the console is quiet until you click **Edit**. Your report is what surfaced the asymmetry.

**The fix.** Normalization now happens at both boundaries where a question list enters the editor — the activity's saved properties and an imported game file — through a single helper:

```js
toQuestionsArray: function (questions) {
    return Array.isArray(questions) ? questions : [];
},
```

An empty list leaves the editor in exactly the state of a brand-new activity, which `addEvents` then seeds with one editable question — so a damaged activity is recoverable in place instead of dead.

Applying it exposed a second latent bug on the same path: `deleteEmptyQuestion()` read `question` as an implicit global that was only assigned when the TinyMCE instance existed, so without one it threw `ReferenceError: question is not defined`. It now treats a missing editor as empty content, which is what that case means.

**Tests.** Ten unit tests in `trueorfalse.test.js` covering both damaged shapes, `null`, the untouched-valid-list control, the import path and the trimming path; plus a Playwright spec, `game-without-questions-edit.spec.ts`, that drives the real editor: it opens the activity, types a question, marks it true, saves, and asserts a proper question array lands on the component. Verified in both directions — reverting the fix makes the spec fail with your error (`Cannot read properties of undefined (reading 'length')`, Chrome's wording of the same TypeError).

Steps to re-check on your side, same as your report: open the content, go to **Activities without questions**, click **Edit**. The console should stay clean, the activity should open with one empty question, and saving should work.

ignaciogros (approved):

Issue #2028, opened 2026-07-02
[Proposal] Remove the “Editable content” (`exportSource`) export option


greynoa:
## Proposal

Remove the **“Editable content”** export option (`exportSource`) from the eXeLearning export flow and stop using it to omit `content.xml` from generated packages.

The current option is unclear for users, has already caused accidental exports without `content.xml`, and creates a misleading sense of copyright protection. In practice, it does not solve copyright or licensing concerns; it mainly prevents legitimate re-editing, reuse, adaptation, and maintenance of educational resources created with eXeLearning.

@erseco @ignaciogros @cristinavaldera, could you please share your opinion on this proposal?

## Background

The `exportSource` / “Editable content” option appears to control whether the export includes the editable source representation of the project, especially `content.xml`.

This is problematic because `content.xml` is not an optional cosmetic file. In the current `.elpx` format documentation, it is described as the root-level file used by importers to reconstruct the project. The container documentation also marks `content.xml` as required by the importer; without it, the package cannot round-trip back into the editor.

This means that a user can accidentally generate an artefact that looks like a normal eXeLearning export but is no longer editable as an eXeLearning resource.

### 1. It is unclear and easy to misuse

The label “Editable content” is ambiguous. Users may disable it without understanding that this can remove the source needed to reopen or adapt the resource later.

We have already found real cases where this option was selected incorrectly and the resulting package did not include `content.xml`. From the user’s perspective, this behaves like accidental data loss: the content can be viewed, but not properly recovered as an editable eXeLearning project.

### 2. It does not provide meaningful copyright protection

Omitting `content.xml` is not a reliable copyright protection mechanism.

The exported HTML, media files, images, CSS, JavaScript, and other assets are still distributed. Anyone receiving the package can still inspect, copy, extract, or reuse the published content at different levels. What is lost is mainly the structured editable representation that allows legitimate users to maintain, translate, adapt, correct, or remix the resource.

Copyright and licensing concerns should be handled through proper licensing metadata, attribution, terms of use, and legal notices, not through an unclear export checkbox that silently removes the editable source.

Creative Commons also recommends sharing content in downloadable and editable formats when reuse is intended, and CC licenses explicitly discourage adding technical restrictions that prevent others from exercising the rights granted by the license.

### 3. It goes against the purpose of Open Educational Resources

[...]

PR #2196, opened 2026-07-20
fix(export): include content.xml in the download-source-file .elpx manifest

### LINKED ISSUE #2195: Download source .elpx from the download-source-file iDevice is missing content.xml (not re-importable)
### Summary

When a resource contains a **Download source file** iDevice, clicking **"Download the .elpx file"** in the published/served output produces an `.elpx` that is **missing `content.xml`** (and also `content.dtd` and the root `screenshot.png`). Because `content.xml` is the file the importer uses to reconstruct the project, the downloaded package **cannot be reopened or edited** in eXeLearning — it behaves like a view-only publication even though the user explicitly asked for the editable source. From the recipient's point of view this looks like data loss.

### How it was found

A resource shared through exeviewer (`https://exeviewer.intef.es/?url=...&download=1`) was downloaded with the in-content **"Descargar el archivo .elpx"** button. The downloaded `.elpx` was missing `content.xml`, while the original `.elpx` (served by the viewer) had it.

Comparing both packages:

- **original .elpx** → 421 entries, root contains: `search_index.js`, `index.html`, **`content.xml`**, **`content.dtd`**, **`screenshot.png`**
- **downloaded .elpx** → 418 entries, root contains: `search_index.js`, `index.html` — the three source files above are gone (exactly the 3-file difference)

### Root cause

The download-source-file iDevice rebuilds the `.elpx` **client-side** from a manifest baked into the package: `libs/elpx-manifest.js` (`window.__ELPX_MANIFEST__`), consumed by `public/libs/exe_elpx_download/exe_elpx_download.js`, which fetches exactly the files listed in the manifest and re-zips them.

In `src/shared/export/exporters/ElpxExporter.ts`, that manifest is generated in Section 1 (HTML5 content), **before** Section 2 adds the ODE re-import files. `content.xml`, `content.dtd` and `screenshot.png` are written with `this.zip.addFile(...)` (untracked) *after* the manifest already exists, so those three files are present in the saved `.elpx` but **not listed in the manifest**. The client-side re-download therefore silently drops them.

Confirmed by inspecting the manifest embedded in a real saved `.elpx`: it lists `theme/screenshot.png` but **not** root `content.xml` / `content.dtd` / `screenshot.png`.

(For comparison, the SCORM and IMS exporters already push `content.xml` / `content.dtd` into their common-files list before generating the manifest, so they are **not** affected. This is specific to `ElpxExporter`.)

### Relationship to #2028

This is **related to but not the same as** #2028. #2028 is about the `exportSource` / "Editable content" option intentionally omitting `content.xml`. Here `exportSource` is enabled and `content.xml` **is** present in the saved `.elpx`; it is only dropped from the download-source-file manifest used to reconstruct the package on re-download. Fixing this does not depend on the outcome of #2028.

### Impact

**High.** Users who share editable resources through the download-source-file iDevice unknowingly distribute non-re-importable packages. This has been reported more than once.

### Steps to reproduce

1. Create a project and add a **Download source file** iDevice.
2. Export/save it as `.elpx` and open or serve the published output.
3. Click **"Download the .elpx file"**.
4. Inspect the downloaded `.elpx`: `content.xml` is missing, and importing it back into eXeLearning fails.

### Proposed fix

Generate the ELPX download manifest **last**, from the actual ZIP contents (`this.zip.getFilePaths()`), after every file — including `content.xml`, `content.dtd` and `screenshot.png` — has been added. This keeps the manifest a faithful reflection of the package, so the manifest can never drift from the ZIP again and the re-download stays re-importable.

---

Reported by @cristinavaldera (also requested as reviewer of the fix).

erseco:
## What

Fixes #2195: the **Download source file** iDevice produced an `.elpx` **missing `content.xml`** (plus `content.dtd` and root `screenshot.png`), so the downloaded package could not be reopened/edited.

This PR now fixes the same defect on **both** download paths:

1. **Direct ELPX export** (`ElpxExporter`) — the re-download manifest dropped the ODE source files.
2. **In-app Service Worker preview** (`Html5Exporter.generateForPreview()`) — the preview omitted `content.xml` entirely, so a download triggered from the preview was never editable.

## Why

The iDevice rebuilds the `.elpx` client-side from `libs/elpx-manifest.js` (consumed by `public/libs/exe_elpx_download/exe_elpx_download.js`), fetching exactly the files the manifest lists.

- In `ElpxExporter`, the manifest was generated in Section 1, **before** Section 2 added the ODE re-import files with `this.zip.addFile(...)`. Those files were in the saved `.elpx` but absent from the manifest, so the re-download silently dropped them.
- In `Html5Exporter.generateForPreview()`, `content.xml` was skipped unconditionally ("not needed for viewing"). But the preview is also a real download source for the iDevice, so a preview-originated `.elpx` was non-re-importable even when editable source was enabled.

Related to but **not** the same as #2028: here `exportSource` is enabled and `content.xml` **is** in the saved `.elpx` — it was only missing from the re-download manifest / preview.

## Fix

**`ElpxExporter`** — generate the manifest **last**, from the actual ZIP contents (`this.zip.getFilePaths()`), after every file (including `content.xml`, `content.dtd`, `screenshot.png`) has been added. The manifest can no longer drift from the package. `ElpxExporter` always ships `content.xml`, so a direct ELPX export stays fully editable.

**`Html5Exporter`** — extract a single shared helper `addEditableContentXml(pages, meta, addFile)` that decides whether `content.xml` is generated (skipped only when `exportSource === false`) and use it in **both** `export()` and `generateForPreview()`. Because the preview `addFile()` already tracks files in the manifest `fileList`, `content.xml` is now automatically listed in `libs/elpx-manifest.js` when editable source is on — no manual manifest append, and no duplicated `exportSource` conditional between export and preview.

Also fixed a stale comment in `ElpxExporter.ts`.

### Resulting behaviour

| Context | `exportSource` | `content.xml` |
| --- | ---: | ---: |
| HTML5 ZIP export | `true` or undefined | Included |
| HTML5 ZIP export | `false` | Omitted |
| Service Worker preview | `true` or undefined | Included |
| Service Worker preview | `false` | Omitted |
| ELPX direct export | any value | Included |

## Tests (TDD)

`ElpxExporter.spec.ts`:

- lists `content.xml` and `content.dtd` in the manifest so the re-download is re-importable;
- lists `screenshot.png` in the manifest when the project has a screenshot;
- **completeness invariant**: every file present in the ZIP is listed in the manifest (a faithful copy) — this guards against the whole class of "file added but not tracked".

`Html5Exporter.spec.ts` (added a `parseElpxManifest()` test helper and preview regression tests):

- preview includes `content.xml` when `exportSource` is `true`;
- preview includes `content.xml` when `exportSource` is undefined (editable by default);
- preview omits `content.xml` when `exportSource` is `false`;
- with a Download source file iDevice, the preview `libs/elpx-manifest.js` lists `content.xml` when enabled and does **not** list it when `exportSource` is `false`.

## How to verify

- `bun test src/shared/export/exporters/ElpxExporter.spec.ts src/shared/export/exporters/Html5Exporter.spec.ts` — all pass.
- `make test-unit` / `make test-integration` — green.
- Manual: add a Download source file iDevice → open the in-app **preview** → click **"Download the .elpx file"** → the downloaded `.elpx` now contains `content.xml` and re-imports cleanly. Setting **Export source: off** (`exportSource === false`) omits it, as before.

## Files changed

- `src/shared/export/exporters/ElpxExporter.ts`
- `src/shared/export/exporters/Html5Exporter.ts`
- `src/shared/export/exporters/Html5Exporter.spec.ts`
- `src/shared/export/exporters/ElpxExporter.spec.ts`

### DIFF
--- a/src/shared/export/exporters/ElpxExporter.spec.ts
+++ b/src/shared/export/exporters/ElpxExporter.spec.ts
@@ -144,6 +144,24 @@ class MockZipProvider implements ZipProvider {
     }
 }
 
+/**
+ * Parse the embedded `window.__ELPX_MANIFEST__` JSON from the generated
+ * `libs/elpx-manifest.js`. The download-source-file iDevice rebuilds the
+ * package client-side from this manifest, so its `files` list must faithfully
+ * reflect the ZIP contents.
+ */
+function parseElpxManifest(zip: MockZipProvider): { version: number; files: string[]; projectTitle: string } {
+    const manifestJs = zip.files.get('libs/elpx-manifest.js') as string;
+    if (!manifestJs) {
+        throw new Error('libs/elpx-manifest.js not found in ZIP');
+    }
+    const match = manifestJs.match(/window\.__ELPX_MANIFEST__=(\{[\s\S]*?\});/);
+    if (!match) {
+        throw new Error('Could not parse __ELPX_MANIFEST__ from manifest file');
+    }
+    return JSON.parse(match[1]);
+}
+
 // Sample pages for testing
 const samplePages: ExportPage[] = [
     {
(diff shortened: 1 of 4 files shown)


erseco:

- Verified the core fix on both paths. `ElpxExporter.ts` now builds the download manifest last from `this.zip.getFilePaths()` (Section 2.4), after `content.xml`/`content.dtd`/`screenshot.png` are added, so the manifest can no longer drift from the package. `Html5Exporter.ts` shares one `addEditableContentXml()` helper between `export()` and `generateForPreview()`, keyed on `exportSource !== false`; the preview registers the file via `addFile`, so it lands in `libs/elpx-manifest.js` when a download-source-file iDevice is present. Behavior matches the PR's table and the added specs.
- Not a defect: the only proposed finding (preview always generates `content.xml` regardless of `needsElpxDownload`) describes intentional, tested behavior — the specs assert `content.xml` is present in previews built from pages with no download iDevice, and the shared helper is the PR's stated single-source-of-truth. Gating preview on `needsElpxDownload` would reintroduce export/preview divergence and break two passing tests; the cost is a minor, unquantified serialization on the preview path, not a correctness issue.

Verified: static read of the diff and full files at head e1e10f8; CI: all checks green on e1e10f8. Residual risk: the manifest/preview round-trip and re-import were not exercised at runtime (no local build/E2E run); coverage rests on the added unit specs plus green CI.

ignaciogros (approved):

PR #2197, opened 2026-07-21
fix: render LaTeX correctly inside FX tabs

### LINKED ISSUE #2170: Bug: FX elements inside Form iDevice work in the editor but fail to generate HTML upon export

### LINKED ISSUE #2191: bug: LaTeX rendering issue in FX Tabs
## Description
There is a rendering issue when attempting to display LaTeX equations inside `FX Tab` components. 
## Steps to Reproduce
1. Create an `FX Tabs` component.
2. Pass a string containing LaTeX notation as the tab's content.
3. Save iDevice.

Image

[error-latex-tab_web.zip](https://github.com/user-attachments/files/30086135/error-latex-tab_web.zip)

erseco:
## Summary

LaTeX equations placed inside an **FX Tabs** component rendered correctly everywhere
except in the **tab labels**, which showed garbled text such as `sumn=110n`
(see the original report). Tab content, plain text, and the Accordion effect were
never affected.

This PR makes FX tab labels keep the rendered equation, so a heading like
`\(sum_{n=1}^{10}n\)` now shows the typeset formula in the tab, in the workarea,
on reload, in Preview and in exports.

## Root cause

This is **not** a MathJax/DOM race — it is a structural bug in label building.

`$exeFX.tabs.rft()` copies each tab heading into a separate navigation entry
(`<ul class="fx-tabs"><li><a>…</a></li>`) and built that label from
`h2.text()`. By the time the effect runs, the heading's LaTeX has already been
pre-rendered to `<span class="exe-math-rendered">` containing an SVG **and** an
assistive MathML `<math>` block. `.text()` throws away the SVG and returns the
concatenated MathML characters:

```
<h2><span class="exe-math-rendered" data-latex="\(sum_{n=1}^{10}n\)">
      <svg>…</svg>
      <math><mi>s</mi><mi>u</mi><mi>m</mi><mi>n</mi><mo>=</mo><mn>1</mn><mn>10</mn><mi>n</mi></math>
    </span></h2>

h2.text()  ->  "sumn=110n"          (rendered SVG discarded)
```

The **Accordion** effect is immune because it *wraps* the existing `<h2>` in an
`<a>` (keeping its markup) instead of extracting `.text()`. Paginated/Carousel
use `.text()` only for the `title="…"` tooltip attribute (their visible label is
a number), so they show no visible breakage and are intentionally left unchanged.

In a fully pre-rendered export MathJax is not shipped (the selective
pre-rendering optimisation), so the label *must* carry the already-rendered
markup — it cannot be re-typeset later. The fix therefore preserves the
heading's HTML rather than trying to re-run MathJax.

## Relationship with #2170 and #2171

Same subsystem (`exe_effects.js`), **different root cause**.

- **#2170 / PR #2171** fix *"FX not initialised in dynamically-inserted Form/Magnifier
  content"* and *"the export library detector misses FX markup stored in nested
  iDevice properties"*. PR #2171 changes `$exeFX.init()` to accept a scope
  container; it does **not** touch the tab-label building.
- **#2191** (this PR) fixes the tab **labels** losing the rendered equation.

Verified empirically — running this PR's failing test against each revision:

| Revision | `keeps pre-rendered equations visible in tab labels (#2191)` |
|---|---|
| `main` | ❌ fails (label has no rendered math) |
| PR #2171 head (`fix/2170-form-fx-export`) | ❌ fails (identical — never touches `.text()`) |
| **this PR** | ✅ passes |

This PR is **independent** of #2171 (it changes different lines of
`exe_effects.js`), does not depend on it, and does not supersede it. It does
**not** close #2170.

## Changes

- `public/app/common/exe_effects/exe_effects.js`
  - New `$exeFX.labelHtml(h2, salt)` helper that returns the heading's inner
    HTML and suffixes any element `id`/`#id` references so a copied
    self-referential MathJax SVG never produces duplicate ids or broken
    `<use>` links.
  - `tabs.rft()` builds the label with `labelHtml()` instead of `h2.text()`.
  - Plain-text titles and the TinyMCE `title`-attribute path are unchanged.

### Failing test before the fix

`public/app/common/exe_effects/exe_effects.test.js` →
`keeps pre-rendered equations visible in tab labels (#2191)`.

It builds a tab whose heading is a pre-rendered `exe-math-rendered` span and
asserts the navigation label still contains `.exe-math-rendered` / `svg`. On
`main` it fails with `AssertionError: expected null not to be null` (the label
was flattened to `sumn=110n`).

### Passing test after the fix

```
$ npx vitest run public/app/common/exe_effects/exe_effects.test.js
 ✓ public/app/common/exe_effects/exe_effects.test.js (19 tests)
 Test Files  1 passed (1)
      Tests  19 passed (19)
```

New tests cover: pre-rendered LaTeX in the label, raw `\(…\)` LaTeX preserved
for MathJax-shipping contexts, **no duplicate ids** after copying, and
idempotent re-initialisation (no duplicate tab controls).

An end-to-end regression, `test/e2e/playwright/specs/fx-tabs-latex.spec.ts`,
drives the full user workflow: create the FX-tabs iDevice, save, verify the
rendered label, switch to the initially hidden tab and verify its content,
switch back, reload for persistence, verify Preview and the HTML5 export, and
assert no relevant page errors.

## Manual verification

Reproduced and verified in a real browser using the attached `.elpx`
(equivalent to the reporter's project):

- **Before (unmodified `exe_effects.js`)** — tab labels show orange `sumn=110n`
  while text/accordion render fine.
- **After (this PR)** — tab labels render `sum₁₀ₙ₌₁ n`.
- Workarea after importing the `.elpx`: both tab labels rendered; switching to
  the initially hidden second tab renders its equation; no `MJX-*` duplicate ids
  in the document.
- The real exported package (only `exe_effects.js` swapped) confirms the fix in
  exported HTML, where MathJax is not shipped.

Screenshots and the reproduction `.elpx` are attached below.

## Validation

```
make fix            # biome — clean (exe_effects legacy path is lint-exempt; new spec formatted)
npx vitest run public/app/common/exe_effects/                       # 19 passed
npx vitest run .../ideviceNode.test.js .../idevicesEngine.test.js   # 754 passed (FX-init callers)
```

The Playwright browser binary could not be downloaded in this environment, so
`fx-tabs-latex.spec.ts` was validated manually via the equivalent real-browser
workflow above; it follows the existing spec patterns and runs in CI.

Fixes #2191
Related to #2170
Related to #2171


erseco:
## Before / after (#2191)

Same project, same equation (`\(sum_{n=1}^{10}n\)`), same viewport and tab state.

### Before — broken (unmodified `exe_effects.js`)

The **FX Tabs** labels show the flattened MathML text `sumn=110n` (orange), while the text above and the Accordion below render fine.

2191-before-broken-tabs

### After — fixed (this PR)

**Exported HTML** (MathJax is not shipped — the label uses the pre-rendered SVG):

2191-after-fixed-tabs

**Workarea** after importing the reproduction `.elpx` (first tab):

2191-after-workarea-fixed

**Initially hidden second tab**, after switching to it — its equation renders too:

2191-after-workarea-hidden-tab

### Reproduction project

Unzip and open the `.elpx` in eXeLearning (**File → Open → Select a file**). It reproduces the reported "En texto / En FX" page (Tabs, Accordion and Paginated blocks with LaTeX headings). On `main` the tab labels are broken; with this PR they render.

[error-latex-fx-tabs-elpx.zip](https://github.com/user-attachments/files/30219478/error-latex-fx-tabs-elpx.zip)

erseco:

- The production fix is sound. `labelHtml()` (public/app/common/exe_effects/exe_effects.js:130) suffixes copied `id="…"`/`href="#…"` with exact closing-quote boundaries, so no prefix collisions, and the MathJax `<use xlink:href="#id">` is correctly rewritten because the regex matches the `href="#id"` substring inside it. The `rft()` change (line ~409) preserves heading markup only on the visible-label path; the plain-text and TinyMCE `title` paths are untouched. The colocated unit test `keeps pre-rendered equations visible in tab labels (#2191)` is a deterministic regression guard (seeds a pre-rendered `exe-math-rendered` span, asserts the SVG survives).
- **Test gap (minor):** test/e2e/playwright/specs/fx-tabs-latex.spec.ts:214-242 — the HTML5 export test asserts the export contains `exe-math-rendered` and bundles `exe_effects.js`, but tab labels are built at runtime; the static export only ever contains the panel `<h2>` headings, which carry `exe-math-rendered` on both buggy and fixed code, and `exe_effects.js` is always bundled. So this test is a packaging smoke test, not a #2191 regression guard. It is not misleading (the spec comments say the labels are "built at runtime"), and the unit test covers the regression — no change required.

Verified: static inspection of the diff and full files at head; unit-test authorship reviewed. `CI: all checks green on 6ce7d44 ✅`.
Residual risk: I did not run Playwright or Vitest locally. The workarea e2e assertion (spec lines 118-156) injects raw `\(…\)` and its strength depends on pre-render-vs-FX-init ordering that is not observable statically; the deterministic unit test remains the authoritative guard regardless of that ordering.

franmate (approved):
Using labelHtml() properly resolves the markup loss when rendering LaTeX inside tabs. Tested and approved. LGTM!

ignaciogros (approved):

PR #2198, opened 2026-07-21
fix: allow desktop import of ELPX files with large assets

### LINKED ISSUE #2193: Desktop cannot reopen ELPX files exported by the web editor when an asset exceeds 200 MiB
## Summary

The desktop application cannot import an ELPX file exported by the web editor when the package contains a single asset larger than 200 MiB.

The web editor currently allows the project to contain and export this asset, but the downloaded ELPX cannot subsequently be opened in the desktop application.

This creates an interoperability problem between the web and desktop versions.

## Example error

```text
Entry 'content/resources/<video>.mp4' in ELP/ELPX archive is too large when decompressed
(359357639 bytes > 209715200 byte limit).
````

## Steps to reproduce

1. Create or open a project in the cloud editor.
2. Add an MP4 file larger than 200 MiB.
3. Save the project.
4. Export or download the project as ELPX.
5. Open the downloaded ELPX in the desktop application.

## Actual behaviour

The import is aborted and the desktop application displays a generic error indicating that the ZIP entry exceeds the 200 MiB decompressed-size limit.

The complete project cannot be opened.

## Expected behaviour

A project accepted and exported by the web editor should be reopenable in the desktop application.

Large files must not disable ZIP-bomb protections, but the application should either:

* support legitimate large assets safely; or
* warn the user before producing an ELPX that cannot be reopened in another supported eXeLearning environment.

## Technical context

`ElpxImporter` currently defines the following default limits:

* 500 MiB total uncompressed size;
* 200 MiB per ZIP entry;
* 10,000 entries.

The browser importer creates `ElpxImporter` without overriding these defaults. The desktop import flow therefore uses the same 200 MiB per-entry limit.

The current implementation uses `fflate.unzipSync`, which materialises entries in memory. Simply removing the limit or increasing it globally could reintroduce memory-exhaustion and ZIP-bomb risks.

### Short-term

* Keep the current conservative limits for server-side and hosted imports.
* Make decompression limits configurable by runtime.
* Allow the desktop runtime to use a higher explicit per-entry limit, optionally requiring user confirmation.
* Keep a cumulative archive limit.
* Add tests covering runtime-specific limit overrides.
* Add a pre-export compatibility warning when an asset exceeds the desktop import limit.

A desktop-only limit such as 512 MiB per entry could cover this case as an interim measure, but it should not be considered the final solution because synchronous decompression may still require a large amount of memory.

### Long-term

Replace full synchronous in-memory extraction with a streaming or lazy extraction path:

* parse only the project metadata and XML in memory;
* stream large assets directly to persistent storage;
* enforce actual decompressed byte counts while streaming;
* abort cleanly when configurable per-entry or total limits are exceeded;
* preserve ZIP-bomb and resource-exhaustion protections.

## User experience

The current error should be replaced with a translated and actionable message containing:

* the affected filename;
* its formatted size;
* the configured limit;
* instructions to open the File Manager and sort by “Largest first”;
* a suggestion to check whether an unused older version exists;
* a recommendation to optimise or replace the asset.

An explicit “Import without this asset” option could be considered, but it must clearly warn that the resulting project will contain missing media. Oversized assets should never be skipped silently.

## Acceptance criteria

* An ELPX containing a legitimate MP4 of approximately 360 MB can be opened by the desktop application through a controlled mechanism.
* ZIP-bomb protections remain enabled.
* Hosted/server imports retain conservative defaults.
* Runtime-specific limits are covered by unit tests.
* The web editor warns before exporting a project that is incompatible with the current desktop limits.
* The displayed error is translated and provides actionable remediation.

erseco:
## Summary

Allows the Electron desktop application to open legitimate ELPX projects containing assets above the conservative hosted-import threshold while preserving ZIP-bomb and resource-exhaustion protections.

Adds runtime-specific import policy handling and prevents the web editor from silently producing ELPX files that exceed the supported desktop compatibility policy.

## Root cause

`ElpxImporter.safeUnzip` enforces three conservative decompression caps (200 MiB/entry, 500 MiB total, 10 000 entries) read from the ZIP central directory before inflation. The optional per-runtime `zipLimits` constructor argument existed but **no production caller passed it** — server routes, CLI, and the browser adapter (`src/shared/import/browser/index.ts:141`) all used 3-argument construction — so every runtime, including Electron, inherited the 200 MiB per-entry cap.

Electron and the static PWA both resolve to `RuntimeConfig` `mode === 'static'`, so the runtime mode alone cannot distinguish "desktop" from "static web". The desktop app therefore rejected a legitimate ~360 MB video with `Entry '…' too large when decompressed (359357639 bytes > 209715200 byte limit)`.

## Security model

- Conservative defaults **unchanged** for hosted web, server, CLI, static PWA, and embedded imports (200 MiB/entry, 500 MiB total, 10 000 entries).
- Explicit **desktop** policy (1 GiB/entry, 2 GiB total, 10 000 entries), injected only when `window.electronAPI` is present.
- Per-entry, cumulative-size, and entry-count caps all remain enforced under every policy.
- Invalid limits (non-finite, non-positive, non-integer entry count, entry cap > total cap) are rejected at one boundary (`validateZipLimits`, called in the importer constructor).
- Preflight inspection reads the central directory **without inflating** any entry; an over-limit archive is rejected before any Y.Doc / asset-store mutation.
- No `Infinity` / `MAX_SAFE_INTEGER`, no exception suppression, no silent asset skipping.
- The declared-size trust limitation of `safeUnzip` is pre-existing and documented; it is unchanged here.

## Changes

- **New** `src/shared/import/importPolicy.ts` — single source of truth for limits, runtime selection, structured `ZipLimitError`, `ImportCancelledError`, the desktop export-compatibility check, and byte formatting.
- `ElpxImporter` — structured error details, constructor validation, and a `inspectZipArchive` preflight; limit primitives moved here and re-exported for compatibility.
- Browser adapter `importFromFile` — preflight → reject / confirm → post-gate `beforeImport` hook → import; fresh importer per call (no stale policy); exposes `window.ExeImportPolicy`.
- `YjsProjectBridge` — runtime policy selection, desktop confirmation, actionable translated error, and the pre-export desktop-compatibility warning.
- `modalOpenUserOdeFiles` — static open flow delegates the previous-project clear to the gated import (`clearPreviousProject`).
- ADR-0001 documents the decision.

### Failing tests before the fix
- `importPolicy.spec.ts` — module did not exist (RED).
- `ElpxImporter.spec.ts` — `inspectZipArchive` export missing; structured `details` absent (RED).
- `browser/index.spec.ts` — `createBrowserImporter` export / preflight+confirm absent (RED).
- `YjsProjectBridge.test.js` — `_resolveImportPolicy`/confirmation/export-warning methods absent (13 of 16 RED).

### Passing tests after the fix
- `bun test src/shared/import/` — 759 pass.
- `npx vitest run public/app/yjs/YjsProjectBridge.test.js` — 440 pass (incl. 16 new).
- `npx vitest run public/app/workarea/modals/modals/pages/modalOpenUserOdeFiles.test.js` — 102 pass.
- `bun x playwright test --project=static desktop-large-asset-import` — 3 pass.

## Runtime behavior

| Runtime | Per-entry policy | Confirmation | Result |
|---|---:|---|---|
| Hosted web | Conservative | No override | Reject above hosted limit |
| Server/CLI | Conservative | No override | Reject above hosted limit |
| Static PWA | Conservative | No override | Reject above hosted limit |
| Embedded | Conservative | No override | Reject above hosted limit |
| Electron desktop | Explicit desktop policy | Required in controlled range | Import when accepted |

## Manual verification

Verified in the Electron desktop app (`make run-app`) against a real ~436 MB ELPX containing a ~211 MiB MP4 entry (over the 200 MiB conservative per-entry cap, under the 500 MiB total cap) — the same failure class as the reported ~360 MB file:

- **Before the fix** (conservative policy, reproduced against the real file with the production code): the entry is rejected with `Entry '…' too large when decompressed (221238448 bytes > 209715200 byte limit)`.
- **After the fix** (desktop policy): the confirmation appears, and on accept the project imports (20 pages / 26 blocks / 25 components / 96 assets) with the large video asset materialised at its full size — preserved, not skipped.

### Desktop confirmation for a large asset

2193-confirmation

_The "project opened" and "large asset preserved" outcomes are verified by the automated E2E test (`test/e2e/playwright/specs/desktop-large-asset-import-static.spec.ts`) and the real-file headless run described above. Those screenshots are intentionally omitted here to avoid publishing third-party personal data (author name, a person's likeness) present in the sample project._

## Validation

- `make fix` / `make lint` — clean (exit 0).
- `bun test src/` — 7473 pass; `bun test test/integration/` — 721 pass.
- Frontend Vitest for changed files — pass.
- `bun x playwright test --project=static desktop-large-asset-import` — 3 pass.
- Patch line coverage: `importPolicy.ts` 100%, `browser/index.ts` 100%, `ElpxImporter.ts` 99.44%.

## Follow-up considerations

Streaming / lazy extraction that aborts on the actual inflated byte count remains future work; the desktop path still uses synchronous in-memory `unzipSync`, bounded by the desktop caps. This PR does not claim synchronous extraction is safe for arbitrarily large archives. Applying the shared limits to the two raw `fflate.unzipSync` calls in the Electron main process (`app/main.js`) is a separate follow-up.

Fixes #2193

## Fix — rejected/cancelled import still showed "Completed successfully"

`importFromElpx` now returns `{cancelled:true}` (optionally `error:'zip-limit'`) instead of throwing when an import is rejected/cancelled, but only one of five callers honored it — so a rejected or over-limit import showed a green "Completed successfully (0 pages)" and could re-save the unchanged project. Added a shared `isImportCancelled(result)` helper and made every caller (File-menu append, platform/URL bootstrap, structure append, static open) skip the success UI, save, and refresh on the cancelled shape. Commit `f4dc7b86`; tests added (609 passing across the touched suites).


erseco:

- **Important:** `public/app/yjs/YjsProjectBridge.js` `importFromElpx` was changed from throwing `ZipLimitError`/`ImportCancelledError` to returning `{cancelled:true[, error:'zip-limit']}` (lines ~3725-3735), but only `modalOpenUserOdeFiles` was updated to honor the new shape. The File-menu append-import at `public/app/workarea/menus/navbar/items/navbarFile.js:1128-1139` ignores the return and unconditionally runs `progressModal.setComplete(true, _('Completed successfully') + ' (0 pages)')`. So a hosted user importing an ELPX with a >200 MiB entry (or a desktop user cancelling the large-file confirmation) sees the bridge's "File too large to open" modal **and** a green "Completed successfully (0 pages)". Before this PR the thrown error hit the surrounding `catch` (~line 1264) and showed an error, not a success. Have every caller honor the new contract (skip the success UI when `result?.cancelled`), or keep `importFromElpx` throwing.

- **Minor (same root cause):** the platform/URL import paths at `public/app/workarea/project/projectManager.js:466` and `:556`, and `public/app/workarea/menus/structure/menuStructureBehaviour.js:589`, likewise ignore `{cancelled:true}` and proceed to `saveToServer()`/refresh and log "complete". No data loss — the preflight and confirm callback in `src/shared/import/browser/index.ts:232-282` throw *before* `importFromBuffer` clears the doc — but a rejected import still persists the unchanged doc and reports success. Worth guarding with the same `cancelled` check.

- The third proposed finding (export-compat cancel dropping the transition save) is **not confirmed**: `transitionToProject` (projectManager.js:673-675) already awaits and ignores the export result pre-PR, the user is explicitly warned at the compat prompt, and it only triggers for >1 GiB assets — a pre-existing save-cancel semantic, not a regression introduced here.

Verified: static open flow correctly consumes the new return shape; importer ordering rules out data loss on rejection. CI: all checks green on 6b48e26 (green CI does not exercise the navbar/append-import success/error messaging, so the contradictory toast is not caught). Residual risk: no local build/E2E was run by this reviewer; the append-import symptom is code-substantiated, not runtime-reproduced.

---
**Update — resolved (`f4dc7b86`):** added a shared `isImportCancelled(result)` helper; all `importFromElpx` callers now short-circuit on the `{cancelled}` contract (no success UI/save/refresh); tests added.

ignaciogros (approved):
Great job! Thanks, @erseco.

PR #2204, opened 2026-07-22
Fix unsaved changes warning after saving

### LINKED ISSUE #2200: Unsaved changes warning appears even after saving
In both the desktop and online versions, the application shows an unsaved changes warning when trying to close the project, even if it has just been saved.

The warning indicates that changes may be lost, although there are no pending changes.

Image

Image

erseco:
Fixes #2200.

Successful local saves left asset metadata pending, so close guards still reported unsaved changes. Saved local assets are now tracked correctly, and online saves only become clean after uploads finish.

**Test:** modify a project with an asset, save it, then close it; no warning should appear. Failed or cancelled saves should still warn.

## Packaging fix (added after review)

@cristinavaldera's review found the desktop installer crashing on startup:

```
Error: Cannot find module './editor-window-close-guard'
Require stack:
- /Applications/eXeLearning.app/Contents/Resources/app.asar/main.js
```

That was a packaging bug, not a problem with the fix itself. `app/main.js` requires the new `editor-window-close-guard.js` module, but the electron-builder `files` allowlist in `package.json` enumerated the `app/` modules one by one, so the new file was never copied into `app.asar`. `make run-app` loads `app/` straight from the source tree, which is why it only failed in the installer. The same trap already bit us in #1708 with `save-utils.js`.

`a093d2c4` replaces the enumeration with a non-recursive `*.js` glob (excluding spec/test files), so new root modules are packaged automatically instead of needing a manual registration nobody remembers.

Verified by repacking and listing the asar:

| | `9103fbc83` (before) | `02094bdeb` (after) |
|---|---|---|
| `editor-window-close-guard.js` in asar | absent | present |
| App on launch | error dialog | workarea loads |

## `app/` now runs in CI

While digging, it turned out `test:unit` globbed only `./src ./test/helpers ./scripts`, so **the Electron main-process specs under `app/` never ran in CI** — this PR's 60+ close-guard tests plus the older `save-utils` ones, 77 in total.

`02094bde` adds `./app` to the glob, mirroring what #2217 did for `./scripts`. The per-file coverage gate only parses `.ts` rows, so the `app/` `.js` modules do not affect it; they land in the lcov report at 92% and 99%.

## Follow-up

CI screenshotted that crash dialog and still reported success on all three platforms, so the bug reached review instead of dying in CI. #2226 fixes that blind spot separately.


cristinavaldera (dismissed):
Tested in online version.

Couldn´t test it in desktop version because after generating the PR desktop installer, when trying to open it, it showed this warning:

Image

erseco:
@cristinavaldera thanks for catching this — the crash was a **packaging bug**, not a problem with the fix itself.

`app/main.js` requires the new `editor-window-close-guard.js` module, but electron-builder's `files` allowlist in `package.json` enumerated the `app/` modules one by one, so the new file never made it into `app.asar`. `make run-app` loads `app/` straight from the source tree, which is why it only blew up in the installer. Same trap as #1708 with `save-utils.js`.

Fixed in a093d2c4 by replacing the enumeration with a `*.js` glob, so future root modules get packaged automatically (spec/test files stay out). Verified by repacking locally and listing the asar — `editor-window-close-guard.js` is in there now.

While digging, I also found that `test:unit` globbed only `./src ./test/helpers ./scripts`, so the **77 tests under `app/` were never running in CI** — including this PR's close-guard specs and the older `save-utils` ones. Added `./app` to the glob in 02094bde, mirroring what #2217 did for `./scripts`.

Dismissing the review so a fresh desktop installer gets built. @ignaciogros could you take a look too?

cristinavaldera (approved):
tested in ubuntu and mac!!

ignaciogros (approved):

PR #2208, opened 2026-07-23
fix(link-validator): stop treating browser-blocked fetches as broken links

### LINKED ISSUE #2207: Link validator reports "Failed to fetch" for valid YouTube channel URLs
When running the link validator on a OER, YouTube channel/user URLs such as https://www.youtube.com/@freddcosmos always fail with the message "Failed to fetch", even though the link is valid. The URL works correctly when accessed from the OER, but the validator always reports it as "Failed to fetch".

Example:  [fallo-autor-video-youtube.zip](https://github.com/user-attachments/files/30298565/fallo-autor-video-youtube.zip)

erseco:
## Summary
Fixes #2207. Link validation reported reachable URLs (e.g. YouTube channel pages) as broken in static/desktop mode — and silently reported plain 404s as valid. Each flavor now does the most honest thing it physically can:

| Flavor | Who checks the link | Report |
|---|---|---|
| Cloud / server installs | Backend (`link-validator.ts`) | ✓ / ✗ with real HTTP statuses |
| Desktop app | **Electron main process (new)** | ✓ / ✗ with real HTTP statuses — same as cloud |
| Static web (demos, embeds, `make up-static`) | Nobody — a browser page cannot | ⚠ manual review + up-front notice |

## Why a browser page cannot check links
CORS makes every cross-origin response opaque (`status: 0`): a 200 and a 404 are indistinguishable, and hosts with cross-site redirects (YouTube's consent wall) throw `Failed to fetch` for live and dead pages alike. There is no workaround — Service Workers and Web Workers obey the same rules.

## Changes
- **Electron**: new `app/link-check.js` + IPC `app:checkLink`. The main process (no CORS) checks links with Node's `fetch` (undici): measured from an EU IP, Chromium's `net.fetch` receives Google's cookie-consent interstitial as an inline 200 for existing and deleted channels alike, while undici resolves the consent chain to the real status. `net.fetch` stays as a fallback for proxy-only networks (a 2xx obtained through it is ⚠, since the answering host cannot be attributed). Verified from Spain, Electron 43: `@bbc` → ✓, `@thisisjustanexample` → ✗ 404.
- **Final-host rule**: a 2xx that landed on a different host (consent gates, captive portals, login walls, shorteners) is ⚠ *"Redirects to another site (host)"* — green means the *requested* host answered. An unresolved 3xx is never valid.
- **Validation policy (server + desktop)**: one single `GET` with `Range: bytes=0-0` and `Accept-Encoding: identity` per link — no HEAD (too many hosts reject it, spoof it or hang on it). Body cancelled unread; 2xx/3xx and 416 classify as valid.
- **Desktop never probes local/private addresses** (loopback, RFC1918, link-local/metadata, CGNAT — mirrors `ssrf-guard.ts`): they appear as ⚠ *"Local or private address: open the link to review it manually"*, so an untrusted OER cannot turn validation into a LAN scan. The server keeps its strict SSRF guard.
- **Static web**: no probing at all (zero network traffic). Every external link is ⚠ and the modal explains: *"Due to browser security restrictions, this version of eXeLearning cannot check external links automatically. Open each link to review it manually."*
- **Report UI**: three-state legend (now noting that ✓ means "the server responded" — soft-404s exist), every URL clickable (`target="_blank" rel="noopener noreferrer"`, http(s) only), review counter, CSV export (`link-report.csv`, named after its broken + review content) and repeated URLs checked once per run. Browser-limited flavors stop promising validation: progress reads "Listing links… / Links listed: N to review manually" and the menu entry warns before opening.

## Tests
Colocated unit specs for the Electron checker (policy + private-address detection), adapter, manager, modal and server validator; E2E `link-validation.spec.ts` is mode-aware (server: ✓/✗, no notice; static: ⚠ for reachable *and* dead URLs, notice visible). `make fix`, unit, integration and E2E (chromium + static) green.

## Screenshots
Sample OER from the issue: [fallo-autor-video-youtube.zip](https://github.com/user-attachments/files/30298565/fallo-autor-video-youtube.zip)

### Static web mode — up-front notice, everything reviewable in one click

The desktop app matches the server-mode screenshot (`@bbc` ✓, `@thisisjustanexample` ✗ 404).

## Follow-up
Server-side, inconclusive answers (403/429 anti-bot walls, timeouts) are still reported as broken and could reuse the review state.


mnunezcedec (approved):
Thaks a lot!

ignaciogros (dismissed):
Thanks for this change, @erseco.

The only issue I see is that it assumes those links are valid, even though they may not be. For example: `https://www.youtube.com/@thisisjustanexample`. That URL returns a 404 error, but the tool reports it as a valid link.

It might be better to handle this type of link differently: rather than marking it as valid or invalid, classify it as requiring manual review. Including a link to the URL itself (with `target="_blank"`) in the report would make the review process easier.

What do you think?

Thanks!

erseco:
@ignaciogros you're right, and the problem is actually wider than the example — I've pushed a change that adopts your suggestion.

> **Note:** I have rewritten the PR description and replaced the screenshots, since the original text described the previous behaviour ("treat a resolvable host as valid") that you objected to. Worth re-reading it rather than the version you reviewed.

**What I measured.** In static/offline mode (static builds, embedded editor and the Electron desktop app) the check runs in the browser with `fetch(..., { mode: 'no-cors' })`, which returns an **opaque** response: `type: "opaque"`, `status: 0`. Run from the branch preview in headless Chrome:

| URL | Real status (server-side, follows redirects) | What the browser sees |
|---|---|---|
| `https://www.youtube.com/@Google` | 200 | `TypeError: Failed to fetch` |
| `https://www.youtube.com/@thisisjustanexample` | 404 | `TypeError: Failed to fetch` |
| `https://example.com/this-page-does-not-exist` | 404 | opaque response, `status: 0` |

Two conclusions:

1. A live channel and a 404 channel fail in **exactly** the same way, so the DNS fallback could never tell them apart — your example is the rule, not an edge case.
2. Any plain 404 was already reported as valid before this PR, because an opaque response cannot expose a status code. "Valid" in offline mode has never meant "this link works", only "this link is not obviously dead".

**What the PR does now.** Three outcomes instead of two:

- ✓ **Valid** — the status is actually known (server-side validation, which follows redirects and reads the real code).
- ✗ **Broken** — proven dead: the host does not resolve, the request times out, mixed content is blocked.
- ⚠ **Requires manual review** — the check was inconclusive. This is what a browser-side check now returns for any reachable external URL, including the YouTube case, with a short reason in the Error column.

Every URL in the table is now a link with `target="_blank" rel="noopener noreferrer"` so the review is one click away, there is a legend above the table explaining the three icons, the summary counts them ("Complete: no broken links, 12 to review"), and the CSV export includes the review rows with the status spelled out in its own column.

Nothing changes in online mode: the server follows the consent redirect and reads the real code, so `@thisisjustanexample` is still reported as broken (404) and `@Google` as valid. Both screenshots in the description show exactly your example.

**Follow-ups I'd rather not squeeze into this PR** — tell me if you agree and I'll open the issues:

1. The desktop app is treated as static mode, so it inherits the ⚠ for every external link. Electron can check links from the main process with no CORS restriction at all and get exact statuses, same as the server. That would turn most of those ⚠ back into ✓/✗ for desktop users.
2. Server-side, an inconclusive answer (403/429 anti-bot walls, LinkedIn's 999, timeouts) is still reported as broken. That is the same "needs a human" case and could reuse the new state.

ignaciogros (dismissed):
The Electron and the static version return an error for `https://www.youtube.com/@bbc`, but the Cloud version finds no error (that's right: there's no error).

**Cloud version (correct):**

Image

**Electron and online (`make up-static`) versions (wrong):**

Image

Please confirm if that can be fixed.

Thank you.

erseco:
@ignaciogros answering your "can that be fixed?" — the answer is different for each version:

**Desktop app (Electron): yes — fixed in this PR.** The desktop editor runs inside a browser window, and browser security (CORS) hides the HTTP status of every external site — that is why it could never tell a live channel from a deleted one. The check now runs in the Electron **main process**, where CORS does not apply, so the desktop app reports exactly what your Cloud screenshot shows:

| URL | Desktop app (this PR) |
|---|---|
| `https://www.youtube.com/@bbc` | ✓ valid |
| `https://www.youtube.com/@thisisjustanexample` | ✗ broken (404) |

**Static web version (`make up-static`, demos, embeds): no — it is a browser security limitation, not a bug we can code around.** Instead of guessing, the report now says it up front (⚠ per link, clickable, plus a notice: *"Due to browser security restrictions, this version of eXeLearning cannot check external links automatically. Open each link to review it manually."*). It sends no network traffic at all.

**Cloud: unchanged** — as your first screenshot shows.

Two robustness improvements came out of testing with real content:
- Validation now uses a **single ranged GET** per link. HEAD proved unreliable in the wild: some CDNs answer HEAD with `404` for pages that exist (educa.madrid), and some hosts hang on it.
- The desktop app **never probes local or private addresses** (they are flagged ⚠ for manual review), so an untrusted `.elpx` cannot turn validation into a scan of your local network.

One note on your test: the "wrong" screenshot shows the behaviour *before* this branch (red ✗ with the raw browser error, no legend, URLs not clickable). Please re-test static on the branch preview — https://fix-2207-link-validator-yout.exelearning.pages.dev — and the desktop behaviour on a build that includes this branch. Fresh screenshots of both modes are in the PR description.

ignaciogros (dismissed):
Tested on branch `fix/2207-link-validator-youtube-channel-urls`, from Spain (relevant: Google serves an EU cookie-consent wall), Electron 43.2.0.

## Summary

The direction is right — a browser genuinely cannot read the status of a cross-origin response, so reporting those links as `broken` was wrong. But the desktop path introduces a **false negative that is worse than the bug it fixes**: a dead YouTube channel is now reported as valid. That should be fixed before merging.

## What each flavor does now

| Flavor | Result | Verdict |
|---|---|---|
| Cloud (server validation) | `@bbc` valid, non-existent handle broken | Correct |
| Static (`make up-static`) | Every external link → amber "Not checked automatically" | Correct by design, see note below |
| Electron (`make run-app`) | **Both** YouTube links → valid | **Bug** |

## The Electron false positive

`net.fetch` from the main process never reaches the channel page: it gets Google's cookie-consent interstitial, served with **HTTP 200**. Reproduced with a standalone Electron main process using the exact headers from `app/link-check.js`:

```
https://www.youtube.com/@bbc                 -> status=200  <title>Antes de ir a YouTube</title>   (consent page)
https://www.youtube.com/@thisisjustanexample -> status=200  <title>Antes de ir a YouTube</title>   (consent page)
https://github.com/nonexistent-repo-xyz123   -> status=404   correct
https://es.wikipedia.org/wiki/no-existe-xyz  -> status=404   correct
```

Both YouTube URLs return the *same* 200 whether the channel exists or not, so `classifyHttpStatus()` reports valid in both cases. With `curl` the chain resolves differently — `302 → consent.youtube.com → 303 → www.youtube.com?ucbcb=1 → 404` — which is why the cloud version (requesting from a datacenter IP, no consent wall) gets the right answer.

Aggravating detail: in Electron 43, `net.fetch` returns `redirected=false` and an **empty `response.url`**, so the current code cannot even tell that the request was redirected to another host.

## Requested changes

1. **Follow redirects manually and check the final host**, the way `safeFetch` already does server-side. If the final URL's host is not the requested one (consent gates, captive portals, login walls), report `unknown` → "requires manual review", not `valid`. Green must mean *the requested host answered 2xx*, not *someone answered 2xx*.
2. **`classifyHttpStatus()` treats every 3xx as valid.** With manual redirect following this should only apply to a resolved chain; an unresolved 3xx does not prove the target exists.
3. **Test coverage does not exercise the failing case.** `LinkValidationAdapter.test.js` asserts `{status:'broken', error:'404'}` for `@thisisjustanexample` against a mocked `checkLink`, so it passes while the real behavior is the opposite. Add a case for "redirected to a different host → unknown".
4. Optional, not recommended as the fix: sending `SOCS=CAI` / `CONSENT=YES+` bypasses the wall (verified: `curl` with those cookies gets a direct 404). It is Google-specific and does not address the general problem.

## Soft-404s — worth documenting, not a bug

`https://exelearning.dev/no-such-page-here` returns **200** (confirmed with curl too), so status-code checking cannot classify it. The legend should be explicit that a green check means "the server responded", not "the page exists".

## Static mode — keep the tool, relabel it

Reporting every external link as amber is the only honest outcome in a plain browser, and the notice added by this PR (`isBrowserLimited()` + `createBrowserLimitedNoticeHtml()`) is the right call. Disabling the tool would be worse: the extraction still produces the valuable part — a full inventory of links with page, block, iDevice, occurrence count, clickable URLs and CSV export.

What should change is the wording, so the feature promises what it delivers in that flavor:

- The menu entry and progress text still say "Link validation" / "Validating…" / "Complete" when nothing is validated.
- The export is still named `BrokenLinks.csv` although no row is broken.
- The notice appears only after opening the dialog and extracting; anticipating it (menu tooltip or modal header) avoids the wait for a result that never comes.

## Note on the "Failed to fetch" report in the issue

That message cannot come from this branch — the new adapter no longer calls `fetch` and falls back to "Not checked automatically". It is the old adapter's Chromium error, i.e. a build without these changes (the PR commits are only on `origin/fix/2207-link-validator-youtube-channel-urls`; the local `2207-…` branch sits at `main`). Changes to `app/main.js` and `app/preload.js` also require a full Electron restart. Quick check in DevTools: `typeof window.electronAPI.checkLink === 'function'`.

erseco:
@ignaciogros thank you for this review — the standalone repro made the fix straightforward. All points addressed:

1. **False positive fixed, and with real statuses rather than ⚠.** I reproduced your experiment and went one step further: `redirect: 'manual'` is not even usable in Electron 43's `net.fetch` ("Redirect was cancelled"), but **Node's `fetch` (undici) from the same main process resolves the consent chain** (`302 → consent.youtube.com → 303 → ?ucbcb=1`) to the real page status. The checker now uses undici as its primary stack, keeping `net.fetch` only as a fallback for proxy-only networks — and a 2xx obtained through it is reported ⚠, since its empty `response.url` means the answering host cannot be attributed. Measured on this branch (Electron 43, from Spain):

   | URL | Desktop app |
   |---|---|
   | `https://www.youtube.com/@bbc` | ✓ valid |
   | `https://www.youtube.com/@thisisjustanexample` | ✗ broken (404) |
   | `https://youtu.be/…` | ⚠ Redirects to another site (youtube.com) |

2. **Final-host rule implemented** (with undici's reliable `response.url`): a 2xx that landed on another host is ⚠ with the host named; an unresolved 3xx is never valid.
3. **Tests now exercise the failing case at policy level**: consent-like cross-host 200 → ⚠, resolved consent chain → real status, proxy-fallback cases, redirect loops.
4. **Soft-404s documented in the legend**: "A check mark means the server responded: some sites answer 200 even for pages that do not exist."
5. **Browser-limited flavors relabelled**: progress reads *"Listing links…" / "Links listed: N to review manually"* (no validation language), the menu entry warns about the limitation before opening the dialog, and the export is now `link-report.csv` in every flavor, named after what it contains.

ignaciogros (commented):
Perfect. Thank you.

I've pushed a few small presentation-related changes:

* Removed the yellow background when the limitation is due to the browser (static online version). In that case, there's already a yellow warning message at the top. I tried to make presentation lighter.
* Added CSS so that links inside tables with yellow or red backgrounds have enough contrast.

Please tell us if you're OK with the changes.

erseco:
Hi @ignaciogros I agree with the changes, much better now, thanks!

Captura de pantalla 2026-08-05 a las 12 25 38

ignaciogros (approved):

Issue #490, opened 2025-10-21
Check all hidden options in the main menu


ignaciogros:
Some options, like "Export metadata" or "SCORM 2004" are currently hidden, but there's code related to them.

After releasing v3.0.0 we should check (and maybe remove part of) that code.

Look for the `d-none` CSS class in the *.twig files.

ignaciogros:
After reviewing it with the Development Team, it has been decided to remove SCORM 2004 from the code.

PR #2209, opened 2026-07-24
refactor(scorm): rewrite SCORM 1.2 runtime with clean provenance and regression tests

erseco:
## Summary

Replaces the two SCORM 1.2 files every exported package carries — `libs/SCORM_API_wrapper.js` and `libs/SCOFunctions.js` — with a clean-provenance implementation.

**It is a drop-in replacement.** Same filenames, same public functions, same score arithmetic. What it does differently is stop telling the LMS things that were never true.

## Why

The legacy `SCOFunctions.js` descends from ADL's SCORM 1.2 sample code distributed under the Concurrent Technologies Corporation (CTC) licence; that grant — the only one the file ever carried — was removed on 2013-03-21 (iteexe `15b55411c`), leaving a warranty disclaimer and a CC BY-SA 3.0 notice that refers to a file that does not exist. The bundled wrapper is a modified pipwerks 1.1.7 fork (~970 added lines) with three functions attributed to Rustici under CC BY 3.0 US (added 2013-04-29, iteexe `78125d127`) under a header that declares MIT for the whole file. Nothing in either file is proven *incompatible* with AGPL-3.0; the point is that their provenance and licence terms cannot be reliably established or defended for AGPL-3.0 distribution, and the files are small enough that a spec-based rewrite is cheaper than establishing the status of each line.

Writing the contract down first surfaced a set of real defects. They are fixed here.

## What a learner's teacher sees

Measured on a real project with **33 iDevice types over 51 pages**, exported by each revision's own exporter and played in a live Moodle 5.0.7 `mod_scorm`.

**Opening all 51 pages and touching nothing:**

| what the package sends the LMS | before | after |
|---|---|---|
| `cmi.core.score.raw` | **36 pages** | 0 |
| `cmi.core.lesson_status = "failed"` | **36 pages** | 0 |
| `cmi.core.lesson_status = "not attempted"` | **62 writes, every one rejected by Moodle** | 0 |
| `cmi.core.lesson_status = "incomplete"` | 51 | 51 |
| `cmi.core.score.min` / `.max` | 45 | 45 |

**36 of the 51 pages came back with a 0 and a fail just for being opened.**

**And the full factorial matrix**, on packages built from clean worktrees: 25 declared scenarios × 2 producers × 2 hosts (`mod_scorm`, `mod_exescorm`) × 4 Moodle grading methods = **400 cells per browser engine**, judged against a hand-written oracle. This runtime: **1224 pass, 0 fail** in Chromium and the same in Firefox. The old one: 401 failures in Chromium, 460 in Firefox. Every cell records the SHA-256 of the package it used, of each runtime file inside it, the producing commit and the Moodle release.

The two engines matter here because the end of session moved from `unload` to `pagehide`. Of the 2576 paired assertions, 59 differ between Chromium and Firefox — **all 59 belong to the old runtime**, all of them the same rejected `not attempted` write. This runtime behaves identically in both.

**Then playing every gradable iDevice with an identical click script.** 48 of the 51 pages are click-for-click comparable; the other three shuffle their own content on every load. On those 48:

- **no real score is lost** — 0 scores lost, 0 gained, 0 with a different value;
- the 16 scores that only the old runtime produces are **all zeros it seeded itself**;
- `passed` 2 = 2, `failed` 5 = 5, `score.min`/`.max` written on 43 = 43 pages;
- 27 pages move from `failed` to `incomplete` — activities the learner started and did not finish, which the old runtime reported as a fail;
- 14 move from a rejected `not attempted` to `incomplete`.

## The defects this fixes

| | before | after |
|---|---|---|
| **Opening a page** | an activity the learner never touches is seeded at 0, so the page reports `score.raw = 0` and `failed` on load | nothing is reported until the learner produces a score |
| **Status on load** | `incomplete` → `unknown` → `not attempted`: progress erased on every page view, and Moodle rejects the write | preserved; only `""` / `not attempted` becomes `incomplete` |
| **Half-finished pages** | one activity answered and another untouched still reports `passed`/`failed` | stays `incomplete` until every required activity is finished |
| **End of session** | `onunload`/`onbeforeunload` — unreliable on mobile, and they disable the back/forward cache | `pagehide`/`pageshow`/`visibilitychange`; zero unload handlers anywhere in the package |
| **Termination** | a boolean: a failed `LMSFinish` looked successful on the next call | a state machine; a failed finish stays failed and is never retried |
| **Score bounds** | one failing optional `.min`/`.max` discarded the whole write | the mandatory `.raw` is tracked separately, and the bounds have a single owner, so content and runtime cannot each send their own copy |
| **Resume** | game iDevices register before `loadPage()`, so a resumed attempt could overwrite a restored score with 0 | LMS writes wait until the entry policy has run |
| **Legacy getters** | `GetExit()`, `GetSessionTime()`, `SetMode()` issued calls SCORM 1.2 forbids | answered from a local cache, or documented no-ops |
| **API binding** | preferred `API_1484_11`, so a 1.2 package could bind a SCORM 2004 API | pinned to SCORM 1.2 before discovery |
| **Missing runtime** | silently shipped a different, broken inline runtime | the export fails loudly |

Example of what an LMS now sees for a page with one quiz the learner completes:

```
LMSInitialize("")
LMSSetValue("cmi.core.lesson_status", "incomplete")   ← entry, no longer "not attempted"
… learner answers …
LMSSetValue("cmi.core.score.raw", "80")
LMSSetValue("cmi.core.lesson_status", "passed")       ← only now, not on the first answer
-- page hidden --  session_time + LMSCommit           ← never LMSFinish
-- page closed --  LMSCommit + LMSFinish, exactly once
```

## What stays exactly the same

- Both filenames. iDevices and the Moodle plugin hard-code them.
- Every documented global and facade member: `loadPage`, `unloadPage`, `doQuit`, `setScore`, the `scorm` object, the eXe extension methods on `pipwerks.SCORM`.
- The score aggregation. Same weighting, same rounding, to the decimal. Same 50 % pass threshold.
- **SCORM 2004** keeps the legacy pair untouched. Its packages do pick up the shared `unload` → `pagehide` change, which is inert for them because their `onunload` attributes already kept them out of the back/forward cache.

On a SCORM 1.2 page the activity registry is the **single owner of `cmi.suspend_data`**: the gamification helpers in `common.js` route through it, the previous line format is migrated on entry (score inherited, completion decided by the live iDevice — the old format never stored it), and the registry is persisted on `hidden`/bfcache entry too, not only at exit.

## Content changes

Thirty-two iDevice export runtimes were touched, and only two kinds of change:

- `unload`/`beforeunload` → `pagehide` in every export runtime that had one. This also fixes a typo in `flipcards`, which registered its handler under one name and removed it under another, so it was never removed.
- Six iDevices — `beforeafter`, `challenge`, `complete`, `flipcards`, `map`, `scrambled-list` — now say when their activity is finished. They already knew; they just never told the runtime, so a page carrying one stayed `incomplete` in the LMS even at 100 %. Each fix sets the flag where the activity's own end condition is already computed, never on an intermediate report.

Measured, on the same 51-page project: pages that end with a real score but no final status — the signature of "the activity never says it finished" — go from **2** to **0**. (`main` scores 0 there only because it calls everything `passed`/`failed` as soon as a score exists.)

For the record, because the fix is easy to over-read: an activity is also finished when the learner presses its own send-score control (32 of the 35 gradable iDevices ship one), which is why the exposure was limited to learners who just play the activity through. What these six now do is report their end condition **on their own**, without requiring that press.

No scoring logic in any iDevice was changed.

## Which runtime is this

The assembled `libs/SCOFunctions.js` is now stamped with the eXeLearning version that produced it, both as a header line and as `exeScorm12.runtimeVersion`:

```
 * eXeLearning-SCORM12-Runtime: 0.0.0-alpha
```

There is one runtime per eXeLearning version. The stamp is what lets another project — `moodle-mod_exelearning` above all, which vendors the same file — say which one it carries and prove it has not drifted from the release it claims to match. Paired with exelearning/moodle-mod_exelearning#105.

## Licensing

`libs/SCORM_API_wrapper.js` is the **unmodified upstream pipwerks wrapper**, byte-identical to commit `82e455b` (MIT, as stated in its header, readme and package.json — the upstream repository has no separate LICENSE file), vendored and pinned by SHA-256 in `THIRD-PARTY-NOTICES.md`; its own header acknowledges that the API-finder functions derive from ADL/Rustici code. `libs/SCOFunctions.js` is assembled from five project-owned AGPL-3.0-or-later layers under `public/app/common/scorm/scorm12/`, written from the SCORM 1.2 RTE specification and sharing no code with the legacy files.

The legacy pair remains in the repository and is still shipped by the **SCORM 2004** exporter; its provenance is not addressed here (see the SCORM 2004 decision below).

Contract: [`doc/development/scorm12-runtime-contract.md`](doc/development/scorm12-runtime-contract.md) · Decisions: [ADR-2209-01](doc/architecture/adr/ADR-2209-01-scorm12-runtime-rewrite.md), [ADR-2209-02](doc/architecture/adr/ADR-2209-02-scorm12-activity-completion-registry.md).

## Two engines

The 51-page project was played in a live Moodle under **both Chromium and Firefox**, because the move from `unload` to `pagehide`/`visibilitychange` is exactly the kind of change that can behave differently in Gecko. Opening all 51 pages: 48 of 51 produce a byte-identical write sequence, and the three that differ differ only in the **order** of the same three writes (the entry status before the score bounds instead of after) — the CMI the LMS ends up holding is identical. The old runtime is the one that varies by engine: Gecko draws one extra rejected `not attempted` out of it (63 instead of 62).

## Tests

`make fix` clean; `make test-unit`, `make test-frontend` and `make test-integration` green; SCORM 1.2 runtime line coverage above 99 %.

New: a strict LMS simulator with conformance profiles that validates access rules, vocabularies, ranges, array indices and error codes; an activity matrix; seeded event-sequence invariants; a scanner proving no `unload`/`beforeunload` survives in a real exported ZIP; and E2E tests that run a real exported SCO in an iframe against a strict parent `window.API`, including suspension and resume through `cmi.suspend_data` and a register-before-`loadPage` resume that must not overwrite a restored score with 0.

Every iDevice fix ships with a colocated test that was verified to fail without its line.

Compact fixture used by that last path and by Moodle: [`test/fixtures/scorm12/resume-race.scorm.zip`](test/fixtures/scorm12/resume-race.scorm.zip) (regenerate with `bun test test/helpers/scorm12-resume-package.spec.ts`).

Live Moodle check (alpine-moodle, not CI):

```bash
MOODLE_URL=http://localhost MOODLE_USERNAME=user MOODLE_PASSWORD=1234 \
  bun x playwright test --project=chromium test/e2e/playwright/specs/scorm12-moodle.spec.ts
```

Skipped unless `MOODLE_URL` is set.

## Not covered here

- The score aggregation turns an authored weight of 0 into 1, so a zero-weight activity counts. **This PR preserves that deliberately.** The old runtime does it as `parseFloat(activity.weighted) || 1` then `clamp(…,1,100)`; this one writes the same rule explicitly as `weight > 0 ? weight : 1`. A zero is out of the authoring contract anyway — the editor's control is `<input type="number" min="1" max="100">` and clamps on keyup — so it can only arrive from legacy content, an import or a hand edit. Whether an explicit zero should instead exclude the activity from the aggregate (as Moodle Quiz does for `maxmark` 0 and SCORM 2004 for `objectiveMeasureWeight` 0) is a separate product decision: it would change marks in existing packages, and it is not what a runtime replacement should decide.
- Four gradable iDevices still never signal completion **from their automatic reports**: `padlock`, `identify`, `interactive-video`, `rubric`. Three of them — `identify`, `interactive-video`, `rubric` — do finish the activity when the learner presses their own send-score control, because a manual report completes it whatever the flag holds, so their exposure is a learner who never presses it. Their real end condition is a product question (what finishes an interactive video, or a rubric the learner can keep adjusting?), so they are left for a follow-up.
- **`padlock` is a separate, older defect and is not fixed here.** `padlock.js:170` is the only assignment to `gameStarted` and it writes `false`, and `gameOver` is never assigned, so the guard in `common.js` rejects every report it makes — automatic or manual. It can publish no score at all, in this runtime or the previous one. Worth its own issue.
- Five gradable iDevices (`padlock`, `identify`, `interactive-video`, `rubric`, `select-media-files`) were not reached by the measurement. Reading their source says they behave, but they are not measured.
- SCORM 2004 packages were not run against an LMS; the 2004 path is covered as a regression control only.
- Real bfcache entry/restore is unreachable in Playwright, so the E2E harness dispatches `PageTransitionEvent`s with the real `persisted` flag and the semantics are asserted at unit level. WebKit is not configured in this repo.

## Refresh after the integration audit (2026-08-27)

Merged `main` (13 commits; the `beforeafter`/`map` test suites that #2279 and this branch had created independently are folded into one file each) and applied the audit's required fixes:

- **Hide handlers (correctness).** Three iDevices kept a `pagehide` handler from the `unload` era: `az-quiz-game` finished the game on hide (a learner leaving mid-rosco was reported `failed` with a durable completion flag, on the bfcache path this lifecycle made reachable), and `trivial`/`trueorfalse` sent a score on hide, racing the runtime's own persistence. Deleted, with a test per iDevice.
- **Session ownership.** `exeScorm12.session.open` records who owns the page lifecycle on the first successful open (first caller wins), and `applyEntryPolicy` is idempotent — a host that declined ownership (the Moodle plugin) is no longer overridden by a later `loadPage()`, and the entry policy never runs twice.
- **Docs = code.** The contract, ADR-2209-01/-02 and the header comments no longer describe a four-layer Moodle subset (the plugin vendors all five layers and parses the versioned `exe12/` format); the bounds-once-per-session rule, the injector change, the downgrade rule and the hand-over order now match the implementation; §9.2 states what an **empty score field** means today (counted at 0 with full weight in the SCO's aggregate — whether a consumer grading per item must mirror that is an open product decision, see below).
- **Second pass.** The 23 remaining `pagehide` handlers whose only effect was the no-op `endScorm` are gone (the six that persist state or tear down viewers stay); `exe_export.test.js` pins the legacy bridge (`unloadPage()` once on a real `pagehide`, nothing on a bfcache freeze) and the scored-flag hand-over before `loadPage()`, and its harness no longer leaks window listeners between tests.
- **Hygiene.** `/api/resources/libs/scorm` no longer lists `*.test-util.js`; the export and integration specs assert all five layer markers; `test/fixtures/scorm12/resume-race.scorm.zip` is built deterministically and regenerated from this head.

Open product decisions this PR deliberately does not take (tracked for the team): the weight of an evaluable activity with an empty score (SCO counts it at 0; mod_exelearning #105 currently drops it and recomputes), the status semantics while a learner is working (`incomplete`-until-finished here vs a live verdict), the order-dependent aggregate (kept for parity; see ADR-2209-02 §9), and whether SCORM 2004 remains a supported export target (#490).

## Rollback

A single `git revert` restores the previous behaviour: the legacy files were never modified, and already-exported packages carry their own runtime copy.


ignaciogros (approved):
Thank you very much, @erseco.

As agreed after reviewing the report we received, we will merge the different related PRs in this order, following your recommendations:

1. #2209: new SCORM 1.2 runtime.
2. #2299: independent contract tests for that runtime.
3. #2310: tests with real packages, grading traces, and validation against Moodle.

Once the changes have been merged into `main`, @mnarvaezm will make usability improvements in line with the standard, regarding when and how grades are saved in the LMS.

PR #2215, opened 2026-07-28
Fix Slide iDevice text cursor drift

### LINKED ISSUE #2214: Cursor drifts out of position when typing text in the "Slide" iDevice
When using the **Slide** iDevice and inserting text, the cursor is placed correctly when you start typing. However, as you keep typing, the cursor gradually **drifts out of position** and ends up **before the position where it should be**. The longer the text, the more evident and pronounced the offset becomes.

This makes editing very difficult: because the visual position of the cursor no longer matches its real position within the text, it is hard to tell which letter or character the cursor is actually acting on.

## Steps to reproduce

1. Open eXeLearning (tested on https://online.exelearning.net/).
2. Create or edit a resource and insert the **Slide** iDevice.
3. Insert a text inside the slide.
4. Start typing and keep adding more and more text.
5. Observe the cursor position relative to the character where you are actually typing.

## Expected behavior

The cursor should always stay aligned with the real typing/editing position within the text, regardless of the length of the content.

## Actual behavior

The cursor progressively shifts and appears **before** the correct position. The offset grows with the length of the text, making it hard to identify which character the cursor is acting on while editing.

## Environment

- **Platform:** eXeLearning Online (https://online.exelearning.net/)
- **iDevice:** Slide
- **Browser:** _(specify browser and version, e.g. Chrome 126, Firefox 127)_
- **Operating system:** _(specify OS and version)_

## Additional information

- The issue is consistently reproducible.
- The offset is proportional to the length of the text: the more text there is, the larger the cursor displacement.
- In the attached screenshot you can see how, in the word "probar", the cursor appears right on top of the "a", when it should be placed after the "a".

Image

erseco:
## Summary

- Keep the Slide text caret aligned with the actual editing position at scaled canvas sizes.
- Add regression tests covering long text at non-default canvas scale (Chromium and Firefox), per-character styled font sizes, and shared-Fabric cache invalidation.

## Root cause

The app ships Inter as a variable font with an `opsz` (optical sizing) axis, and browsers apply automatic optical sizing in canvas 2D, so glyph advances are non-linear in font size. Fabric.js measures text once at `CACHE_FONT_SIZE` (400 px, where auto optical sizing clamps `opsz` to the axis maximum) and scales linearly down, while glyphs are painted at the character's real font size, where a wider optical instance applies below that maximum. The caret, selection and hit-testing geometry are built from the measured advances, so each typed character added a constant fractional error (6–10 % per advance at sizes below 32 px) and the caret drifted further left as the text grew.

## Fix

Text objects on the Slide canvas get a per-instance `_measureChar` override that measures each character with the exact font declaration it is painted with (its own `charStyle.fontSize`, no `CACHE_FONT_SIZE` scaling), so measured and painted advances resolve the same optical size in every browser. Cached widths carry the size in the style/weight segment of Fabric's cache key, keeping the top-level family key — and the documented `cache.clearFontCache(fontFamily)` invalidation — intact. The shared `window.fabric` instance is never mutated; objects outside the Slide canvas keep stock behavior.

## Tests

- `bun run bundle:slide-editor` — passed
- `bun x playwright test --project=chromium test/e2e/playwright/specs/idevices/slide.spec.ts` — 13 passed (includes new caret, styled-span and cache-invalidation regressions)
- `bun x playwright test --project=firefox test/e2e/playwright/specs/idevices/slide.spec.ts` — passed (caret regression verified on Firefox)
- `bun x playwright test --project=chromium test/e2e/playwright/specs/idevices/` — 188 passed, 1 skipped, 1 unrelated flake (`download-source-file`, passes in isolation)
- `bun test ./src ./test/helpers` — 7496 passed
- `bun test ./test/integration` — 721 passed
- `make test-frontend` — 13717 passed, 1 pre-existing flake (`focusedEditMode.test.js`, passes in isolation, unrelated)
- `make fix` — clean (2 pre-existing warnings in an unrelated helper)

Each regression test failed against the code it guards: the caret test with a ~56 design-px divergence on a 94-character line, the styled-span test with a ~250 px corrupted line width, and the cache test with size-suffixed family keys surviving `clearFontCache`.

## Evidence

| Scenario | Screenshot |
| --- | --- |
| Before (caret several glyphs early, tail clipped) | before |
| Fit zoom (51 %) | fit zoom |
| 100 % zoom | 100 percent |
| Multiline, caret mid-text | multiline |
| Firefox | firefox |

Closes #2214


greynoa:
Thanks @erseco! I tested the preview build with my original repro: a long text in the Slide iDevice using Inter at 32px with the canvas scaled to 50%. The caret no longer drifts — the cursor and the selection highlight now match the real editing position, even for characters deep into a long line. The bug is fixed on my end!

ignaciogros (approved):

PR #2216, opened 2026-07-28
fix(media): set referrerpolicy on YouTube/Vimeo iframes (Error 153)

erseco:
## Problem

YouTube's embedded player returns **Error 153** (`embedder.identity.missing.referrer`) when it
cannot read the HTTP `Referer` header — for example when the embedding host page sends a
restrictive `Referrer-Policy` (`no-referrer` / `same-origin`) and the iframe carries no
`referrerpolicy` attribute of its own to override it.

eXeLearning made this worse from both sides: the collaborative sanitizer **stripped**
`referrerpolicy` (it was missing from DOMPurify's `ADD_ATTR`), and the export/preview renderer
never **added** it. Content authored in eXe therefore hit Error 153 in the wild once embedded in
an LMS — and WordPress **Jetpack**'s embed handling strips the attribute too.

## Fix

- **Sanitizer** (`public/app/utils/sanitizeHtml.js`): preserve `referrerpolicy` by adding it to
  `COLLABORATIVE_HTML_CONFIG.ADD_ATTR`, so an author-supplied attribute survives a round trip.
- **Export/preview** (`IdeviceRenderer.addReferrerPolicyToEmbeds`): stamp
  `referrerpolicy="strict-origin-when-cross-origin"` onto YouTube/Vimeo iframes that lack it.
  Applied in `renderIdevice` right after `escapePreCodeContent`, so it covers pre-existing
  content and every export format at once.
- **Docs**: new "YouTube Error 153 and `referrerpolicy`" section in
  `doc/development/embedding.md`, including the host-side `Referrer-Policy` response header
  integrators should set and the Jetpack caveat.

Only the two providers that actually need the Referer are matched — YouTube (including
`youtube-nocookie.com` and `youtu.be`) and Vimeo. Matching parses the URL and compares the
**hostname**, accepting an exact host or a subdomain, so a lookalike such as
`vimeo.com.example.org` is not treated as the provider. Non-provider iframes are left untouched,
and an existing `referrerpolicy` is never overwritten.

Two details worth calling out, since naive versions of this get them wrong:

- The `<iframe` presence check is case-insensitive, matching the case-insensitive replacement.
  HTML tag names are case-insensitive and pasted embed codes vary, so `<IFRAME SRC=...>` must be
  handled too.
- The tag pattern consumes quoted attribute values whole
  (`/<iframe\b(?:"[^"]*"|'[^']*'|[^"'>])*>/gi`) instead of using `[^>]*`, which would stop at a
  `>` inside a quoted value and corrupt tags like `<iframe src="…" title="1 > 2">`.

## Tests

- 10 cases for `addReferrerPolicyToEmbeds` in `IdeviceRenderer.spec.ts`: adds the attribute,
  covers nocookie/youtu.be/Vimeo, does not duplicate an existing value, leaves non-provider
  iframes alone, no-op for content without iframes, uppercase tags and attributes, `>` inside a
  quoted attribute value, lookalike hostname, protocol-relative URLs, and relative `src`.
- 2 new assertions in `sanitizeHtml.test.js`: the attribute survives sanitising, and the exported
  config lists it.

`make fix`, `make test-unit` (7506 pass, coverage gate green on 221 files) and
`make test-integration` (721 pass) are all green.

## Known limitations

- An iframe `src` written without a scheme and without a leading `//` (a bare
  `www.youtube.com/embed/x`) no longer matches, because matching now resolves the URL to compare
  hostnames. Such a value is a document-relative path, so a real browser would not load the
  provider from it either.
- Against already-invalid HTML — an unterminated quote inside an attribute — the tag pattern can
  absorb text past the intended tag. This is a strictly narrower failure mode than the `[^>]*`
  pattern it replaces, and DOMPurify never emits HTML shaped like that.


cristinavaldera:
I´ve tested this PR with a youtube video embed in a text idevice. 

- In edition it shows ok
- In Preview it shows ok
- In html exportation (offline) it shows error 153
- Inserted in Moodle with the plugin exeScorm and as scorm activity it shows ok.

I´ve made the same tests with exe 4.0.2 and it works exactly the same way. Maybe I havent understood whsata this PR fixes ??

erseco:
@cristinavaldera, this fix targets the web-hosted case: for example, an eXeLearning export embedded in WordPress or another website whose hosting or security layer suppresses the `Referer` header, often through an automatically applied WAF or restrictive `Referrer-Policy`.

The local `file://` Error 153 is a different case and cannot be solved from the export itself, since there is no valid HTTP referrer. The content must be served over HTTP(S).

ignaciogros (approved):

PR #2219, opened 2026-07-29
feat(slide): keyboard undo/redo, drag-to-select marquee and arrow-key nudge in the editor

### LINKED ISSUE #2218: Slide iDevice: usability improvements for the editor (undo/redo keyboard shortcuts, drag-to-select, move with the keyboard)
While building a cover page with the **Slide** iDevice, I ran into a few editing limitations that make the tool harder to use than it needs to be. None of them are bugs exactly — the editor works — but fixing them would make the everyday experience much smoother, especially for teachers who are not used to design tools. I believe they are also relatively small, self-contained changes.

This report groups three main requests plus a couple of smaller related suggestions.

## Environment

- **Platform:** eXeLearning Cloud (`cloud.exelearning.dev`)
- **Build tested:** `v0.0.0-alpha-build202607281245`
- **iDevice:** Slide
- **Browser:** Chrome (desktop)
- **Note for developers:** the Slide editor is built on the Fabric.js canvas (version 7.4.0). The three requests below map to standard capabilities of that library, so they should be feasible without a major rewrite.

---

## 1. Undo/redo keyboard shortcuts do not work inside the Slide editor

**What happens now**

The undo and redo **arrows in the toolbar work correctly** (moving an element and then clicking the undo arrow puts it back). The problem is the **keyboard shortcuts**: pressing Ctrl + Z / Cmd + Z (undo) or Ctrl + Y (redo) while editing a slide does **not** undo or redo anything — the shortcuts appear to be ignored by the slide editor.

In some situations this message is also shown instead:

> "Hay cambios sin guardar. Guarda el iDevice antes de continuar."
> ("There are unsaved changes. Save the iDevice before continuing.")

So the shortcuts seem to be captured by a different, page-level handler instead of the slide editor's own undo/redo.

**What I would expect**

- While the slide editor is focused, Ctrl/Cmd + Z (undo) and Ctrl/Cmd + Y or Ctrl/Cmd + Shift + Z (redo) do exactly the same as the toolbar arrows.
- The unsaved-changes message does not interrupt these shortcuts.

**Why it matters**

Almost everyone uses these shortcuts by reflex. Having them do nothing — and pop up an unrelated warning instead — is confusing and slows people down, even though the feature itself already exists in the toolbar.

**How to reproduce**

1. Open the Slide iDevice and make any change (move, add or delete an element).
2. Press Ctrl + Z / Cmd + Z (or Ctrl + Y).
3. Notice that nothing is undone/redone. In some cases the message "Hay cambios sin guardar. Guarda el iDevice antes de continuar." appears instead.
4. For comparison, click the toolbar undo/redo arrows: those do work as expected.

---

## 2. Cannot select several elements by dragging the mouse

**What happens now**

I can select more than one element (for example, with Shift + click), but I cannot draw a selection box by clicking on an empty area and dragging the mouse over several elements at once. When I try, no selection rectangle appears; instead the drag grabs and moves whatever object is under the pointer (for instance a full-slide background shape), which can move things by accident.

**What I would expect**

Clicking on an empty part of the slide and dragging should draw a rectangle ("rubber-band" / marquee selection) and select every element inside it, so I can move or align a group in one go.

**Why it matters**

Selecting elements one by one is slow and error-prone when a slide has many pieces (shapes, texts, icons). Drag-to-select is the fastest, most intuitive way to group things, and most people expect it by habit from other editors.

**How to reproduce**

1. Open a slide that has several elements.
2. Click on an empty area and drag the mouse across them.
3. Notice that no selection rectangle appears and nothing gets selected.

---

## 3. Cannot move elements with the keyboard

**What happens now**

Once an element is selected, I can only move it by dragging with the mouse. The arrow keys do nothing.

**What I would expect**

- With an element selected, the arrow keys nudge it a few pixels in each direction.
- Holding Shift + arrow moves it in larger steps.

**Why it matters**

The mouse is not precise for fine adjustments. Nudging with the arrow keys is the easiest way to line things up exactly, and it is much more comfortable and accessible than trying to place items pixel by pixel with the mouse.

**How to reproduce**

1. Select any element on the slide.
2. Press the arrow keys.
3. Notice that the element does not move.

---

## Smaller related suggestions

These are optional, but they would round off the same set of improvements:

- **Keyboard shortcuts for common actions:** Delete/Backspace to remove the selected element(s), and Ctrl/Cmd + D to duplicate.
- **Text cursor drift when typing:** this is already reported separately, but it belongs to the same "editing comfort" family, so I mention it here for context.

## Priority (from a user's point of view)

If they have to be ordered, **drag-to-select (item 2)** and **move with the keyboard (item 3)** would make the biggest everyday difference, since undo/redo already works from the toolbar; wiring the **undo/redo keyboard shortcuts (item 1)** would remove a common source of confusion.

Thank you for considering these. They would noticeably improve the day-to-day experience of creating resources with the Slide iDevice.

erseco:
Closes #2218

Implements the Slide editor usability improvements reported by @greynoa — thanks for the detailed report; the repro steps made every fix easy to target.

## What changed

1. **Undo/redo keyboard shortcuts work inside the Slide editor.** The project-level (Yjs) handler in `YjsProjectBridge` captured `Ctrl/Cmd+Z`, `Ctrl/Cmd+Shift+Z` and `Ctrl/Cmd+Y` at document level and ran the *project* undo, which popped the "Unsaved changes detected…" modal via `checkOpenIdevice()` whenever an iDevice was open. The handler now yields silently while an iDevice is in edition mode, so the editor's own history (already wired to those keys) owns the shortcuts. The toolbar arrows and the project-level shortcuts outside edition mode behave exactly as before.
2. **Drag-to-select (rubber-band marquee).** Fabric only starts a marquee on an empty pixel, so a cover page with a full-bleed background shape could never start one — the drag moved the background by accident. Now:
   - Holding **Ctrl/Cmd while dragging** always starts a marquee, even on top of objects. Implemented with `skipTargetFind` toggled from a DOM capture-phase listener — Fabric caches the hit target *before* it emits `mouse:down:before`, so a Fabric-event hook is too late.
   - Marquee selection uses **fully-contained semantics** (`selectionFullyContained`, PowerPoint behaviour), which also keeps a full-bleed background out of the selection unless the marquee encloses the whole slide.
   - Plain drags on objects still move them; plain drags on empty areas still marquee.
3. **Arrow-key nudge.** With a selection active, the arrow keys move it 1 px and `Shift`+arrow moves 10 px. A burst of presses coalesces into a single history entry (like a drag), so one `Ctrl+Z` reverts the whole nudge.
4. **`Ctrl/Cmd+D` duplicates the selection** (same action as the toolbar button). `Delete`/`Backspace` already worked.
5. **The navbar (project-level) undo/redo buttons are disabled while an iDevice is in edition mode.** They act on the *project* history, which the iDevice editor owns while open — clicking them used to pop the same "unsaved changes" modal (found while testing the static preview build). A `MutationObserver` on the `mode` attribute re-enables them as soon as the edition closes; it is torn down on `disconnect()`.

The editor's keyboard mapping was extracted into a pure module (`keyboardActions.ts`) so it is unit-testable without a DOM or a Fabric canvas.

## Tests

- 13 new unit tests for `YjsProjectBridge`: the keydown guard (yields while an iDevice is editing: no `undo()`, no `preventDefault`, no modal), the button-disabling logic, the edition-mode observer wiring and its teardown.
- 20 unit tests for the `keyboardActions.ts` resolver (nudge steps, undo/redo/duplicate mapping, crop/text-editing gates, forced-marquee modifier).
- 5 new Playwright specs in `slide.spec.ts`: keyboard undo/redo without the unsaved-changes modal, disabled project undo/redo buttons during edition (re-enabled after saving), 1 px / 10 px nudge, `Ctrl+D` duplicate, and `Ctrl/Cmd`+drag marquee over a full-bleed background (asserts the background is neither selected nor moved).

`make fix`, `make test-unit`, `make test-integration` and `make test-e2e` are green. Patch coverage on the changed `YjsProjectBridge.js` lines is 100 % except the three observer-teardown lines in `disconnect()`, which the lcov converter misattributes — the "tears down the edition-mode observer" unit test proves they execute (the observer mock is called and the reference nulled) while neighbouring lines report 8 hits. The slide TS sources are covered by their colocated Vitest specs plus the E2E flows (frontend coverage scope excludes `public/files/perm/**`).

## Before

`Ctrl+Z` pops the project-level "unsaved changes" modal, dragging over a full-bleed background moves it by accident, and the arrow keys do nothing:

<details><summary>Stills</summary>

</details>

## After

`Ctrl/Cmd`+drag draws a marquee on a fully covered slide and selects the enclosed elements, the arrow keys nudge the selection, and `Ctrl+Z` drives the editor history with no modal. Note the project-level undo/redo arrows in the navbar staying greyed out while the editor is open, and coming back after saving the iDevice (last frame):

Project-level undo/redo while editing vs. after saving:

 → 

<details><summary>Stills</summary>

</details>

---

Note: the text-cursor drift mentioned in the issue is tracked separately in #2214.


erseco:
While verifying this branch on the static preview build we found a **second, previously unnoticed path to the same "unsaved changes" modal**: the project-level undo/redo arrows in the navbar stayed active while an iDevice was open in edition mode, and clicking them ran the *project* undo — whose `checkOpenIdevice()` guard pops the modal. The original issue focused on the keyboard shortcuts (the toolbar arrows @greynoa mentioned as working are the Slide editor's own), so this button path had gone undetected until now.

Fixed in 31e55fe2b: the navbar undo/redo buttons are now **disabled while any iDevice is in edition mode** — the iDevice editor owns the history at that point — and a `MutationObserver` on the `mode` attribute re-enables them as soon as the edition is saved or discarded. Covered by new unit tests and a Playwright spec (buttons disabled during edition, no modal, re-enabled after saving).

The **After** media in the description has been regenerated so it reflects the current behaviour: the project arrows appear greyed out during edition and come back after saving (last GIF frame and the navbar close-ups).

Since this affects every iDevice edition session (not just the Slide editor) and popped a misleading modal, the PR label has been bumped from `low priority` to `high priority`.

greynoa:
Thanks so much for the work on this, @erseco! I tested the branch on the preview deployment and can confirm the improvements work as described:

- Keyboard undo/redo inside the editor no longer triggers the "unsaved changes" warning that used to pop up.

- Nudging elements with the arrow keys works well, and if I press several arrows in a row, a single Ctrl+Z undoes the whole movement at once (no need to undo step by step).

- The undo/redo buttons in the top toolbar stay disabled while I'm editing a slide, and become active again as soon as I save. That makes sense and avoids confusion.

- I also tested holding Ctrl/Cmd while dragging to select over a full-bleed background: it selects without moving or resizing the background, which is exactly what's needed. Without this, it would be impossible to select anything when there's a background in place, so this is a great fix. 

Overall, everything works as expected.

cristinavaldera (approved):

ignaciogros (approved):

PR #2224, opened 2026-08-03
fix(import): tell the author which activities reference missing files

### LINKED ISSUE #2223: Unresolved asset references are shown to the author as raw {{context_path}} placeholders
## Description

When a project references an asset that is not present in the package, the editor shows the author the raw placeholder token instead of telling them a file is missing.

`{{context_path}}/…` is the placeholder the exporter writes for an asset reference. On import, `AssetManager.convertContextPathToAssetRefs` resolves each one against the assets actually extracted from the package and rewrites it to an `asset://` URL. When no asset matches, it logs a console warning and returns the reference verbatim:

```js
// public/app/yjs/AssetManager.js:3670
convertedHTML = convertedHTML.replace(contextPathRegex, (fullMatch, assetPath) => {
    const url = findAssetUrl(assetPath);
    if (url) return url;
    console.warn(`[AssetManager] Asset not found for path: ${assetPath}`);
    return fullMatch;
});
```

Preserving the reference is the right call — the asset may be recoverable, and rewriting it would destroy the only record of what the activity pointed at. The problem is that nothing surfaces this to the author. The only trace is a `console.warn` nobody sees, so the unresolved token ends up rendered in the UI as if it were content.

## Steps to reproduce

1. Import a package whose stored properties reference an asset the package does not contain.
2. Edit the affected activity.

## Actual result

Form fields show the literal placeholder, e.g. `{{context_path}}/rabbit.svg`. Nothing in the interface says a file is missing; the only diagnostic the code offers is the `console.warn` above.

## Expected result

The author is told which activities reference files that are not in the package, in the interface rather than the console. The reference itself must still be preserved — the ask is a notice, not a rewrite.

## Notes

Surfaced while reviewing #2194: a hand-built test package referenced media that had never been bundled with it, and the unresolved tokens showing up in the Classify editor form read as a product defect rather than as a missing file. The reviewer's reaction — "I don't think it's a bug, but it did seem unusual to me" — is the gap this issue is about.

This is the missing-asset counterpart to #2190, which covers the same lack of user-facing feedback for `jsonProperties` that cannot be parsed. Worth solving with one shared notice mechanism rather than two.

erseco:
Closes #2223. Found while reviewing #2194, where a test package referenced media that had never been bundled with it and the unresolved placeholders showing up in the Classify editor read as a product defect rather than as missing files.

## What it does

`{{context_path}}/…` is the placeholder the exporter writes for an asset. On import the handler rewrites each one to an `asset://` URL and, when the package does not carry the file, leaves the reference verbatim — that part is deliberate, since rewriting it would destroy the only record of what the activity pointed at. Nothing surfaced it though, so the token ended up rendered in a form field as if it were content.

The importer now reports the affected activities on its result and the browser raises a dialog naming them. References are untouched: restoring the files restores the activity.

A surviving placeholder is the signal a file is missing, and it is the same signal whether the handler found no match or never ran at all — a package with no assets skips conversion entirely, which is the case that started this.

## How to test

Import the fixture added in this PR: [missing-asset-refs.elpx.zip](https://github.com/user-attachments/files/30662974/missing-asset-refs.elpx.zip) — rename it to `.elpx` first (GitHub does not accept the bare extension). It is a Classify activity whose package carries no `content/resources/` at all, so its eight references have nothing to resolve against.

- On `main`: the import finishes silently; open the activity for editing and the form fields show `{{context_path}}/rabbit.svg` and friends.
- Here: the import raises **Missing files**, naming the `classify` activity and the eight files. Editing still shows the preserved references, which is intended — the point is that you are now told why.

## Tests

- `src/shared/import/unresolvedAssetRefs.spec.ts` — the detector, including the escaped-JSON form the properties arrive in and the bare-placeholder edge case.
- `src/shared/import/ElpxImporter.spec.ts` — the fixture above produces the report; a package whose assets all resolve produces an empty one.
- `public/app/workarea/project/missingAssetsNotice.test.js` — message building, escaping (a crafted package must not inject markup), and the cap on runaway lists.
- `public/app/workarea/project/projectManager.test.js` + `public/app/yjs/YjsProjectManagerMixin.test.js` — the notice is raised from the single funnel every browser import goes through.
- `test/e2e/playwright/specs/missing-asset-notice.spec.ts` — imports the fixture through the real UI and asserts the dialog.

Verified in both directions: disabling the detector fails the importer test, and the E2E fails without the fix.


ignaciogros (approved):
Good improvement! Thanks.

ignaciogros:
It works well, and I think it should be merged so we can release the next version as soon as possible.

As a possible improvement, the iDevice name and location could be localized.

For example, in [classify-content-2166.zip](https://github.com/user-attachments/files/30782347/classify-content-2166.zip), it says that files are missing in "classify", but telling the user that files are missing in **iDevice 1** on the **"Classify content"** page would be more helpful.

PR #2231, opened 2026-08-04
Fix #2227 - Could not edit rubric cells using the "Edit" button.

### LINKED ISSUE #2227: Can't edit a rubric
- Add a rubric.
- Try to use the pencil to edit the values of any of its cells.

You'll se a modal window that can be edited or closed.

ignaciogros:
You could not edit a rubric cell using the "Edit" button because Bootstrap modals must be direct children of `<body>`, at least for now.

How to test: just add a rubric and click on the pencil to edit any cell.

### DIFF
--- a/.agents/skills/idevice/SKILL.md
+++ b/.agents/skills/idevice/SKILL.md
@@ -39,6 +39,18 @@ public/files/perm/idevices/base/<name>/
     └── <name>.css          # Export styles (optional)
 ```
 
+## Edition Lifecycle
+
+The workarea drives `$exeDevice` (the object exported by `edition/<name>.js`) through three calls:
+
+| Hook | Required | When |
+|------|----------|------|
+| `init(element, previousData, path)` | yes | Edition mode starts. Build the form inside `element`. |
+| `save()` | yes | The user saves. Return the HTML/JSON to persist, or a falsy value to block the save after showing a validation message. |
+| `destroy()` | no | This edition class stops being the active one — **save, discard and delete alike**. Tear down anything mounted outside `element`. |
+
+`destroy()` exists because some editors must mount UI outside their own container: a Bootstrap dialog only stacks correctly as a direct child of `<body>`, since the edition box is itself a stacking context (`assets/styles/layout/_idevice-focus.scss`). Such nodes survive the removal of the edition form, so the iDevice has to take them down itself. See `removeEditModals` in `base/rubric/edition/rubric.js`.
+
 ## Testing — Round-Trip Is Mandatory
 
 Every iDevice **must** have a round-trip test: save data → load data → verify all fields are preserved. This is the #1 source of iDevice bugs — fields silently disappear when `loadData()` doesn't restore what `save()` stored. _Example: PR #1581._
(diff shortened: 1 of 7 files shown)


erseco (changes requested):
Thanks for working on this, @ignaciogros. The original cell-editing issue is fixed, but I can reproduce a blocking problem with the row and column editors.

Steps to reproduce:

1. Add or edit a Rubric iDevice.
2. Open the row or column editor.
3. Change any value.
4. Close the editor without saving.

The unsaved-changes confirmation is opened while the rubric modal is still visible, but it appears behind it, making the confirmation buttons inaccessible and leaving the interface blocked.

The global workarea CSS applies:

```css
.modal {
    z-index: 20009 !important;
}
```

This overrides the rubric-specific `z-index: 1060`. Both dialogs therefore have the same z-index, and because the rubric modal is appended later to `<body>`, it is rendered above the confirmation modal.

I also noticed that, after opening a rubric dialog and using **Discard changes**, the body-level rubric modals remain attached to the DOM.

Could you please check both cases and maybe add E2E coverage for:

* The unsaved row and column confirmation being displayed above the rubric modal and remaining clickable.
* Removing the rubric dialogs and backdrops when leaving edition mode through save, discard, or delete.

Thanks!

mnarvaezm:
Hi everyone,

I fixed the Rubric iDevice issue in a separate branch, using a different approach.

The dialogs for editing cells and rows are now part of the Rubric iDevice, rather than application modal windows, so they are no longer driven by the Bootstrap Modal API: that would trap the focus, lock the page scroll and claim `aria-modal` on behalf of the whole document.

Each dialog is mounted inside `#ri_IdeviceForm`, next to a backdrop that covers only that form. The rubric controls underneath stay inert while a dialog is open, and the rest of the work area remains fully available. The panel is centred horizontally on the iDevice and placed over the visible part of the form, which is usually taller than the window. The Esc key goes through the same unsaved-changes check as the Close and Cancel buttons.

Living inside the form also means that the dialog and backdrop are removed along with it when editing ends, so there are no orphaned elements and no cleanup bookkeeping that has to be kept in sync with every exit path.

The branch includes unit and E2E tests covering the scope of the blocking—the rubric is inert, while the rest of the app is not—the centring and the absence of leftover elements after saving, discarding or deleting.

Please have a look and let me know whether you think this is a suitable solution.

Here is the branch: https://github.com/exelearning/exelearning/tree/rubric-hide-modal
@ignaciogros  @erseco

erseco:
Thanks, @mnarvaezm. This approach looks cleaner: it avoids the global modal stacking issue, keeps the blocking scoped to the Rubric iDevice, and removes the orphaned-dialog cleanup problem naturally.

I did notice a few points that may need reviewing:

* Using the main iDevice **Save** button while a row or column dialog is still open may silently lose the pending dialog changes.
* The backdrop blocks pointer clicks, but the controls underneath may still be reachable with the keyboard or assistive technologies.
* Focus does not seem to return to the control that opened the dialog after closing it.

It would be better to open a separate PR from `rubric-hide-modal` so we can track the implementation, tests, and these follow-up points there. Then @ignaciogros can decide which PR should be discarded.

ignaciogros:
I clearly prefer @mnarvaezm's solution.

In fact, for the solution I had proposed (which I believe has already been fixed) I was planning to add a follow-up task to define a mechanism that allows Bootstrap modals to be created from iDevices. I think this should be part of the iDevices API, which is still pending.

I'm closing this PR.

@mnarvaezm,

**Please open a new PR from your branch.**

Thanks!

PR #2238, opened 2026-08-05
Hide empty footer

### LINKED ISSUE #2145: Visible empty footer
Export as SCORM a content with no footer or license using the Universal style, and then open it with your browser.

`#siteFooter` is visible (it's a box), although it has no contents. That may be happening in other styles too.

ignaciogros:
The page footer is now hidden by default when it contains no content or license information.

**How to test:** 

- Choose a style with visible footer, like Flux or Universal.
- Preview or export. The footer should be visible.
- Check that there's no content in the "Page footer" field.
- Check that there's no license (or "Not appropriate").
- Preview or export. The footer should be hidden.

### DIFF
--- a/doc/development/styles.md
+++ b/doc/development/styles.md
@@ -131,6 +131,32 @@ Example:
 
 ---
 
+## Empty Site Footer
+
+Every page ends with `<footer id="siteFooter">`, which holds the license and the
+content of the **Page footer** project property. When a project has neither — no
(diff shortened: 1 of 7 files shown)


erseco (approved):
Approved. I tested it and it works as expected.

As a minor note, an additional test covering the SCORM export itself could have been useful, since that was the original reported scenario.

PR #2240, opened 2026-08-05
fix(rubric): cell dialogs could not be edited, scope them to the iDevice (#2227)

mnarvaezm:
The cell, row and column edition dialogs were mounted inside `#ri_TableEditor`
while their backdrop was appended to `<body>`. The focused iDevice editor forms a
stacking context of its own (`position: absolute; z-index: 5`, see
`assets/styles/layout/_idevice-focus.scss`), so the backdrop, living outside it,
always painted above the dialog: it looked open, but every click landed on the
backdrop and nothing could be edited.

## The approach

The dialogs are **part of the iDevice**, not application modal windows, and they
now behave accordingly:

- Dialog and backdrop are mounted together inside `#ri_IdeviceForm`. Sharing a
  stacking context, the dialog paints above its backdrop and is usable again.
- The backdrop only covers the rubric form. Its controls are inert while a dialog
  is open; the rest of the workarea remains fully available.
- Living inside the form means both are destroyed with it when editing ends: no
  orphaned dialogs, and no cleanup bookkeeping that has to be kept in sync with
  every exit path.
- The panel is centred horizontally on the iDevice and placed over the visible
  part of the form, which is usually taller than the window.

This meant dropping the Bootstrap Modal API, which trapped the focus, locked the
whole page scroll and claimed `aria-modal` on behalf of the entire document:
behaviours that belong to an application modal, not to an iDevice panel. The Esc
key and the unsaved-changes guard are now handled by the iDevice itself, with
`role="dialog"` and `aria-labelledby`.

## Review points addressed

- **Saving with a dialog open no longer loses anything.** The iDevice Save button
  sits outside the form, so it can be pressed while a dialog is open; `save()` now
  writes its pending edits back to the table before serialising.
- **The controls underneath are out of reach of the keyboard and assistive
  technology.** The form is marked `inert` while a dialog is open, since the
  backdrop on its own only stops the pointer.
- **The focus returns to the control that opened the dialog.** The trigger is
  recorded on open and focused again on close, if it is still in the DOM.

## Tests

103 rubric unit tests and 13 E2E tests, all green, along with `make lint`. The E2E
tests cover the dialog stacking, the scope of the blocking (the rubric is
unreachable, the rest of the app is not), the centring, saving with a dialog open,
the focus restoration, and the absence of leftovers when leaving edition through
save, discard or delete.

Each one was checked as a real regression guard: with the corresponding fix
disabled, the test fails.

## How to test

1. Add a Rubric iDevice and click "New rubric".
2. Click the pencil on a cell: the dialog should open and be usable.
3. With the dialog open, check that the rubric responds to neither the mouse nor
   the Tab key, and that the rest of the interface does.
4. Close with Esc: the focus returns to the pencil.
5. Type in a row dialog and press Save without accepting it: the text must appear
   in the saved rubric.


erseco (approved):
LGTM. The fix works as expected.

One minor observation: the edit dialog stays fixed in its original position when scrolling. I’m not sure whether that behaviour is intentional, but it may be worth checking.

Issue #63, opened 2025-06-11
Remove all hidden classification fields


ignaciogros:
Many classification fields are hidden by these lines: 

```
#properties-node-content-form .exe-form-tabs,
.property-row[property="pp_learningResourceType"],
#properties_package_description,
#properties_package_usage {
    display: none !important;
}
```

They should be removed from the code instead of hidden.

Previous information from eXe < 3 will be deleted.

PR #2244, opened 2026-08-06
 SCORM sibling navigation with nested pages

### LINKED ISSUE #2222: Navigation issues in SCORM with nested pages
When using Moodle's standard SCORM activity (not the eXe-specific one), SCORM packages generated with eXeLearning 2.9 allow correct navigation using the buttons at the bottom of the screen. Both the **Previous** and **Next** page buttons, as well as the links to the previous or next page at the same hierarchy level, work as expected.

With SCORM packages generated by eXeLearning 4, this navigation breaks. You can still move to the previous or next page, but the links to sibling pages at the same level do not work.

I've attached two SCORM packages:

* [nested-pages-29-scorm.zip](https://github.com/user-attachments/files/30657259/nested-pages-29-scorm.zip), generated with eXeLearning 2.9
* [nested-pages-scorm-test_scorm.zip](https://github.com/user-attachments/files/30657258/nested-pages-scorm-test_scorm.zip), generated with eXeLearning 4

When using the second package with Moodle's standard SCORM activity player, it is not possible, for example, to use `button#nav_skipnext` to navigate from **"Topic 1"** to **"Topic 2"**.

**Could this be an issue with the navigation structure of the SCORM package generated by eXe?** When using the `mod_exescorm` plugin, `button#nav_skipnext` works correctly in this and other similar cases.

mnarvaezm:
# SCORM sibling navigation with nested pages

`Closes #2222`

## Summary

SCORM and IMS packages exported by eXeLearning 4 left the sibling navigation buttons (`<<` and `>>`) of Moodle's **stock** SCORM player unusable whenever the project had nested pages. Packages exported by eXeLearning 2.9 worked fine.

The cause is the shape of the `<organization>` element in `imsmanifest.xml`. This PR wraps every page in a non-launchable root cluster item, which is what Moodle needs in order to recognise them as siblings.

### Symptom

With Moodle's standard SCORM activity (`mod_scorm`, not `mod_exescorm`), sitting on "Topic 1" the `button#nav_skipnext` is disabled and cannot reach "Topic 2". The previous/next page buttons (`nav_prev` / `nav_next`) still work and the table of contents renders correctly. Only sibling jumps break, and only at the top level.

### Difference between the two packages

Both packages attached to the issue contain the same 17 pages and the same logical hierarchy. What differs is the top level of the manifest:

**eXeLearning 2.9** — a single top-level `<item>`; everything hangs from "Home":

```xml
<organization>
  <item identifier="ITEM-...home" identifierref="RES-...">
    <title>Home</title>
    <item ...Topic 1...> ... </item>
    <item ...Topic 2...> ... </item>
    <item ...Topic 3...> ... </item>
  </item>
</organization>
```

**eXeLearning 4.0.2** — four top-level `<item>` elements; "Home" is a sibling, not a parent:

```xml
<organization>
  <item ...Home... />
  <item ...Topic 1...> ... </item>
  <item ...Topic 2...> ... </item>
  <item ...Topic 3...> ... </item>
</organization>
```

The eXe 4 manifest faithfully mirrors the document model: in `content.xml` all four pages have an empty `<odeParentPageId></odeParentPageId>`. eXe 4 models the outline as a *forest* of root pages, whereas eXe 2.x modelled it as a *tree* with a single root node. **The eXe 4 manifest is valid SCORM**; the problem is that `mod_scorm` cannot digest it.

### Full failure chain in Moodle

Verified against the real `mod_scorm` sources (`MOODLE_405_STABLE`):

1. **`datamodels/scormlib.php`** assigns `parent = <organization identifier>` to top-level `<item>` elements and `parent = <parent item identifier>` to nested ones.

2. **`player.php:284-285`** builds the navigation JSON with `scorm_get_toc_object(...)` **without** passing `$organizationsco`, unlike the call that renders the table of contents (`player.php:148`, which does pass it). Without that anchor, the organization SCO is not part of the tree.

3. **`scorm_get_toc_get_parent_child()`** (`locallib.php:1688`) can only re-attach to a parent that is already present in the current level. With a single root item everything lands in `level 0`; with four, it increments `$level` for each extra root and returns **four disconnected trees**. Those roots **receive no `parentscoid`**.

4. **`scorm_get_adlnav_json()`** (`locallib.php:2056`) only links `prevsibling`/`nextsibling` when two SCOs are **adjacent in depth-first order and share a parent**:

   ```php
   if (isset($adlnav['prevparent']) && $adlnav['prevparent'] == $sco->parent) {
       $adlnav[$sco->id]['prevsibling'] = $adlnav['prevscoid'];
       $adlnav[$adlnav['prevscoid']]['nextsibling'] = $sco->id;
   }
   ```

   Since "Section 1.3" sits between "Topic 1" and "Topic 2" in that traversal, **a page that has sub-pages never gets a `nextsibling`**. This happens identically in 2.9 and in 4.

5. **`scorm_update_siblings()`** (`module.js:55`) is the client-side patch that repairs the previous step: it groups SCOs that **share a `parentscoid`** and fills in their `prevsibling`/`nextsibling`.
   - eXe 2.9: Topic 1/2/3 have `parentscoid = Home` → they get grouped → **link repaired** ✅
   - eXe 4: Topic 1/2/3 are TOC roots and **have no `parentscoid`** → they are never grouped ❌

6. **`scorm_skipnext()`** (`module.js:474`) then fails both of its branches: the first needs `nextsibling` (undefined) and the second needs `node.parent.parent` (undefined for a tree root) and `parentscoid` (undefined). It returns `null` and `scorm_fixnav()` disables the button. The same applies to `#nav_skipprev`.

`#nav_up` is a different story and this PR does **not** change it. A cluster item renders as a `<span>` without `data-scoid` (`locallib.php`, `TOCJSLINK` branch), `scorm_parse_toc_tree()` leaves `node.scoid` undefined (`module.js:126-138`), and `scorm_fixnav()` disables the up button whenever the target has no `scoid` (`module.js:260-263`). Top-level pages therefore keep it disabled before and after; sub-pages keep it enabled in both cases. See "Known effects" for the resulting gap against 2.9.

### Evidence

The table below is produced by **replaying `mod_scorm`'s own algorithms** (steps 3–6, ported line by line from `MOODLE_405_STABLE`) over the two manifests attached to the issue. It is a deterministic reproduction of Moodle's logic rather than a live session; the behaviour it predicts has since been **confirmed by hand in a real Moodle** — see "Verification status" below.

| `#nav_skipnext` on | eXe 2.9 + `mod_scorm` | eXe 4 + `mod_scorm` | eXe 4 + `mod_exescorm` |
|---|---|---|---|
| Home | disabled | → Topic 1 | → Topic 1 |
| **Topic 1** | **→ Topic 2** ✅ | **DISABLED** ❌ | → Topic 2 ✅ |
| **Topic 2** | **→ Topic 3** ✅ | **DISABLED** ❌ | → Topic 3 ✅ |
| Section 1.1 | → Section 1.2 | → Section 1.2 | → Section 1.2 |
| Section 2.1 | → Section 2.2 | → Section 2.2 | → Section 2.2 |

This matches the report exactly.

### Details that made this confusing

- **The table of contents renders perfectly** (17/17 pages in both packages): that code path does pass `$organizationsco`. Only the buttons break.
- **`<` and `>` kept working**: they use `prevscoid`/`nextscoid`, a flat chain that ignores the hierarchy.
- A **completely flat** eXe 4 project does not fail. The failure appears **as soon as a top-level page has sub-pages**.
- **`mod_exescorm` already patched step 5** (its issue #63): it buckets parentless roots under a sentinel key. That is why it works there.

## The solution

Every page now hangs from a single root `<item>` with **no `identifierref`** (a *cluster* in SCORM terminology), titled with the project title:

```xml
<organization identifier="eXe-20260803111525LY6FEU" structure="hierarchical">
  <title>Nested pages SCORM test</title>
  <item identifier="ITEM-ROOT-20260803111525LY6FEU" isvisible="true">
    <title>Nested pages SCORM test</title>
    <item identifier="ITEM-page-...home" identifierref="RES-page-...home" isvisible="true">
      <title>Home</title>
    </item>
    <item identifier="ITEM-page-...t1" identifierref="RES-page-...t1" isvisible="true">
      <title>Topic 1</title>
      <item ...Section 1.1...> ... </item>
    </item>
    <item ...Topic 2...> ... </item>
    <item ...Topic 3...> ... </item>
  </item>
</organization>
```

Three decisions about that root item:

1. **It carries no `identifierref`.** It is a non-launchable organizational node, so the `<resources>` section needs no changes.
2. **Its identifier is `ITEM-ROOT-<projectId>`**, derived from the project id and therefore stable across re-exports, matching the manifest identifier convention (#1785).
3. **Its title is the project title**, falling back to `eXeLearning`.

With that, Moodle assigns a `parentscoid` to every top-level page and its `scorm_update_siblings()` links them as siblings with no plugin patch required.

### Rejected alternatives

- **Nest top-level pages under the first page** (exact 2.9 parity): fixes the bug and adds no extra node to the TOC, but misrepresents the document model — pages the author placed at the same level would appear indented under "Home" in every LMS and in the IMS CP export.
- **Change nothing and treat it as a Moodle defect**: the manifest is already spec-valid, but every user of the standard SCORM activity would stay broken with no workaround under their control.

## Advantages

- **Fixes the reported case**: `#nav_skipnext` and `#nav_skipprev` work again in unmodified Moodle, with no plugin to install and no core patching.
- **Improves on eXeLearning 2.9**: sibling jumps between pages *that have sub-pages* (for example "Section 1.1" → "Section 1.2") now work too; they were broken in 2.9 as well, because of step 4 of the chain.
- **Preserves the authored hierarchy**: "Home" remains a sibling of "Topic 1", not its parent. The tree the learner sees is the one the author designed.
- **Fixes all three formats at once**: SCORM 1.2, SCORM 2004 and IMS CP.
- **Benefits any LMS**, not just Moodle: the fix lives in the package, so it covers any platform that parses the manifest just as naively.
- **Better SCORM 2004 conformance**: a cluster without `identifierref` is the canonical shape for an aggregation node.
- **Removes triplication**: one implementation of the item tree instead of three identical copies, plus one `escapeXml` instead of four.
- **More robust**: the new generator builds a `parent → children` index in a single pass (it previously filtered the whole page list at every level) and guards against cyclic hierarchies that used to hang the export.

### New: `src/shared/export/generators/ManifestItems.ts`

Shared item-tree generator. It exposes:

```ts
export function generateManifestItems(options: ManifestItemsOptions): string
export function buildRootItemIdentifier(projectId: string): string
export function escapeXml(str: string): string
```

```ts
export interface ManifestItemsOptions {
    pages: ExportPage[];          // pages already filtered for visibility, in reading order
    projectId: string;            // used to build the root item identifier
    rootTitle?: string;           // project title; falls back to "eXeLearning"
    renderClusterExtras?: (indentStr: string) => string;   // per-format hook
}
```

Implementation notes:

- Builds a `Map<parent, children[]>` once instead of filtering the full list at every recursion level.
- Normalises `parentId: ''` as a root page.
- Keeps a `Set` of already-rendered ids that prevents infinite recursion when duplicate pages declare each other as descendants.
- Returns an empty string — no dangling root item — when there are no pages, or when every page points at an ancestor that is not in the list.
- **Warns the author instead of staying silent** when pages are left out of the organization (unknown parent, cyclic hierarchy) or reuse another page's id. Their files still ship in the ZIP, so the omission is otherwise invisible.
- `renderClusterExtras` fires only on items that **actually rendered** children, so a skipped duplicate cannot leave an empty item carrying `<imsss:sequencing>`.
- The `renderClusterExtras` hook is the single point of variation between formats.

### Modified: the three manifest generators

`Scorm12Manifest.ts`, `Scorm2004Manifest.ts` and `ImsManifest.ts`:

- `generateItemRecursive()` is removed (nothing else called it; verified across `src/` and `public/`).
- `generateItems()` stays a public method with the same signature — the three classes are re-exported to the browser from `browser/index.ts` — and now simply delegates:

  ```ts
  generateItems(): string {
      return generateManifestItems({
          pages: this.pages,
          projectId: this.projectId,
          rootTitle: this.metadata.title,
      });
  }
  ```

- **SCORM 2004** is the only format that adds `<imsss:sequencing>` to items with children. To keep SCORM 2004 knowledge out of the shared file, it becomes a hook:

  ```ts
  renderClusterExtras: indentStr => this.generateItemSequencing(indentStr),
  ```

  `generateItemSequencing()` is unchanged. In the test project the manifest goes from 2 `<imsss:sequencing>` blocks to 3 (organization + root item + "Chapter 1").

- `escapeXml()` was byte-identical in all three classes; they now delegate to the shared implementation. They remain public methods because the existing specs call them directly.

### Tests

- **New** `src/shared/export/generators/ManifestItems.spec.ts` — 23 tests: root item shape, non-launchability, stable identifier, top-level pages sharing indentation, deep nesting, preserved ordering, XML escaping, the cluster hook, configurable indentation, empty project, orphan pages, empty `parentId` and cycle protection.
- **Updated** the three existing specs with a regression test asserting the non-launchable root item, plus one in SCORM 1.2 checking that top-level pages stay at the same level. Corrected the `<imsss:sequencing>` count in the SCORM 2004 spec (2 → 3).
- **New E2E** `test/e2e/playwright/specs/scorm-manifest-navigation.spec.ts` — creates a project with "Topic 1 > Section 1.1 / Section 1.2" and "Topic 2", exports to SCORM 1.2, SCORM 2004 and IMS CP, and validates the manifest tree in all three.

### Documentation

- **New** `doc/architecture/adr/ADR-2222-01-scorm-manifest-root-cluster-item.md` with the full failure chain, the alternatives and the consequences. It follows the issue-based identifier scheme adopted in ADR-2232-01; there is no committed index to update.
- `public/CHANGELOG.md`: one entry under **Fixed** for the restored sibling buttons, and one under **Changed** recording that every exported package — SCORM *and* IMS CP — now groups its pages under a root entry, and that re-uploading moves existing entries one level deeper.

## Known effects

- Moodle's table of contents shows **one extra root node** with the project title. Since the TOC panel header already displays that same title, there is slight visual duplication. That is the price of preserving the authored hierarchy.
- The root item is not launchable: LMSs render it as text rather than a link. This is standard SCORM cluster behaviour.
- Manifests change shape compared with earlier 4.x releases: an LMS that tracked the old top-level items will now see them one level deeper.
- **`#nav_up` on a top-level page stays disabled**, and the gap against eXeLearning 2.9 is not closed. In 2.9 the parent of "Topic 1" was the launchable "Home" page, so "up" led there; the cluster is not launchable, so it cannot be an up target. This is not a regression against current 4.x — the button was already disabled — but it is not a fix either. Closing the gap would mean making the root item launchable, which would reintroduce the very problem this PR solves.
- IMS CP packages get the root node as well, even though the reported bug is specific to Moodle's SCORM player. Keeping one item tree for the three formats was preferred over branching the behaviour per format.

## How to test

```bash
make fix
bun test src/shared/export/generators/
make test-unit
make test-integration
bun x playwright test --project=chromium test/e2e/playwright/specs/scorm-manifest-navigation.spec.ts
```

Manual:

1. ✅ `make up-local`, create a project with "Home", "Topic 1" (with sub-pages) and "Topic 2", export to SCORM 1.2.
2. ✅ Check in `imsmanifest.xml` that `<item identifier="ITEM-ROOT-...">` exists without `identifierref` and that the top-level pages are its children.
3. ✅ Upload the ZIP to Moodle using the **standard** SCORM activity (`mod_scorm`, not `mod_exescorm`), with the navigation bar visible.
4. ✅ Confirm that `#nav_skipnext` moves from "Topic 1" to "Topic 2" and `#nav_skipprev` the other way round.

## Verification status

| Check | Status |
|---|---|
| `make fix` | ✅ clean |
| `make architecture-check` | ✅ clean |
| Unit tests (`make test-unit`) | ✅ no new failures |
| Integration tests | ✅ no new failures |
| E2E spec for this change | ✅ passes |
| Full Playwright suite | ⚠️ not run to completion |
| Replay of `mod_scorm`'s algorithms over both manifests | ✅ reproduces the bug and the fix |
| **Manual navigation check in a real Moodle 5.x, stock SCORM activity (steps 3–4)** | ✅ **`#nav_skipnext` / `#nav_skipprev` work between top-level pages** |
| Manual re-upload over an activity with existing attempts (step 5) | ❌ outstanding |

The navigation check was run on Moodle 5.x rather than 4.5. The player logic this PR depends on is byte-identical in both: `scorm_update_siblings()` (`module.js:55-69`) and the nav-button gating (`module.js:260-263`) are unchanged between `MOODLE_405_STABLE` and `MOODLE_500_STABLE`, so the confirmation applies to 4.5 as well.

On step 5 specifically: `scormlib.php:598-601` matches existing SCO rows by `identifier` and keeps their row id (`// Update the Sco sortorder but keep id so that user tracks are kept against the same ids.`), and this PR changes no page identifier — only the `parent` of the top-level ones, plus one new `ITEM-ROOT-…` row. Reading the code, tracking should survive; that has not been observed in a running instance.

## Follow-up (out of scope for this PR)

- Report the defect upstream to Moodle: `player.php` builds the navigation JSON without `$organizationsco`, which is why `scorm_get_toc_get_parent_child()` cannot anchor top-level items to the organization.
- Separate issue for SCORM 2004 conformance: an `<item>` with children should not carry `identifierref`. eXeLearning 2.9 inserted a "fake node" for this in `scormexport.py`.
- Separate issue for adding `xmlns:xsi` and `xsi:schemaLocation` to the manifests, which 2.9 emitted and 4.x does not.


erseco (changes requested):
The fix is correct and the code is well built — the root cluster is the right call, the shared generator is a net improvement, and the test coverage is solid. Requesting changes for two items, both cheap:

**1. The `nav_up` claim is incorrect (PR body + ADR lines 64/141).**
Verified against `MOODLE_405_STABLE`: a non-launchable item renders as a `<span>` with no `data-scoid` (`locallib.php`, TOCJSLINK branch), and `scorm_fixnav` disables the up button when `upnode.scoid` is undefined (`module.js:260-263`). So this PR does not change `#nav_up` at all: sub-pages already worked before (nested items always had `parentscoid`), and top-level pages stay disabled after (their up-target is the non-launchable cluster). There is also a residual gap vs 2.9 worth stating: in 2.9, "up" from Topic 1 reached Home (a launchable parent); with the cluster it cannot. Please correct the claim in the PR description and in ADR-2222-01 — it is the right trade-off, just describe it accurately.

**2. Confirm the manual Moodle test was actually run.**
The evidence table reads as a replay of the algorithm over the two manifests, not a live session. Please confirm the "How to test" steps were executed in a real Moodle (4.5, stock `mod_scorm`), including attempt restore/progress after re-uploading the reshaped package.

Non-blocking recommendations:
- The shared module now formalizes silently dropping pages with duplicate ids or dangling/cyclic `parentId` (in the all-orphan case the organization is empty while `<resources>` still ships everything). Consider surfacing a warning to the author instead of silence — silent export degradation has bitten us before.
- `renderClusterExtras` fires on the pre-`visited`-filter children list, so a duplicate-id edge case can emit `<imsss:sequencing>` on an item that rendered zero children. Gate it on actually-rendered children.
- Add a CHANGELOG note: IMS CP TOCs gain a root node too (a format that had no bug), and existing LMS activities will see items one level deeper on re-export.
- Minor: the `|| 'eXeLearning'` fallback is duplicated at every call site and inside the shared builder; `baseIndent` has no production consumer; the E2E parses `organizationTitle` but never asserts it, and `decoder.decode(files['imsmanifest.xml'])` should throw if the entry is missing rather than decode `undefined` to `''`.

erseco (changes requested):
The follow-ups from the last round all landed: `#nav_up` is now described accurately in both the PR body and the ADR, the manual Moodle check was actually run (and the 5.x → 4.5 carry-over argument is sound), dropped pages are no longer silent, `renderClusterExtras` is gated on actually-rendered children, and the E2E asserts the organization title.

What is genuinely good here, for the record:

- **ADR-2222-01's evidence section.** Replaying `scormlib.php` → `player.php` → `locallib.php` → `module.js` step by step over both attached manifests, and identifying that `player.php` builds the nav JSON without `$organizationsco`, is a root-cause analysis rather than a guess. The three options are weighed honestly and the negative consequences (extra TOC node, `#nav_up` gap vs 2.9) are stated instead of hidden.
- **One item builder instead of three.** The old `generateItemRecursive()` was triplicated, did `this.pages.filter()` inside the recursion (quadratic), and had no cycle protection — a self-referencing `parentId` would blow the stack. `ManifestItems.ts` fixes all three at once, and the three class-level `escapeXml()` copies now delegate to one function.
- **`ITEM-ROOT-<projectId>` is stable across re-exports**, consistent with the manifest identifier from #1785, which is what keeps LMS tracking rows matched by `identifier`.
- **Coverage is real**: degenerate inputs (empty, orphan, cycle, duplicate id) are all exercised, plus an E2E across the three formats. And CI on `9c6e6b0` is fully green — every Playwright shard on chromium, firefox and static, plus `build_and_test` and the codecov patch gate — so the "full suite not run to completion" caveat in the PR table is now stale; please just update that row.

Requesting changes for two items, both small.

## Required

**1. Both changelog entries are under the already-released `v4.0.3` heading.**

`public/CHANGELOG.md`: `## v4.0.4 – 2026-09-10` starts at line 3, `## v4.0.3 – 2026-08-06` at line 100, and the two new entries are at lines 116 and 120 — inside the v4.0.3 block. The branch was cut before v4.0.4 was prepared, and the merge of `main` inserted the new section above them. As it stands, v4.0.4 ships a package-format change with no release note, while a release from a month ago claims it. Move both bullets under `## v4.0.4`.

**2. The duplicate-id warning promises something the manifest does not deliver.**

The message says the pages "were written to the manifest only once", but only the `<item>` tree deduplicates. `generateResources()` in all three generators (`Scorm12Manifest.ts:224`, `Scorm2004Manifest.ts:251`, `ImsManifest.ts:251`) still iterates `this.pages` unconditionally, so two pages sharing an id emit `<resource identifier="RES-a">` twice. `identifier` is `xsd:ID` in `imscp_rootv1p1p2.xsd:83`, which must be document-unique, so the package becomes schema-invalid — and the author is told it was handled. Either dedupe the resource loop against the same set, or reword the warning to say the duplicate page still emits a duplicate resource.

## Non-blocking

- **All-orphan projects still emit an organization with zero items.** Pre-existing (the old `this.pages.filter(p => !p.parentId)` had the same hole), so not a regression — but the root cluster now makes the fix trivial: hang unreachable pages off it instead of returning `''`. Today `test/helpers/xml-validator.helper.ts:357` would flag the result as "No items defined in organization" while `<resources>` still ships every file.
- **Falsy-id asymmetry.** `buildChildrenMap()` keys on `page.parentId || null`, but `renderItem()` looks up `childrenMap.get(page.id)`. For a page with `id: ''` the two disagree: its children get bucketed as roots and render as siblings of their own parent, while the parent renders childless (and in SCORM 2004 loses its `<imsss:sequencing>`). `childrenMap.get(page.id || null)` makes it symmetric.
- **`escapeXml` is still duplicated at module level.** The PR removed three class copies — good — but `OdeXmlGenerator.ts:350` already exports an `escapeXml`, in the same directory, and it emits `&apos;` where `ManifestItems.ts:96` emits `&#039;`. Two exported functions with the same name and divergent output in one folder is worse than four private copies. Import the existing one, or hoist a single implementation and have `LomMetadata.ts:308` and `BaseExporter.ts:433` delegate too.
- **`DEFAULT_ROOT_TITLE` still duplicates the `|| 'eXeLearning'` fallback** computed in the three `generateOrganizations()` methods. Passing the already-resolved title down would make "root title == organization title" structural instead of a coincidence that the new E2E assertion happens to check.
- **`console.warn` is not an author-facing channel.** The doc comment says it "warns the author", but exports run in the browser bundle and in the CLI/API server — the message reaches devtools or a server log, never the person exporting. Surfacing it through the export result the caller already returns would make the guarantee real. Related: only the `describe('author warnings')` block installs the stub, so `bun test src/shared/export/generators/ManifestItems.spec.ts` prints three stray `[ManifestItems] …` lines from the tests at `:255`, `:290` and `:308`.
- **`reportDroppedPages()` builds the label string for every page on every export**, plus a second `Set` that duplicates `visited`, to produce nothing in the normal case. An early return when `rendered.size === pages.length` removes the cost.
- **The root-level `childrenXml !== ''` guard is dead** — the `rootPages.length === 0` early return means `visited` is empty on the first child, so it can never be skipped. It is meaningful in `renderItem()` (which is where I asked for it last round); at the root it reads as if a childless root were reachable.
- **Two spec assertions are weaker than they look.** `Scorm2004Manifest.spec.ts:126` and `ImsManifest.spec.ts:181` compare `indexOf` positions, which passes even if the root item is opened and closed immediately, leaving the pages as top-level siblings again — the exact bug. `Scorm12Manifest.spec.ts` gets it right with indentation; `ManifestItems.spec.ts:66-71` gets it right with the containment check. Also `scorm-manifest-navigation.spec.ts:122`: `expect(Object.keys(manifests)).toEqual(['scorm12','scorm2004','ims'])` is tautological, since the helper builds that object by iterating the same literal.
- **`ManifestItems.spec.ts:124` locks in escaping for an `xsd:ID`.** `identifier="ITEM-ROOT-proj &amp; 1"` is well-formed but not schema-valid: `xsd:ID` must be an NCName, so no spaces and no `&` even after entity expansion. Ids come from `getBareProjectIdentifier()` and are safe today, but the test documents escaping as sufficient for whoever widens that source next. Sanitizing to an NCName is the correct treatment, and it matters because the identifier is also the LMS tracking key.

## For @ignaciogros — two product calls, not code

1. **Every project exported from 4.0.4 onwards gains a visible root node in the LMS table of contents**, titled with the project name, duplicating the TOC panel header. It is the accepted cost of Option 1 in ADR-2222-01 and I agree with the trade-off, but it is user-visible in every SCORM export and worth your explicit sign-off before release.
2. **IMS CP is included even though the bug is specific to Moodle's `mod_scorm`.** The ADR argues consistency across the three generators and files the "LMS requiring a launchable first item" case under low risk; no IMS CP viewer was exercised, and `renderClusterExtras` already exists as the per-format hook if we would rather leave a format that never had the bug untouched. Your call on whether consistency or minimal blast radius wins.

Also still outstanding from the PR's own table: re-uploading a reshaped package over an activity that already has attempts. The `scormlib.php:598-601` reading is convincing, but it is a reading.

ignaciogros (approved):
> **Every project exported from 4.0.4 onwards gains a visible root node in the LMS table of contents**

As long as this is standard-compliant, and considering that this is the option accepted in the ADR, I think we should keep it, at least for the moment. We can always adjust the presentation in `mod_exescorm` and `moodle-mod_exelearning` if necessary.

> **IMS CP is included even though the bug is specific to Moodle's `mod_scorm`.**

I’m OK with including the change in IMS CP in favor of the consistency. I’ve tested it with Moodle’s IMS viewer, and it works correctly.

**Other issues:**

* **CHANGELOG item positioning issue:** I’ve removed the entries. I’ll complete the CHANGELOG as part of #2378.
* **duplicate-id issue:** I believe this is fixed in `3ba3fcfb8c06f00526dc50df565907c58b21ab8a` and `91fab88a2c7aa40fbe1287dd45caac6b46129ef4`.

If you don’t find new issues, I think it’s ready to be merged into `main`. That will allow us to continue working on the SCORM improvements in #2371.

erseco:
## Code review — root cluster item in SCORM/IMS manifests

Reviewed `main...HEAD` (9 commits, 10 files). The root-cluster decision itself looks right, the shared builder is a real de-duplication win, and `bun test src/shared/export/generators/` (212 pass), `bunx biome check` and `make architecture-check` are all green. Findings below, most severe first.

---

### 1. Pages with a dangling `parentId` are dropped instead of re-parented — and can produce a manifest with zero items

`src/shared/export/generators/ManifestItems.ts:283-288`

When no page has `parentId === null`, `generateManifestItems()` warns and returns `''`. `generateOrganizations()` then emits an `<organization>` with **no `<item>` at all**, and Moodle rejects such a package ("Could not find any SCO"). The same drop happens per-page at `renderItem` for any page whose parent id is not in the list: its `<resource>` still ships, but nothing in the TOC points at it.

Reachable state: a page whose parent was removed from the Yjs navigation without re-parenting, or a partially repaired import. `isPageVisible()` cannot cause it (it hides descendants too), but nothing else guarantees referential integrity of `parentId`.

Now that a root cluster exists, there is an obvious home for orphans: treat a page whose `parentId` is not present in the list as a root page. That is the deeper fix — the current code detects the broken state and then ships the broken package anyway, with only a `console.warn` that nobody sees during a browser-side export (the primary export path).

### 2. Deduplication is applied at three symptom sites instead of once at the boundary

`src/shared/export/generators/Scorm12Manifest.ts:226`, `Scorm2004Manifest.ts:253`, `ImsManifest.ts:253`

`dedupePagesById()` is called in the three `<resources>` loops, with the same three-line comment copy-pasted. The rest of each generator still sees the raw list:

- `categorizeFilesForCommon()` (`Scorm12Manifest.ts:122`, `Scorm2004Manifest.ts:120`, `ImsManifest.ts:118`) still iterates `this.pages`, so a duplicated id with a different title adds its default `html/<slug>.html` to the exclusion set. After dedup that file is excluded from `COMMON_FILES` **and** has no `<resource>` — an unreferenced file in the package.
- `generatePageResource()` computes `this.pages.indexOf(page) === 0` against the undeduped array.

Normalising once in the constructor (`this.pages = dedupePagesById(pages || [])`) fixes all three call sites and removes the triplicated comment.

### 3. The dedupe docstring's safety claim is narrower than stated

`src/shared/export/generators/ManifestItems.ts:110-113`

> when the orders disagree they can keep different page objects. That is safe because both identifiers derive from `page.id` alone

True for identifiers, not for everything derived from the page. Counter-example:

```
[ X1(id=x, parent=p1), X2(id=x, parent=p2), P2(id=p2, parent=null), P1(id=p1, parent=null) ]
```

Root order is `[P2, P1]`, so the depth-first walk renders **X2** as the `<item>` (and takes its `<title>`), while `dedupePagesById` keeps **X1** for the `<resource>` (and, when `pageFiles` has no entry for that id, derives `href` from X1's title). The TOC label and the resource then come from different page objects. Worth narrowing the claim to identifiers, or making both mechanisms pick the same object (see #2).

### 4. The duplicate-id warning states the wrong outcome, and the spec pins the wrong wording

`src/shared/export/generators/ManifestItems.ts:255-257`, asserted verbatim at `ManifestItems.spec.ts:392`

> Only the first occurrence gets an `<item>` and a `<resource>`

In the case above the *later* occurrence gets the `<item>` and the *first* gets the `<resource>`. Also, `reportDroppedPages` classifies a repeated id as `duplicated` and `continue`s, so a first occurrence that is itself unreachable is never reported as unreachable. Asserting the sentence verbatim ("the wording is the promise made to the author") locks in a promise the code does not keep.

### 5. A fifth `escapeXml` — `OdeXmlGenerator` already exports one

`src/shared/export/generators/ManifestItems.ts:89` vs `src/shared/export/generators/OdeXmlGenerator.ts:350`

Same directory, same job, and they differ only in the apostrophe entity (`&#039;` here, `&apos;` there). AGENTS.md: *"Single source of truth. When two code paths need the same data, extract a shared function. Duplicated logic is a bug waiting to happen."* A PR whose stated goal is collapsing triplicated generator code adds a fifth copy of this helper (the three class methods now delegate, but `LomMetadata.ts:308` and `OdeXmlGenerator.ts:350` still hold their own). Suggest one exported `escapeXml` in a shared util that all five use, and pick one apostrophe entity.

### 6. `identifier` values are XML-escaped, not sanitised — and the new spec asserts the invalid result

`ManifestItems.ts:293` and `:205`; `ManifestItems.spec.ts:143`

`identifier` is `xsd:ID`, i.e. an NCName: no spaces, no `&`, must not start with a digit. `escapeXml()` does not produce an NCName; it only stops the document being ill-formed. The new test asserts

```js
expect(xml).toContain('identifier="ITEM-ROOT-proj &amp; 1"');
```

as the expected behaviour, which is a schema-invalid ID. The escaping of `ITEM-${page.id}` predates this PR, but the root identifier is new and the test now codifies the wrong contract. Either sanitise to NCName in `buildRootItemIdentifier()` or drop that assertion so it does not read as a guarantee.

### 7. Cluster items still carry `identifierref` while gaining `<imsss:sequencing>`

`ManifestItems.ts:205` + `Scorm2004Manifest.ts:234`

Generated output:

```xml
<item identifier="ITEM-p2" identifierref="RES-p2" isvisible="true">
  <title>Topic 1</title>
  <item identifier="ITEM-p3" identifierref="RES-p3" .../>
  <imsss:sequencing>...</imsss:sequencing>
</item>
```

SCORM 2004 CAM: only leaf activities may reference a resource. The ADR flags this as follow-up work, but this PR is the one that shows the correct shape (the root item) and attaches sequencing to exactly the offending items. Applying the same "cluster = no `identifierref` + a synthetic leaf" treatment here would close it now rather than leaving a known-nonconformant tree with new sequencing hung off it.

### 8. `buildChildrenMap` re-implements an existing index

`ManifestItems.ts:133-147` vs `src/shared/export/renderers/PageRenderer.ts:438-451`

`PageRenderer.buildNavRenderContext()` already builds `childrenByParent: Map<string | null, ExportPage[]>` with identical semantics (`page.parentId ?? null`, insertion order preserved). Two implementations of the same page-forest index in the same export package.

### 9. The same two facts are derived three times per export

`ManifestItems.ts:228-243`

`reportDroppedPages` rebuilds the duplicate-id set that `dedupePagesById` just computed, and `dedupePagesById` runs once more in each generator's `<resources>` loop. Three passes over the page list for one piece of information. Folding dedup into the constructor (#2) and passing the result to the reporter collapses it to one.

### 10. The integration "export" lane gives this change no coverage, and its reference fixtures now contradict it

`test/integration/export/really-simple-export.spec.ts:32-40`, `test/fixtures/export/really-simple/really-simple-test-project_scorm/imsmanifest.xml`

`readFixtureFile()` reads pre-generated files from `test/fixtures/export/`; nothing in that spec runs an exporter. So `should have hierarchical organization with all pages` would still pass if the exporter regressed to the old flat tree. Meanwhile the checked-in fixtures still show the pre-#2222 shape (`<organization>` with sibling top-level `<item>`s), so the repo's "reference" output now disagrees with what the code emits. Either regenerate the fixtures or note in the spec that they are historical snapshots rather than a reference for current output.

### 11. Tautological assertion in the E2E spec

`test/e2e/playwright/specs/scorm-manifest-navigation.spec.ts:122`

```ts
expect(Object.keys(manifests)).toEqual(['scorm12', 'scorm2004', 'ims']);
```

`manifests` is populated by a literal `for (const format of ['scorm12', 'scorm2004', 'ims'])` a few lines above, so this asserts the loop against itself. It cannot fail unless the helper throws first — which it already does, with a better message.

### 12. `ITEM-ROOT-<projectId>` shares the `ITEM-` namespace with page items

`ManifestItems.ts:80`

A page whose id is `ROOT-<projectId>` produces two `<item>` elements with the same `xsd:ID`. Astronomically unlikely with the current id generator, but a separator the page-id space cannot contain (or a prefix like `ORG-ROOT-`) removes the class of bug for free.

---

**Not findings, for the record:** the item/sequencing element order matches the CP content model (`title`, `item*`, extensions); the depth-first `visited` guard correctly fixes the unbounded recursion the old `generateItemRecursive` had on a cyclic hierarchy; `generateItemRecursive` has no remaining callers; the E2E bridge API (`_yjsBridge`, `structureBinding.createPage(title, parentId)`, `window.fflate`) all check out; and `exporters.bundle.js` is gitignored, so no stale artifact ships.

ignaciogros (changes requested):
Although, as an end user, I would prefer not to have the additional top-level node proposed in this PR, I think that if it is essential for the LMS to provide robust navigation, we should accept it. As I mentioned before, I think we can revisit the usability/presentation side of the plugins later if necessary.

In any case, issues related to overlapping icons should be fixed:

Image

I think this PR should be merged into `main` after #2371, which is the more substantial change.

Thanks.

erseco:
@ignaciogros thanks — agreed on the root node, and that settles the product question from my side.

On the overlapping icons: I don't think this PR introduces them, but it is worth five minutes to be sure. The overlap in your screenshot is on *"1. El contrato de trabajo"*, a page that is **launchable and has sub-pages** at the same time (`<item identifierref="RES-…">` with child items). This PR does not change that shape — those items looked the same before, one level higher. Tellingly, the node this PR *does* add ("Tema 8…") renders cleanly, precisely because it has no `identifierref`.

The competing hypothesis, which would blame this PR: the root cluster pushes everything one level deeper, and Moodle's TOC indents per level, so at greater depth the status icon can collide with the collapse triangle.

Could you export the **same project from `main`** and upload that package? If the overlap reproduces, it is a pre-existing issue and should be tracked separately rather than blocking here. If it only appears with the root cluster, it is a presentation regression and we deal with it in this PR.

Either way the underlying cause is the one I flagged in my review: an `<item>` that carries `identifierref` *and* has children is nonconformant under the SCORM 2004 CAM (only leaf activities may reference a resource), which is why Moodle paints both a SCO icon and an aggregation icon on it. The proper fix is the "fake node" eXeLearning 2.9 used in `scormexport.py` — every page with children becomes a non-launchable cluster with a leaf child for its own content. That is already listed in this PR's follow-ups as a separate issue, and doing it here would roughly double the change.

On merging after #2371: no objection, but there is no technical dependency — the two PRs share zero files (#2371 is `public/app/common/scorm/**` plus the iDevices, this one is `src/shared/export/generators/**`), so neither order causes conflicts.

ignaciogros:
@erseco,

> On the overlapping icons: I don't think this PR introduces them...

You might be right, and I just can't reproduce the problem right now in either branch.

The reason I suggest merging the other one first is that some users are already testing it, and it's an important change. This one should only require a short discussion within the development team, so that everyone involved is aware of the reasons behind this PR. It has some presentation consequences in the viewer that may come as a surprise to more than one, and could be perceived as a bug.

I suggest waiting until we get the OK from @eXeLearningProject and @cristinavaldera.

cristinavaldera:
If these presentation issues are the price to pay for solving the navigation problems as soon as possible , for me it is ok to go ahead. Thank you for the investigation and the solution!

ignaciogros (approved):
I’ve been reviewing the details, and I think it makes sense to merge this PR: it’s the universal solution and is aligned with the standard, fixes the issue in any LMS, and doesn’t depend on the organization being able to install the eXeLearning plugins.

In my opinion, the potential tracking issue when re-uploading a package over an activity that already has grades is something we can accept: it’s a residual behaviour that has already been flagged in that workflow and is up to the teacher’s discretion. Moodle’s own code matches the entries by identifier, which this PR does not change.

If you agree with this approach, I suggest merging the solution into `main` (before or after #2371, depending on @mnarvaezm's preference), but I think we should open an issue for `mod_exescorm` to replicate the warning in the embedded workflow, because the `editdialogcontent:caution` warning is only displayed when the plugin is used with eXeLearning Cloud.

@erseco, @mnarvaezm:

Do you agree?

Thanks.

erseco (approved):
Both blocking items are resolved: the changelog entries were reverted out (`148f56618`, deferred to #2378) and duplicate ids can no longer emit duplicate `xsd:ID` resources (`3ba3fcfb8`, `91fab88a2`). CI is green on `91fab88a2`, all 19 checks including every Playwright shard. Approving.

The 12 items in my [follow-up review](https://github.com/exelearning/exelearning/pull/2244#issuecomment-5601566576) are non-blocking — none is a regression introduced here, and the heaviest one (`identifierref` on items that also have children, nonconformant under the SCORM 2004 CAM) is pre-existing and already listed in this PR's follow-ups.

One cheap change worth folding in before merge, if you want it: `this.pages = dedupePagesById(pages || [])` in the three constructors. `dedupePagesById()` currently runs inside each `<resources>` loop, so `categorizeFilesForCommon()` still walks the raw list and a duplicated id can leave an HTML file in the ZIP that is excluded from `COMMON_FILES` *and* referenced by no `<resource>`. Normalising at the boundary fixes all call sites and removes the triplicated comment. Fine as a follow-up too.

Merge order is @mnarvaezm's call — no technical dependency on #2371, the two share zero files. Agreed on opening the `mod_exescorm` issue for the `editdialogcontent:caution` warning in the embedded workflow.

mnarvaezm:
Hi!

If both branches are going to be merged one after the other, with both already approved, the order is practically irrelevant. If, however, one is going to be merged and tested for a while before merging the other, I think it would be better to merge #2244 first and #2371 afterwards.
@erseco  @ignaciogros  @cristinavaldera

mnarvaezm:
What we should make sure of is to leave a couple of days after both branches are merged to properly test everything related to the SCORM packages before releasing the new version.

@erseco @ignaciogros @cristinavaldera

cristinavaldera:
So, merge them and we start testing!

PR #2249, opened 2026-08-12
Support wildcard subdomains in `PROVIDER_URLS`

### LINKED ISSUE #2248: Support wildcard subdomains in `PROVIDER_URLS` for multi-tenant platform integration
## Summary

`PROVIDER_URLS` only accepts literal URL prefixes, so a deployment serving many Moodle instances under a shared parent domain must enumerate every single host. When the variable-length part of the URL comes *before* the fixed part (`https://<anything>.example.net/`), the current matching cannot express the rule at all.

This issue proposes single-label wildcard support (`https://*.example.net`) and moves the allow-list check from raw string prefixes to the parsed hostname.

## Current behaviour

`isAllowedProviderUrl()` in `src/utils/platform-jwt.ts` performs a plain string prefix test:

```ts
return config.urls.some(allowedUrl => url.startsWith(allowedUrl));
```

There is no wildcard, pattern or regular-expression support, and no alternative environment variable. The whole platform-integration path reads only four variables — `PROVIDER_URLS`, `PROVIDER_TOKENS`, `PROVIDER_IDS` and `APP_SECRET` — and none of them offers a way to relax the URL check.

`PROVIDER_IDS` / `PROVIDER_TOKENS` are not an alternative: `isValidProvider()` is an independent check, and `isAllowedProviderUrl()` is still evaluated separately in `getPlatformIntegrationParams()`, so the callback is rejected regardless of how those are configured.

## Motivating use case

A single eXeLearning instance serves many platforms whose return URLs look like:

```
https://<tenant>.example.net/
https://<tenant>.example.net/<path>/
```

`<tenant>` is unbounded and changes over time; `example.net` is fixed. The desired rule is: **the hostname ends with the fixed part and has exactly one label before it.** The path is irrelevant.

Enumerating every tenant in `PROVIDER_URLS` is not operationally viable — the list changes whenever a tenant is added, and each change requires an `.env` edit plus a restart.

## Why this became blocking

Before #1896 ("Security hardening", released in **v4.0.2**, 2026-07-07), an empty `PROVIDER_URLS` meant *allow everything*:

```diff
-    // If no providers configured, allow all
+    // No providers configured => empty allow-list => nothing is allowed.
     if (config.urls.length === 0) {
-        return true;
+        return false;
     }
```

Failing closed is the correct fix — the previous behaviour let a holder of a valid platform JWT drive server-side requests to arbitrary hosts. But it removed the only configuration that made multi-tenant deployments possible, leaving no supported replacement. Affected operators see:

```
401 Invalid token or unauthorized provider
[PlatformJWT] Return URL not in allowed providers: <url>
```

Two aspects of the rollout made this hard to diagnose:

1. The v4.0.2 changelog entry groups 32 fixes into one generic line and does not mention `PROVIDER_URLS` or the behaviour change. The migration note exists only in `.env.dist`, which deployed instances never re-read.
2. The one-time startup warning in `src/index.ts` is only emitted when `PROVIDER_TOKENS` or `PROVIDER_IDS` are set. Deployments that sign with `APP_SECRET` and leave both empty get **no warning at all** — callbacks simply start failing.

### 1. Single-label wildcard

Accept `*` as the leftmost label of an entry, matching exactly one DNS label (no dots), following the same semantics as TLS wildcard certificates:

```
PROVIDER_URLS=https://*.example.net
```

- matches `https://tenant.example.net/` and `https://tenant.example.net/path/`
- does not match `https://a.b.example.net/`
- does not match the bare `https://example.net/` (see open questions)

Entries without `*` keep working as today, so no existing deployment needs to be reconfigured.

### 2. Match on the parsed hostname

The check should parse the URL and compare `hostname` (case-insensitively) instead of testing the raw string. Prefix matching on the raw string does not actually validate the host component, so an allow-list entry can be satisfied by a URL whose effective host is a different domain. Since the matched URL is subsequently used as a server-side request target, host-accurate matching is required for the allow-list to mean what operators expect.

`isSafeReturnUrl()` already parses the URL in the same code path, so this adds no new dependency or parsing cost.

### 3. Regular expressions: considered and rejected

Regex was evaluated as the more general alternative and is **not** recommended:

- `parseEnvArray()` splits on commas, which collides with regex quantifiers such as `{2,3}`. Supporting regex would require changing the separator (breaking existing deployments) or inventing an escaping scheme.
- Unanchored patterns are the classic source of allow-list bypasses, and unescaped dots match more than the author intends.
- The `returnurl` is attacker-controlled input, so an operator-authored pattern introduces ReDoS exposure.
- `.env` files make backslash escaping error-prone, and the resulting values are hard to audit.

The wildcard covers the known use cases. If a case ever appears that it cannot express, regex can be added later as a separate opt-in variable without disturbing existing configuration.

## Open questions

1. Should a bare domain entry (`https://example.net`) also match its subdomains, or only the exact host? Proposal: exact host only — `*.` is explicit about intent.
2. Should the scheme be mandatory in an entry, or should a bare `*.example.net` be accepted?
3. How should non-standard ports be handled — inherited from the entry, or ignored?

## Acceptance criteria

- [ ] `PROVIDER_URLS` accepts `*` as the leftmost label, matching exactly one DNS label.
- [ ] Allow-list matching is performed against the parsed hostname, case-insensitively.
- [ ] Entries without `*` retain their current behaviour; upgrading requires no configuration change.
- [ ] An empty `PROVIDER_URLS` still fails closed.
- [ ] Unit tests in `src/utils/platform-jwt.spec.ts` cover: wildcard match, multi-label rejection, bare-domain behaviour, case-insensitivity, host-accurate matching, empty allow-list, and malformed URLs.
- [ ] Patch coverage ≥ 90%.
- [ ] `make fix`, `make test-unit` and `make test-integration` pass.
- [ ] `.env.dist` documents the wildcard syntax with an example.
- [ ] An ADR records the matching semantics, using this issue number as its identifier (`ADR-<issue>-01-...`).
- [ ] The changelog entry explicitly names `PROVIDER_URLS` so operators can find it.

## Out of scope

- Regular-expression or full glob support.
- Changing the fail-closed behaviour of an empty `PROVIDER_URLS`.
- Runtime/API-based provider registration (no restart required) — worth a separate issue if enumeration remains painful.

## References

- `src/utils/platform-jwt.ts` — `isAllowedProviderUrl()`, `isSafeReturnUrl()`, `getPlatformIntegrationParams()`
- `src/routes/platform-integration.ts` — endpoints returning `Invalid token or unauthorized provider`
- `src/index.ts` — one-time startup warning for empty `PROVIDER_URLS`
- `.env.dist` — "Multiple Platform Providers Configuration" section
- #1896 — security hardening that introduced the fail-closed behaviour (v4.0.2)

ignaciogros:
# fix(platform): match `PROVIDER_URLS` on the parsed host and support wildcard subdomains

Closes #2248

## Problem

`PROVIDER_URLS` is the SSRF allow-list that decides whether a platform JWT's
`returnurl` may become a server-side request target. It was matched as a raw string
prefix:

```ts
return config.urls.some(allowedUrl => url.startsWith(allowedUrl));
```

That caused two distinct problems.

**1. Multi-tenant deployments are inexpressible.** Operators running many LMS tenants
under a shared parent domain have return URLs shaped like
`https://<tenant>.example.net/[path]`, where `<tenant>` is unbounded. Because the
variable part comes *before* the fixed part, no prefix can express the rule.
Enumerating every tenant is not viable, and the only alternative that "worked" —
`PROVIDER_URLS=https://` — matched every https URL by prefix and reinstated the SSRF
the allow-list exists to prevent.

This became blocking in **v4.0.2** (commit `54373758`), which correctly changed an
empty `PROVIDER_URLS` from *allow all* to *deny all* but left prefix matching in
place, removing the only configuration that made these deployments possible.
Affected operators saw `401 Invalid token or unauthorized provider` with no
migration signal: the v4.0.2 changelog folded 32 fixes into one generic line that
does not name `PROVIDER_URLS`, and the startup warning only fires when
`PROVIDER_TOKENS` or `PROVIDER_IDS` are set — deployments signing with `APP_SECRET`
got no warning at all.

**2. Prefix matching does not validate the host.** With `https://moodle.example.com`
configured, the entry was also satisfied by `https://moodle.example.com.evil.com/`
and `https://moodle.example.com@evil.com/`, whose effective hosts are `evil.com`.
Since the matched URL becomes a server-side request target, the allow-list did not
constrain what operators believed it constrained.

### `src/utils/platform-jwt.ts`

`isAllowedProviderUrl()` now matches structurally against the parsed URL instead of
comparing string prefixes. Four small helpers were added:

- **`normalizeEntryHost()`** — normalizes an entry host to punycode via the URL
  parser; rejects `*` anywhere but the leftmost label, a bare `*`, and userinfo.
- **`parseProviderEntry()`** — parses `[scheme://]host[:port][/path]`. Malformed
  entries and non-http(s) schemes are discarded **individually**, so one bad entry
  does not disable the rest of the list. A query or fragment in an entry is dropped:
  the path is compared against `URL.pathname`, which carries neither, so keeping
  them would make the entry impossible to satisfy.
- **`hostMatchesEntry()`** — `*.` matches **exactly one** DNS label, the semantics of
  TLS wildcard certificates.
- **`matchesProviderEntry()`** — parts the entry omits are unconstrained; ports
  compare effective values (`:443` ≡ no port on https) and the path is a prefix.

Two defensive branches guarding an empty parsed host were removed: `http:`/`https:`
are special schemes, so the parser either throws or yields a non-empty host. They
were unreachable, and unreachable code cannot meet the patch-coverage gate.

### `.env.dist` and `doc/deployment.md`

Documented the entry syntax, the wildcard rule with an allowed/not-allowed table,
the fail-closed behaviour, the positional mapping with `PROVIDER_TOKENS` /
`PROVIDER_IDS`, and troubleshooting for both `[PlatformJWT]` log lines.
`doc/deployment.md` had no coverage of these variables at all before this PR; it
also carries the upgrade note for deployments coming from v4.0.1 or earlier.

### ADR

Adds `doc/architecture/adr/ADR-2248-01-match-provider-allow-list-on-parsed-host-with-wildcards.md`
(`status: Proposed`), recording the matching semantics, the options considered, and
why regular expressions were rejected.

## Configuration

Existing entries keep working unchanged. Multi-tenant deployments can now write:

```env
PROVIDER_URLS=https://*.example.net
```

| URL | Allowed |
|---|---|
| `https://tenant.example.net/` | Yes |
| `https://tenant.example.net/course/view.php?id=1` | Yes |
| `https://a.b.example.net/` | No — two labels |
| `https://example.net/` | No — bare domain not covered |
| `https://evilexample.net/` | No |

## Why not regular expressions

Regex was evaluated as the more general option and rejected:

- `parseEnvArray()` splits entries on commas, which collides with quantifiers such as
  `{2,3}`. Supporting regex means changing the separator — breaking every existing
  deployment — or inventing an escaping scheme.
- The `returnurl` is attacker-controlled input matched against an operator-authored
  pattern, adding ReDoS exposure on a request path.
- Unanchored patterns are the classic allow-list bypass, and an unescaped `.` matches
  more than intended. Both fail silently.
- Backslash escaping in `.env` files is error-prone and the values are hard to review.

The wildcard covers the known cases with fixed semantics that cannot be written
insecurely. If a case ever needs more, regex can be added later as a separate opt-in
variable. Full reasoning in the ADR.

## Behaviour change

Every difference between old and new matching was enumerated by running both
implementations side by side over each entry shape. Only one turns a previously
working deployment into a rejected one:

**`PROVIDER_URLS=https://` (or `http://`) no longer matches anything.** It now parses
to an empty host and is discarded as malformed. It was the only way to express "allow
everything", and it is precisely the SSRF this allow-list prevents. Deployments
relying on it must list their platforms. Intentional, and documented in `.env.dist`,
`doc/deployment.md` and the ADR.

The remaining differences either fix the host-accuracy bug (`…example.com.evil.com`
and `…example.com@evil.com` are no longer accepted — no real platform lives at those
hosts) or accept callbacks that were previously rejected by mistake: a host differing
only in case, and an entry with a trailing slash against a URL without a path. Both
concern the exact host already configured, so neither widens the allow-list to any
new domain.

## Decisions worth reviewing

Three semantics were chosen deliberately; all are isolated in `matchesProviderEntry()`:

1. `*.example.net` does **not** cover the bare `example.net` — `*.` is explicit about
   intent, matching TLS wildcard behaviour.
2. The scheme is optional. With a scheme it is enforced; without one, both http and
   https match.
3. A port in the entry is enforced exactly; no port means any port.

## Testing

- 30 new unit tests in `src/utils/platform-jwt.spec.ts`: wildcard matching (single
  label, multi-label and bare-domain rejection, misleading suffixes), host-accurate
  matching (longer-domain, userinfo and query-string cases), scheme/port/path
  constraints, case-insensitivity, IDN normalization, malformed entries, the rejected
  bare `*`, and the rejected scheme-only entry.
- `platform-jwt.spec.ts` + `platform-integration.spec.ts` + `pages.spec.ts`:
  **200 pass, 0 fail**. `platform-integration.spec.ts` is unchanged, confirming the
  integration path is unaffected for existing configurations.
- **Patch coverage: 100%** of changed lines (`platform-jwt.ts` at 98.81%; the
  remaining uncovered lines are pre-existing, in `isSafeReturnUrl()` and
  `decodePlatformJWT()`).
- `make fix` clean on the changed files; `make architecture-check` OK.

Pre-existing failures on the branch, unrelated to this change and not touched here:
82 unit failures in `Config` / `Games` / `Resources` / `Themes` routes (none of which
import `platform-jwt` — in `src/routes` only `pages.ts` and `platform-integration.ts`
do, and both pass), and 2 integration failures in `integration-app.spec.ts` directory
cleanup, which are Windows `EBUSY` file locks.

`make test-e2e` was not run: no E2E spec covers platform integration, since exercising
it requires a real LMS issuing signed JWTs.

### DIFF
--- a/.env.dist
+++ b/.env.dist
@@ -278,13 +278,21 @@ REDIS_PASSWORD=
 # "*" on its own is rejected, and so is a scheme-only entry such as "https://":
 # neither is a supported way to allow every host, because that would reinstate
 # the SSRF this allow-list exists to prevent.
+#
+# NOTE: this is ONE global allow-list, not a per-provider one. A callback is
+# allowed when it matches ANY entry, whichever provider signed the token, so every
+# entry is authorized for every provider. Only PROVIDER_IDS and PROVIDER_TOKENS are
+# bound to each other by position (see below).
 PROVIDER_URLS=
 
 # Provider tokens - authentication tokens/secrets for each platform
+# The token at index n verifies JWTs whose provider_id is PROVIDER_IDS[n], so keep
+# both lists in the same order and of the same length.
 # Example: PROVIDER_TOKENS=moodle_lms_token,workplace_token,custom_lms_token
 PROVIDER_TOKENS=
 
-# Provider IDs - unique identifiers for each platform
+# Provider IDs - unique identifiers for each platform, matched against the JWT
+# provider_id to select the signing secret from PROVIDER_TOKENS
 # Example: PROVIDER_IDS=moodlelms,workplace,custommoodle
 PROVIDER_IDS=
 
--- a/.env.dist
+++ b/.env.dist
@@ -260,6 +260,24 @@ REDIS_PASSWORD=
 
 # Provider URLs - base URLs for each platform (SSRF allow-list; fails closed when empty)
 # Example: PROVIDER_URLS=https://moodle_lms.com,https://moodle_workplace.example.com,https://custom_lms.example.com
+#
+# Entries are matched on the parsed host, not as text prefixes. Each entry is
+# "[scheme://]host[:port][/path]"; the parts you omit are unconstrained:
+#   https://moodle.example.com   exact host, any port, any path
+#   moodle.example.com          same, over either http or https
+#   https://example.com:8443    only that port
+#   https://example.com/moodle  only URLs under /moodle
+#
+# A leading "*." wildcard covers subdomains and stands for EXACTLY ONE label,
+# like a TLS wildcard certificate. Use it for multi-tenant deployments instead of
+# enumerating every tenant host:
+#   PROVIDER_URLS=https://*.example.net
+#     matches https://tenant.example.net/ and https://tenant.example.net/course/
+#     does NOT match https://a.b.example.net/ nor the bare https://example.net/
+#
+# "*" on its own is rejected, and so is a scheme-only entry such as "https://":
+# neither is a supported way to allow every host, because that would reinstate
+# the SSRF this allow-list exists to prevent.
 PROVIDER_URLS=
 
 # Provider tokens - authentication tokens/secrets for each platform
--- a/doc/architecture/adr/ADR-2248-01-match-provider-allow-list-on-parsed-host-with-wildcards.md
+++ b/doc/architecture/adr/ADR-2248-01-match-provider-allow-list-on-parsed-host-with-wildcards.md
@@ -1,0 +1,217 @@
+---
+id: ADR-2248-01
+title: "Match the provider allow-list on the parsed host, with single-label wildcards"
+status: Proposed
+date: 2026-08-12
+tracking_issue: 2248
+deciders:
+  - "@erseco"
+reviewers:
+  - "@erseco"
+related:
+  prs: []
+  changes: []
+  adrs: []
+supersedes: []
+superseded_by: []
+ai_assistance:
+  tool: "Claude Code"
+  model: "claude-opus-5"
+---
+
+# ADR-2248-01: Match the provider allow-list on the parsed host, with single-label wildcards
+
+## Context
+
+`PROVIDER_URLS` is the SSRF allow-list that decides whether a platform JWT's
+`returnurl` may become a server-side request target. It is consulted from
+`getPlatformIntegrationParams()` in `src/utils/platform-jwt.ts`, which every
+platform-integration endpoint in `src/routes/platform-integration.ts` calls
+before contacting the platform.
+
+Until now the check was a raw string prefix test:
+
+```ts
+return config.urls.some(allowedUrl => url.startsWith(allowedUrl));
+```
+
+Commit `54373758` ("Security hardening", released in v4.0.2) changed the empty
+allow-list from *allow all* to *deny all*, closing an SSRF in which a holder of
+a valid platform JWT could drive the server to arbitrary hosts. That change was
+correct, but it left prefix matching in place and removed the only configuration
+that made multi-tenant deployments possible.
+
+Two consequences surfaced in issue #2248, reported by an operator running many
+Moodle tenants under a shared parent domain:
+
+1. Tenants have the shape `https://<tenant>.example.net/[path]`, where `<tenant>`
+   is unbounded. Because the variable part precedes the fixed part, **no prefix
+   can express the rule**. Enumerating every tenant is not operationally viable,
+   and the only "working" alternative — `PROVIDER_URLS=https://` — matches every
+   https URL by prefix and reinstates the SSRF.
+2. Prefix matching does not validate the host component at all. With
+   `https://moodle.example.com` configured, the entry is also satisfied by
+   `https://moodle.example.com.evil.com/` and by
+   `https://moodle.example.com@evil.com/`, whose effective hosts are `evil.com`.
+
+## Problem
+
+How should `PROVIDER_URLS` entries be matched so that a deployment can authorize
+an open-ended set of subdomains under a fixed parent domain, without weakening
+the fail-closed guarantee that the allow-list exists to provide?
+
+## Decision drivers
+
+- **Security**: the matched URL becomes a server-side request target, so the
+  check must be host-accurate and must not be looser than the operator believes.
+- **Operability**: multi-tenant deployments must not require an `.env` edit and a
+  restart per tenant.
+- **Auditability**: an operator reading `.env` must be able to tell exactly which
+  hosts are authorized, without reasoning about pattern semantics.
+- **Backward compatibility**: existing single-host deployments must keep working
+  with no configuration change.
+- **Fail closed**: an empty or malformed allow-list must still deny.
+
+## Options considered
+
+### Option 1: Keep prefix matching, document enumeration
+
+No code change; operators list every tenant.
+
+- Pros: no work, no new semantics.
+- Cons: does not solve the reported case at all — the rule is inexpressible when
+  the variable part is a subdomain. Leaves the host-accuracy weakness in place.
+
+### Option 2: Regular expressions in `PROVIDER_URLS`
+
+Treat each entry as a regex against the URL.
+
+- Pros: maximum expressiveness; covers cases no fixed syntax anticipates.
+- Cons, in order of weight:
+  - `parseEnvArray()` splits entries on commas, which collides with regex
+    quantifiers such as `{2,3}`. Supporting regex requires changing the
+    separator — breaking every existing deployment — or inventing an escape.
+  - The `returnurl` is attacker-controlled input matched against an
+    operator-authored pattern, introducing ReDoS exposure on a request path.
+  - Unanchored patterns are the classic allow-list bypass, and an unescaped `.`
+    matches more than the author intends. Both failures are silent.
+  - Backslash escaping in `.env` files is error-prone, and the resulting values
+    are hard to review.
+
+### Option 3: Match on the parsed host, with a single leftmost `*` label
+
+Parse both the entry and the URL, compare `hostname`, and let an entry use `*`
+as its leftmost label standing for exactly one DNS label — the semantics of TLS
+wildcard certificates.
+
+- Pros: expresses the reported rule exactly; fixed semantics cannot be written
+  insecurely; no ReDoS surface; no separator conflict; auditable at a glance;
+  and moving to `hostname` fixes the host-accuracy weakness as a side effect.
+- Cons: does not cover hypothetical rules more complex than "one label under a
+  fixed parent"; adds a small parser for entries.
+
+## Evidence
+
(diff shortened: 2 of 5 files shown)


erseco (changes requested):
Thanks @ignaciogros — overall, I think the direction of this PR is good. Moving away from raw `startsWith()` matching is the right approach for an SSRF allow-list, and the single-label wildcard semantics are clear and well tested. I also like that malformed entries fail closed individually instead of disabling the whole list.

I found two points that I think should be addressed before merging:

1. **An explicit port can be silently dropped when the entry contains a query or fragment without a path.** For example, `PROVIDER_URLS=https://example.com:8443?foo=bar` ends up being parsed with `host = example.com` and `port = null`, so it also allows `https://example.com:9000/`. The same happens with fragments and wildcard entries. Since this is part of the SSRF boundary, a configured constraint should never be silently relaxed. It would be safer to remove query/fragment before extracting path/port, or ideally let `URL` parse the full structure and only special-case the wildcard hostname. A regression test for this case would be useful.

2. **The deployment documentation says that `PROVIDER_URLS`, `PROVIDER_TOKENS`, and `PROVIDER_IDS` all map by position, but the implementation does not bind URLs to provider IDs.** Only `PROVIDER_IDS[n]` and `PROVIDER_TOKENS[n]` are positionally associated; `PROVIDER_URLS` is checked independently with `some(...)`. I think the docs should make that distinction explicit, unless positional URL/provider binding is actually intended.

Apart from those two points, the wildcard matching, host normalization, malformed-entry handling, and the additional test coverage look solid. Thanks for the thorough work on this.

ignaciogros:
Thanks for the review. Both points were valid and have been addressed.

**1. Dropped port.** Confirmed: `PROVIDER_URLS=https://example.com:8443?foo=bar` also allowed `:9000`. I took your second suggestion and let `URL` parse the entry, special-casing only the wildcard hostname. Two constraints are read explicitly because they don't survive that parse:

- the **port** comes from the authority isolated before any `?`/`#`, and from the raw text rather than `URL.port` — the latter drops a scheme-default port, which turned an explicit `:80` on an http entry into "any port" (same class of bug, also fixed);
- a **path** constrains matching only when the entry writes one, since `URL.pathname` is `/` both for no path and for a lone trailing slash.

`normalizeEntryHost()` is gone, now redundant. Six regression tests cover query-after-port, fragment-after-port, the wildcard variant, explicit `:80`, query-without-path, and `..` normalization in the entry path.

**2. Positional mapping.** The docs were wrong. `doc/deployment.md` and `.env.dist` now state that only `PROVIDER_IDS[n]` ↔ `PROVIDER_TOKENS[n]` are positional (the id selects the signing secret) and that `PROVIDER_URLS` is a single global allow-list, so every entry is authorized for every provider.

Positional URL binding was not intended, and I've left it as a follow-up in the ADR rather than implementing it here: tokens signed with `APP_SECRET` carry no `provider_id`, so they'd need a fallback to the whole list. That's a separate decision deserving its own ADR. The residual exposure is a confused deputy between hosts the operator already authorized, not the SSRF this change targets.

The ADR also records the underlying rule now: a constraint the operator writes is never silently relaxed, and an entry the parser cannot decompose is discarded whole.

erseco (approved):
Thanks @ignaciogros — the two issues from my previous review are addressed and the changes look good to me now.

One small follow-up to consider: the docs currently describe `PROVIDER_IDS[n]` / `PROVIDER_TOKENS[n]` as selecting the signing secret by `provider_id`, but the production integration flow appears to call `decodePlatformJWT(jwtToken)` without passing a provider ID, so it may still fall back to `APP_SECRET`. This is not a blocker for this PR, but it would be worth either clarifying the documentation or tracking/fixing that behavior in a separate issue.

Approved. Thanks for the improvements.

PR #2254, opened 2026-08-13
fix(preview): serve PDF.js as .js so nginx does not break the PDF preview

erseco:
## Problem

Previewing a Text iDevice with an embedded PDF fails on https://online.exelearning.net:

```
Failed to load module script: Expected a JavaScript-or-Wasm module script but the
server responded with a MIME type of "application/octet-stream".
[Preview] PDF.js load failed: TypeError: Failed to fetch dynamically imported
module: https://online.exelearning.net/libs/pdfjs/pdf.min.mjs
```

The preview iframe is sandboxed, so the browser refuses to run its native PDF plugin
("Failed to load ... as a plugin, because the frame into which the plugin is loading
is sandboxed"). PDF.js is the designed fallback — and it is the one that cannot load,
so the embed renders as nothing.

## Root cause

nginx has no `.mjs` entry in its bundled `mime.types` — still true on the current
1.31 image — and falls back to `default_type application/octet-stream`. Browsers
apply strict MIME checking to module scripts, so `import('.../pdf.min.mjs')` is
rejected outright.

`Dockerfile.static` uses stock `nginx:alpine` with no MIME configuration, so every
deployment of that image is affected. Every other serving layer already maps `.mjs`
correctly, which is why this never showed up in development:

| Serving layer | `.mjs` Content-Type | PDF preview |
|---|---|---|
| Preview Service Worker | `application/javascript` | ok |
| Electron `app://` | `application/javascript` | ok |
| Bun/Elysia (dev and server mode) | `text/javascript` | ok |
| Cloudflare (online.exelearning.dev) | `application/javascript` | ok |
| nginx static image (online.exelearning.net) | `application/octet-stream` | **broken** |

## Fix

Two layers, because they solve different halves of the problem.

**1. Rename the vendored PDF.js bundles to `.js`.** The static build is also published
as a release ZIP and self-hosted on servers we do not control, so the artifact has to
be portable on its own instead of depending on server configuration. The extension
does not affect parsing — module-ness comes from `import()` / `type="module"`, not
from the filename. Verified against stock `nginx:alpine` with no config: both bundles
are now served as `application/javascript`.

**2. Add the missing MIME mappings to the static image.** `.mjs` was not the only gap:
`.ttf` and `.otf` (about 150 files) were served as `application/octet-stream` too. The
table also covers `.vtt`, `.glb`, `.gltf`, `.stl` and `.webmanifest` — content types
eXeLearning produces, which the same nginx would get wrong the moment they show up.
A missing entry fails silently and is miserable to diagnose; an unused one costs a line
of config. A `types` block in `conf.d/` merges with the base `mime.types`, so existing
mappings are preserved.

## Scope checked

- `public/app/common/exe_math/sre/require.mjs` is the only other `.mjs` shipped. It is
  imported only in speech-rule-engine's Node branch, never in the browser, so it was
  not a live failure — and the nginx mapping now covers it anyway.
- Exported packages do not bundle PDF.js; exports are not sandboxed and use the native
  viewer, so they were never affected.
- Subtitles are unaffected in the editor: `AssetManager` hands `<track>` a `text/vtt`
  blob URL, so no server MIME table is involved. They do go through the hosting server
  once a package is exported, which is part of why the table above is not minimal.

## Verification

- `make fix`, `make test-unit` (8007), `make test-frontend` (14121),
  `make test-integration` (727) — all green.
- E2E: `pdf-noext-iframe-no-download.spec.ts` on chromium, plus the full static suite.
- Real nginx serving the built `dist/static`: `pdf.min.js` and `pdf.worker.min.js`
  return `application/javascript`; `.ttf` returns `font/ttf`, `.otf` returns
  `font/otf`, `.mjs` returns `text/javascript`.

## Tests

`scripts/static-assets-mime.spec.ts` (new) fails the build if a `.mjs` reappears under
`public/libs/pdfjs/`, if a consumer imports one, or if the MIME config stops being
copied into the image. `public/libs/pdfjs/README.md` documents the rename so the next
PDF.js update does not silently undo it.

## Deploying

online.exelearning.net needs a rebuild of the static image to pick this up.

Heads-up, unrelated to this change and not fixed here: `Dockerfile.static` runs
`bun install --frozen-lockfile`, and that currently fails on `main` because `bun.lock`
is behind `package.json` (`ioredis` is declared `^6.0.0`, the lockfile still resolves
`5.10.0`). The static image cannot be rebuilt until the lockfile is refreshed, so that
needs to land first for this fix to reach production.


juanda (approved):
Revisado con codex y comprobado manualmente, los PDF se ven bien en modo edición y previsualización tanto en Firefox como en Chrome

ignaciogros (approved):
Thanks, @erseco.

PR #2260, opened 2026-08-14
perf(static): shed dead vendored weight and repack incompressible payloads (-8.5 MB zip / -25 MB installed on top of the perf series)

erseco:
## Summary

Next installment of the effort to make the static eXeLearning distribution practical to package inside the WordPress plugin and publish through WordPress.org (related to #1542). The WordPress Plugin Review Team's position is explicit: the editor must stay bundled in the plugin (no runtime code download), and the path forward is to "resolve those optimizations which will significantly reduce the size of the plugin". The plugin ZIP is ≈99% `dist/static`, and reviewers flagged both the download size and the unzipped footprint on constrained hosting.

Every byte class was traced to its consumers before touching it (build provenance, static/dynamic reference analysis, real-browser verification), and each optimization ships with regression tests.

> **The MathJax part of this PR has moved out**, into PR #2351, which is stacked on this branch. See *Relationship with the MathJax PR* below for why. The figures here are remeasured without it.

- **TikZJax repack (electrical-circuits): −1.4 MB zip from bytes deflate couldn't touch.** The vendored `tikzjax.js` embeds 6.4 MB of base64-of-gzip data URIs (base64-of-gzip defeats the outer ZIP), plus 3.6 MB of TeX fonts. At build time the dist copy is repacked into two zstd sidecars (payloads stored *decompressed* so zstd sees real TeX data), loaded lazily on first editor use via the already-vendored `fzstd`. The pristine vendored file stays in git and keeps serving server mode; structural guards fail the build loudly if upstream ever changes shape.
- **2.2 MB of TinyMCE no configuration can load** (22 plugin dirs incl. the 1.4 MB emoticons databases, the legacy mobile theme, unused skins) — the editor's plugin surface is closed (one main init + five iDevice mini-configs, no `external_plugins`, no dynamic loads), now pinned by tests in both directions.
- **~2 MB of `app/` sources already compiled into `app.bundle.js`** shipped loose next to the bundle. The removal list is derived from esbuild's module graph at build time (never hardcoded), with the two ES-module entry points the page loads loose kept, and a plausibility guard.
- **`data/bundle.json` → `bundle.json.zst`** (3.2 MB pretty-printed JSON precached by the service worker on every install → 0.4 MB), dropping an embedded copy of the bundle manifest that had zero consumers. Same fzstd pattern as #2037, byte-identical vendored decoder.
- **Fonts:** editor UI fonts (Inter variable, Catamaran, Material Icons) converted TTF/OTF→WOFF2 (lossless, glyph-order-verified); OpenDyslexic (both copies, incl. the a11y toolbar) →WOFF2 with export generators updated; legacy `.woff` fallbacks dropped where a `.woff2` sibling ships (pre-2016 browsers fall back to system fonts in exported content).
- **Duplicate/oversized default images:** three default images shipped byte-identical in `edition/` and `export/` (the editor now previews the export copy via the established redirect pattern); the magnifier's 443 KB sample photo and the complete iDevice's 379 KB flat-art PNG were recompressed (1024px q85 JPEG / WebP q85, visually verified including at the magnifier's 600% zoom).
- **Server-only and development files** pruned from the static dist with a loud-failure list (admin page assets, login styling, test helpers, inlined schemas, dev docs); first-party editor JS minified in the dist copy only (sources stay readable in git; export-copied libraries excluded so exported packages remain byte-identical between server and static modes).
- **Bug fix:** the four parallel base-library lists still shipped the two bootstrap sourcemaps into every exported package (~0.9 MB each) and would have 404'd during static assembly once #1910 and #2036 landed together. A follow-up commit also removed the `sourceMappingURL` comments the dist files still carried, so the 404 does not move from the build into published content — the same class of note @ignaciogros left on #2036.

## Measurements (exact bytes, reproducible)

Method: `make build-static`, then the release workflow's exact packaging (`cd dist && zip -r … static/`). Both builds from clean trees on the same machine, same day.

Since this PR was opened, #1910, #2036, #2037 and #2038 have all merged, so the old A/B/C table collapses into a single before/after against current `main`:

| State | files | `dist/static` raw | release ZIP |
|---|---:|---:|---:|
| `main` @ `049051e62` | 3,463 | 75,029,008 | 33,667,602 |
| **This PR** @ `0b01b61e7` | **3,068** | **55,843,919** | **26,188,450** |

- **ZIP −7,479,152 B (−22.2 %); installed −19,185,089 B (−25.6 %); 395 files fewer.**
- For #1542: the WordPress plugin ZIP tracks the release ZIP + ≈0.44 MB of plugin files → **≈26.6 MB plugin ZIP / ≈56 MB unzipped**, from ≈34.1 MB / ≈75 MB on `main` today.

## Relationship with the MathJax PR (#2351)

This PR originally removed 6.2 MB of MathJax 4 leftovers from `public/app/common/exe_math/`, making that tree consistently 3.2.2. That commit has been **reverted here and the work moved to #2351**, which is stacked on this branch. Three reasons:

1. **It answered @ignaciogros's review by hiding UI rather than fixing it.** Removing the unreachable files left three contextual-menu entries pointing at files that were no longer there — Accessibility → Activate, Collapsible Math, and Math Settings → Math Renderer → CHTML. Two of them also fire at page load from a setting MathJax persists in `localStorage`, without anyone opening the menu, so the 404 moved out of the static build and into published content. The stacked PR fixes those same entries by shipping the components they need.
2. **It coupled this branch to a product decision.** The stacked PR carries ADR-2259-03, an open question about whether to ship the Speech Rule Engine at all. Static size should not wait on that.
3. **One source of truth for `exe_math/` at every point in the history.** The alternative — merge this, then let the stacked PR empty and refill the same directory — produces a 100-file conflict and an unreadable history for that path.

Cost: 6,155,759 B raw / ≈1.06 MB of the release ZIP that this branch no longer claims. The stacked PR more than replaces the *correctness* half of that work and lands a self-contained MathJax tree; the size half returns partly, depending on the SRE decision.

Both PRs touch `scripts/static-bundle/static-index.html`, in different regions, and nothing else: the stack rebases cleanly.

## Relationship with the rest of the optimization series

| PR | Status | This PR's relation |
|---|---|---|
| #1910 | **merged** | Untouched; this PR fixes the `BASE_LIBS` sourcemap entries that would 404 during its static assembly |
| #2036 | **merged** | Untouched; this PR extends the same philosophy to docs/dev files via a new loud-failure prune list (different mechanism, different bytes) |
| #2037 | **merged** | This PR reuses its pattern and the byte-identical `fzstd` UMD for `bundle.json.zst`. Dataset bytes untouched |
| #2038 | **merged** | Orthogonal: it does not change the release ZIP at all |
| #1593 | open | Deliberately avoided its territory: simplelightbox variants and the html2canvas triplication are left for it; `libs/yjs/build` is excluded from the dist but kept in git as its build inputs |

## Why these bytes are safe (evidence classes)

- **TinyMCE prune:** exhaustive enumeration of every `plugins:`/`skin:`/`theme:`/`external_plugins` config in the repo; exports never include TinyMCE. Tests assert every configured plugin exists on disk and pruned payloads stay gone.
- **Dead-dup app sources:** the exclusion list is *derived from the esbuild metafile of the real `app.js` graph at build time*; e2e-static runs the entire suite against the pruned dist.
- **Prune/minify lists:** every listed path must exist or the build throws (no silent drift); the pruner refuses any path containing "license".
- **TikZJax:** unit specs cover the transform, the pack format, the known upstream-corrupt asset (`tex_files/pgfplotscoordprocessing.code.tex.gz` ships with a broken gzip stream on `main` today — its "File not available" behavior is reproduced byte-for-byte), and stale-layout guards; a static E2E spec opens the circuits editor in a real browser and resolves a TeX asset and a parsed font through the sidecars.
- **Fonts:** conversions done with fontTools 4.60.2 + brotli, glyph order asserted identical before deleting sources; every `@font-face`/generator/export-list reference updated in the same commit and grepped for remnants.

## Compatibility

- **Static PWA / offline:** the service worker precache list is updated (`bundle.json.zst`, fzstd) and asserted by spec; everything else was already lazy (network-first + IndexedDB). Full static E2E suite green.
- **Server mode:** untouched by design — the TikZJax repack, pruning and minification act on the dist copy only; `public/` keeps serving pristine files (pack loaders fall back to loose files).
- **Exports (HTML5/SCORM/EPUB/IMS):** exported packages *shrink* (OpenDyslexic, theme fonts, default images, no more bootstrap sourcemaps) but remain byte-identical between server and static modes — export-copied libraries were explicitly excluded from minification.
- **Electron:** ships the same dist; every reduction applies; no Electron-only file was touched.
- **iDevices:** electrical-circuits covered by new unit + E2E specs; magnifier/complete/crossword default-image behavior covered by colocated tests; no iDevice removed, none lose functionality.
- **Localization:** all interface locales still ship (inside `bundle.json.zst`); no translation files touched.

## WordPress.org considerations

- The Plugin Review Team confirmed (Aug 2026 thread) that the 10 MB form limit is not a hard policy ceiling and that plugins in the 20–30 MB range have been processed; the request is to optimize as far as defensible before re-review.
- Minified first-party JS is compliant with detailed guideline 4 ("A link in the readme to the development location"): sources are in this repository; the wp-exelearning `readme.txt` should link it (tiny follow-up in that repo).
- The `.zst` payloads are data-only, generated reproducibly by `scripts/` in this repo, with the decoder vendored locally (no external services, per guideline 8). Should a reviewer object to compressed data files, the loaders keep a plain-file fallback tier, so a build flag that skips compression remains a cheap plan B.
- No third-party license/notice file was removed; the runtime-fetched attribution documents (`libs/README.md`, `libs/LICENSES.md`) are exempted from pruning by test.

## Tests

- `make fix` — clean.
- `npx vitest run` — 15,346 pass / 288 files.
- `bun test ./src ./scripts` — 8,323 pass; 2 fail in `Assets API v1`, which pass 31/31 when that spec runs alone: pre-existing cross-spec pollution, unrelated to this branch (it touches no file under `src/routes/api`).
- `make test-integration` — 772 pass, 0 fail.
- `make test-e2e-static` — full suite against the built dist.
- New/extended specs: `prune-dist.spec.ts`, `minify-dist.spec.ts`, `zstd.spec.ts`, `repack-tikzjax.spec.ts`, `vendored_assets.test.js`, TinyMCE surface tests, apiCallManager zst tiers, electrical-circuits font-pack tiers, edition-dedup guards, and `electrical-circuits-tikzjax-static.spec.ts` (real-browser sidecar round-trip).
- Adversarial audit: full-diff review, removal falsification pass, license sweep, dist reference-integrity scan (manifest entries, css url() targets, SW precache) and an independent clean-tree rebuild reproducing the size numbers.

## Risk

- The dead-dup prune derives the removal list with esbuild's resolver while `app.bundle.js` is built by Bun; the graphs agree today (verified: dynamic-import inlining, workers, ES-module entries) and the full static E2E suite guards divergence, but a shared-bundler derivation would be stronger.
- The TikZJax repack forks *behavior* of a vendored file at build time. Mitigations: pristine file kept in git, exact-match structural guards (build fails loudly on upstream change), unit + real-browser E2E coverage, server mode unaffected. Residual: a transient network failure on first pack fetch caches the failure for the session (same class as today's per-URI behavior).
- Legacy-browser rendering of exported content (IE11/Safari<10) falls back to system fonts where `.woff` fallbacks were dropped — cosmetic, on platforms that cannot run the interactive content anyway.
- The magnifier/complete default images are now lossy (visually verified, incl. 600% zoom); authors who kept the defaults get slightly softer pixels.

## Rejected alternatives (investigated, not shipped)

- **zstd-at-rest for large JS libs** (mermaid 2.75 MB, pdf.worker): they already deflate well inside the ZIP; zstd would save only ~100–150 KB each at the cost of loader indirection everywhere. Not worth it.
- **Unifying the three three.js copies** (three-d-viewer r160 ESM, model-viewer's embedded r163, three-sixty r147 UMD): a real refactor with regression surface for ~150 KB zip. Deferred.
- **quext shared-asset dedup across ~30 iDevices** (~0.5–1.2 MB): the export path requires per-package self-containment; the edition-side half is mechanical but touches ~30 editors — left as a documented follow-up.
- **html2canvas/simplelightbox dedup:** already covered by #1593; duplicating it here would conflict.
- **TikZ font subsetting / pruning "exotic" TeX fonts:** the reachable font set equals the compiled font table and user labels can request any family — graceful degradation was not worth the risk.
- **Splitting translations per locale:** ~0 net ZIP effect (all locales still ship); only an install-time win. Deferred.
- **Removing the `example` iDevice or LOMLOE regional datasets:** product decisions, not build waste.

Each optimization is a separate commit with its own rationale, exact numbers and tests — the commit series is the detailed change log.

## Follow-ups

- #2351 — the MathJax work, stacked on this branch.
- wp-exelearning: add the editor repository + build instructions link to `readme.txt` (guideline 4).
- Optional next wave: quext edition-side dedup, manifest-alias export dedup, three.js unification.


cristinavaldera (approved):
Nothing to report so far! Thank you

erseco:
Pushed `e0245e3b9` (on top of a fresh merge from `main`, `00d67d579`) closing a gap this PR opens.

Removing `bootstrap/bootstrap.bundle.min.js.map` and `bootstrap/bootstrap.min.css.map` from `BASE_LIBRARIES` and from `BASE_LIBS` in `build-resource-bundles.js` is only half the change: the dist files still ended with `//# sourceMappingURL=bootstrap.bundle.min.js.map` and its CSS counterpart. So the 0.9 MB does come out of every exported package, but each package then 404s there the moment a reader opens DevTools — the 404 moves out of the static build and into published content, which is worse than where it started. Same class as the note @ignaciogros left on #2036.

The fix is a deletion in the two vendored dist files (48 and 45 bytes) rather than a runtime transform:

```
public/libs/bootstrap/bootstrap.bundle.min.js   80,743 → 80,695
public/libs/bootstrap/bootstrap.min.css        232,018 → 231,973
```

`fetchLibraryFiles` has two implementations (`FileSystemResourceProvider` server-side, `ResourceFetcher.js` in the browser) and `libs.zip` is assembled in a third place, so a strip-on-copy would be three edit sites, three sets of tests and a cost on every export. All four consumers — exports, the resource bundle, `dist/static` and server mode — read the same two files, so deleting the line once keeps them consistent for free. The `.map` files stay in git: this is what jQuery settled on in 1.11/2.1 — drop the comment from the distributed build, keep the maps available to attach by hand.

Pinned in `vendored_assets.test.js`, since a Bootstrap upgrade that drops in fresh dist files brings the comment straight back. Verified it fails when the comment is restored:

```
AssertionError: bootstrap/bootstrap.bundle.min.js still announces a source map
```

This pairs with #2036, whose tripwire in `constants.spec.ts` fails if these `.map` entries are ever removed without this half.

### Tests

- `make fix` — clean.
- `bun test ./src/shared/export ./scripts` — 2,383 pass / 0 fail.
- `npx vitest run` — 15,276 pass / 286 files.
- `make test-integration` — 772 pass / 0 fail.
- `make test-e2e-static` — 432 passed, 1 failed, 134 skipped. The failure is in `idevices-draganddrop.spec.ts` and is not related: three consecutive runs of that file failed a **different** test each time (`move iDevice from one block to another`, `when iDevice is moved out`, `not corrupt block references`), 7 of 8 passing every run, and the third of those runs was a control with the `sourceMappingURL` comment restored in the built dist — i.e. the pre-change state fails the same way. Flaky spec on this branch, worth its own issue rather than a retry bump.

ignaciogros (changes requested):
Thanks for the proposal. I think it should be accepted because:

* The PR keeps everything consistent on 3.2.2 and adds a test that prevents versions from being mixed again.
* No functionality that currently works is lost. Rejecting it would only leave 6 MB of dead code.
* It fixes another bug: `input/mml/entities.js` was on version 4.

Just two details, being the second one the most important:

1. **`src/routes/config-params.ts:350-357` — update the help text.** It still says "about 8 MB" (now ~2.5 MB) and mentions "accessibility tools". Please do not remove the original text from the code, so we don't have to translate it again later. Maybe we could just leave an `old_help` text.
2. **The context menu still offers "Accessibility → Activate".** It will result in a **404**. I suggest hiding those entries until #2259.

The definitive solution will come with #2259.

erseco:
@ignaciogros thanks — both points were right, and the second one turned out to be bigger than either of us thought. That is why the answer is not a patch here but a change of shape: **the MathJax part of this PR has been reverted and moved to #2351**, stacked on this branch.

### Why moving it answers you better than patching it

Your point 2 was about entries the context menu still offers after the removal. Tracing it, there were **three**, not one:

| Menu entry | Requests | Result |
|---|---|---|
| Accessibility → **Activate** | `a11y/explorer.js` | 404 — the one you spotted |
| Accessibility → **Collapsible Math** / Auto Collapse | `a11y/complexity.js` | 404 |
| Math Settings → **Math Renderer → CHTML** | `output/chtml.js` | 404 — this PR deleted `output/` too |

And worse than a menu click: MathJax persists those settings in `localStorage`, scoped to the **origin**, and acts on them while the document is being built. `checkLoadableItems()` reads `explorer` and `collapsible`; `applySettings()` acts on `renderer`. So anyone who ever pressed "Activate" on `main` — where the setting was saved and *then* the code crashed — would get the 404 **on every page load, without opening the menu**, and in an LMS the key travels between packages.

I could have hidden those entries here. But hiding them is covering a hole this PR digs, and the follow-up PR removes the hiding again, because it fixes the same entries by **shipping the components they need** instead. Reverting keeps one source of truth for `public/app/common/exe_math/` at every point in the history, and it unblocks this branch, which otherwise waits on a product decision about the Speech Rule Engine that has nothing to do with static size.

### Your point 1, and your concern about the translations

Also moved, and solved the way you asked — with one change I think you will prefer.

You suggested keeping the old English text in the code as an `old_help` so `make translations-cleanup` would not drop the ten translated `<target>`s. That works, but `config.ts` translates the params object recursively, so an `old_help` key would be translated and served to every client. The extractor is a plain text scan over the source, so a **comment** does the same job with nothing in the payload:

```ts
/*
 * The retired text stays here on purpose: `make translations-cleanup` drops any
 * trans-unit whose key is not found in the sources […]
 *   `${TRANS_PREFIX}Formulas are rendered even if this is disabled. […] (about 8 MB) […]`
 */
help: `${TRANS_PREFIX}Formulas are rendered even if this is disabled. Enable this option to include the full MathJax library in exports, which adds its contextual menu and interactive features.`,
```

Verified the extractor picks up both strings, so the ten existing translations stay parked.

On the wording itself I went further than you asked, and I would like your opinion: the new text names **neither a size nor a feature list**. The directory has been 8 MB, 2.4 MB and 6.8 MB inside a single release cycle, and since the English text *is* the XLF key, every edit orphans ten locales. Quoting "~2.5 MB" would have been accurate for about a week. If you would rather keep a figure, say so and I will put it back — it is one string, in one PR.

### What changed in the numbers

`main` has moved a lot since this was opened: #1910, #2036, #2037 and #2038 have all merged, so the old A/B/C table no longer describes anything. Remeasured from clean trees on the same machine today, with the MathJax part removed:

| | files | `dist/static` raw | release ZIP |
|---|---:|---:|---:|
| `main` @ `049051e62` | 3,463 | 75,029,008 | 33,667,602 |
| this PR @ `0b01b61e7` | 3,068 | 55,843,919 | 26,188,450 |

**−7,479,152 B of ZIP (−22.2 %), −19,185,089 B installed (−25.6 %), 395 files fewer.** For #1542 that is a plugin ZIP of ≈26.6 MB and ≈56 MB unzipped, from ≈34.1 MB / ≈75 MB today. The description is updated.

The revert costs this branch 6,155,759 B raw / ≈1.06 MB of ZIP. Part of it comes back in the stacked PR; how much depends on the SRE decision there, which is worth ≈470 KB of the plugin ZIP on its own and which I have put to you and @cristinavaldera in ADR-2259-03.

ignaciogros (approved):
Thank you, @erseco!

PR #2264, opened 2026-08-14
feat(auth): change password from the user menu, admin panel and CLI (#2261)

### LINKED ISSUE #2261: New Make command: change password
Add a new Make command to facilitate changing a user’s password.

Perhaps also add a new option to the control panel.

erseco:
Closes #2261

Users who sign in with an eXeLearning password can now change it themselves, administrators can reset it, and there is a Make/CLI command for recovery. Accounts that authenticate elsewhere (CAS, OpenID Connect, SAML, guest) are refused everywhere — eXeLearning never gives an SSO account a local password.

## How it works

**User menu → Change password.** The entry is rendered only when the server says the session may change its password. The dialog asks for the current password, the new one and a confirmation.

**Admin → Users → ⋮ → Reset password.** The three-dot menu now groups the per-row actions (*Log in as* moved there too). The administrator sets a new password without knowing the current one. The entry only appears for local accounts.

**Command line**, for recovery. Asks twice, hidden, so the password never lands in the shell history or a process listing:

```bash
make change-password EMAIL=user@example.com   # → bun cli user:password user@example.com
```

`--password-stdin` is available for automation. The password is never accepted as a command-line argument.

## Screenshots

**User menu**

User menu with Change password

**Change password dialog** — with the length hint and an advisory strength meter

Change password dialog

**After a successful change** — fields cleared, session kept

Password changed successfully

**Admin: per-row actions menu**

Admin users table with the actions menu open

**Admin: reset dialog**

Admin reset password dialog

## Endpoints

| Method | Path | Purpose |
|---|---|---|
| `PATCH` | `/api/user/password` | Change your own password. Body `{ currentPassword, newPassword }`. No user id — the session decides whose password changes. |
| `PATCH` | `/api/admin/users/:id/password` | Admin reset. Body `{ newPassword }`. Requires `ROLE_ADMIN`. |

Statuses: `200` success · `401` not authenticated / wrong current password · `403` account or session not allowed · `404` unknown user (admin) · `422` invalid body or password.

`GET /api/admin/users[/:id]` now also returns a derived `can_reset_password` boolean. Password hashes are never returned.

## Caveats

- **Only eXeLearning-managed passwords.** CAS, OIDC, SAML and guest accounts are refused; those users change their password with their identity provider. Eligibility is *not* read from the `password` column — those accounts also store a random bcrypt hash there — nor from `APP_AUTH_METHODS`, which describes the installation and not the account.
- **Impersonation is excluded.** An impersonated token inherits the administrator's `authMethod`, so `!isImpersonated` is checked explicitly. Neither the menu entry nor the endpoint works while impersonating.
- **Offline/desktop installations** have no real login, so the feature is hidden and the endpoint refuses.
- Hidden UI is not the boundary: every rule is re-checked server-side.
- The minimum stays at **4 characters**, unchanged and consistent with account creation. The strength meter is advisory and never blocks.
- bcrypt (cost 10) is unchanged — no hashing migration in this PR.

## Docs

- [`doc/development/authentication.md`](doc/development/authentication.md) — new "Changing Passwords" section covering the UI, the admin reset and the CLI.
- [`doc/development/environment.md`](doc/development/environment.md) — `make change-password` in the command table.
- New `ADR-2261-01` recording why eligibility is derived from account metadata plus session claims.

## Tests

Backend unit + route tests for the full account/session matrix (local, guest, CAS, OIDC, SAML, offline, impersonated), CLI command and hidden-prompt tests, Vitest specs for the modal and the strength meter, and a Playwright spec covering the local flow end to end plus the guest refusal.


ignaciogros (approved):
It works great.

Thanks.

PR #2265, opened 2026-08-15
Configure spell checker languages in desktop builds

### LINKED ISSUE #1878: Allow configuring spell checker languages in desktop builds
In the desktop versions of eXeLearning, the spell checker language is currently not configurable from the application.

This creates problems for multilingual users, because the spell checker may not match the actual language being used for writing. For example, users working in Catalan may still get suggestions in Spanish or English even if Catalan is installed in the operating system.

It would be useful to allow configuring the spell checker languages explicitly in desktop builds where Electron supports it.

Suggested behavior:
- add a desktop preference for spell checker languages,
- include an `Automatic` option to preserve the current behavior,
- allow selecting one or several active languages,
- use the list of languages supported by Electron on each platform.

This would improve support for multilingual authoring and make spell checking more predictable.

Technical note:
Electron supports configuring spell checker languages on Windows and Linux through session APIs such as `setSpellCheckerLanguages(...)`. On macOS this may need to remain system-controlled.

milos-plavsic:
## What changed

- expose Electron's available and selected spell-checker languages to the renderer
- add a reusable multi-select preference control for choosing one or more languages
- provide an explicit **System default** option that resolves the OS locale to an available dictionary, with Electron's documented `en-US` fallback when available
- keep **System default** exclusive from explicitly selected languages
- filter unsupported and duplicate language selections before applying them
- persist both the selection mode and explicit language list, then apply the persisted state once at startup
- keep settings reads side-effect free and prevent spell-checker failures from blocking other preference saves
- leave macOS language selection to the native operating-system spell checker
- add focused unit, preload, preference-modal, and Playwright regression coverage

## Why

TinyMCE delegates spell checking to Chromium/Electron. Windows and Linux do not automatically detect the language being typed, so desktop users need a way to configure the existing Electron spell checker independently of the interface language. Multiple dictionaries are useful for authors who create multilingual educational content.

Closes #1878.

## Validation

- `make fix`
- focused spell-checker controller, preload bridge, renderer preference, and modal regression tests
- GitHub Actions packaging and installation tests pass on Windows, macOS, and Ubuntu
- GitHub Actions E2E suites pass for Chromium, Firefox, and the static build
- manually tested successfully on Windows by @erseco

The artifact-cleanup `Resource not accessible by integration` failure is an expected permission limitation for contributor/fork pull requests and is unrelated to this implementation.


milos-plavsic:
CI note: all E2E matrix test jobs completed successfully. The workflow failed only in the final `merge-report` cleanup step, after the merged Playwright report had been created and uploaded. The cleanup attempted to delete intermediate artifacts and received `403 Resource not accessible by integration` (`actions=write` required), which is consistent with the restricted `GITHUB_TOKEN` permissions for a fork-originated PR. This is unrelated to the spell-checker implementation or test results.

erseco (changes requested):
The implementation looks good overall, but the `Automatic` / empty-selection behavior needs to be corrected before merging.

`setSpellCheckerLanguages([])` does **not** restore the operating-system default for the current session. Electron documents that an empty language list falls back to `en-US`; Electron only tries to populate an empty spell-checker language setting from the OS locale during application startup.

This means the current help text — `Select no languages to use the operating system default.` — is misleading and can switch a non-English Windows/Linux user to English after clearing the selection.

Please represent **System default** explicitly and restore/resolve the current OS locale when that mode is selected, instead of calling `setSpellCheckerLanguages([])` as the automatic/default behavior.

The rest of the approach looks good: using `availableSpellCheckerLanguages`, filtering unsupported values, supporting multiple languages on Windows/Linux, and leaving macOS language selection to the native OS spell checker.

milos-plavsic:
Thanks @erseco — addressed in `27b7231`.

The updated implementation now:
- represents **System default** explicitly in the preferences UI;
- persists whether System-default mode is active separately from Electron’s selected dictionaries;
- resolves `app.getSystemLocale()` against `session.availableSpellCheckerLanguages`, including base/regional matching;
- reapplies the resolved system dictionary when settings are loaded after restart; and
- never calls `setSpellCheckerLanguages([])` (including when no dictionaries are available).

Explicit multi-language selection and macOS behavior remain unchanged. Focused verification passes: 7/7 main-process spell-checker tests and 31/31 frontend preferences tests.

erseco (changes requested):
Thanks for addressing the previous review. The explicit **System default** mode is a good improvement and the original issue is mostly resolved.

There is still one fallback case I think should be corrected before merging.

In `resolveSystemSpellCheckerLanguages()`, the current fallback is effectively:

```js
exactMatch || baseMatch || regionalMatch || current[0] || availableLanguages[0]
```

Using `current[0]` means that switching to **System default** can keep a previously selected manual language even when it has no relation to the current OS locale. For example, if the previous manual language is `es` and the system locale has no matching dictionary, System default may continue using Spanish.

Likewise, `availableLanguages[0]` is not a documented Electron fallback; the ordering of `availableSpellCheckerLanguages` should not be given semantic meaning.

Please avoid both of those fallbacks for System default. If the system locale cannot be matched, use Electron's documented fallback semantics instead (notably `en-US` when available), without carrying over the previous manual selection.

The test that currently expects an unknown locale to fall back to the current `es` selection should be adjusted accordingly.

One unrelated note: please don't worry about the artifact-cleanup / `Resource not accessible by integration` failures in Actions. Those are expected for contributor/fork PRs because the token has restricted artifact write/delete permissions; they are not caused by this implementation.

milos-plavsic:
Thanks, @erseco. Addressed in 44a26b6.

`resolveSystemSpellCheckerLanguages()` no longer considers the current manual selection or the ordering of `availableSpellCheckerLanguages`. It now resolves an exact/base/regional locale match, then uses `en-US` when that documented fallback is available; otherwise it returns no arbitrary language selection.

I also updated the unknown-locale regression test so a previous manual `es` selection no longer carries over, and added coverage for the case where neither the locale nor `en-US` is available.

Focused verification: `bun test app/spell-checker-settings.spec.js` — 7 passed, 0 failed.

Understood regarding the contributor/fork artifact-cleanup permission failures; I will treat those as unrelated to this implementation.

erseco (dismissed):
Thanks for your patience through these review rounds, @milos-plavsic — the fallback semantics from the previous round look good now. I ran a more intense validation pass on the branch (unit suites, packaging checks, and a full trace of the renderer/modal flow) and found some additional issues. The main one is a blocker, although the fix itself is small.

## Blocker

**The preference row is unreachable in the UI.** `spellCheckerLanguages` is added without a `category`, while every other preference carries `category: 'General settings'` (see `src/routes/config-params.ts`). With two distinct categories present, `makeCategoriesTabs()` (`modalProperties.js:419`) turns the previously tab-less Preferences dialog into a tabbed one with a blank second tab, and `makeRowElement()` stringifies the missing category into `category="undefined"`, which no tab's `row.getAttribute('category') == categoryTitle` comparison can ever match. Net effect on Windows/Linux — the target platforms — is a blank extra tab and a spell-checker row that is never visible.

The fix is to give the new entry a category that resolves to the **same runtime string** as the existing preferences (mind the translation: the server sends its categories through the i18n catalog, so the renderer-side entry has to end up identical at runtime or a second tab still appears). The unit tests pass because they cover the preference object and the IPC mapping but not the modal's category/tab rendering — please add a regression test that would have caught this.

## Should fix

1. **Sentinel exclusivity.** Selecting *System default* together with concrete languages silently discards the languages: the submitted array contains `__system_default__`, `languages.includes(SYSTEM_DEFAULT)` wins (`app/main.js:1552-1554`), and the UI resets with no feedback. Make the sentinel exclusive in the control (or prefer the explicit languages when both are present). Related: an empty selection re-enables system default rather than offering any way to disable spellcheck — if that is intended, the help text should say so.
2. **Save-path robustness.** The new block in `apiSaveProperties` (`userPreferences.js:236-238`) runs before any persistence, with no try/catch and an unguarded `this.preferences.spellCheckerLanguages.value` dereference. `load()` rebuilds `this.preferences` on every project open and only re-adds the entry when the IPC succeeds with `supported: true`, so one failed reload while the modal is open aborts saving *all* preferences (locale, advanced mode, …) while the modal closes as if it saved. Please mirror the try/catch the load path already has.
3. **Apply once, read pure.** `app:getSpellCheckerSettings` mutates state on every call — a synchronous `readSettings()` plus, in system-default mode, an unconditional `session.setSpellCheckerLanguages()` — and it runs on every project open (via `UserPreferences.load()`), not just at startup. Prefer applying the persisted state once at `app.whenReady()` and making the getter a pure read.
4. **Persistence.** Only `spellCheckerUseSystemDefault` is stored in settings.json; the explicit language list lives solely in Chromium's delayed-write Preferences file. Persisting `{ mode, languages }` and re-applying at startup makes the selection survive a hard kill or a Chromium prefs recovery.
5. **Tests.** The added lines in `app/main.js` / `app/preload.js` have no coverage, and there is no Playwright spec although this is a user-visible workarea change; the in-page `electronAPI` stub pattern from `desktop-large-asset-import-static.spec.ts` makes one cheap — and it would also have caught the blocker above.

## Minor

- `__system_default__` is hardcoded three times in the renderer while `app/spell-checker-settings.js` exports `SYSTEM_DEFAULT`, and the settings→value mapping is duplicated between the load and save paths with already-divergent fallbacks — one renderer-side constant plus a shared mapping helper would prevent drift.
- The `useSystemDefault` predicate is duplicated verbatim in `app/main.js:1552-1554` and `spell-checker-settings.js:45-47`; the `platform !== 'darwin'` guard in the get handler is dead code (the module already early-returns the same shape), and the darwin path omits `systemDefault`, giving the IPC payload two shapes.
- The new `multiselect` case in `modalProperties.js` copy-pastes the adjacent `select` loop and already diverges from it (`innerHTML` vs `textContent` for option labels, `setAttribute('selected')` vs `.selected`); a shared option-building helper parameterized by `multiple` would keep them in sync.
- When resolution yields an empty list the session apply is skipped but the requested mode is still persisted and reported (`spell-checker-settings.js:51`, `main.js:1556`). Hard to reach in practice (en-US is effectively always available), but the flag should reflect whether the apply actually happened.

The core approach remains the right one: the pure logic is well extracted and tested, `session.defaultSession` is the correct target (the windows use no partition), the new module is packaged and its spec runs in CI, and macOS is correctly left to the OS spell checker. With the blocker fixed and the robustness points addressed this will be in good shape.

milos-plavsic:
Thanks for the thorough trace, @erseco. I addressed the blocker, all five robustness/test points, and the minor cleanup items in `9509cfe`, with the remaining integration edge cases completed in `6663800`.

Point by point:

- **Category/UI blocker:** the spell-checker preference now reuses `locale.category` exactly when it exists; when fallback preferences are uncategorized, the new row remains uncategorized too. Unit and Playwright regressions verify that the row is visible, no `category="undefined"` row is produced, and no extra tab appears.
- **Sentinel exclusivity:** the shared select builder keeps **System default** exclusive, and normalization prefers concrete languages if a mixed value nevertheless reaches the main process. The help text now explicitly states that an empty selection uses the operating-system language.
- **Save robustness:** the Electron save is guarded, optional, and wrapped in `try/catch`; an IPC failure no longer prevents locale or other preferences from being persisted.
- **Apply once / pure read:** persisted state is applied from `app.whenReady()` only. The getter reads and reports state without calling `session.setSpellCheckerLanguages()`.
- **Persistence:** `{ mode, languages }` is stored in `settings.json` and reapplied at startup; the legacy boolean is removed after a successful write.
- **Main/preload/UI coverage:** the main-process behavior is now behind an injected controller with tests for IPC registration, pure reads, startup restoration, successful persistence, and failed application. Preload, frontend, modal, and static Playwright coverage remain in place.
- **Renderer duplication:** the renderer uses one `SYSTEM_DEFAULT` constant and one settings-to-value helper; select and multiselect option construction share one builder.
- **Payload/branch cleanup:** macOS returns the same payload shape, the dead getter platform branch and duplicated mode predicate are gone.
- **Empty resolution:** an unapplied request is no longer persisted or reported as the requested mode; invalid persisted explicit-language state with an empty list is ignored at startup.

Validation completed locally:

- `make fix`
- `make test-unit`
- `make test-integration`
- `make test-frontend`
- `bun test app/spell-checker-settings.spec.js app/preload.spec.js` — 15 passed
- focused frontend preference/modal suites — 54 passed
- `bun x playwright test --project=static test/e2e/playwright/specs/desktop-spell-checker-preferences-static.spec.ts` — 1 passed
- `git diff --check`

I also attempted the complete Chromium E2E suite using the system Brave binary because Playwright does not provide its bundled Chromium for this Ubuntu 26.04 environment. It reached 262 passes before unrelated Slide/Text editor cases failed under that substituted browser; the focused spell-checker Playwright regression passed independently.

milos-plavsic:
Ernesto and the eXeLearning community — I apologize for the hasty pushes earlier in this review cycle. I should have completed the deeper end-to-end and failure-path audit before asking you to review successive revisions. Thank you for the patient, precise feedback.

I have now done a fresh adversarial pass over the complete change and pushed `bca1169`. In addition to addressing every item in the review, this revision:

- keeps persisted-state reads inside the guarded startup restoration path, so settings failures cannot prevent application startup;
- tracks the effective runtime mode separately from persisted intent;
- contains and reports write failures after a successful session change;
- avoids mutating the modal submission object;
- keeps production, mock, and E2E response contracts aligned with `applied` and `persisted` state;
- extends the Playwright regression through the real Save button and verifies the exact selection sent to Electron.

Validation completed successfully: 8,018 unit tests, the repository 90% per-file coverage gate across 229 files, 20 focused app/preload tests, 63 focused frontend tests, and the focused Playwright UI-to-Electron save-path test. Formatting and diff checks also pass; the remaining lint warnings are pre-existing and outside this change.

I believe this version now addresses the review comprehensively and should be sufficient for another review. I appreciate the scrutiny and your time.

erseco:
I have tested the latest version on Windows and, from my side, the spell checker works perfectly. Technically, I think the implementation is in very good shape and I would be happy to approve it.

Captura de pantalla 2026-08-16 a las 21 58 25
Captura de pantalla 2026-08-16 a las 22 02 23

Before merging, I would like @ignaciogros to take a look at the Preferences UI and give his opinion on the UX.

I am not completely sure that having a separate multi-select field for spell-checker languages is the best approach from the user's perspective. It may be worth considering whether:

* a single language selector would be simpler, or
* the spell-checker language could be derived from the existing interface language preference, avoiding an additional field in Preferences altogether.

There are certainly cases where the interface language and the language used for spell checking need to be different, but I suspect they may be uncommon enough that adding another preference could introduce more complexity than necessary.

So, from the technical side and based on my Windows testing: **it works perfectly and looks ready to approve**.

I would just like @ignaciogros to review the UI/UX and decide whether keeping a separate multi-language selector is the most appropriate solution, or whether it would be better to simplify or integrate it with the existing language preference.

milos-plavsic:
Thanks very much for testing it on Windows and for the detailed feedback. I’ll wait for @ignaciogros’s UX assessment before changing the language-selection design. I’ve also corrected the remaining Biome formatting failure in the new Playwright test and updated the PR description to reflect the current implementation and validation. The artifact-cleanup failure is the expected fork-permission issue.

milos-plavsic:
Rebased to resolve the out-of-date warning. @erseco Should I mark this as ready for review, or wait for @ignaciogros's UX feedback first?

Issue #1842, opened 2026-05-22
Export: FileSystemAssetProvider throws ENOENT when assets dir is missing (noisy [BaseExporter] warning)


erseco:
## Summary

During server-side HTML5 export/preview, this warning is logged:

```
[BaseExporter] Failed to add assets to ZIP: ENOENT: no such file or directory, scandir 'data/assets/<uuid>'
    path: "data/assets/<uuid>",
 syscall: "scandir",
   errno: -2,
    code: "ENOENT"
```

The export still **succeeds** — this is non-fatal log noise — but it pollutes logs and looks like a failure.

> Note: this is **unrelated to PR #1425** (public viewer URL / `public_view_id`). The public preview export builds the asset path exactly as the previous code did (`path.join(filesDir, 'assets', project.uuid)`), and the warning predates those changes (it appears in baseline test runs). Filing separately as requested.

## Root cause

`FileSystemAssetProvider` reads the root `basePath` directory **without first checking that it exists**, unlike every subdirectory read in the same methods, which are guarded with `fs.pathExists(...)`:

- `src/shared/export/providers/FileSystemAssetProvider.ts:286` — in `forEachAsset()`:
  `const entries = await fs.readdir(this.basePath, { withFileTypes: true });` (unguarded)
- `src/shared/export/providers/FileSystemAssetProvider.ts:167` — in `collectRootAssets()` (called by `getAllAssets()`): same unguarded `fs.readdir(this.basePath, ...)`.

Compare with the guarded reads just above each (e.g. lines 143, 151, 273, 280) which wrap `collectAssetsFromDirectory` / `forEachAssetInDirectory` in `if (await fs.pathExists(...))`.

`basePath` is `FILES_DIR/assets/<projectUuid>`. Asset directories are created **lazily** (see `AGENTS.md` §7.3), so a project with no filesystem-stored assets (assets only in the DB, or none yet) has no such directory. The unguarded `readdir` then throws `ENOENT`, which propagates up to `BaseExporter.addAssetsToZip()` (`src/shared/export/exporters/BaseExporter.ts:477` and `:526`), where it is caught and logged as `console.warn('[BaseExporter] Failed to add assets to ZIP:', e)`.

## When it happens

- Any server-side HTML5 export/preview of a project whose `FILES_DIR/assets/<uuid>` directory does not exist (no filesystem assets). Common for fresh/empty projects and DB-only assets.
[...]

PR #2266, opened 2026-08-15
refactor(storage): shard project assets and make storage paths portable

### LINKED ISSUE #2250: Asset storage does not scale
**Asset storage does not scale:** flat `assets/` directory and absolute `storage_path`

## Summary

Every project's assets are stored in a single flat directory, `FILES_DIR/assets/`, with
one subdirectory per project and no partitioning of any kind. On a server with many
users this grows without bound in one directory. Additionally, the `assets.storage_path`
column stores **absolute** filesystem paths, which pins the database to one specific
`FILES_DIR` location and makes any change to the layout a data migration.

The two problems are linked: the second one is what makes the first one hard to fix, so
they should be addressed together.

## Current behaviour

`src/services/file-helper.ts:142`:

```ts
const getProjectAssetsDir = (projectUuid: string): string => {
    return path.join(getFilesDir(), 'assets', projectUuid);
};
```

Resulting layout:

```
FILES_DIR/
├── assets/{projectUuid}/          ← flat, permanent, no partitioning
├── tmp/YYYY/MM/DD/{sessionId}/    ← partitioned by date, cleaned up
├── dist/YYYY/MM/DD/{sessionId}/   ← partitioned by date, cleaned up
└── exelearning.db
```

Notes on what is and is not stored on disk:

- The project document itself is **not** on disk. It lives in the database as a BLOB in
  `yjs_documents.snapshot_data` (one row per project, `project_id` is `UNIQUE`) —
  `src/db/migrations/001_initial.ts:142`.
- Only assets (images, media, attachments) hit the filesystem.
- The folder tree the user sees in the file manager is **virtual**: it is only the
  `folder_path` column, not real directories (`src/routes/assets.ts:402-403`).
- There is **no per-user directory** anywhere. Ownership exists only in the database
  (`projects.owner_id`).
- `tmp/` and `dist/` are already partitioned by date (`src/utils/url.util.ts:74-83`) and
  are cleaned up by `src/services/cleanup-scheduler.ts`. They are **not** part of this
  issue.

## Impact

With 1,000 users and ~20 projects each, `assets/` holds 20,000 subdirectories; at 10,000
users it reaches the hundreds of thousands.

**Runtime is not the problem.** ext4 with `dir_index` (htree) and NTFS both resolve
lookups in O(log n), so opening a specific asset stays fast even with a million
directories. All `readdir` / `listFiles` calls in the codebase were reviewed and **none
enumerates the root of `assets/`** — the ones in
`src/shared/export/providers/DatabaseAssetProvider.ts` (lines 240, 317, 373) and
`src/routes/admin.ts:1413` are always scoped to a single project directory.

**Operations is the problem.** Anything that enumerates the directory degrades:

- backups with `rsync` (one `stat` per entry),
- `du` / `ls` / disk-usage reporting,
- Docker volume snapshots,
- any future maintenance or migration script.

There is also a distant hard limit: modern ext4 with `dir_nlink` has no 65,000
subdirectory cap, but needs the `large_dir` feature to exceed roughly 10M entries.

Separately, absolute paths in `storage_path` mean `FILES_DIR` cannot be moved, remounted,
or restored to a different path without rewriting every row in the `assets` table.

### 1. Shard `assets/` by UUID prefix

Two levels of two hex characters give 65,536 buckets — the scheme used by Git and by
CDNs. With 1M projects that is ~15 subdirectories per bucket.

```ts
const getProjectAssetsDir = (projectUuid: string): string => {
    return path.join(
        getFilesDir(),
        'assets',
        projectUuid.slice(0, 2),
        projectUuid.slice(2, 4),
        projectUuid,
    );
};
```

All asset paths funnel through this single function, so the code change is contained.

### 2. Make `storage_path` relative to `FILES_DIR`

Store `ab/cd/{uuid}/{clientId}.png` instead of `/mnt/data/assets/{uuid}/{clientId}.png`,
and resolve against `getFilesDir()` at read time. This decouples the database from the
deployment path and means a future layout change no longer requires touching every row.

Write sites: `src/routes/assets.ts:427`, `:443`, `:690`, `:704`, `:1159`, `:1173`,
`:1314`; `src/services/folder-manager.ts:417`, `:643`.
Read sites: `src/routes/assets.ts:780`, `:833`, `:895`, `:923`, `:958`;
`src/services/folder-manager.ts:403`, `:491`.

### 3. Migration

A single migration that, per asset row:

1. computes the new sharded destination,
2. moves the file (idempotent — skip if already at the destination),
3. rewrites `storage_path` as a `FILES_DIR`-relative path.

It must be resumable and safe to re-run, since it touches user data on disk. It must also
handle rows whose file is already missing (log and continue rather than abort), and the
ZIP-extraction sublayout `assets/{uuid}/{clientId}/{filename}`
(`src/services/folder-manager.ts:582`).

Cross-platform care is required: on Windows a rename across an open handle raises `EBUSY`.

## Acceptance criteria

- [ ] `getProjectAssetsDir()` returns a sharded path.
- [ ] `assets.storage_path` holds paths relative to `FILES_DIR`; all read sites resolve
      through a single shared helper.
- [ ] Migration moves existing files and rewrites paths; it is idempotent and resumable.
- [ ] Export, import, download, delete, duplicate-project and ZIP-extraction flows all
      still work — covered by unit and E2E tests.
- [ ] Moving `FILES_DIR` to a different absolute path no longer breaks asset resolution
      (regression test).
- [ ] Patch coverage ≥ 90% per `AGENTS.md`.

## ADR

This changes the storage model, so it needs an ADR under `doc/architecture/adr/` per
`AGENTS.md` §7.11, using this issue's number as the tracking identifier
(`ADR-<issue>-01-shard-project-asset-storage.md`). A change directory under
`doc/architecture/changes/` is likely warranted too, given the data migration.

## Out of scope / related

`src/routes/assets.ts:544` hardcodes the chunk upload root as
`path.join(process.cwd(), 'data', 'chunks')`, ignoring `FILES_DIR`. Under Docker or with
`FILES_DIR=/mnt/data`, upload chunks land outside the intended volume. This is a
one-line fix unrelated to sharding and should be a **separate issue**.

erseco:
## Summary

Project asset storage moves from a flat, absolute-path model to a sharded, portable one:

- **Layout:** `FILES_DIR/assets/<projectUuid>/…` → `FILES_DIR/assets/<2-hex-shard>/<projectUuid>/…` (one shard level, 256 lazily created buckets).
- **Database:** `assets.storage_path` now stores POSIX-style paths **relative to `FILES_DIR`** (`assets/ab/<uuid>/<clientId>.png`) instead of absolute host paths.
- **Migration:** existing installations converge automatically at startup via an idempotent, resumable, never-overwriting reconciliation.
- All path construction/resolution is centralized in `src/utils/asset-paths.ts` + `src/services/file-helper.ts`; documented in [ADR-2250-01](doc/architecture/adr/ADR-2250-01-shard-project-asset-storage.md) and the [change design/admin guide](doc/architecture/changes/2250-asset-storage-sharding/design.md).

## Problem

Issue #2250 raises two distinct concerns, and this PR deliberately treats them differently:

1. **Absolute `storage_path` — a concrete architectural defect.** Every asset row was pinned to one specific `FILES_DIR` mount (`/mnt/data/assets/…`), so moving, remounting or restoring the data directory invalidated every row. Themes and templates already store FILES_DIR-relative paths; assets were the outlier. This is fixed outright.
2. **Flat `assets/` root — a preventive concern, not a demonstrated bug.** ext4 with `dir_index` resolves known-name lookups through an htree index, and nothing in the codebase enumerates the assets root at runtime (all `readdir`s are per-project — verified in #2250, consistent with #1842). We shard **now** because it is cheap now: fixing (1) decouples the database from the physical layout, and retrofitting a layout change after years of data accumulation would be far more expensive.

## Decision

```
FILES_DIR/assets/<shard>/<projectUuid>/<clientId>.<ext>
FILES_DIR/assets/<shard>/<projectUuid>/<clientId>/<filename>   (ZIP extraction / duplication)
```

`<shard>` = lowercased first two hex characters of the canonical project UUID (v4, uniformly distributed — the bucket is readable straight off the UUID). Because `projects.uuid` is not format-validated at every creation site (`createProjectWithUuid` accepts client-supplied ids verbatim), non-canonical identifiers fall back to a deterministic 8-bit FNV-1a hash, so the mapping is total and stable. **One shard level, 256 possible buckets, created lazily** — never pre-created.

## Why one level

Average bucket-namespace occupancy (lazy creation — unused buckets never exist on disk):

| Projects | 1 level (256) | 2 levels (65,536) | 3 levels (16,777,216) |
|---:|---:|---:|---:|
| 20,000 | ≈ 78 | ≈ 0.31 | ≈ 0.0012 |
| 200,000 | ≈ 781 | ≈ 3.05 | ≈ 0.0119 |
| 1,000,000 | ≈ 3,906 | ≈ 15.26 | ≈ 0.0596 |
| 10,000,000 | ≈ 39,063 | ≈ 152.59 | ≈ 0.5960 |

One level already caps per-bucket counts very effectively at any plausible eXeLearning scale, with minimal depth and minimal intermediate directories. Two levels (the issue's original proposal) would leave buckets ~15 entries deep at one **million** projects — the extra level does nearly nothing except add depth; three levels is pure overhead. There is no measured evidence that 65k or 16M buckets are needed, and the now-portable `storage_path` makes a future re-shard cheap if measurements ever justify it. Precedent: **Git itself uses exactly one two-hex-character level** for loose objects ("splayed over 256 subdirectories using the first two characters of the sha1 object name" — [gitrepository-layout](https://git-scm.com/docs/gitrepository-layout), [gitformat-loose](https://git-scm.com/docs/gitformat-loose)); the issue's description of two levels as "the scheme used by Git" was not accurate. We do not claim one level is universally optimal — it is the proportional choice for the expected scale and current evidence (full rationale and rejected options in the ADR).

## Migration

Runs automatically inside `bootstrap()` on **every** startup (after schema migrations, before `app.listen()`); it is a fast no-op once converged: one legacy-row query plus one readdir of the assets root, plus bounded content checks for existing two-decimal shard buckets (`00`–`99`), which are ambiguous with legacy numeric project-id directories. Re-checking every boot is deliberate (it lets restored legacy data converge automatically). During a large migration a progress line (`[AssetStorage] Migration progress: 1,000/3,250 legacy row(s) processed, 2,250 pending.`) is printed every 1,000 processed rows. Deliberately **not** a Kysely migration: filesystem moves and row updates cannot share a transaction, so this is a startup reconciliation instead (`src/services/asset-storage-migration.ts`).

- **Files:** `assets/<uuid>/…` → `assets/<shard>/<uuid>/…` via rename (same filesystem: metadata-only); `EXDEV` falls back to copy-to-temp + rename + size-verify + remove, so the destination never holds a partial file. Handles the flat layout, the nested ZIP-extraction layout, and the legacy numeric-directory quirk (`assets/<numericId>/`, from routes that accepted numeric project ids).
- **Rows:** legacy absolute values are rewritten to the canonical relative form, derived from the row's own `assets/…` suffix (the stale absolute prefix is ignored).
- **Orphans:** files on disk without rows are swept into the sharded layout; emptied legacy directories are removed; unrecognized entries are reported and left alone.
- **Temporary legacy-read compatibility exists**: the shared resolver re-roots legacy absolute values under the *current* `FILES_DIR` via their `assets/…` suffix, so unmigrated/conflict rows keep working (and become FILES_DIR-portable even before migration). All new writes emit **only** the new format — there is no dual-write, and removal of the fallback is tracked as ADR follow-up. Permanent dual-layout support and a hard cut-over were both evaluated and rejected (ADR, "Migration / compatibility strategy").

## Safety

- **Idempotent & resumable:** per-row protocol is *move file, then rewrite row* under an optimistic `WHERE id = ? AND storage_path = <old>` guard; a crash between the two steps leaves a state the next boot reconciles. Individual file errors are counted and logged, never abort the run, and are retried on the next startup.
- **Restart/interruption states handled explicitly:** source-only → move; destination-only → rewrite row (crash recovery); both identical (size + SHA-256) → drop legacy copy; **both different → never overwrite**: keep both files, log the conflict, park the row on its legacy location in portable relative form, re-report on every boot until an operator resolves it; neither → log missing, converge the row (reads keep returning 404 exactly as before); uninterpretable values (no `assets/…` suffix, traversal) → left untouched and reported, and the migration never dereferences arbitrary absolute paths from the database.
- **Concurrent multi-instance startup** (Redis HA): the optimistic guards and existence checks make the race benign — one instance wins each row.
- **Path security:** every stored-value → physical-path conversion goes through one resolver that rejects `..`, separators in segments, control characters, and anything escaping `FILES_DIR/assets` (separator-aware containment via the existing `isWithinBase`).
- **Cleanup paths** (cleanup scheduler, admin user deletion, `projects-cleanup` CLI) remove both the sharded and the legacy directory during the transition.

## Tests

All executed locally on this branch (macOS, Bun 1.3.14):

- `make fix` / `make lint` — clean (exit 0; the two Biome warnings also appear on a clean checkout of `main`).
- `make test-unit` — **8,104 pass, 0 fail** (backend, 232 files); per-file 90% coverage gate **passed**, including every new/modified file.
- `make test-frontend` — 14,120 pass, **1 fail**: `focusedEditMode.test.js › removes all focus-mode state` — a known pre-existing flake that fails on clean `main` in full runs and passes in isolation (re-verified both here: passes 20/20 in isolation, and `git diff origin/main -- public/` is empty — this PR touches no frontend files).
- `make test-integration` — **727 pass, 0 fail**.
- `make test-e2e` — **528 passed, 6 skipped, 0 failed** (16.9 m, full Playwright suite: file manager incl. special-characters, upload, import/export, project clone/duplicate, save flows) — run twice: once on the base implementation and once again on the final branch including the review-hardening commit. `make test-e2e-static` was not run: this change touches only server-side storage code — no `public/`, embedding or static-build files are modified, and the static app runs without a server.
- New/updated suites: `src/utils/asset-paths.spec.ts` (shard determinism, golden FNV values, grammar, traversal rejection, directory-level-value rejection), `src/services/asset-storage-migration.spec.ts` (26 tests: every migration state, idempotency, batching, EXDEV fallback, orphan sweep incl. two-digit numeric-dir/bucket disambiguation, unmounted-volume latch, conflicts, unicode filenames, traversal safety), `src/services/file-helper.spec.ts` (**FILES_DIR portability regression test**: the same stored path resolves under two different `FILES_DIR` values without rewriting the row), plus updated specs for upload/download/delete/duplicate/ZIP-extract/sync/stream/export-provider/cleanup/coordinator flows asserting relative sharded writes, the legacy-read fallback, and partial-failure reporting.
- An adversarial multi-agent review pass was run on the branch before opening this PR; the confirmed findings (unmounted-volume rewrite hazard, numeric-directory sweep gap for project ids 10–99, unvalidated project identifiers reaching `getProjectAssetsDir`, directory-level legacy values, swallowed partial cleanup errors, superseded-file orphans on relocating re-uploads, missing legacy-dir fallback in server-side exports, non-canonical UUID in upload sessions, case-blind bucket-collision guard) are fixed in the final commit, each with a regression test.

## Benchmarking

This change is **preventive**; it is not justified by a current runtime failure, and no benchmark was required to adopt it. A reproducible benchmark ships anyway: `scripts/benchmark-asset-storage.ts` compares 0/1/2/3 levels (creation, known-path `stat`, root/bucket `readdir`, full recursive traversal) at configurable scale with seeded, reproducible datasets; `--json` captures environment metadata. It is not part of CI (its spec checks harness correctness only — no timing thresholds).

One real run was executed while developing (20,000 projects × 2 assets, **macOS 25.4.0 / APFS / Apple M5 / Bun 1.3.14 — not the production ext4 target**): 5,000 known-path stats: 184 ms flat vs ≈ 57 ms at every sharded depth; root readdir 8.2 ms/20,000 entries flat vs 0.2 ms/256; full traversal 1.21 s flat, **1.04 s at one level, 1.64 s at two, 2.25 s at three** — deeper trees make the full walk *slower*, consistent with sharding redistributing entries rather than removing work. These numbers must not be extrapolated to ext4 or network storage; the ADR recommends running the benchmark on local ext4 SSD/NVMe, on the actual production backend, and on NFS/SAN if used.

## Operational considerations

- Persistent `FILES_DIR` data must live outside the container writable layer. Docker named volumes and bind mounts can both expose host/external storage, named volumes are not inherently a performance problem (they bypass the storage-driver union-filesystem overhead — [Docker storage docs](https://docs.docker.com/engine/storage/)), and volumes can themselves be NFS-backed. NFS/SAN/network storage behaves differently from local ext4 and should be measured on the actual deployment.
- ext4 is **not** unsuitable for large installations: with `dir_index` lookups are htree-indexed, `dir_nlink` lifts the old ~65k subdirectory limit, and `large_dir` raises directory size/htree height further ([ext4(5)](https://man7.org/linux/man-pages/man5/ext4.5.html)). Very large deployments should choose ext4/XFS/Btrfs/OpenZFS/NFS/SAN-backed storage from operational requirements and measured workloads, not from folklore.
- Full-tree operations (backup, `du`, `find`, `tar`, `rsync` list generation) still scale with total object count — sharding bounds directory sizes; it does not shrink the tree.
- Upgrade/downtime/verification/rollback details for administrators: [change design doc](doc/architecture/changes/2250-asset-storage-sharding/design.md) ("Administrator guide"). Short version: back up DB + `FILES_DIR` first; first boot renames existing files before listening (seconds to a few minutes locally, longer on network storage); success is `[AssetStorage] Asset storage layout is up to date.` on the next boot; rollback = restore the pre-upgrade backup.

## ADR

[ADR-2250-01 — Shard project asset storage using a single 8-bit bucket level](doc/architecture/adr/ADR-2250-01-shard-project-asset-storage.md) (Accepted). Change document: [`doc/architecture/changes/2250-asset-storage-sharding/design.md`](doc/architecture/changes/2250-asset-storage-sharding/design.md). `make architecture-check` passes.

## References

- Linux kernel ext4: [directories](https://docs.kernel.org/filesystems/ext4/directory.html), [inodes](https://docs.kernel.org/filesystems/ext4/inodes.html); [ext4(5)](https://man7.org/linux/man-pages/man5/ext4.5.html)
- Git: [gitformat-loose](https://git-scm.com/docs/gitformat-loose), [gitrepository-layout](https://git-scm.com/docs/gitrepository-layout)
- Docker: [storage](https://docs.docker.com/engine/storage/), [volumes](https://docs.docker.com/engine/storage/volumes/), [bind mounts](https://docs.docker.com/engine/storage/bind-mounts/)
- Btrfs: [kernel docs](https://docs.kernel.org/filesystems/btrfs.html)

## Historical context

PR #788 intentionally introduced the UUID-per-project directories and lazy creation this PR builds on — this is an evolution of that design, not a correction of a mistake. Issue #1842 confirms application-level filesystem enumeration is scoped to individual project directories, never the assets root.

Out of scope (as requested in the issue): the hardcoded `data/chunks` upload root is unrelated to sharding and is left for a separate issue.

## Issue

Closes #2250

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -157,7 +157,7 @@ CLIENT (Browser)                        SERVER (Bun/Elysia)
 YjsDocumentManager (Y.Doc)              SessionManager: lightweight metadata
 ├── navigation (Y.Array)                WebSocket: stateless relay
 ├── metadata (Y.Map)                    Database: projects + yjs snapshots
-├── assets (Y.Map)                      Filesystem: FILES_DIR/assets/{uuid}/
+├── assets (Y.Map)                      Filesystem: FILES_DIR/assets/{shard}/{uuid}/
 └── themeFiles (Y.Map)
 ```
 
@@ -185,7 +185,7 @@ Client-side storage:
 
 ```
 FILES_DIR/
-├── assets/{projectUuid}/           # Permanent project assets
+├── assets/{shard}/{projectUuid}/   # Permanent project assets (shard = first 2 hex chars of UUID, ADR-2250-01)
 ├── tmp/{year}/{month}/{day}/{id}/  # Temporary files
 ├── dist/{year}/{month}/{day}/{id}/ # Ready-to-download exports
 ├── chunks/                         # Upload chunks
@@ -196,6 +196,8 @@ FILES_DIR/
 Key rules:
 - Directories created **lazily** (on-demand), never eagerly
 - Assets use project **UUID**, not numeric ID
+- `assets.storage_path` is stored **relative to FILES_DIR** (`assets/<shard>/<uuid>/...`, POSIX separators) —
+  never absolute; always build/resolve through `src/utils/asset-paths.ts` via `src/services/file-helper.ts`
 - Always use `isPathSafe()` for user-supplied paths
 - Use `path.join()` for cross-platform paths
 
--- a/doc/architecture/adr/ADR-2250-01-shard-project-asset-storage.md
+++ b/doc/architecture/adr/ADR-2250-01-shard-project-asset-storage.md
@@ -9,7 +9,7 @@ deciders:
 reviewers:
   - "@ignaciogros"
 related:
-  prs: []
+  prs: [2266]
   changes:
     - "2250-asset-storage-sharding"
   adrs: []
--- a/doc/architecture/adr/ADR-2250-01-shard-project-asset-storage.md
+++ b/doc/architecture/adr/ADR-2250-01-shard-project-asset-storage.md
@@ -274,8 +274,10 @@ We will:
 - One extra directory level in every asset path.
 - A transitional legacy-read fallback exists in the resolver until installations
   converge; it must eventually be removed (see "Follow-up work").
-- Startup performs one cheap check per boot (one indexed-free `NOT LIKE 'assets/%'`
-  query + one readdir of the assets root) even on converged installations.
+- Startup performs a cheap, bounded check per boot even on converged installations: one
+  indexed-free `NOT LIKE 'assets/%'` query, one readdir of the assets root, and a
+  content check (`readdir`) of each existing two-decimal shard bucket (`00`-`99`) to
+  distinguish it from a legacy numeric project-id directory.
 
 ### Neutral
 
@@ -333,8 +335,12 @@ We will:
 ## Follow-up work
 
 - Remove the legacy absolute-path read fallback from the resolver once field
-  installations have converged (tracked for a future major release; the migration's
-  per-boot summary tells operators when their installation is clean).
+  installations have converged (tracked in
+  <https://github.com/exelearning/exelearning/issues/2288> for a future major release;
+  the migration's per-boot summary tells operators when their installation is clean).
+- Provide a CLI subcommand that lists parked migration conflicts and resolves them
+  with an explicit keep-old / keep-new choice
+  (<https://github.com/exelearning/exelearning/issues/2287>).
 - Consider validating client-supplied project identifiers at creation time so the
   FNV-1a fallback becomes purely defensive.
 
--- a/doc/architecture/adr/ADR-2250-01-shard-project-asset-storage.md
+++ b/doc/architecture/adr/ADR-2250-01-shard-project-asset-storage.md
@@ -1,0 +1,356 @@
+---
+id: ADR-2250-01
+title: "Shard project asset storage using a single 8-bit bucket level"
+status: Accepted
+date: 2026-08-15
+tracking_issue: 2250
+deciders:
+  - "@erseco"
+reviewers:
+  - "@ignaciogros"
+related:
+  prs: []
+  changes:
+    - "2250-asset-storage-sharding"
+  adrs: []
+supersedes: []
+superseded_by: []
+ai_assistance:
+  tool: "Claude Code"
+  model: "claude-fable-5"
+---
+
+# ADR-2250-01: Shard project asset storage using a single 8-bit bucket level
+
+## Context
+
+Server-side project assets are stored under one directory per project:
+
+```
+FILES_DIR/assets/<projectUuid>/<clientId>.<ext>          (flat uploads)
+FILES_DIR/assets/<projectUuid>/<clientId>/<filename>     (ZIP extraction / duplication)
+```
+
+This layout was introduced deliberately by PR #788, which replaced session/numeric-id
+directories with UUID-based directories and lazy creation. It is an intentional design,
+not an accident. Application code never enumerates the root of `assets/` at runtime —
+all `readdir` calls are scoped to a single project directory (verified in issue #2250,
+and consistent with issue #1842, where the only root-level reads are per-project
+`FileSystemAssetProvider` scans of `FILES_DIR/assets/<uuid>`).
+
+Two separate concerns were raised in issue #2250:
+
+1. **Portability (a concrete defect):** `assets.storage_path` persisted **absolute host
+   filesystem paths** such as `/mnt/data/assets/<uuid>/<clientId>.png`
+   (`src/routes/assets.ts` @ `3c7c9e821`, write sites at lines 427/443/690/704/1159/1173/1314).
+   This couples every database row to one specific `FILES_DIR` mount point. Moving,
+   remounting or restoring the data directory to a different path silently invalidates
+   every asset row. Notably, the `themes` and `templates` tables already store
+   FILES_DIR-relative paths (`src/routes/admin-themes.ts:603`,
+   `src/routes/admin-templates.ts:223`) — assets were the outlier.
+
+2. **Scalability (a preventive concern):** the flat `assets/` root grows without bound —
+   one subdirectory per project, forever. There is **no measured runtime failure at
+   current scale** (see "Evidence"), but an ever-growing single directory is unnecessary,
+   and changing the layout is far cheaper now than after years of data accumulation —
+   especially because fixing concern 1 first decouples the database from the physical
+   layout.
+
+## Problem
+
+Should project asset storage (a) keep absolute `storage_path` values, and (b) keep the
+flat one-directory-per-project layout — and if not, how many shard levels should the new
+layout use, and how do existing installations migrate safely?
+
+## Decision drivers
+
+- Database rows must survive a change of `FILES_DIR` (backup/restore, remount, Docker
+  volume relocation, desktop-app data moves).
+- Bounded directory sizes at plausible scales (tens of thousands to millions of
+  projects) without adding depth or complexity that no evidence justifies.
+- Lazy directory creation (established by PR #788) must be preserved.
+- The migration must be idempotent, resumable, crash-safe, safe under multi-instance
+  startup, and must never silently overwrite user data.
+- One canonical layout long-term; no permanent dual-write or dual-layout support.
+- Cross-platform correctness: the same backend runs on Linux servers and inside the
+  Electron desktop app on Windows/macOS (`AGENTS.md` §1: EBUSY file locks, `path.join`).
+
+## Options considered
+
+### Sharding depth
+
+Assuming two hexadecimal characters per level, average **bucket-namespace** occupancy
+(directories are created lazily, so unused buckets never physically exist):
+
+| Projects | 0 levels (1 dir) | 1 level (256 buckets) | 2 levels (65,536) | 3 levels (16,777,216) |
+|---:|---:|---:|---:|---:|
+| 20,000 | 20,000 | ≈ 78 | ≈ 0.31 | ≈ 0.0012 |
+| 200,000 | 200,000 | ≈ 781 | ≈ 3.05 | ≈ 0.0119 |
+| 1,000,000 | 1,000,000 | ≈ 3,906 | ≈ 15.26 | ≈ 0.0596 |
+| 10,000,000 | 10,000,000 | ≈ 39,063 | ≈ 152.59 | ≈ 0.5960 |
+
+**Option A — 0 levels (status quo):** no migration cost; but the root keeps growing
+unbounded, and retrofitting later means migrating much more data.
+
+**Option B — 1 level (`assets/aa/<uuid>/`), chosen:** caps per-bucket project counts
+very effectively at plausible scales (≈ 781 per bucket at 200k projects; even at 10M
+projects a bucket averages ≈ 39k entries — a size ext4's htree handles for lookups, and
+far from any structural limit). Minimal path depth, minimal intermediate directories,
+one readable hop (`assets/ab/ab12…`). Precedent: Git loose objects use exactly one
+two-hex-character level (see "Evidence").
+
+**Option C — 2 levels (`assets/aa/bb/<uuid>/`, the issue's original proposal):**
+65,536 possible buckets is disproportionate: at 1M projects the average bucket would
+hold ≈ 15 directories, i.e. the intermediate level does almost nothing except add
+depth and intermediate directories. No measurement indicates it is needed. (The issue
+described two levels as "the scheme used by Git"; Git actually uses **one** level —
+see "Evidence".)
+
+**Option D — 3 levels:** ≈ 0.06 projects per leaf bucket at 1M projects; pure overhead
+at any plausible eXeLearning scale.
+
+We do **not** claim one level is universally optimal — it is the proportional choice for
+the expected scale and current evidence, and the path portability work (below) makes a
+future re-shard cheap if measurements ever justify it.
+
+### Stored-path representation
+
+**Chosen:** `storage_path` is a normalized POSIX-style path **relative to `FILES_DIR`,
+including the `assets/` component**: `assets/ab/<uuid>/<clientId>.png` (nested form:
+`assets/ab/<uuid>/<clientId>/<filename>`). This matches the existing themes/templates
+convention, is unambiguous (no implicit `assets/` root to remember), and is portable
+across hosts and path-separator conventions.
+
+Rejected: keeping absolute paths (the defect being fixed), and prefix-free relative
+paths (`ab/<uuid>/…`) which hide the storage root and invite `path.join(filesDir, …)`
+mistakes.
+
+### Shard algorithm
+
+**Chosen:** for canonical UUIDs (8-4-4-4-12 hex), the shard is the **lowercased first
+two hex characters**. Project UUIDs are generated with `uuid.v4()`
+(`src/db/queries/projects.ts:302`) or `crypto.randomUUID()`
+(`src/services/session-manager.ts:48`), whose first byte is uniformly random, so the
+256 buckets are uniformly filled and an operator can find a project's directory by
+reading its UUID.
+
+However, `projects.uuid` is **not format-validated at every creation site** — e.g.
+`createProjectWithUuid` accepts a client-supplied identifier verbatim
+(`src/routes/project.ts:1251`). For any non-canonical identifier the shard falls back to
+a deterministic, stable 8-bit **FNV-1a** hash rendered as two lowercase hex characters,
+so the mapping is total and stable for arbitrary strings. Both branches are golden-tested
+in `src/utils/asset-paths.spec.ts`.
+
+Caveat recorded for the future: if project ID generation ever switches to a time-ordered
+scheme (e.g. UUIDv7, whose leading bits are a timestamp), the first-two-hex prefix would
+stop being uniform and this ADR must be revisited (worst case degrades toward today's
+flat layout, not below it).
+
+### Migration / compatibility strategy
+
+Evaluated:
+
+- **(a) Permanent support for both layouts** — rejected: two storage architectures
+  forever, every reader forked.
+- **(b) Temporary legacy-read compatibility + convergent migration** — **chosen**: the
+  shared resolver re-roots legacy absolute values via their `assets/…` suffix under the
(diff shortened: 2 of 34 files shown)


ignaciogros (approved):
Great work!

I’ve done some manual testing with previously created content, including collaboratively edited content, and I haven’t found any issues related to the assets.

The PR fixes a real architectural defect (non-portable absolute paths), takes the opportunity to introduce a proportionate and well-argued sharding scheme, centralises path logic that was previously scattered across the codebase, and ships a remarkably careful data migration: idempotent, resumable, never overwriting, and refusing to act at all when it detects the storage volume is not mounted. Documentation (ADR + administrator guide) and test coverage meet what `AGENTS.md` requires.

The portability fix in particular will matter a lot for institutions running large installations. Those are exactly the deployments that move mount points, restore backups onto different machines and migrate storage volumes, and every one of those operations used to invalidate every asset row.

The observations below are about operations and polish, not design, and **none of them is blocking**. Observation 1 (startup-blocking migration vs. healthchecks) does not apply at the real or expected size of current installations; the only part of it worth implementing now is the **progress log**, purely for diagnostics. What is worth capturing as follow-up work is observation 3 (a CLI tool to resolve conflicts, which at any scale are impossible to resolve today without hand-editing files on disk) and agreeing on a plan to remove the **legacy absolute-path read fallback**, so the transitional debt does not become permanent.

---

## Non-blocking observations

1. **The migration blocks startup, but nothing needs to be done about it today.** It runs before `app.listen()`. In theory, with many projects or on network storage (NFS/SAN), the first boot could take long enough for a Docker healthcheck or a Kubernetes `startupProbe` to kill the container and enter a restart loop (the migration is resumable, so nothing is corrupted, but the symptom would be confusing).

   **In practice, this does not apply:** there are no large installations, and none with more than ~10,000 projects are expected. At that scale, this means only a few tens of thousands of `rename` calls on the same filesystem. These are metadata-only operations, so they should take just seconds on local ext4. No reasonable health check would ever trigger. The real risk would only arise if `FILES_DIR` were mounted on NFS/SAN, where each `rename` incurs network latency.

   **Concrete suggestion (this one is worth doing):** add a **progress log** to phase 1, with one line every N processed rows (e.g. every 1,000) showing the running count and the pending total. It costs three lines of code and turns a slow boot into something legible in the logs, instead of leaving a silent process that nobody can tell is progressing or stuck. Exposing the migration as an explicit CLI command can wait until a network-storage deployment actually shows up.

2. **A nuance about per-boot cost (not a defect).** The PR describes the converged state as "one query and one readdir". In practice, in phase 2, buckets whose name is two decimal digits (`00`–`99`, ~100 of the 256 possible) cannot be told apart a priori from a legacy numeric-id directory, so each of them gets an extra `readdir` on every boot. It is a bounded, cheap check (milliseconds at realistic scale) and **re-verifying on every boot is deliberate and correct**: revalidating the on-disk state is precisely what makes the migration resumable and what lets a restored backup containing legacy data converge on its own. A sentinel file allowing the sweep to be skipped would save very little and could go stale (legacy data copied in by hand, an older backup restored), causing a needed migration not to run. This is only a minor wording imprecision in the PR description.

3. **Conflicts are parked forever and there is no tool to resolve them.**

   *What a conflict is:* the migration finds a file at the old location (`assets/<uuid>/photo.png`) **and a different file already present at the new location** (`assets/ab/<uuid>/photo.png`), with different content (compared by size and SHA-256). It cannot know which one is correct, so it applies the conservative rule: **it overwrites nothing**. Both files are kept, and the database row continues to point to the old one, rewritten to the portable relative form at least.

   *Consequence:* that row **never converges to the sharded layout**. On every subsequent boot the migration detects the same conflict and emits the same warning, indefinitely, until an administrator SSHes in, compares the two files by hand and deletes the redundant one. On top of that, the phase 2 sweep sees the file is still referenced by a row and leaves the legacy directory in place, so the installation stays permanently in a mixed state.

   *Assessment:* not overwriting is the right call (a repeated warning is far better than losing a user's file) and the scenario should be extremely rare (it requires something to already exist at the destination before migrating). What is missing is the other half: a CLI reconciliation subcommand that lists pending conflicts and resolves them with an explicit choice (keep the old one / keep the new one), instead of leaving the administrator to move files around by hand. It would also help if the log warning printed both full absolute paths and each file's size, so the decision can be made without hunting for them.

4. **`storage_path NOT LIKE 'assets/%'` as the "legacy" criterion** depends on the engine's collation: on MySQL/MariaDB with a case-insensitive collation, a value like `Assets/…` would be treated as already migrated. It is a marginal case, but the criterion is not identical across SQLite, PostgreSQL and MariaDB.

5. **PR size.** ~4,000 lines across 34 files is a lot for a single review. The commits are well separated (grammar → writes → migration → benchmark → ADR → hardening), which keeps it reviewable, but the final hardening commit bundles nine heterogeneous fixes. Not a problem for approval.

6. **The benchmark was measured on macOS/APFS/M5**, not on the production ext4 target. The PR states this explicitly and does not extrapolate, which is correct; the numbers just should not be treated as evidence of anything beyond orders of magnitude.

---

Please push a new commit if you consider any of these improvements important to land before merging into `main`. From my side the only candidate is the progress log in observation 1; the rest can be follow-up issues.

Thank you!

erseco:
Thanks for the thorough review, @ignaciogros — and for testing it against existing and collaboratively edited content.

I agree with the assessment and pushed 58894ace2 addressing the two points worth landing in this PR:

* **Progress logging (observation 1):** phase 1 now prints `[AssetStorage] Migration progress: 1,000/3,250 legacy row(s) processed, 2,250 pending.` every 1,000 processed rows, so a slow first startup is visibly making progress instead of staying silent until the final summary. The total is counted lazily — only once actual legacy work is found — so a converged startup still issues a single legacy-row query (regression-tested, including a test that counts the queries issued on a converged boot).
* **Per-boot cost wording (observation 2):** you are right that "one query + one readdir" was too precise. Existing two-decimal shard buckets (`00`–`99`) are ambiguous with legacy numeric project-id directories and get a bounded content check every boot. I corrected ADR-2250-01, the change design document and this PR's description. Re-checking every boot stays deliberate — it is what lets restored legacy data converge on its own — so no sentinel was added.

For the follow-up work you flagged, I opened two issues rather than growing this PR:

* #2287 — conflict-resolution CLI: list parked conflicts showing both physical paths and file sizes, and resolve them explicitly with `keep-old` / `keep-new`, never touching either copy without an operator choice. It also covers improving the per-boot conflict warning to print both absolute paths and sizes so the decision can be made straight from the logs.
* #2288 — removal plan for the legacy absolute-path read fallback, with explicit preconditions (field convergence signal, migration window, future major release), so the transitional debt has operational tracking instead of only an ADR note.

Both issues are now linked from the ADR's "Follow-up work" section.

On `storage_path NOT LIKE 'assets/%'` (observation 4): the collation nuance is technically correct across the supported engines, but application-generated paths always use the lowercase canonical `assets/` prefix, so a mixed-case `Assets/…` value would already be anomalous, hand-edited data. I would rather not add dialect-specific SQL to every startup without evidence that such rows exist.

No changes for observations 5 and 6 (PR size, and the benchmark caveats are already stated in the PR).

Thanks again!

ignaciogros (approved):
Thanks, @erseco!

ignaciogros (commented):
Hi @erseco,

> This stack is out-of-date with its base branch

Could you please take a look at this when you have a chance, so we can merge it as soon as possible? I’m not entirely sure about the differences involved here, although the test suite has passed successfully.

Thank you!

erseco:
@ignaciogros I was testing GitHub's stacked PR functionality with #2266 and #2290. The stack itself is now up to date, but GitHub still marks it as not ready because #2290 also needs to be approved before the stack can be merged.

So, if you can review/approve #2290 as well, the stack should become ready. Thanks!

PR #2268, opened 2026-08-15
fix(import): preserve malformed iDevice properties

### LINKED ISSUE #2190: Importing a project with damaged iDevice properties silently discards the activity data

erseco:
## Summary

- `ElpxImporter` no longer replaces an unparseable `jsonProperties` payload with `{}`: the raw payload is stored verbatim in the Y.Doc, where the existing #2178 machinery detects the parse failure, blocks editing and keeps the data recoverable. Affected activities are reported in `ElpxImportResult.malformedProperties` and shown to the author as a damaged-activities notice (sharing one dialog with the missing-files notice from #2224, since the alert modal is a singleton).
- The export side round-trips the preserved payload: `YjsDocumentAdapter` carries the unparseable string as `malformedProperties` and `OdeXmlGenerator` / `ComponentExporter` write it back inside CDATA, so a save or export no longer normalizes the damaged data to `{}`.
- No repair guessing: the payload is never parsed leniently, rendered as HTML or evaluated. Adds `ADR-2190-01` documenting the decision.

Known follow-up (recorded in `ADR-2190-01`): the single-`.idevice`/`.block` re-import path in `public/app/yjs/ComponentImporter.js` still discards unparseable payloads.

## How to test

```bash
bun test src/shared/import src/shared/export
npx vitest run public/app/workarea/project public/app/yjs
make bundle && bun x playwright test --project=chromium test/e2e/playwright/specs/malformed-idevice-json.spec.ts test/e2e/playwright/specs/missing-asset-notice.spec.ts
```

Or manually: import `test/fixtures/damaged-trueorfalse-json.elpx`, confirm the "Damaged activities" alert names the trueorfalse activity, the page still renders all blocks, editing the damaged activity stays blocked, and a re-exported `.elpx` still carries the original malformed payload in `content.xml`.

Closes #2190


erseco:
Verified in Sentry: the malformed-`jsonProperties` failure this PR addresses is present in production telemetry (project `cloud-exelearning-dev`, last 30 days):

- **CLOUD-EXELEARNING-DEV-2** — `SyntaxError: Expected ',' or '}' after property value in JSON at position 2509` in `JSON.parse` — **66 events** ([Sentry](https://erseco.sentry.io/issues/?project=4511542266953808&query=issue%3ACLOUD-EXELEARNING-DEV-2))
- **CLOUD-EXELEARNING-DEV-2W** — same signature surfacing in `setParams` (`app/app.bundle`), Firefox wording (`…at line 1 column 2510…`) — **9 events** ([Sentry](https://erseco.sentry.io/issues/?project=4511542266953808&query=issue%3ACLOUD-EXELEARNING-DEV-2W))

Both point at the same byte offset (position 2509 / column 2510), i.e. the same corrupted document(s) being reopened repeatedly. All events come from alpha builds `202607160306` / `202607281245`, which predate the #2194 root-cause fix (merged 2026-08-04) — confirming that documents corrupted before that fix keep failing at parse time, exactly the gap this PR's import-side preservation covers (#2190).

ignaciogros (approved):

PR #2269, opened 2026-08-15
fix(markdown): render single-dollar LaTeX formulas in the Markdown iDevice

### LINKED ISSUE #1990: LaTeX formulas in Markdown iDevice not rendered in 4.0.1
Latex formulas between Dollar signs in the Markdown idevice are not rendered - neither in the editor, nor in the preview, nor in the export as folder, even if Mathjax export option is activated.
Environment: 4.0.1, Win 11, Edge

erseco:
## Summary
- `$...$` inline math in the Markdown Text iDevice was never rendered: the MathJax configuration only enables `\(...\)` inline delimiters (so a lone `$` never triggers loading or typesetting), and Showdown mangled formula underscores first (`$a_1 + b_2$` became `$a<em>1 + b</em>2$`).
- Normalize `$...$` to `\(...\)` during markdown-to-HTML conversion — in the shared `eXe.app.common.markdownToHTML` helper and in the iDevice's standalone Showdown fallback — so `hasLatex` detection, the LaTeX pre-renderer, workarea, preview and exports all work unchanged. The global MathJax configuration is untouched, avoiding `$` false positives in existing non-markdown content.
- Pandoc-style rules keep plain dollar amounts literal (`It costs $5 and $10` is not math); escaped dollars never pair (a formula cannot end in a backslash, so `\$` before a closing delimiter rejects the pair); fenced code blocks and inline code spans are skipped; `$$...$$` display math keeps working as before.

Known heuristic limit (shared with Pandoc and markdown-it-texmath): a line carrying both a leading and a trailing dollar amount, e.g. `Cuesta $5 y vale 10$`, satisfies the pairing rules and is treated as math.

## How to test
```bash
npx vitest run public/app/common/app_common.test.js public/files/perm/idevices/base/markdown-text/edition/markdown-text.test.js
make bundle
bun x playwright test --project=chromium test/e2e/playwright/specs/idevices/markdown-text.spec.ts
```
Manually: add a Markdown Text iDevice, write `The formula $a^2 + b^2 = c^2$ costs $5 and $10`, save — the formula renders via MathJax in the workarea and preview while the prices stay literal.

## Screenshots
Workarea rendered view after the fix — the formula typeset by MathJax, the prices literal:

Markdown iDevice rendering an inline LaTeX formula while dollar prices stay literal

Closes #1990

### DIFF
--- a/public/app/common/app_common.js
+++ b/public/app/common/app_common.js
@@ -67,21 +67,26 @@ export default class Common {
    * stashed before Showdown runs so that markdown processing does not eat
    * underscores, asterisks or backslashes inside formulas. They are restored
    * verbatim afterwards so MathJax can pick them up at render time.
+   * Single-dollar inline math ($...$) is normalized to \(...\) because the
+   * MathJax configuration shipped with eXeLearning does not enable the $
+   * inline delimiter (#1990).
    */
   markdownToHTML(content) {
     var src = String(content == null ? '' : content);
     var store = [];
+    var stash = function (match) {
+      store.push(match);
+      return 'EXELATEXBEGIN' + (store.length - 1) + 'EXELATEXEND';
+    };
     [
       /\\\[[\s\S]*?\\\]/g,
       /\$\$[\s\S]*?\$\$/g,
       /\\begin\{[^}]+\}[\s\S]*?\\end\{[^}]+\}/g,
       /\\\([\s\S]*?\\\)/g,
     ].forEach(function (re) {
-      src = src.replace(re, function (match) {
-        store.push(match);
-        return 'EXELATEXBEGIN' + (store.length - 1) + 'EXELATEXEND';
-      });
+      src = src.replace(re, stash);
     });
+    src = this.normalizeDollarMath(src, stash);
 
     var converter = new showdown.Converter({
       noHeaderId: true,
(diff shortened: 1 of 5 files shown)


ignaciogros (commented):
Thanks, @erseco.

I'm not entirely sure this is the right approach, as `$...$` could lead to false positives.

If we want to support single-dollar delimiters, It might be better to enable `$...$` as an inline MathJax delimiter instead of converting it to `\(...\)` during processing.

I'd also like @mnarvaezm to take a look, as he's more familiar than I am with how LaTeX is currently used in the application. It might be more practical to document the supported syntax instead, perhaps even in the application UI, and keep using the current `\(...\)` / `\[...\]` inline/display delimiters.

ignaciogros (changes requested):
I tested this code in the preview URL:

```
The price is $10 and the total is $20

$4^2 + 1 = 17$

$$4^2 + 1 = 17$$

\(4^2 + 1 = 17\)

\[4^2 + 1 = 17\]
```

**Expected:** 4 formulas.

**Result:**

Image

I'm concerned that this could lead to additional errors.

Thanks.

mnarvaezm:
Hi everyone,

Previously, `$$...$$` was also used as a LaTeX delimiter, but it was eventually dropped precisely because of the kind of problems Ignacio mentions.

Currently, only two types of delimiters should be considered valid: `\(...\)` for inline formulas and `\[...\]` for display formulas.

I think we should keep it that way rather than reintroducing `$...$` or `$$...$$`. If necessary, we can make the supported syntax clearer in the documentation or in the application UI.

erseco (commented):
Totally agree, @ignaciogros and @mnarvaezm. Thanks both for the review and for taking the time to test this.

The reporter in #1990 confirmed that the currently supported `\(...\)` syntax works, and the expected syntax has already been explained there.

Given the ambiguity and risk of false positives with `$...$`, I don't think it's worth introducing this additional processing.

Closing this PR. Thanks again for the review.

PR #2281, opened 2026-08-15
fix(workarea): fall back to append when the drag reference node is stale

### LINKED ISSUE #2274: addIdeviceNodeToContainer throws NotFoundError when the drag reference node is stale
## Sentry report

Project `cloud-exelearning-dev`: [CLOUD-EXELEARNING-DEV-42](https://erseco.sentry.io/issues/?project=4511542266953808&query=issue%3ACLOUD-EXELEARNING-DEV-42) — `NotFoundError: Failed to execute 'insertBefore' on 'Node': The node before which the new node is to be inserted is not a child of this node.` in `B0.addIdeviceNodeToContainer` (app.bundle) — **1 event**.

## Root cause (verified)

`public/app/workarea/project/idevices/idevicesEngine.js:1159` — `container.insertBefore(ideviceBlockNodeContent, this.draggedElement)` never verifies that `this.draggedElement` is still a child of `container`. During drag-and-drop the reference node can already have been removed or re-parented (drop races, Yjs-driven re-render), making `insertBefore` throw and the drop fail.

## Proposed fix

```js
const ref = this.draggedElement?.parentNode === container ? this.draggedElement : null;
container.insertBefore(ideviceBlockNodeContent, ref); // null ⇒ append at the end
```

Low frequency, but the failure aborts the drop mid-way after the Yjs block was already created, so it is worth the two-line guard.

erseco:
Fixes #2274.

`addIdeviceNodeToContainer` called `container.insertBefore(node, this.draggedElement)` without verifying the reference node was still a child of the target. When a drop race or a concurrent Yjs re-render detached/re-parented the dragged element, the browser threw `NotFoundError` and the drop aborted mid-way — after the Yjs block had already been created. Both `insertBefore` sites in the method now verify parenthood and fall back to a `null` reference (append at the end).

**Sentry** (`cloud-exelearning-dev`, last 30d): [DEV-42](https://erseco.sentry.io/issues/?project=4511542266953808&query=issue%3ACLOUD-EXELEARNING-DEV-42) — `NotFoundError: Failed to execute 'insertBefore' on 'Node'` in `addIdeviceNodeToContainer` — 1 event (Chrome/macOS, production workarea).

**How to reproduce**
1. Drag an iDevice from the menu into the page content.
2. While the drop is in flight, let the content re-render underneath (concurrent Yjs update from a collaborator, or a drop race that detaches the drag placeholder).
3. Release: the Yjs block is created, then `insertBefore` throws `NotFoundError` because the reference node is no longer a child of the container.

**Tests**
`npx vitest run public/app/workarea/project/idevices/idevicesEngine.test.js` — 314/314 green (full frontend suite: 14429/14429). On unfixed `main` the 2 stale-reference tests fail with the exact `DOMException` from Sentry; the normal insert-before-reference path is covered and unchanged. `make fix` clean.

**Order alignment (review follow-up)**

`blockOrder` was computed from `draggedElement.getBoundingClientRect()` *before* the guard. A detached element reports an all-zero rect, so the loop matched the first block and sent `blockOrder = 0` to `bridge.addBlock()`, while the DOM fallback appended the node last — the block rendered at the bottom but was stored first in the Y.Doc until the next render. The reference is now resolved once, above the calculation, and reused for both the order loop and `insertBefore`; when it is stale `blockOrder` keeps its initial `existingBlocks.length`, so DOM and Yjs agree on "append". Covered by a test that fails on the previous commit with `0` instead of `1`.


ignaciogros (approved):
Thanks, @erseco.

One optional, non-blocking suggestion: in the first branch, `blockOrder` is calculated before the `insertBefore` from `this.draggedElement.getBoundingClientRect()`. If the element is detached, that rect is all zeros, so `blockOrder = 0`, while the new fallback appends it to the end of the DOM: the block ends up last on screen but first in the Y.Doc until the next render. Moving the guard above the calculation and using `blockOrder = existingBlocks.length` when the reference is stale would keep the DOM and Yjs aligned.

If you prefer to leave this for another PR, that’s OK. It can also be merged as is.

erseco:
@ignaciogros Applied in 02dff8dbe — good catch on the rect.

Reproduced it first: with a detached reference the loop sent `blockOrder = 0` to `bridge.addBlock()` while the DOM fallback appended the node last, and `syncNewIdeviceToYjs` only recomputes the block order when the block does *not* already exist in Yjs — here `addBlock()` runs before the insertion, so the mismatch stuck until the next render.

The reference is now resolved once, above the calculation, and reused for both the order loop and `insertBefore`; when it is stale, `blockOrder` keeps its initial `existingBlocks.length`, so the DOM and the Y.Doc agree on "append". The new test fails on the previous commit with `0` instead of `1`.

The other branch (iDevice into an existing block) doesn't have this divergence: `syncNewIdeviceToYjs` derives `componentOrder` from the DOM index *after* insertion, so appending is already consistent there.

`idevicesEngine.test.js` 314/314, full frontend suite 14429/14429, `make fix` clean.

ignaciogros (approved):
Thanks for those changes!

PR #2285, opened 2026-08-16
fix(assets): store upload chunks under FILES_DIR

### LINKED ISSUE #2283: Chunk uploads ignore FILES_DIR and write to process.cwd()/data/chunks
## Current behaviour

The chunked upload endpoint (`POST /api/projects/:projectId/assets/upload-chunk`) builds its temporary chunk directory with a hardcoded root, ignoring the configured files directory:

```ts
// src/routes/assets.ts (upload-chunk handler)
const chunksRoot = path.join(process.cwd(), 'data', 'chunks');
const chunkDir = safeJoin(chunksRoot, projectId, identifier);
```

Every other on-disk location (`assets/`, `tmp/`, `dist/`, `themes/`) resolves through `getFilesDir()` in `src/services/file-helper.ts` (`ELYSIA_FILES_DIR` → `FILES_DIR` → `./data` fallback). Chunk uploads are the only path that bypasses it.

## Expected behaviour

Chunk data should live under the configured `FILES_DIR`, e.g.:

```text
FILES_DIR/chunks/<projectId>/<identifier>/chunk_<n>
```

resolved through the existing `getFilesDir()` abstraction — no separate environment-variable parsing or duplicated fallback logic.

## Impact

In Docker or any deployment that sets `FILES_DIR` (e.g. `FILES_DIR=/mnt/data`), in-progress upload chunks are written outside the persistent data volume, into the container filesystem under the process working directory. They can fill the container layer, are lost on restart, and are invisible to volume-level backups and disk accounting.

## Acceptance criteria

- [ ] The chunk root resolves under the configured files directory (`FILES_DIR/chunks/...`) via the existing `getFilesDir()` from `src/services/file-helper.ts`, injectable through the `createAssetsRoutes()` dependency design.
- [ ] Uploading a chunk writes it under `<FILES_DIR>/chunks/<projectId>/<identifier>` and nothing is written under `process.cwd()/data/chunks`.
- [ ] Finalize, cancellation (`DELETE /upload-chunk/:identifier`) and the abandoned-upload sweeper all clean up the chunk directory from the configured location.
- [ ] Existing path traversal protections (`safeJoin`, `isSafePathSegment`) are preserved.
- [ ] Regression tests exercise the real `src/routes/assets.ts` route with a custom files directory (not a mirrored test-only implementation).

## Notes

Identified as an explicit out-of-scope follow-up in #2250 ("hardcodes the chunk upload root … should be a separate issue"); the sharding/portability work in #2266 deliberately leaves this untouched.

Related to #2250.

erseco:
## Summary

- Store resumable upload chunks under `FILES_DIR/chunks/<projectId>/<identifier>` instead of `process.cwd()/data/chunks`, resolving the root through the existing `getFilesDir()` from `src/services/file-helper.ts` (no duplicated env parsing or fallback logic).
- Expose `getFilesDir` through `AssetsFileHelperDeps` so tests can inject a custom files directory; path traversal protections (`safeJoin`, `isSafePathSegment`) are unchanged.
- Add regression coverage against the real route in `src/routes/assets.spec.ts`: chunk staging location, cancellation cleanup, and finalize cleanup all verified under an injected files dir, plus an assertion that nothing lands in the legacy `process.cwd()/data/chunks` location. (The mock file helper now also routes the pre-existing chunk tests through the cleaned-up test dir instead of leaking into `data/chunks`.)

## Tests

- `bun test src/routes/assets.spec.ts` (86 pass, includes 3 new regression tests written first and observed failing on the unfixed code)
- `bun test test/integration/routes/chunked-upload.integration.spec.ts` (6 pass)
- `make test-unit` (8001 tests, 0 fail, 90% per-file coverage threshold green)
- `make test-integration` (727 pass)
- `make fix` / `make lint` (clean; the 2 reported warnings are pre-existing unsafe-fix suggestions in unrelated E2E helpers)

Fixes #2283
Related to #2250


ignaciogros (approved):
Thanks, @erseco.

Non-blocking observation: orphaned chunk directories left behind after a process crash are still not being cleaned up, and they will accumulate on the volume. That's a pre-existing issue we might need to fix later.

PR #2290, opened 2026-08-16
feat(cli): assets:conflicts — list and resolve asset storage conflicts (keep-old / keep-new)

### LINKED ISSUE #2287: Asset storage: CLI to list and resolve migration conflicts (keep-old / keep-new)
Follow-up to PR #2266 (ADR-2250-01), from observation 3 of the review: conflicts detected by the asset storage migration are preserved safely but cannot be resolved without hand-editing files on disk.

## Background

When the startup migration finds a file at the legacy location (`assets/<uuid>/…`) **and** a different file already present at the canonical location (`assets/<shard>/<uuid>/…`) — compared by size and SHA-256 — it applies the conservative rule: it overwrites nothing, keeps both files, and parks the database row on its legacy location (rewritten to the portable relative form). This is the right default, but the other half is missing:

- the row never converges to the sharded layout;
- the same warning is re-emitted on every boot, indefinitely;
- the phase-2 sweep leaves the legacy directory in place, so the installation stays permanently in a mixed state;
- the only resolution today is SSH-ing in, comparing the two files by hand and deleting one.

## Proposal

Add a CLI reconciliation subcommand (alongside the existing `projects-cleanup`-style commands) that:

- **lists** unresolved asset-storage conflicts (rows still pointing at a legacy-form `storage_path` whose canonical counterpart also exists with different content);
- shows, per conflict, the **legacy and canonical absolute paths** plus useful metadata (**file sizes**, mtimes) so the operator can decide without hunting for the files;
- resolves each conflict only with an **explicit choice**: `keep-old` (legacy copy wins, moved to the canonical location) or `keep-new` (canonical copy wins, legacy copy removed);
- **never** overwrites or deletes either copy without that explicit operator choice;
- leaves the row pointing at the canonical sharded location afterwards, so the next startup migration converges and the legacy directory can finally be swept.

Additionally, improve the per-boot conflict warning emitted by `src/services/asset-storage-migration.ts` so it prints **both full absolute paths and each file's size**, making the decision possible straight from the logs.

## Notes

- Conflicts should be extremely rare (they require something to already exist at the destination before migrating), which is why this was deliberately kept out of #2266.
- The never-overwrite rule of ADR-2250-01 stays authoritative; this tool is the explicit-consent path around it.

erseco:
Implements #2287, the conflict-resolution follow-up from the review of #2266.

> [!IMPORTANT]
> **Stacked on #2266** (base branch is `2250-asset-storage-does-not-scale`). Do **not** merge before #2266 lands in `main`; GitHub retargets this PR automatically afterwards. Issue #2287 is marked *blocked by* #2250 accordingly.

## What

- `bun cli assets:conflicts` lists the conflicts parked by the startup migration (ADR-2250-01 never-overwrite rule): asset id, project, and **both absolute paths with file sizes and mtimes**. `--json` prints raw machine-parseable output.
- `bun cli assets:conflicts resolve <asset-id> --keep-old|--keep-new [--dry-run]` resolves **one** conflict with an explicit operator choice — nothing is ever deleted or overwritten without one of the two flags. `--keep-old` moves the legacy copy into the canonical sharded location; `--keep-new` keeps the canonical copy. Copies that turn out identical converge with the migration's own decision-free rule.
- Crash-safe by construction: operation order plus the migration's optimistic `WHERE storage_path = <old>` guard mean any interruption (or a concurrently starting instance) leaves a state the next startup migration reconciles; the emptied legacy directory is tidied by the next boot's sweep.
- No duplicated semantics: the service reuses the migration's `filesAreIdentical` / `moveFile` / `rewriteRow` (now exported), plus a new pure `deriveShardedAssetStoragePath` in the shared path grammar (`src/utils/asset-paths.ts`).
- The per-boot conflict warnings now print both absolute paths and sizes and point at the CLI (the other half of issue #2287).

## How to test

Unit level (real filesystem + real test DB):

```bash
bun test src/services/asset-conflicts.spec.ts src/cli/commands/assets-conflicts.spec.ts
```

Manually, on a dev install (fabricate a conflict, then resolve it):

```bash
# 1. Upload any image to a project, note the asset row:
#    UUID=<project uuid>  FILE=<stored filename>  ID=<asset id>  FILES=./data
# 2. Park the row on a diverging legacy copy:
mkdir -p "$FILES/assets/$UUID"
echo "legacy version" > "$FILES/assets/$UUID/$FILE"
sqlite3 "$FILES/exelearning.db" \
  "UPDATE assets SET storage_path='assets/$UUID/$FILE' WHERE id=$ID;"
# 3. Restart the app: the boot log reports the conflict with both paths and sizes.
# 4. Inspect and resolve:
bun cli assets:conflicts
bun cli assets:conflicts list --json
bun cli assets:conflicts resolve $ID --keep-new --dry-run
bun cli assets:conflicts resolve $ID --keep-new
# 5. Restart again: no conflict warning; the legacy directory is gone.
```

## Docs

- New change document `doc/architecture/changes/2287-asset-conflict-cli/design.md` (scope, resolution semantics, why phase-2 orphan conflicts stay out).
- `doc/deployment.md`: new "Asset storage conflicts" maintenance section.
- Cross-references updated in ADR-2250-01 and the #2250 admin guide.

## Validation

`make fix`, `make architecture-check`, `make test-unit` (8,344 pass; all files ≥ 90 % coverage — new service 98.8 %, new command 99 %), `make test-integration` (727 pass), `make test-e2e` (530 passed, 6 skipped, 0 failed). No translations touched; no schema changes.

### DIFF
--- a/doc/architecture/adr/ADR-2250-01-shard-project-asset-storage.md
+++ b/doc/architecture/adr/ADR-2250-01-shard-project-asset-storage.md
@@ -275,7 +275,7 @@ We will:
 - A transitional legacy-read fallback exists in the resolver until installations
   converge; it must eventually be removed (see "Follow-up work").
 - Startup performs a cheap, bounded check per boot even on converged installations: one
-  indexed-free `NOT LIKE 'assets/%'` query, one readdir of the assets root, and a
+  indexed-free `NOT LIKE 'assets/__/%'` query, one readdir of the assets root, and a
   content check (`readdir`) of each existing two-decimal shard bucket (`00`-`99`) to
   distinguish it from a legacy numeric project-id directory.
 
--- a/doc/architecture/adr/ADR-2250-01-shard-project-asset-storage.md
+++ b/doc/architecture/adr/ADR-2250-01-shard-project-asset-storage.md
@@ -340,7 +340,8 @@ We will:
   the migration's per-boot summary tells operators when their installation is clean).
 - Provide a CLI subcommand that lists parked migration conflicts and resolves them
   with an explicit keep-old / keep-new choice
-  (<https://github.com/exelearning/exelearning/issues/2287>).
+  (<https://github.com/exelearning/exelearning/issues/2287>, implemented as
+  `assets:conflicts` — see the change document for #2287).
 - Consider validating client-supplied project identifiers at creation time so the
   FNV-1a fallback becomes purely defensive.
 
--- a/doc/architecture/changes/2250-asset-storage-sharding/design.md
+++ b/doc/architecture/changes/2250-asset-storage-sharding/design.md
@@ -133,8 +133,8 @@ each boot). Unrecognized entries are reported and left alone.
 - **Verifying success**: after the upgrade boot, the log shows the summary; on the next
   boot it shows `Asset storage layout is up to date.` and
   `SELECT COUNT(*) FROM assets WHERE storage_path NOT LIKE 'assets/%'` returns 0.
-  Remaining warnings identify conflict files to resolve manually (a reconciliation
-  CLI is tracked in issue #2287).
+  Remaining warnings identify conflict files; list and resolve them with
+  `bun cli assets:conflicts` (issue #2287).
 - **Rollback**: restore the pre-upgrade backup (database + `FILES_DIR`) and redeploy the
   previous release. Rolling back only the binary after migration is not supported —
   old releases cannot read relative paths.
--- a/doc/architecture/changes/2287-asset-conflict-cli/design.md
+++ b/doc/architecture/changes/2287-asset-conflict-cli/design.md
@@ -1,0 +1,133 @@
+---
+tracking_issue: 2287
+title: "CLI resolution of asset storage conflicts"
+status: accepted
+date: 2026-08-16
+authors:
+  - "@erseco"
+reviewers:
+  - "@ignaciogros"
+implementation_prs: []
+related_adrs:
+  - "ADR-2250-01"
+supersedes: []
+superseded_by: []
+ai_assistance:
+  tool: "Claude Code"
+  model: "claude-fable-5"
+---
+
+# CLI resolution of asset storage conflicts — design
+
+## Current state
+
+The startup migration introduced by #2250 (ADR-2250-01) applies a strict
+never-overwrite rule: when a row's legacy file (`assets/<uuid>/…`) and its
+canonical sharded destination (`assets/<shard>/<uuid>/…`) both exist with
+different content (size + SHA-256), both files are kept and the row is parked
+on its legacy location in portable relative form. Parked rows never converge:
+the same warning is re-emitted on every boot, the phase 2 sweep leaves the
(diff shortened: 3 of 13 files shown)


ignaciogros (commented):
Thank you for this tool!

Before merging this, could you please take a look at these two points?

1. **Crash recovery for parked rows.** The claim that any interruption is reconciled by the next startup migration does not hold for the rows this command targets: phase 1 selects with `storage_path NOT LIKE 'assets/%'`, but a parked conflict row is rewritten to exactly `assets/<uuid>/<file>`, so later boots skip it, and phase 2 never rewrites rows. If `resolve` is interrupted between the `fs.remove` / `moveFile` and the `rewriteRow`, the row ends up pointing at a file that no longer exists and nothing fixes it later, which also makes the incomplete message ("the startup migration will converge it automatically") misleading for these rows.
Maybe `resolve` could self-heal that state, or the phase 1 filter could be widened to also pick up non-sharded `assets/...` paths (e.g. `NOT LIKE 'assets/__/%'`).
Could you please confirm this point? If you'd rather leave it as it is, that is OK too.

2. **Warning wording.** The phase 2 sweep warning still says "Resolve manually", while the phase 1 warning now points at `bun cli assets:conflicts`. Perhaps it would be worth unifying them.

erseco:
@ignaciogros Both points addressed in 17c64ced9:

1. **Confirmed.** Phase 1's `NOT LIKE 'assets/%'` filter skipped parked rows, so an interrupted `resolve` could leave a row pointing at a missing file with nothing fixing it later. Fixed by widening the filter to `NOT LIKE 'assets/__/%'` (the option you suggested): parked rows are re-examined at every startup, so any interrupted resolution self-heals through the existing crash-recovery branches, and the "converge automatically" message is now accurate. The unmounted-volume safety latch uses the widened filter too, and re-reporting a still-unresolved conflict no longer rewrites the row (no `updated_at` churn). Covered by four new tests.

2. **Half-taken.** The phase 2 sweep conflicts are orphan files with no database row, so `assets:conflicts` cannot list or resolve them — pointing at the CLI there would be misleading. Instead the warning now explains why it is manual: "No database row references this file, so 'bun cli assets:conflicts' cannot resolve it; compare the copies and remove one manually."

ignaciogros (approved):
Thanks, @erseco.

Both points have been fixed.

Three small details that we can address later, none of them blocking:

1. `_` is a SQL wildcard, so `assets/__/%` also excludes two-character legacy directories: a row parked at `assets/12/<file>` (numeric project IDs 10–99) still skips phase 1 and would not self-heal after an interrupted `resolve`. The CLI still lists and resolves conflicts on those directories normally; only the recovery after an interrupted resolve is affected, and such a row stays reported as incomplete.
2. `migrateRow` still computes `alreadySharded` / `relWithinProject` inline instead of calling `deriveShardedAssetStoragePath`, which does the same thing (the derivation is duplicated).
3. There's a redundant `ensureDir` after `fs.remove(dest)` in the keep-old branch, and `resolve` exits with 1 on an already-canonical row rather than treating it as an idempotent no-op.

PR #2297, opened 2026-08-17
feat(preview): add fullscreen fallback when preview extraction fails

### LINKED ISSUE #2296: [UX/Bug] Missing visual fallback when preview extraction to a new tab fails
### Problem Description
When the user triggers `extractToNewTab()` and the Service Worker (SW) is unavailable, the application only logs an error in the console. This leaves the user without any visual feedback or useful view of the content.

### Proposed Solution
1. Implement a visual fallback in `extractToNewTab()` that expands the panel to full width (`previewFullscreen`) when the Service Worker fails.
2. Ensure that the `close()` method always clears the `previewFullscreen` class to prevent inconsistent states when reopening the panel.

### Acceptance Criteria
- [ ] If the SW verification fails in `extractToNewTab()`, the panel toggles to full-screen mode.
- [ ] Executing `close()` removes the `previewFullscreen` class.

franmate:
## Description
When the user attempts to extract the preview to a new tab (`extractToNewTab()`) and the Service Worker is unavailable, the application no longer simply logs an error to the console: the panel now expands to full width (`previewFullscreen`) as a visual fallback. This ensures the user retains a clear, useful view of the content directly within the panel.

### `previewPanel.js`
- **`extractToNewTab()`**: When the Service Worker availability check fails, `this.panel.classList.toggle('previewFullscreen')` is added as a fallback, expanding the panel to `100vw`.
- **`close()`**: Removes the `previewFullscreen` class when closing the panel to prevent the fallback's fullscreen state from incorrectly persisting on the next open.

### `previewPanel.test.js`
- **New test**: Verifies that a SW failure triggers the fullscreen fallback in `extractToNewTab()`.
- **New test**: Verifies `toggle()` behavior across repeated failed attempts (alternates active/inactive state on each call).
- **New test**: Verifies that `close()` correctly resets the `previewFullscreen` class.

## Why
Without this change, an extraction failure left the user with only a console error and no helpful visual feedback. Now, the panel adapts to provide the best possible viewing experience under failure conditions. Additionally, resetting the state in `close()` prevents the fullscreen layout from sticking across subsequent panel opens.

## Testing
- [x] Unit tests added and passing successfully.
- [x] Manual verification: Extraction failure → panel goes fullscreen → panel closed → panel reopens in normal state.

Closes: #2296


franmate:
## Manual Testing (Moodle)

Tested using Moodle plugins (e.g., `mod_exeweb`):

erseco (changes requested):
Thanks for tackling this, @franmate. The premise is right: when the SW is unavailable, `refresh()` already falls back to `refreshWithBlobUrl()` (`previewPanel.js:748`), so the panel *does* hold real content and widening it is a sensible fallback. The problem is how it is wired.

I reviewed this against `ee9fa2fcb` and I'm requesting changes: there are two reproducible bugs, and the fallback only covers one of the three ways extraction can actually fail.

---

## 1. [Blocking] The fallback does nothing when the preview is pinned

**Why it fails.** Both buttons call the same `extractToNewTab()` — `#preview-extract-button` (`workarea.njk:189`) and `#preview-pinned-extract-button` (`workarea.njk:123`). But the fallback only touches `this.panel`, which is the slide-out panel:

- `previewPanel.js:16` → `this.panel = document.getElementById('previewsidenav')`
- `previewPanel.js:1354` → `this.panel?.classList.toggle('previewFullscreen')`

In pinned mode that element is hidden by `_previewpanel.scss:185`:

```scss
#workarea[data-preview-pinned="true"] ~ #previewsidenav-content { display: none; }
```

So the class is applied to the element the user cannot see, while they are looking at `#preview-pinned-container`.

**Repro** (viewport > 992px, otherwise the pin button is hidden by the media query):

1. Open a project, open the preview, click Pin.
2. Make the SW unavailable. Unregistering in DevTools is not reliable here, since `isServiceWorkerPreviewAvailable()` may still find an active worker; the deterministic way is:
   ```js
   window.__origGetSW = window.eXeLearning.app.getPreviewServiceWorker;
   window.eXeLearning.app.getPreviewServiceWorker = () => null;
   ```
3. Click the extract button **in the pinned panel**. Nothing happens visually.
4. ```js
   document.getElementById('previewsidenav').classList.contains('previewFullscreen');       // true
   document.getElementById('preview-pinned-container').classList.contains('previewFullscreen'); // false
   ```
5. Now click **Unpin** — the slide-out panel appears already in fullscreen, because the class added while it was hidden is still there.

**What's needed.** Not just swapping `this.panel` for a ternary: `#preview-pinned-container` has its own `width: 45%; max-width: 60%` (`_previewpanel.scss:170-176`) and there is no `previewFullscreen` rule for it, so a JS-only fix will still render nothing. Both the JS target and a matching SCSS rule are required.

Worth noting this file already has the idiom five times — `this.isPinned ? this.pinnedIframe : this.iframe` at lines 1269, 1307, 1403, and the equivalent in `showLoadingState()` / `hideLoadingState()`. Reusing it (e.g. a small `_setFullscreenFallback(on)` helper) would have avoided this and keeps us on the "single source of truth" rule in AGENTS.md.

---

## 2. [Blocking] The cleanup is unreachable in exactly the states that need it

**Why it fails.** `close()` bails out before reaching the new line:

```js
close() {
    if (this.isPinned) return; // Can't close when pinned   <-- previewPanel.js:653
    ...
    this.panel?.classList.remove('previewFullscreen');      <-- previewPanel.js:658
}
```

So in pinned mode the class is set (bug 1) and never cleared. `pin()` and `unpin()` don't clear it either, and `resetToDefaultState()` (`previewPanel.js:74-86`) removes `active` but not `previewFullscreen`.

**Repro A** — the class survives a pin/unpin cycle even without pinned extraction:

1. Preview open, not pinned. Make the SW unavailable, click extract → panel goes fullscreen.
2. Click Pin, then Unpin → the panel comes back in fullscreen.

**Repro B** — it leaks into a different project:

1. Pinned + SW unavailable + extract (the workarea is still usable in pinned mode).
2. File → New / Open project → `projectManager.js:849-851` calls `previewPanel.resetToDefaultState()`, which does not clear the class.
3. Open the preview in the new project → `open()` adds `active` and the panel opens at `100vw`, `position: fixed`, `z-index: 100001`, covering the whole app.

**What's needed.** Centralise the reset and call it from `close()`, `pin()`, `unpin()` **and** `resetToDefaultState()` — keeping in mind that `close()` alone can never be the place, because of the early return above.

---

## 3. [Blocking] Only 1 of the 3 failure paths gets a fallback

The issue is titled "missing visual fallback when preview extraction to a new tab fails", but extraction can fail three ways:

| Path | Line | Fallback? |
|---|---|---|
| `!isServiceWorkerPreviewAvailable()` | 1352 | yes |
| `window.open()` returns `null` (popup blocked) | 1378 | no — `Logger.warn` + an `a.click()` that is equally inert |
| `refreshWithServiceWorker()` throws | 1386 | no — `Logger.error` only |

The popup-blocked case is the most common one in production (browser popup blocker, LMS iframe with `sandbox` and no `allow-popups`, Electron). From the user's point of view it is indistinguishable from the SW case — the button does nothing — yet it gets no fallback at all.

Also worth keeping in mind: the SW branch is not an edge case. The comment at `previewPanel.js:746` says "Fallback for cross-origin iframes where SW can't register" — in embedded deployments (WordPress / Moodle / Nextcloud) the SW is *never* available, so this is the normal path there, which makes bugs 1 and 2 much more likely to be hit than they look.

---

## 4. [Discussion] The failure path is silent, and the button now means two things

I'm not going to block on `toggle()` vs `add()` — the acceptance criterion in #2296 literally says "toggles", so I read this as deliberate. But two related points:

- **No feedback.** The panel just gets wider. The user has no way to know whether the app hung, the click missed, or the browser blocked something. `showError()` already exists in this class and could carry a short translated message (`_()`), and the issue's own goal is *visual* feedback. Right now nothing is added, so there's also no new string to translate.
- **The button silently changes meaning.** `#preview-extract-button` and `#preview-pinned-extract-button` are labelled `title="Open in new tab"` / `aria-label="Open preview in new tab"`. On failure they become a layout toggle, with no `aria-pressed`, no tooltip change and no live-region announcement. A screen-reader user presses "Open preview in new tab", nothing opens, nothing is announced, the layout changes, and pressing again silently reverts it.

If we do keep `toggle()`, could we document *why* in the code rather than only in a test comment? And if we switch to `add()`, we should give the user an explicit way back (the close button works, but that also throws away the preview).

---

## 5. Tests: 100% line coverage, 0% coverage of the actual bug

The new tests (`previewPanel.test.js:853-871`) only exercise `isPinned === false`, against a detached `<div>`. No test calls `pin()` first, and none checks which container is the visible one. That's exactly why Codecov reports every changed line covered while bug 1 goes unnoticed — line coverage is not behaviour coverage.

Please add:

- an `extractToNewTab()` test with `isPinned === true`, asserting the class lands on the container that is actually visible;
- cases for `close()` while pinned, `unpin()`, and `resetToDefaultState()` with the class already set;
- a Playwright spec — AGENTS.md 5.3 requires one for user-visible workarea/preview behaviour, and this PR doesn't add any.

The `should flip the fullscreen fallback on repeated failed attempts` test also locks the toggle behaviour into the contract; whichever way we settle point 4, that test should reflect a decision we've stated explicitly.

---

## 6. Nits

- `_previewpanel.scss:61-66`: the new rule is indented with a **tab** in a 4-space file, lines 61 and 66 carry trailing whitespace, and it's `width:100vw` instead of `width: 100vw`. Biome doesn't format SCSS, so `make fix` won't catch any of it.
- `previewFullscreen` is camelCase among `.preview-panel-header`, `.preview-pinned-body`, `.preview-device-frame`, etc. The name comes straight from the issue, but with only two usages it's cheap to normalise now (`.preview-fullscreen`).
- Below 992px the rule is a no-op: `#previewsidenav-content .preview-sidenav` is already `width: 100vw` there (`_previewpanel.scss:347`), so the fallback produces no visible change on small viewports.
- The `### _previewpanel.scss` section in the PR description is empty.

---

One last thought on scope: #2296 is written as an implementation recipe (it names the class, `toggle`, and `close()`), and this PR follows it faithfully — which is why it satisfies both acceptance criteria and still leaves the user without a fallback from the pinned panel and from the most common failure mode. I'd rather we fix the underlying behaviour than the checkboxes. Happy to re-review as soon as the pinned path and the cleanup are sorted.

franmate:
Thanks for the thorough review, @erseco  — you were right on all three blocking points, and the pinned/cleanup bugs were exactly the kind of thing line coverage alone wasn't going to catch.

## 1. Pinned fallback not visible

Reworked: instead of toggling a class on two separate containers, the pinned case now unpins first (`unpin()`) and always falls back to the slide-out panel, which is the one `.preview-fullscreen` actually styles. `_wasPinnedBeforeFullscreen` restores the pinned state if the user clicks extract again while still failing. Covered in both `previewPanel.test.js` and a new e2e spec (`preview-fullscreen-fallback.spec.ts`).

## 2. Cleanup unreachable

`close()`, `resetToDefaultState()`, and the unpin path inside `extractToNewTab()` now all clear `isFullscreen` / `_wasPinnedBeforeFullscreen` and the `preview-fullscreen` class.

## 3. Only 1 of 3 failure paths had a fallback

All three (SW unavailable, popup blocked, and the `refreshWithServiceWorker()` catch) now go through the same fallback.

One thing worth flagging explicitly: for the popup-blocked path, we can't detect whether `a.click()` actually succeeded, so the fallback always fires there — meaning if the click *does* open a tab, the user will also see the panel enlarge unnecessarily. Went with your read that "equally inert" from the user's perspective outweighs that occasional false positive.

## 4. Feedback / button meaning

Button now reflects state via:
- `aria-pressed="true"` while enlarged (with visual feedback: brand-primary color, gray background, and an inset border)
- Dynamic `title` / `aria-label` (`_('Enlarge preview')` when collapsed, `_('Restore preview size')` when enlarged, and `_('Open in new tab')` when SW is available)
- The icon remains the same (`aspect_ratio`) for both collapsed and enlarged states. This keeps the UI cleaner and reduces cognitive load.
- A toast explains what happened:

> "Couldn't open in a new tab. Showing an enlarged preview instead."

This deliberately doesn't name a specific cause (SW / popup / error), and avoids "fullscreen" in user-facing copy since the panel doesn't use the actual browser Fullscreen API, just widens within the page.

## 5. Tests

Updated unit and e2e tests:
- `_updateExtractButton` tests now verify the icon class stays `enlarge-icon` in both collapsed and enlarged states, with `aria-pressed` differentiating the state.
- Added verification that `aria-pressed` and tooltips update correctly
- Added unit coverage for pinned extraction, `close()` / `unpin()` / `resetToDefaultState()` with fullscreen state already set, and all 3 failure paths (including the pinned + popup-blocked combination)
- Playwright spec with 9 tests covering the same ground end-to-end, including icon state verification
Added a Playwright spec with 9 tests covering the same ground end-to-end.


## 6. Nits

SCSS formatting fixed (tabs / trailing whitespace / `width: 100vw` spacing); renamed the class to kebab-case for consistency with the rest of the stylesheet.

---

**Note:** Other issues regarding preview generation are out of scope for this PR.

franmate:
### Demo with forced error simulation: ###
https://franmate.github.io/eXeLearning-preview-fallback-demo/

It simulates an environment and demonstrates the visual fallback behavior when the preview can't open in a new tab.

PR #2278, opened 2026-08-15
fix(idevices): guard edition event handlers against cleared $exeDevice

erseco:
Fixes #2271.

The workarea clears the global on editor close (`ideviceNode.js` sets `$exeDevice = undefined`) while DOM handlers bound by the edition scripts stay attached, so late clicks/inputs or pending async callbacks threw `TypeError` — 15 Sentry issues, 56 events. This guards every event-handler dereference of `$exeDevice` in the four affected edition scripts (quick-questions-video, quick-questions-multiple-choice, map, dragdrop): optional chaining for one-liners, early `if (!$exeDevice) return;` for multi-statement and async callbacks (YT player events, image/FileReader loads, `setInterval` ticks, TinyMCE init).

**Sentry** (`cloud-exelearning-dev`, last 30d, 56 events across 15 issues): DEV-1F (10), DEV-11 (8), DEV-21 (7), DEV-3Z (6), DEV-3R (6), DEV-3Y (4), DEV-20 (3), DEV-3P (3), DEV-2G (2), DEV-1Z (2), DEV-2Q (1), DEV-6 (1), DEV-1G (1), DEV-3S (1), DEV-40 (1) — [filtered view](https://erseco.sentry.io/issues/?project=4511542266953808&query=issue%3A%5BCLOUD-EXELEARNING-DEV-11%2CCLOUD-EXELEARNING-DEV-21%2CCLOUD-EXELEARNING-DEV-3Z%2CCLOUD-EXELEARNING-DEV-3Y%2CCLOUD-EXELEARNING-DEV-20%2CCLOUD-EXELEARNING-DEV-2G%2CCLOUD-EXELEARNING-DEV-1Z%2CCLOUD-EXELEARNING-DEV-40%2CCLOUD-EXELEARNING-DEV-1F%2CCLOUD-EXELEARNING-DEV-3P%2CCLOUD-EXELEARNING-DEV-2Q%2CCLOUD-EXELEARNING-DEV-6%2CCLOUD-EXELEARNING-DEV-1G%2CCLOUD-EXELEARNING-DEV-3S%2CCLOUD-EXELEARNING-DEV-3R%5D).

**How to reproduce**
1. Edit a Video test (or Quick multiple choice / Interactive map / Drag and drop) iDevice so its edition form binds its handlers.
2. Close the editor (save or cancel) — the global `$exeDevice` is cleared while form DOM handlers are still attached during teardown.
[...]

Issue #2271, opened 2026-08-15
iDevice edition handlers crash when $exeDevice is already cleared (15 Sentry issues)


erseco:
## Sentry report

Project `cloud-exelearning-dev`, last 30 days — 15 issues, **56 events**, all the same failure mode: an edition-script DOM handler fires and dereferences the global `$exeDevice`, which is `undefined`.

| iDevice (edition) | Sentry issues (events) |
|---|---|
| quick-questions-video | DEV-11 `addQuestion` (8), DEV-21 `timeVIFocus` (7), DEV-3Z `validTime` (6), DEV-3Y `loadVideo` (4), DEV-20 `playQuestionVideo` (3), DEV-2G `previousQuestion` (2), DEV-1Z `showOptions` (2), DEV-40 `questionsGame` (1) |
| quick-questions-multiple-choice | DEV-1F `nextQuestion` (10), DEV-3P `previousQuestion` (3), DEV-2Q `lastQuestion` (1) |
| map | DEV-6 `nextPoint` (1), DEV-1G `placeImageWindows` (1), DEV-3S `showImage` (1) |
| dragdrop | DEV-3R `nextCard` (6) — Firefox wording names the culprit: `can't access property "nextCard", $exeDevice is undefined` |

Search any of them: `https://erseco.sentry.io/issues/?project=4511542266953808&query=issue%3ACLOUD-EXELEARNING-DEV-<id>`

## Root cause (verified)

`ideviceNode.js` sets `$exeDevice = undefined` when the edition closes (`public/app/workarea/project/idevices/content/ideviceNode.js:1647`, `:1768`, `:3215`), but handlers bound by the edition scripts (e.g. `$exeDevice.nextQuestion()` at `quick-questions-video/edition/quick-questions-video.js:1894`) stay attached to DOM that can outlive the editor. A late click/input after close (or during the async teardown window) throws.

## Proposed fix

One shared guard rather than 15 point fixes: route edition handlers through a wrapper that no-ops when `$exeDevice` is falsy (in the shared gamification edition helper), and/or unbind edition-bound listeners in the same code paths that clear `$exeDevice`. Clears the whole family at once.

PR #2301, opened 2026-08-18
fix(idevices): own edition resources with an explicit lifecycle

### LINKED ISSUE #2293: Fix iDevice edition teardown globally before clearing `$exeDevice`
## Problem

PR #2278 fixes the currently reported `$exeDevice` errors by guarding the affected call sites. This prevents the known user-facing errors and was appropriate as an immediate fix.

However, as pointed out by @ignaciogros in the [PR review](https://github.com/exelearning/exelearning/pull/2278#pullrequestreview-4950086204), the underlying lifecycle problem remains.

When an iDevice editor is closed, `ideviceNode.js` clears the global `$exeDevice`, but DOM event handlers, timers, asynchronous callbacks, or other resources created by the edition script may still be active.

This means that:

- Other edition scripts using the same pattern may still trigger the same `TypeError`.
- New handlers added in the future could reintroduce the problem.
- Individual guards such as `if (!$exeDevice) return;` protect specific call sites but do not guarantee that the editor has been properly torn down.

## Expected solution

The iDevice edition lifecycle should be fixed centrally so that resources belonging to an edition instance cannot continue executing after that instance has been destroyed.

As suggested by @ignaciogros, one possible approach would be to remove the edition form DOM and unbind its handlers **before** clearing `$exeDevice` in `ideviceNode.js`.

The exact implementation should be evaluated, but the solution should be generic rather than adding guards to every individual edition script.

The teardown process should cover, where applicable:

- DOM event handlers.
- Timers and intervals.
- Pending asynchronous callbacks.
- Editor/plugin instances.
- Other resources associated with the current iDevice edition instance.

Only after the edition instance has been properly disposed should `$exeDevice` be cleared.

## Acceptance criteria

- Closing or cancelling an iDevice edition performs a centralized cleanup before `$exeDevice` is cleared.
- Event handlers associated with a closed edition cannot execute against a cleared `$exeDevice`.
- Pending asynchronous operations from a closed edition cannot cause `$exeDevice` `TypeError`s.
- The solution applies generically to iDevices rather than maintaining a list of guarded call sites.
- Existing protections introduced in #2278 continue to work while the generic lifecycle handling is introduced.
- Automated tests verify that callbacks/events fired after closing an editor do not access a destroyed edition instance.

## References

- #2271
- #2278
- [Review comment by @ignaciogros](https://github.com/exelearning/exelearning/pull/2278#pullrequestreview-4950086204)

erseco:
Closes #2293.

## Problem

Closing an iDevice editor cleared the global `$exeDevice` but released nothing else. Handlers, timers and pending asynchronous callbacks created by the edition script stayed alive, and the edition form was only replaced later — so anything firing in between dereferenced `undefined`. #2271 recorded 15 Sentry issues and 56 events of exactly that shape; #2278 guarded the four reported scripts.

As @ignaciogros noted [reviewing #2278](https://github.com/exelearning/exelearning/pull/2278#pullrequestreview-4950086204), the guards treat the symptom. The same pattern exists in the other edition scripts, and a newly added handler reintroduces it.

## Root cause

There was no owner. `$exeDevice` was a single mutable global, and three sites cleared it (`loadEditionIdevice`, `restartExeIdeviceValue`, `remove`) with no destructor anywhere. The gap between clearing the global and replacing the form spans a network fetch, and for JSON iDevices up to 5 s of polling.

Worse than the crash: a `$exeDevice?.method()` guard does **not** make a stale callback safe. When it finally runs, the global usually holds a perfectly valid *different* device — the one the user just opened. The guard passes and the callback drives the wrong iDevice.

## Solution

Every edition gets an `EditionLifecycle` (`public/app/workarea/project/idevices/content/editionLifecycle.js`) that owns what it creates: timers, native and jQuery handlers, abortable work, file readers, media elements and third-party instances, plus anything else through a plain `own(disposer)`. `IdeviceNode.initExeDeviceEdition()` publishes it as `$exeDevice.$lifecycle` (and `window.$exeEditionLifecycle` for the shared helpers) *before* calling `init()`.

All four clearing sites now route through one `IdeviceNode.destroyEditionInstance()`:

1. stop the initialization poll;
2. dispose the lifecycle — mark it inactive, cancel timers, abort the signal, run the iDevice's optional `destroyEdition()` hook while its instance and DOM are still alive, then run disposers LIFO (jQuery handlers included, unregistered by a namespace unique to that edition);
3. remove the TinyMCE editors whose target is inside the form, while those targets are still in the document;
4. unbind the form through `jQuery.cleanData` — which native removal and `innerHTML` assignment both skip — leaving the markup on screen until the export view replaces it;
5. release `$exeDevice`, **last**.

The fourth site is `IdevicesEngine.destroyEditionIdevices()`: page switching detaches iDevice DOM directly without ever calling `IdeviceNode.remove()`, so that path previously had no teardown at all. It walks every node rather than disposing the single active lifecycle, because `destroyEditionInstance()` also does per-node work the lifecycle does not own — stopping that node's initialization poll, and releasing a `$exeDevice` whose script loaded but never reached `init()`.

Callbacks registered through the lifecycle are bound to the device that owned them, so **a callback created by edition A can never operate on edition B**.

**Ownership is derived, never mirrored.** The active lifecycle lives in one module slot and names its `ownerNode`, so a node asks "is the open edition mine?" instead of holding a reference that could go stale. A node that owns no edition leaves the open one — and `$exeDevice` with it — untouched. That matters because collaborative sync renders remote iDevices through the export path while a different iDevice is open locally; clearing the global there used to strand that editor. `isSync` keeps its meaning and still skips teardown entirely.

`destroy()` also drops the edition it owned. `bind()` closes over the whole lifecycle, so a bound callback that outlives teardown — parked in an application-lifetime registry, say — would otherwise keep the form's DOM subtree and the owning node's entire engine reachable. The File Manager modal is exactly such a registry, and `file-attachment` now releases its `onSelect` with the edition.

## iDevice audit

All 61 edition files (55 iDevice scripts + shared helpers) were classified.

| Outcome | Files |
|---|---|
| Migrated (owned a resource that can outlive the form) | 44 |
| No owned long-lived resources | 6 |
| Vendored / generated, out of scope | 3 |
| Shared edition helpers migrated | `common_edition.js` |

209 resources were inventoried. Migrated by class: **15** timers, **30** handlers on `document`/`window`/shared targets, **14** YouTube players, **29** FileReaders, **18** media elements, plus observers, object URLs, fetches, injected-script callbacks and promise continuations.

The migration is also what decided the API surface. Helpers that finished it with no call site (`run()`, `one()`, `cancelAnimationFrame()`) were removed, and two that turned out to be exact aliases of `own()`/`ownInstance()` at their two call sites each (`ownObserver()`, `ownObjectUrl()`) were inlined. What stays encodes a cleanup recipe a bare disposer would get wrong at each site — `ownFileReader()` aborts only an *in-flight* read; `ownMedia()` pauses, clears the source, drops child `<source>` elements and calls `load()`; `readFile()` guarantees the promise settles; and `ownMedia()`'s optional slot keeps one live element where the edition rebuilds one per click.

One deliberate exception to "the edition owns its resources": the LOMLOE tooltip controller is a page-scoped singleton whose `__lomloeTipBound` guard is shared with its export script, so releasing it on close would kill every other LOMLOE tooltip on the page with nothing able to reinstall it. It captures no edition state.

**Form-local jQuery handlers were deliberately not rewritten.** Step 4 above already unbinds them, and a test proves a handler bound directly by an edition stops firing after teardown. Rewriting roughly a thousand of them would be a far larger change than the defect. The #2278 guards are kept.

A post-implementation static sweep re-ran the inventory searches; every remaining raw timer, listener and resource constructor is classified — vendored TikZJax, the interactive-video editor's own iframe (destroyed wholesale on close), deliberate no-lifecycle fallbacks, or self-managed by existing teardown.

## Lifecycle guarantees

- Cleanup runs **before** `$exeDevice` is cleared or replaced.
- Teardown is idempotent — save-then-close, delete-during-close, failed init.
- A callback from a closed edition cannot run, and cannot reach the next one.
- Timers, `document`/`window` handlers and abortable work are cancelled; unrelated handlers on those shared targets are untouched.
- A throwing disposer never strands a half-destroyed global: cleanup continues, errors are collected and logged.
- Partial initialization is safe.

## Tests

93 unit tests in the two files dedicated to the lifecycle and the `IdeviceNode` teardown, plus per-iDevice coverage for each migrated resource, and a Playwright spec covering save / discard / delete, open A → open B, repeated cycles and page switching — asserting no page errors throughout, and that a page switch leaves neither the workarea overlay nor a modal scroll lock behind. Several migration groups verified their tests by reverting the source and confirming the new assertions fail.

| Command | Result |
|---|---|
| `npx vitest run` | 304 files, **16 436 passed**, 0 failed |
| `make test-unit` | 8790 tests, coverage threshold met (all 259 files ≥ 90%) |
| `make test-integration` | 770 passed, 0 failed |
| `bun x playwright test --project=chromium test/e2e/playwright/specs/idevices/` | **224 passed**, 1 pre-existing skip |
| `make fix` / `make lint` / `make architecture-check` | clean |

Patch coverage on the changed application files: `ideviceNode.js` and `idevicesEngine.js` 100%, `editionLifecycle.js` 97.2%, `common_edition.js` 97.3%. The lines left uncovered in `editionLifecycle.js` are all pre-existing argument guards; every line this branch adds is covered.

## Review notes

An independent red-team pass raised 7 candidate findings; 6 were rejected on verification. Two of them are worth calling out because they changed the code:

- **Confirmed:** emptying the form left the iDevice visually blank if the export render failed. Fixed by unbinding with `jQuery.cleanData` instead of `.empty()` (step 4 above), which also removes a needless repaint gap on every save.
- **Caught during review:** `destroyEditionInstance()` originally cleared the global even on nodes that owned no edition, so a collaborative sync rendering a remote iDevice could strand a locally open editor. That is what the ownership scoping described above exists for.

A second review pass raised four more items that changed the code:

- **The ownership guard asked the wrong question.** It tested "do I own nothing?" rather than "is the open edition mine?", which only held because one call site happened to clear the previous owner's reference first. Ownership is now derived from the lifecycle's `ownerNode`, and the per-node reference — the thing that could go stale — is gone.
- **A bound callback could retain the whole engine.** See the `file-attachment` case above.
- **Two teardown paths for jQuery.** `bindJQuery()` already built the exact unbind closure, but a second mechanism did a blanket namespace `.off()` per target. The closure now registers through `own()` like every other resource; the nine existing jQuery-teardown tests pass unchanged.
- **An earlier claim in this description was wrong.** It said the diff contained no reformatting. It did: an editor's format-on-save reached the iDevice tree through the Biome LSP and rewrote a `flipcards.js` object method as an arrow — precisely the rewrite that breaks the `this` this PR depends on — reindented `download-source-file.test.js` from the file's own 2-space style, and dropped hand-aligned spacing in `lomloe.test.js`. All three are reverted, `biome.json` now excludes the tree so a directory walk skips it, and AGENTS.md records what that exclusion does and does not cover. A whitespace-only-diff detector run over every changed file now reports zero such lines.

A third pass ([review](https://github.com/exelearning/exelearning/pull/2301#pullrequestreview-5195755298)) found five more, all in the migrated scripts rather than the lifecycle itself:

- **A promise wrapped around an owned `FileReader` never settled.** `abort()` fires no `error` event and a bound `loadend` no-ops, so closing the editor mid-read left `3dmol`'s `readFileAsText()`/`readFileAsArrayBuffer()` and `image-gallery`'s `readFile()` pending for the lifetime of the page, holding the file, the reader and the caller's continuation. The three of them now go through a new `EditionLifecycle.readFile(file, method)`, which rejects with an `AbortError` on teardown — the same way an aborted `fetch()` on the lifecycle's `signal` already did.
- **`trivial.js` did not restore `window.onYouTubeIframeAPIReady`** on teardown, unlike `guess`, `map` and `quick-questions`. It now puts the previous value back, and leaves the global alone if a later edition has already claimed it.
- **`sort.js` forced the shared workarea overlay hidden.** `#load-screen-node-content` is the page loading screen, not part of the form. `lockScreen()` now records the state it found the overlay in and teardown hands that back, so the edition can only undo its own change.
- **`interactive-video.js` disposed the Bootstrap modal without hiding it.** Bootstrap's `dispose()` drops the instance and leaves the scroll lock (`modal-open` on `<body>`) and the backdrop in place, so a page switch with the editor open left the page covered with nothing able to close it. Teardown hides first, then sweeps what a mid-transition `hide()` could not clear — but only once no other modal is on screen, since both are shared.
- **`ownMedia()` stacked one disposer per `playSound()` call.** The preview clip is rebuilt on every click, so an editing session accumulated one live `Audio` per click. `ownMedia()` takes an optional slot name; the 16 `playSound()` implementations own their clip in `previewAudio`, which releases the previous one.

Adds [ADR-2293-01](doc/architecture/adr/ADR-2293-01-own-idevice-edition-resources-with-an-explicit-lifecycle.md) and documents the pattern in the iDevice skill.

## Known limitations

- **TinyMCE editors whose `init()` is still in flight.** Teardown enumerates `tinymce.editors`, so an editor that has not been registered there yet is not removed. TinyMCE 5 documents no cancellation API for a pending `init()` — its `editor.removed` flag is the undocumented mechanism — so this is a gap in the platform, not one the lifecycle can close. In practice `init()` registers early enough that the window is very small.
- **Form-local jQuery handlers are not individually migrated**, by design (see the audit above). They are released by the form teardown, and a test proves a handler bound directly by an edition stops firing afterwards — but that guarantee comes from one central step rather than from per-handler ownership.
- **Two frontend tests are flaky in full Vitest runs**, both pre-existing and unrelated. `focusedEditMode.test.js > "removes all focus-mode state"` reproduces on clean `main` with this branch stashed. `YjsProjectBridge.test.js > "stores export debug timeline when enabled"` failed once in five full runs and passes in isolation (454/454); this branch touches no Yjs, ELPX or export file. If CI shows either red, it is not this change.

## Behaviour worth checking by hand

Teardown now does things that used to be deferred or skipped, so these are the places a manual pass should look at:

- YouTube players are destroyed when the editor closes, instead of being left running.
- Media elements lose their `src` on close; reopening rebuilds them.
- TinyMCE editors are removed on close rather than on the next open.
- `$exeDevice` is no longer cleared by a node that does not own the open edition — the case collaborative sync hits.
- The workarea overlay and the Bootstrap scroll lock survive a page switch taken while an upload or the interactive-video editor is open.

## Relationship to #2293 and #2278

This is the central fix #2293 asked for, replacing the per-call-site approach of #2278. The #2278 guards are intentionally left in place — they are now redundant on the paths the lifecycle covers, but harmless, and removing them is not needed for this change.


ignaciogros (commented):
Thank you for identifying this and proposing a solution.

Although this is a bug fix and it would be good to merge it into `main` as soon as possible, I suggest waiting until #2322 has been merged. That improvement is important, and there are users waiting for our confirmation so they can test the solution before the v4.0.4 is released.

Please confirm if you’re OK with proceeding this way.

If you think this is urgent, we should let @mnarvaezm know so he can take it into account. That PR is expected to be submitted for review tomorrow.

erseco:
Yes, I agree with this approach, I will wait to @mnarvaezm PR and after fixing the possible conflicts I will request a review of both you :)

ignaciogros (commented):
I think the proposal is correct and that the changes can be merged, but **I suggest doing it after releasing v4.0.4**. 

The reason is risk-related: the PR changes the behavior of 44 iDevice editors and the shared helpers in `common_edition.js`, and these changes cannot be manually validated in the few hours remaining before the release.

The benefit for v4.0.4 would be small, as the four scripts that were generating the Sentry errors have already been fixed in main by #2278.

Please confirm if you think it would be good to wait.

Thank you.

ignaciogros (approved):
It's been decided to merge it right after v4.0.4 is released.

A few small things worth a look before that (none blocking):

- `trivial.js` does not restore `window.onYouTubeIframeAPIReady` on teardown, unlike guess, map and quick-questions.
- Promises wrapped with `lifecycle.bind` (3dmol `readFileAsText`/`readFileAsArrayBuffer`, image-gallery reader) never settle if the editor closes mid-read.
- `sort.js` registers a disposer that hides `#load-screen-node-content`, which is the shared workarea overlay; a page switch with an upload in flight could hide the page loading screen.
- `interactive-video.js` calls `dispose()` on the Bootstrap modal; if teardown happens with the modal open (page switch), `modal-open` stays on `<body>`.
- `ownMedia` adds one disposer per `playSound` call within an edition. Negligible, but easy to guard.
- Behaviour changes to keep in mind for manual testing: YouTube players are destroyed on close, media elements lose their `src`, TinyMCE editors are removed on close instead of on the next open, and `$exeDevice` is no longer cleared by a node that does not own the open edition.

**Could you please take a look at those details in case you want to update the PR content?**

Thank you.

erseco:
@ignaciogros All five are real; all five are fixed in 089cdefe9, each with tests. Details and what I found while verifying them:

**1. `trivial.js` did not restore `window.onYouTubeIframeAPIReady`.** Confirmed — it also assigned the bare global rather than `window.*`. It now captures the previous value and restores it on teardown, but only if the global is still the callback this edition installed, so a later edition that has already claimed it is not undone. That guard is `map.js`'s; the restore is `guess`/`quick-questions`'. Two tests: the restore, and the "leave a later callback alone" case.

**2. Promises wrapped with `lifecycle.bind` never settle.** Confirmed, and worse than it looks: `abort()` fires `abort` then `loadend`, never `error`, so the `onerror` branch was unreachable on teardown — and the bound `onload` no-ops. The promise stayed pending for the lifetime of the page, holding the file, the reader and the caller's continuation, and `loadModelFromFile()`/`processFile()` never returned.

Rather than patch the three sites, this is now `EditionLifecycle.readFile(file, method)`, which rejects with an `AbortError` on teardown — the same failure an aborted `fetch()` on `lifecycle.signal` already produced, so callers that handle one handle the other. 3dmol's two helpers and image-gallery's `readFile()` are three lines each now. All existing consumers already had a `catch`; `image-gallery.processFile()` gained an `isActive()` check so an interrupted read is not logged as an error.

The old 3dmol test asserting the promise does *not* settle was encoding the defect, so it is replaced by one asserting the rejection.

**3. `sort.js` hides `#load-screen-node-content`.** Confirmed as a design problem, though I could not make it bite today: in the page-switch path `cleanNodeContent()` — and with it `destroyEditionIdevices()` — runs *before* `showNodeContainerLoadScreen()`, so the disposer fires first. That ordering is exactly the kind of implicit coupling this PR exists to remove, and the `screenLocked` guard only proved the edition had locked the overlay, not that it was safe to force it hidden.

Fixed at the root: `lockScreen()` now records the overlay's `className` and inline style, and `hideLoadScreen()` restores that snapshot. The edition can only undo its own change — if the overlay was up for another reason, it stays up. Three tests, including the first-lock-wins case.

**4. `interactive-video.js` calls `dispose()` with the modal open.** Confirmed, and the backdrop survives too, not just `modal-open`: disposers run LIFO, so `dispose()` ran before the markup was removed, and neither touches `<body>` or `.modal-backdrop`. A page switch left the page scroll-locked behind an overlay with no modal able to close it.

Teardown now goes through `editor.destroyModal()`: `hide()` first so Bootstrap unwinds its own state, then `dispose()`, then the markup, then a sweep of the scroll lock and backdrops — but only when `document.querySelector('.modal.show')` finds nothing, since both are shared with every other modal. Three tests, including the mid-transition `hide()` no-op and the "another modal is open" case.

**5. `ownMedia` adds one disposer per `playSound` call.** Confirmed, and the disposer was the smaller half: each click also left a live `Audio` with its stream loaded until the editor closed.

`ownMedia(media, slot)` takes an optional single-occupancy slot; owning a new element under a slot releases what the slot held. The 16 `playSound()` implementations use `previewAudio`. Re-owning the same element is a no-op, so it cannot stop the clip it was just asked to keep. This also aligns them with `common_edition.js`'s shared `playSound`, which already stopped the previous clip explicitly.

One thing left deliberately: each call still registers the `canplaythrough` listener's remover, which retains the old element's JS object. The heavy part — playback and the buffered stream — is released by the slot, and that residue is a closure per click, so I left the ~16 sites alone rather than adding a second mechanism. Say the word if you would rather have it.

**6. Behaviour changes for manual testing.** Added as a "Behaviour worth checking by hand" section in the PR description, with your five points plus the overlay/scroll-lock one from items 3 and 4.

**Verification**

| Command | Result |
|---|---|
| `npx vitest run` | 304 files, 16 436 passed, 0 failed |
| `make test-unit` | 8790 passed, all 259 files at or above the 90% threshold |
| `make test-integration` | 770 passed, 0 failed |
| `bun x playwright test --project=chromium test/e2e/playwright/specs/idevices/` | 224 passed, 1 pre-existing skip |
| `make fix` / `make lint` / `make architecture-check` | clean |

`make test-e2e` in full reports 550 passed and 15 failed. None are this change: 11 of them (`theme-persistence`, `theme-import`, `theme-import-basic`, `theme-import-collaborative`, `theme-yjs-cleanup`) reproduce unchanged on the parent commit, and the other 4 (`save-legacy-elp` x1, `scorm12-sco-runtime` x3) pass in isolation on this branch — flaky under full-suite load. The theme ones look worth a separate issue if they are not already known.

The E2E lifecycle spec also asserts that a page switch leaves neither the workarea overlay visible nor a modal scroll lock or backdrop behind, and `ADR-2293-01` and the iDevice skill record `readFile()` and the slot.

No whitespace-only lines in the diff, and `biome.json` still excludes the iDevice tree.

PR #2302, opened 2026-08-19
refactor(xapi): emit statements without a package verdict, and fix the emitter's packaging

erseco:
## Summary

The xAPI emitter shipped in every export is an **analytics/LRS feed**, not a grading channel. This makes the code say that: the package-level verdict goes away entirely, and four real packaging and hygiene bugs are fixed.

## What the emitter emits now

Per page: `initialized`, one `answered` per gradable iDevice, `terminated`. **No package-level statement at all.** A consumer that needs a package result derives it from the `answered` statements — which is what `doc/elpx-format/tracking-emission.md` already tells consumers to do.

## Why the verdict goes

Each page of a multipage export is a separate JavaScript context, so the emitter's per-page state only ever holds the current page's scores. It still computed a package-level `completed` + `passed`/`failed` from it, against the package's Activity IRI. **One attempt could emit `passed` raw 100 from page 1 and `failed` raw 40 from page 2, for the same activity** — a self-contradicting record for any consumer.

Suppressing it only on multipage was the earlier plan. But the single-page verdict existed to serve the Moodle plugin's grading, and [the plugin is retiring its xAPI ingestion](#) — measured: per-iDevice grading is byte-identical between the two channels, while the overall is a true weighted mean on SCORM (25/75 → 25) against an unweighted 50 on xAPI. With no grading consumer, a verdict the emitter can only compute correctly for single-page packages is machinery without a purpose.

Removing it makes `_state`, `_packageScore()`, `registerEvaluable()`, `_isMultipage()`, `pageCount` and the `registerEvaluable` forward in `common.js` all dead, and they go with it — about 60 % of the JS side of this branch.

## The four bugs, which are worth landing regardless of any of that

1. **Four export formats loaded the emitter with no identity.** On `main` only `Html5Exporter` and `PageExporter` injected a `window.exeXapi` config, so SCORM 1.2, SCORM 2004, IMS and EPUB3 packages ran the emitter with an empty `odeId`: the activity IRI fell back to the per-page document URL and no statement carried `package-id`. `BaseExporter.buildXapiConfig()` is now the single source and all six formats call it. **Broken for a bare LRS too, not just for Moodle.**
2. **The print preview requested a 404.** The loader `<script>` was unconditional while `PrintPreviewExporter` copies no base libraries. The tag is now gated on the config being present.
3. **Out-of-range scores.** The `answered` statement advertises `min 0 / max 10`; the raw score is now clamped onto that scale. An out-of-range raw — or a `scaled` above 1 — is exactly the shape a strict LRS must reject.
4. **`keepalive` on the LRS POST, and no `unload` listener.** Without `keepalive` the `terminated` statement sent at `pagehide` is cancelled during unload. The `unload` listener was a duplicate under the once-guard and disables the back/forward cache — the same rule #2209 applies.

`pageId` / `pageTitle` stay in the config: only the exporter knows which page a document renders, so a statement cannot carry page identity otherwise, and it is genuinely useful to an analytics consumer.

## Documentation

`ADR-2302-01` is rewritten in place — it is `status: Proposed`, and `doc/architecture/adr/README.md` makes only *Accepted* records append-only — and renamed to match the decision it now records. It also **contained a factual error**: it justified the reverts with "the consumer grades through its SCORM-compatible shim (site setting `xapiprimaryenabled=0`)" while the plugin defaults that setting to `1`. That claim is gone.

`doc/elpx-format/tracking-emission.md` is updated to the reduced contract, and `public/CHANGELOG.md` notes that exported packages no longer emit a package-level verdict.

## Tests

Tests for the removed behaviour were **deleted, not rewritten into assertions that a feature is absent** — except one that pins the new contract: two answers plus a `pagehide` produce two `answered` statements and zero `completed`/`passed`/`failed`.

```
npx vitest run public/app/common          →  28 files, 1793 tests passed
bun test src/shared/export test/integration →  86 files, 2774 tests passed
make lint                                  →  exit 0
```

## Compatibility

This affects **new exports**. Packages already published keep their bundled emitter and keep emitting a package verdict; nothing consumes it, and the Moodle plugin ignores it either way.

---


erseco:
Closing this PR without merging.

Since this work started, the tracking architecture in `moodle-mod_exelearning` has changed substantially. `exelearning/moodle-mod_exelearning#124` retired the xAPI grading channel entirely and made SCORM 1.2 the single tracking/grading path.

That removes the concrete consumer that originally motivated this work. Keeping the xAPI emitter as an always-on analytics/LRS feed would mean continuing to maintain its exporter integration, statement contract, transports and tests without a current product requirement or known consumer.

Rather than fixing and extending that unused path here, the cleaner direction is to remove the xAPI emitter from eXeLearning for now in a separate cleanup PR.

This does not rule out xAPI in the future. If we get a concrete analytics/LRS requirement, we can revisit it with a defined consumer and contract instead of maintaining the current implementation speculatively.

The investigation in this PR was still useful: it identified several correctness and packaging issues in the existing emitter and helped clarify the distinction between analytics and grading. Since the emitter itself will be removed, those fixes do not need to land independently.

Superseded by the upcoming xAPI cleanup.

PR #2305, opened 2026-08-20
Delete instrucciones.md

ignaciogros:
This file should've never been in this repository.


eXeLearningProject (approved):

erseco:
Hi @ignaciogros this another file https://github.com/exelearning/exelearning/blob/main/%24PIDFILE should be removed too

PR #2307, opened 2026-08-20
fix(trueorfalse): close the container so it stops swallowing the next iDevice

erseco:
## Summary

`createInterfaceTrueOrFalse()` opens `.TOFP-MainContainer` and never closes it, so the markup it returns is one `<div>` deep. **Any iDevice placed after a trueorfalse in the same block disappears from every export.**

## What goes wrong

The exporter concatenates each iDevice's markup as a sibling inside the block, so the unclosed tag does not stay inside the trueorfalse. The HTML parser nests whatever follows into it, and `exe_export.js:435` then removes it outright:

```js
if (htmlIdevice) ideviceNode.innerHTML = htmlIdevice;
```

Silent content loss — no console error, no placeholder.

## Why it went unnoticed since the 3.0 import

`git blame` puts the whole region on `0b18ec3043` ("eXeLearning 3.0 first beta release with source code"). No pull request introduced it. Three things kept it invisible:

- **`</article>` contains the damage.** The block wrapper generates the implied end tags, so a trueorfalse that is alone in its block, or last in it, renders correctly.
- **No fixture has more than one component per block**, in either this repo or the Moodle plugin's — so CI never exercises it.
- **The editor is unaffected.** It assigns the markup per element, so the parser balances it there. Authors see a correct workarea and a broken export.

Multi-iDevice blocks are a supported feature (`assets/styles/layout/_idevice-focus.scss:73`).

## Where the closing tag goes

After `.TOFP-GameContainer`, not at the end of the template — so the score button and the after-text become siblings of `#tofPMainContainer` rather than its children. Three things agree on that placement:

- **The code already documents it.** The docblock above `updateLatexInView` states that `.TOFP-instructions` and `.TOFP-After` "are rendered as SIBLINGS of `#tofPMainContainer`". `.TOFP-instructions` already is one; `.TOFP-After` was not.
- **The `form` iDevice lays out the same three parts that way** — main container, then `Games-BottonContainer`, then the after-text, all siblings.
- **It changes rendering for the better.** `.TOFP-MainContainer p { margin: 0 !important; padding: 0 !important }` was flattening the author's after-text paragraphs.

The score button is unaffected either way: it is `width: auto` inside a `width: 100%` container with no padding or border, so losing the inherited `box-sizing: border-box` cannot change its layout.

`trueorfalse` is the only offender: every other iDevice type across the fixture corpus (dragdrop, guess, scrambled-list, form, casestudy, image-gallery, magnifier and 30+ more) is balanced.

## Scope

All exporters share `IdeviceRenderer`, so HTML5, SCORM 1.2/2004, EPUB3, IMS, the browser-side export, the CLI/server export and the Service-Worker preview were all affected equally.

Existing `.elpx` files still carry the bad markup. `ideviceNode.exportProcessIdeviceJson()` regenerates `htmlView` from `renderView()` on every workarea render and persists it in edition mode, so opening and saving an old project self-heals it — no migration needed.

## Tests

Three regression tests colocated in `trueorfalse.test.js`, all three confirmed to fail against pristine `main`:

| test | before | after |
|---|---|---|
| `returns balanced markup` | 13 vs 14 `<div>` | equal |
| `keeps the score button and the after-text outside the main container` | both nested inside | both siblings |
| `closes its main container, so a following iDevice stays a sibling` | parent was `idevice_node trueorfalse` | parent is `box-content` |

```
make lint           → exit 0
make test-frontend  → 266 files, 14509 tests passed
```

Cross-checked in Chromium, Firefox and WebKit: before the fix the following `.idevice_node` is a child in all three engines; after it, a sibling in all three.


ignaciogros (approved):
Thanks for detecting and fixing this. It’s strange that nobody noticed the broken presentation in any of the exports.

PR #2299, opened 2026-08-17
test(scorm): validate legacy and rewritten SCORM 1.2 runtimes with scorm-again

erseco:
## Independent SCORM 1.2 contract tests with scorm-again (#2210)

**Relationship with #2209.** This PR is a test layer on top of #2209. It adds no runtime code and no production code: five files under `public/app/common/scorm/contract/`, one Playwright smoke, one change record, and `scorm-again@3.3.0` (MIT) as a devDependency (never imported outside tests, never packaged into exports). The rewrite variant drives the in-tree `public/app/common/scorm/scorm12/` runtime — the layers the exporter assembles into `libs/SCOFunctions.js`; the legacy variant drives `SCOFunctions.js` + `SCORM_API_wrapper.js`, the pre-#2209 SCORM 1.2 runtime, which #2209 no longer packages for SCORM 1.2 (the SCORM 2004 exporter still ships it). Both run under scorm-again's `Scorm12API`: **this suite is a SCORM 1.2 oracle only** and says nothing about SCORM 2004 behaviour.

**Why a second oracle.** #2209's own tests run against a project-owned fake LMS written from the same contract as the runtime, so a defect shared by both sides would be invisible. scorm-again is an LMS-side implementation written independently of eXe; it validates writes with the SCORM 1.2 vocabularies, ranges and read-only/write-only rules and answers with the official error codes. Every normative assertion is still grounded in the SCORM 1.2 RTE Book, not in scorm-again's answers — the oracle's own quirks (lenient CMITimespan parsing, 64000-char suspend_data, 402 for skipped interaction indexes, …) are documented in the research note and asserted around, never mistaken for the specification.

[...]

PR #126, opened 2025-06-26
55 internal links from old elp files are not working

franmate:
Changes:
eXeLearning 2.x internal link conversion.
Modification of the path of internal links in preview and export.

PR #2310, opened 2026-08-21
test(scorm): generate real graded packages and record what an LMS receives

erseco:
Stacked on #2299 (its base branch). Because the base is not `main`, `ci.yml` does not run for this PR yet; the E2E workflow was dispatched by hand on the branch and is green (build + 12 shards + merge-report), and the unit/integration suites it adds run locally. It will be retargeted to `main` after #2299 merges (`git rebase --onto`), which runs the normal workflows.

## What this is

The SCORM 1.2 contract suite in #2299 certifies the runtime against a data model. It does not
answer the question a teacher actually asks: **does a package built from real iDevices,
answered by a real learner, end up with the right grade?** Nothing in the repo did, and every
SCORM grading defect found while auditing the export was found by hand.

This adds the two halves that make that question testable, and the recorded evidence they
produce.

## A package generator

`buildGradingStructure()` turns a declaration — pages, per-iDevice weights, question counts —
into a real exportable project and feeds it to any exporter:

```ts
buildGradingStructure({ pages: [
  { name: 'P1', idevices: [{ id: 'ide-a', type: 'trueorfalse', weighted: 25, questions: 4 }] },
  { name: 'P2', idevices: [{ id: 'ide-b', type: 'dragdrop',    weighted: 75, questions: 4 }] },
]})
```

Four gradable types are supported (`trueorfalse`, `dragdrop`, `scrambled-list`, `form`), each
with `jsonProperties` and container markup **cloned from a real ELPX** rather than
hand-written, so a fixture cannot drift from what the editor actually saves. Each type
documents how its score reaches `cmi.suspend_data`, which settings make it deterministic, and
which controls a test must drive.

Two properties are asserted because they are easy to get wrong: every gradable iDevice gets
its **own block** (a shared block is a different scenario — see #2307), and `sameBlock` opts
into that scenario deliberately.

`trueorfalse` is authored with `isTest: true, time: 0` by default — the only pair in which
that type can score at all (with `isTest: false` both the start and the check button are
hidden, `startGame()` is unreachable and `sendScoreNew()` drops every write; main's #2308
warns about exactly that pair). `isTest: false` remains available as the control case.

## A recorder

`moodle-serving-model.ts` reproduces how **mod_exelearning** serves a package: the export
served from a clean origin, the plugin's own SCORM loader injected exactly as
`scorm_injector.php` does, the two iDevice save guards patched as `idevice_patch.php` does,
`libs/` supplied from the plugin's own runtime pair, one iframe under a parent `window.API`
that records every call against a real `cmi` map. Page navigation is an `iframe.src`
assignment, so the parent and its `cmi` survive it — which is what makes the cross-page
`cmi.suspend_data` behaviour observable at all, and that is where the grading defects live.

**Which plugin it mirrors.** The injector is reproduced byte for byte from
`classes/local/scorm/scorm_injector.php` at two plugin revisions, selectable per run:

- `105` (default) — mod_exelearning#105 at `09b3f4ce7`: `exeScorm12.session.open({ ownsLifecycle: false })`,
  with `pipwerks.SCORM.init()` as a two-second fallback, and the dedupe of the package's own
  runtime tags. This is the runtime the plugin will ship, and the only injector under which
  the rewritten runtime (#2209) records anything — under the `main` one its client stays idle
  and refuses every write with 301.
- `main` — plugin main at `3b6a7cd45`: `pipwerks.SCORM.init()` only. Kept to reproduce the
  traffic of packages already deployed (the serving matrix's R0 cell).

Readiness under `105` is gated on the runtime's own `exeScorm12.client.isActive()`, not on the
pipwerks connection flag, which turns active in exactly the failure mode #105 exists to avoid.
`test/helpers/moodle-serving-model.spec.ts` carries an independent transcription of both PHP
snippets and diffs them against what the helper injects, so drift on either side fails under
`bun test`.

The runtime pair itself is not vendored: `MOD_EXELEARNING_SCORM_ASSETS` must point at the
plugin checkout's `assets/scorm`. Without it the recorder **skips** — visibly, with the reason,
in every Playwright project.

**The recorder fails when it records nothing.** After writing its trace it requires at least
one `LMSSetValue(cmi.suspend_data)`, a non-empty `finalCmi`, no `pageerror`, and
`Number(cmi.core.score.raw) === expected.overall`. Measured on the integrated tree with
#105's runtime: M2 30, M4 25, M3 50, M3-control 56.25, in Chromium and Firefox; with the
`main` injector the same run stops at "M2: no LMSSetValue(cmi.suspend_data) reached
window.API". On this branch alone (whose package writer predates #2209) M4 stops at
`"0" vs 25` — the legacy stale-slot defect — so the recorder is green only on the rebased
stack, which is the point.

**Scope, stated plainly:** this models *this plugin's* serving model. It does **not** exercise
`mod_scorm` or `mod_exescorm` from the recorder; those are what the live lanes below are for.

## The trace contract and the recorded traces

`test/fixtures/grading/TRACE-CONTRACT.md` is at **v2**. A trace is named
`<scenario>.<engine>.trace.json` and records the bytes that produced it: the sha256 of the
exported zip, the sha256 and `eXeLearning-SCORM12-Runtime` stamp of the served runtime pair,
the injector variant and the plugin commit it came from, the engine, and an oracle with a
`policyId` and an `ungraded` list.

**No consumer of these files exists today.** mod_exelearning#126 replays its own three
hand-authored traces from its `tests/fixtures/traces/` and reads no other directory; only
the contract is shared. Wiring a plugin-side replay of the core traces is future work.

**The seven committed traces are v1 snapshots and are not regenerated here.** They were
recorded from pre-#2209 packages (`recordedFrom.ref` `64df99f22` for m2/m3/m3-control/m4,
`642738cd0` for s2/s2-control/s6) with the plugin-main injector and the legacy runtime pair;
all carry `pageerror: $exe_i18n is not defined` (recorded without built i18n bundles);
`m3`, `m3-control` and `m4` pin the legacy stale-slot defect (`37.5` / `50` / `0` where the
oracle says `50` / `56.25` / `25`); and **`s2`, `s2-control`, `s6` have no producer in the
repository** (their `balanceHtml` repair is set by no shipped spec). They stay as evidence of
the traffic deployed packages still produce, documented as such in the contract.

## The oracle rule

Every scenario's expectation is hand-computed in the declaration, never derived from the code
under test. One recorded case shows why: a three-item scenario produced the arithmetically
correct overall **while silently dropping an iDevice**, because `(75+50+25)/3` happens to equal
`(75+50+25+50)/4`. Assertions are on the per-item map, not the total.

The overall is the weight-normalised mean over every gradable iDevice (`policyId:
weighted-mean-v1`). Every scenario in the recorded matrix answers every iDevice, so the rule
for an **unanswered** iDevice is not exercised — see "Still needs a product decision".

## And a live LMS, in two browser engines

The recorder above drives a **simulated** serving model. That answers "what does the runtime
write", not "what does an LMS store, and what mark does a teacher end up seeing". So this PR
also carries the live lane:

- `test/e2e/moodle/` — a Docker Moodle (`docker compose -p scormaudit up -d`) plus six CLI
  helpers that run inside its bootstrap, so an activity is created from a package without
  driving the admin UI, and the tracking rows and gradebook value can be read back as JSON;
- `playwright.moodle.config.ts` + `test/e2e/playwright/specs-moodle/` — **ten specs**
  (`probe`, `all-idevices-walk`, `all-idevices-play`, `all-idevices-controls`,
  `scorm-grading-matrix`, `package-oracles`, `resume-across-attempts`, `exelearning-live`,
  `exelearning-matrix`, `exelearning-serving-matrix`) that open a real package in the real
  player and record **both** sides: the ordered SCORM API traffic captured in the browser, and
  what Moodle stored for it.

What they cover: a plumbing probe (login, player URL, iframe, API instrumentation — it skips
with the command to run when its hand-created activity descriptor is missing); opening every
page of a project without interacting; playing every gradable iDevice with a fixed click
script; the declared scenario matrix across hosts and grading methods; resume across attempts;
and the mod_exelearning plugin itself, live and under the serving model. The
`exelearning-live` assertions target #105 (runtime stamp, registry present) and #126
(grading off ⇒ nothing recorded) behaviour, not plugin main.

**Two engines, on purpose.** The end-of-session path moved from `unload` to
`pagehide`/`visibilitychange` in #2209, and those do not fire on the same schedule in Gecko as
in Chromium, so a grading result that only holds in one engine is not a grading result.
`--project=firefox` runs the same lane in Gecko; every artefact is tagged with its engine, so
two runs are comparable instead of overwriting each other.

**Not part of `make test-e2e`.** `playwright.config.ts` only picks up
`test/e2e/playwright/specs`; the live lanes live in `specs-moodle`, need an external LMS by
definition, and declare no `webServer` because starting one from the test run would hide which
instance the results came from. Three lanes skip with a message when the scenario catalogue is
not staged; the catalogue, the package manifests and the `allidevices-*` packages they consume
are inputs that are **not generated by anything in this repository** yet.

## What runs in CI, and what does not

- `test/integration/grading-fixtures.spec.ts` — `make test-integration`, 31 tests.
- `test/helpers/moodle-serving-model.spec.ts` — `make test-unit` (`bun test ./test/helpers`),
  18 tests: injector pins, HTML rewrite, provenance digests, asset resolution.
- `test/e2e/playwright/specs/grading-matrix-recorder.spec.ts` — **collected by the default
  Playwright config in chromium, firefox and static** (it sits in `specs/`), 12 entries.
  It needs `MOD_EXELEARNING_SCORM_ASSETS`; CI does not set it, so it **skips** there.
  A green CI run is not evidence that the recorder passed — run it locally against the plugin
  checkout, or set the variable in a dedicated job.
- `specs-moodle/*` — never collected by CI; manual, with a Moodle stack.

## Tests

```
bun test test/integration/grading-fixtures.spec.ts          →  31 pass, exit 0
bun test test/helpers/moodle-serving-model.spec.ts           →  18 pass, exit 0
bun x biome check <harness files>                            →  0 warnings
bun x playwright test --list …/grading-matrix-recorder.spec.ts
                                                             →  12 tests (chromium, firefox, static)
bun x playwright test --project={chromium,firefox,static} …/grading-matrix-recorder.spec.ts
   (MOD_EXELEARNING_SCORM_ASSETS unset)                      →  4 skipped each, exit 0
MOD_EXELEARNING_SCORM_ASSETS=<mod_exelearning#105>/assets/scorm
bun x playwright test --project={chromium,firefox} …/grading-matrix-recorder.spec.ts
   on main + #2209 + #2299 + #2310                           →  4 passed each (30 / 25 / 50 / 56.25)
   on this branch alone                                      →  M4 fails "0 vs 25" (pre-#2209 writer, expected)
```

Output paths default to `test/temp/` inside the repo (gitignored); `GRADING_FIXTURE_OUT_DIR`
and `GRADING_TRACE_DIR` override them, which is how the two halves run on different machines.

## Still needs a product decision

- **The unanswered-iDevice rule** (audit B10): core's `s6` oracle counts it as 0 (overall
  31.25), the plugin's `synthetic-stale-slot` fixture (#126) excludes it as ungraded (50), and
  `exelearning-matrix.spec.ts` asserts a null raw score. Decide once, give it a `policyId`, and
  assert it from both repositories.
- **Trace regeneration.** The committed v1 traces are legacy evidence; the v2 traces of the
  #2209/#105 stack are recorded only after that rule is decided, per engine, against a tree
  with built i18n bundles. Whether `s2`/`s2-control`/`s6` get a shipped producer or are dropped
  is part of the same decision.
- **A single scenario catalogue** shared by the integration spec, the recorder and the live
  lanes (the declarations are currently duplicated by hand), and the package-producer script
  for the live matrix.

## Shared scenario catalogue

The scenarios are now declared once, in `test/helpers/grading-scenarios.ts` — each with
its `ProjectSpec` for `buildGradingStructure()`, the learner script the drivers follow,
and the hand-authored oracle (per-item map, weights, overall, `ungraded`, and the
arithmetic in a note). The integration spec builds its package matrix from it, the
recorder takes each scenario's spec and `expected` from it, and the scenario runner's
`Scenario` is an alias of its `GradingScenario`, so the "verbatim" copies between the
integration spec and the recorder are gone (single source of truth, AGENTS.md). A
colocated `bun` spec checks the hand-authored numbers against each other with its own copy
of the weighted-mean arithmetic — never the runtime's `getFinalScore` — so a typo cannot
hide while the oracle rule holds.

`scripts/build-grading-catalogue.ts` is the producer: it exports every declared scenario
with the production exporters (`Scorm12Exporter` / `Html5Exporter` / `ElpxExporter`) and
writes the `scenarios/catalogue.json`, the per-scenario zips and the
`manifest-<producer>.json` (with `head`, `sha256`, SCORM runtime digests and page lists)
the live lanes read from `$AUDIT_ROOT` — so `scorm-grading-matrix`, `package-oracles`,
`exelearning-serving-matrix` and `exelearning-matrix` no longer depend on inputs "not
generated by anything in the repository". Its own `bun` test builds the catalogue into a
temp dir and asserts every declared scenario has a zip whose digest matches the manifest.
This settles the third bullet under "Still needs a product decision" above. The
`allidevices-*` packages are still out of scope — they come from a real project, not a
declared scenario — and the README says so. The serving matrix's own evidence writer is
still `traceVersion: 1`; a shared v2 writer is left as a TODO in the trace contract.

## Merge sequencing

Stacked on #2299 and reviewed as such; after #2299 merges, this branch is rebased onto `main` and the PR retargeted so that `ci.yml` (`build_and_test`) and the E2E workflow run on it as pull-request checks.


ignaciogros (approved):
Thank you, @erseco!

This PR will be merged into `main` after #2299. Once it is merged, the main changes related to SCORM functionality will have been integrated, and @mnarvaezm will make usability improvements in line with the standard, specifically regarding when and how grades are saved in the LMS.

PR #2311, opened 2026-08-21
fix(trueorfalse): keep legacy True/False texts in the package language

### LINKED ISSUE #2252: Bug:"True-false" iDevice texts remain in English when opening a legacy ELP
## Bug Description
When opening or importing a `.elp` file originally created with eXeLearning 2, the **"True-false"** iDevice fails to translate its default interface texts to the language configured in the resource properties, keeping them in English instead.

## Steps to Reproduce
1. Open a `.elp` package created with **eXeLearning 2** that contains at least one **True-false** iDevice.
2. Check the project/resource properties to confirm that a non-English language is set (e.g., Spanish).
3. Preview or review the content of the True-false iDevice.

## Expected Behavior
The default texts of the True-false iDevice should automatically update and translate to the language assigned in the resource properties.

## Current Behavior
The texts and interface elements of the iDevice remain in **English**, ignoring the language configuration of the resource.

### Screenshots

Image

### Sample File (.elp)

[verdaderofalso.elp.zip](https://github.com/user-attachments/files/30985701/verdaderofalso.elp.zip)

erseco:
Fixes #2252. Thanks @franmate for spotting it and for attaching a sample package.

## Problem

A True/False iDevice imported from an eXeLearning 2.x `.elp` kept its interface texts in English ("True", "False", "Suggestion") even when the package declared another language.

## Cause

eXe 2.x packages don't store those texts — the old exporter rendered them from its own catalogue — so the importer has to supply them. `TrueFalseHandler` always supplied the English set, and `LegacyXmlParser` never handed the handler context (which carries the package language) to `extractProperties()`.

## Fix

- `TrueFalseHandler` resolves the messages from the package language, and `LegacyXmlParser` now passes the context to `extractProperties()` — which also lets `DropdownHandler` and `CaseStudyHandler` use the package language for their feedback captions instead of always defaulting to Spanish.
- The translations come from `translations/messages.<lang>.xlf` — the very strings the editor resolves through `c_()` — projected into `trueFalseMessages.generated.ts` by `scripts/generate-truefalse-messages.ts`. Locales without a catalogue fall back to English, exactly like `c_()` does at edit time.

## Why generate the catalogue instead of reading it

The obvious approach — look the strings up in the XLF catalogue at import time — doesn't work here, and neither does `c_()`:

- **No run-time file access.** `src/shared/import/**` is bundled for the browser (`importers.bundle.js`), so it cannot read `translations/messages.es.xlf`.
- **The parse path is synchronous.** `LegacyXmlParser.parse()` runs the handlers inline, so it cannot `await` a catalogue fetch for an arbitrary language.
- **`c_()` would give the wrong language.** It is bound to the content language *currently loaded*, which during an import is still the previous project's. `loadContentTranslationsStrings(pp_lang)` only runs once `pp_lang` reaches Yjs — after the parser has already created the iDevice. That would be intermittently wrong, which is worse than a fixed English fallback.

So the XLF catalogue stays the single source of truth and the generated module is a build-time projection of it. It is committed so `bun test` and esbuild need no extra step, and `generate-truefalse-messages.spec.ts` rebuilds the table from `translations/` and fails if the committed module ever drifts — a changed translation can't silently diverge from what legacy imports write. Regenerate with `bun run generate:truefalse-messages`.

This also keeps the same shape already used for legacy iDevice titles (`IDEVICE_TITLE_TRANSLATIONS`) and the feedback button caption (`FEEDBACK_TRANSLATIONS`), except the values are no longer hand-copied.

## Note for translators

Generating the table surfaced a pre-existing defect in `translations/messages.de.xlf`: a run of entries around `Memory cards` / `True` / `False` has its targets shifted by one, so German gets `True` → "You're wrong. That's not the card." The generated table mirrors the catalogue faithfully (and the editor already renders the same string through `c_()`), so this PR does not paper over it — `translations/**` is managed by the translation process. Worth a separate fix there.

## Tests

Written first, all failing before the fix:

- `trueFalseMessages.spec.ts` and a `TrueFalseHandler` case in `handlers.spec.ts`
- `LegacyXmlParser.spec.ts` — parse-level check for an `es` and an `en` package
- `test/integration/legacy-truefalse-language.spec.ts` — the sample package attached to the issue
- `test/e2e/playwright/specs/idevices/trueorfalse-legacy-language.spec.ts` — the imported activity renders "Verdadero" / "Falso" / "Sugerencia"
- `scripts/generate-truefalse-messages.spec.ts` — the generator, plus the drift guard


erseco:
Review addressed in e908c63a1. Both blocking points were real:

**1. Generator depended on its own output.** Confirmed by reproducing it — `rm trueFalseMessages.generated.ts && bun run generate:truefalse-messages` failed on module resolution before the generator could write anything.

Split as suggested: `trueFalseMessages.definition.ts` is now an import-free leaf holding `TrueFalseMessages` and `TRUE_FALSE_DEFAULT_MESSAGES`, and the chain is one-way — definition → generator → generated → runtime. Deleting the generated file and regenerating now works. A test asserts the generator never imports the runtime module and the definition stays import-free; verified it fails when the old import is put back.

**2. `quote()` → `JSON.stringify()`.** The hazard is not theoretical. Test first, with a target carrying a newline:

```
msgSuggestion: 'first
second',
```

— a raw line break inside a single-quoted literal, i.e. a syntax error. `JSON.stringify` fixes it; Biome rewrites the quotes on `make fix`, which is fine since the drift test compares data. The stale "byte-for-byte" comment you spotted is gone too.

**3. Follow-ups — two opened, one I'd argue against.**

- #2312 — the German catalogue. Scanned the whole file: exactly one run of nine consecutive shifted targets, self-correcting at `Click on the card`. Of 3922 units only 7 have a target that is verbatim another English source, 6 of them inside that block.
- #2313 — the `trueorfalse.js` ↔ `TRUE_FALSE_DEFAULT_MESSAGES` gap. Verified it is unguarded and that both sides match today (35 keys, no mismatch), so a guard starts clean. Prototyped the extraction: it handles the multi-line `c_()` calls and the double-quoted string, 35/35.
- **Generic `generate-legacy-i18n.ts`: I did not open this, because the two candidate consumers do not hold up.** `IDEVICE_TITLE_TRANSLATIONS` cannot be generated at all — `Case Study`, `Reading Activity`, `True-False Question` are legacy eXe-2 iDevice titles that do not exist in the XLF catalogue (and its values are deliberately unaccented, e.g. `Caso practico`). `FEEDBACK_TRANSLATIONS` covers 15 languages, 7 of which (`fr`, `nl`, `pl`, `ru`, `zh`, `ja`, `ar`) have no XLF file, so generating it would *lose* languages. With no viable second consumer the abstraction would be speculative; happy to revisit if the catalogue grows to cover them.

Re-verified after the refactor: `make fix` clean, 8379 unit tests with all 242 files ≥90% coverage (generator 97.4%, the three modules 100%), 730 integration, 14422 frontend, and the targeted Playwright spec green against a rebuilt bundle.

franmate (approved):
Verified that the text correctly reflects the configured content language across both the **web interface** and the **CLI**. 

Everything works as expected

Ready to merge.

LGTM!

PR #2315, opened 2026-08-24
docs+perf(load-test): C10k benchmark for eXeLearning Cloud, fix bcrypt event-loop block and WebSocket connection leak

### LINKED ISSUE #2255: Measure and optimize eXeLearning Cloud for C10k-class concurrency
## Description

Evaluate, measure, and optimize eXeLearning Cloud to support a large number of concurrent users and long-lived WebSocket connections, using the [C10k problem](https://en.wikipedia.org/wiki/C10k_problem) as a reference.

The goal is **not simply to reach 10,000 HTTP requests or open TCP connections**, but to determine how many concurrent eXeLearning users can be supported under realistic workloads and identify the limiting components.

A typical eXeLearning Cloud usage pattern is favorable for this scenario:

- Usually one user is editing each project.
- Collaborative projects commonly have 2–4 simultaneous users.
- More than 10 simultaneous users editing the same project should be considered an uncommon/extreme scenario.

Because of this, the primary scalability scenario should be **many independent projects with a small number of users per project**, rather than thousands of users connected to the same Yjs room.

## Existing architecture

The current High Availability deployment is the appropriate baseline for this work:

- Nginx load balancer.
- Multiple eXeLearning/Bun instances.
- PostgreSQL shared between instances.
- Redis Pub/Sub for Yjs synchronization across instances.
- Shared project assets/storage.
- Stateless JWT authentication.
- WebSocket-aware reverse proxy configuration.

Documentation:

- [`doc/high-availability.md`](https://github.com/exelearning/exelearning/blob/main/doc/high-availability.md)
- [`doc/deploy/docker-compose.redis.yml`](https://github.com/exelearning/exelearning/blob/main/doc/deploy/docker-compose.redis.yml)
- [`doc/deploy/nginx-ha.conf`](https://github.com/exelearning/exelearning/blob/main/doc/deploy/nginx-ha.conf)

There are already several design decisions that are useful for high concurrency:

### Stateless application instances

Persistent application state is stored outside individual eXeLearning instances, allowing requests to be distributed between multiple instances.

### Redis-based Yjs synchronization

Redis Pub/Sub allows users connected to different application instances to collaborate on the same project.

This makes horizontal WebSocket scaling possible without requiring all collaborators to remain connected to the same instance.

### Lightweight WebSocket relay

The Yjs WebSocket server acts primarily as a relay and does not keep a full server-side `Y.Doc` for every project.

This is important because the memory cost of thousands of connected projects should be substantially lower than architectures where every document is fully materialized server-side.

### Native Bun WebSockets

eXeLearning uses Bun's native WebSocket implementation rather than a traditional thread/process-per-connection architecture.

This should be suitable for large numbers of mostly idle long-lived connections, but the actual capacity must be measured.

## Current limitation to review

The example HA Nginx configuration currently contains:

```nginx
worker_processes auto;

events {
    worker_connections 1024;
    use epoll;
    multi_accept on;
}
```

`worker_connections` also includes upstream connections. A proxied WebSocket normally requires a client-side connection and an upstream connection to the eXeLearning instance.

Therefore, the current example configuration may become a bottleneck well before 10,000 concurrent WebSockets.

For load testing, values such as the following should be evaluated:

```nginx
worker_processes auto;
worker_rlimit_nofile 200000;

events {
    worker_connections 32768;
    use epoll;
    multi_accept on;
}
```

The container/host file descriptor limits must also be configured appropriately, for example:

```yaml
ulimits:
    nofile:
        soft: 200000
        hard: 200000
```

The actual values should be determined from measurements rather than adopted blindly.

## Load-balancing consideration

The current HA WebSocket upstream uses `ip_hash`.

This may distort load tests executed from a single load-generator machine because most or all connections could be routed to the same eXeLearning instance.

For HA scalability tests, compare `ip_hash` with an algorithm such as:

```nginx
upstream exelearning_websocket {
    least_conn;

    server exelearning-1:8080;
    server exelearning-2:8080;
}
```

Redis already provides cross-instance Yjs synchronization, so WebSocket affinity should not be required for correctness.

This should be verified before changing the production configuration.

## Proposed test methodology

Use a dedicated load-testing tool such as [k6](https://grafana.com/docs/k6/latest/) with WebSocket support.

Tests should be executed from a separate machine so the load generator does not compete for CPU, memory, sockets, or network resources with eXeLearning.

### Phase 1: Single-instance connection capacity

Measure the baseline capacity of one eXeLearning/Bun instance without horizontal scaling.

Suggested progression:

```text
100
500
1,000
2,500
5,000
10,000
20,000
```

Connections should remain open for several minutes rather than only testing the connection handshake.

This provides a useful metric:

```text
maximum stable WebSocket connections per eXeLearning instance
```

### Phase 2: HA connection capacity

Repeat the test using:

```text
Nginx
  |
  +-- eXeLearning instance 1
  +-- eXeLearning instance 2
  +-- ...
  |
PostgreSQL
Redis
Shared storage
```

Compare the results against the single-instance baseline.

This should show whether adding application instances produces approximately linear improvements or whether another component becomes the bottleneck.

### Phase 3: Realistic eXeLearning workload

The primary test should model expected production behavior.

For example:

```text
10,000 concurrent users
10,000 projects
1 user/project
```

and:

```text
10,000 concurrent users
2,500–5,000 projects
2–4 users/project
```

A further collaborative scenario could use:

```text
10,000 concurrent users
1,000 projects
10 users/project
```

The last scenario should already be considered significantly more collaborative than normal usage.

### Phase 4: Per-project collaboration limits

Run a separate test to determine how collaboration fan-out affects performance:

```text
1 project × 10 users
1 project × 50 users
1 project × 100 users
1 project × 500 users
```

This is a different scalability problem from C10k.

Every Yjs update may need to be relayed to all other clients in the room, so the cost of a highly collaborative project grows with the number of participants.

The result should establish a separate recommendation such as:

```text
Recommended maximum simultaneous collaborators per project: N
```

rather than mixing this limit with the global concurrent-user capacity.

## Workload types

At least three workloads should be tested.

### Idle WebSockets

Clients connect to a project and remain connected.

This measures the basic C10k-style connection capacity.

### Normal editing

Clients periodically send realistic Yjs updates representing normal editing activity.

This measures CPU, network traffic, Redis traffic, and message fan-out under realistic conditions.

### HTTP/API activity

Connected users should also perform representative application operations such as:

- Opening projects.
- Saving projects.
- Loading project metadata.
- Uploading/downloading small assets where appropriate.

A system with 10,000 established WebSockets but unusable API latency should not be considered capable of supporting 10,000 active users.

## Metrics

Collect metrics from all layers during every test.

### Load generator

- Attempted WebSocket connections.
- Successful WebSocket upgrades.
- Failed connections.
- Unexpected disconnects.
- Connection establishment latency.
- Messages sent/received.
- HTTP request latency.
- HTTP error rate.

### Nginx

- Active connections.
- File descriptor usage.
- Connections per worker.
- Requests per second.
- 4xx/5xx responses.
- Upstream failures.
- CPU and memory.

### eXeLearning/Bun

- Active WebSocket connections.
- Connections per instance.
- Active Yjs rooms.
- CPU.
- RSS/heap memory.
- Memory per additional connection.
- Event-loop responsiveness.
- Unexpected WebSocket closures.
- Message throughput.

### PostgreSQL

- Active connections.
- Query latency.
- Queries per second.
- CPU.
- I/O.
- Lock contention.

### Redis

- CPU.
- Memory.
- Pub/Sub messages per second.
- Number of channels/subscriptions.
- Network throughput.

### Host

- CPU.
- RAM.
- Network throughput.
- Open file descriptors.
- Socket counts.
- TCP errors/retransmissions.

## Suggested success criteria

Initial target:

```text
10,000 simultaneous users
```

with a realistic project distribution, rather than 10,000 users in one project.

A test should only be considered successful if the target concurrency can be maintained for a sustained period.

Example initial criteria:

- 10,000 concurrent WebSocket connections maintained for at least 10 minutes.
- No unexpected systematic WebSocket disconnections.
- WebSocket connection failure rate below 1%.
- HTTP/API error rate below 1%.
- No progressive/unbounded memory growth.
- No file-descriptor exhaustion.
- No sustained CPU saturation.
- Acceptable API latency under load.
- Stable Redis and PostgreSQL behavior.

Exact latency and resource thresholds should be defined after obtaining an initial baseline.

## Deliverables

- [ ] Create a reproducible load-testing environment.
- [ ] Add k6 or equivalent WebSocket load tests.
- [ ] Measure a single eXeLearning instance.
- [ ] Measure the HA deployment.
- [ ] Increase Nginx/file-descriptor limits required for high concurrency.
- [ ] Verify whether `ip_hash` is appropriate for HA WebSocket balancing.
- [ ] Test mostly independent projects.
- [ ] Test 2–4 collaborators per project.
- [ ] Test 10 collaborators per project.
- [ ] Measure high-collaboration fan-out separately.
- [ ] Identify the first bottleneck at each concurrency level.
- [ ] Determine stable WebSocket connections per application instance.
- [ ] Determine stable concurrent users for the HA deployment.
- [ ] Determine a recommended maximum number of simultaneous collaborators per project.
- [ ] Document the benchmark methodology.
- [ ] Document the hardware/resources used for each benchmark.
- [ ] Document the resulting capacity and scaling recommendations.

## Expected outcome

At the end of this issue we should be able to document measurable statements such as:

```text
A single eXeLearning instance supports approximately X concurrent
WebSocket users under workload Y using resources Z.

An HA deployment with N application instances supports approximately
X concurrent users under the same workload.

The recommended maximum number of simultaneous collaborators in a
single project is Y.
```

This would provide a more meaningful capacity specification than simply claiming that eXeLearning "supports C10k", while still using the [C10k problem](https://en.wikipedia.org/wiki/C10k_problem) as the concurrency target and architectural reference.

erseco:
## Summary

Load-tested eXeLearning Cloud against the C10k benchmark methodology described in #2255, to measure (not assume) how many concurrent users and long-lived WebSocket connections one instance — and a 2-instance HA deployment — can sustain.

Along the way this found and fixed two real server bugs:

- **Password verification was blocking Bun's event loop** under concurrent logins (`bcryptjs.compare()` in pure JS instead of native `Bun.password.verify()`). A 500-login burst went from 36% success / up to 138s stalls on an unrelated `/healthcheck` endpoint, to 100% success / 117ms.
- **WebSocket connections leaked forever from the room manager on every close** (not just crashes), because Elysia wraps each socket in a fresh object per event and the room tracked connections by object identity. Fixed by keying by the stable `clientId` instead.

Both fixes ship with regression tests that fail without the fix and pass with it.

## What was measured

- Single instance: **10,000 concurrent idle WebSocket connections** held for 10 minutes at 99.92% success, 1.15% CPU, 257 MiB RAM.
- Realistic editing workload (1/2/4/10 users per project) and **500 simultaneous collaborators on one project** (HA, 2 instances): both essentially clean (100% and 99% success).
- HA tuning: confirmed `ip_hash` is unsuitable for the WebSocket upstream (routes ~100% of a single-IP load generator's connections to one instance) and that the default `worker_connections 1024` is a real bottleneck (58.4% login success at 1000 WS connections vs. 100% after tuning). `doc/deploy/nginx-ha.conf` and `docker-compose.redis.yml` updated with the validated values.
- Real-browser validation against the public domain while background load ran.

## Reports and reproducibility

- Technical report (English): [`doc/development/c10k-benchmark.md`](https://github.com/exelearning/exelearning/blob/2255-c10k-load-testing/doc/development/c10k-benchmark.md)
- Informe técnico (español): [`doc/development/c10k-benchmark.es.md`](https://github.com/exelearning/exelearning/blob/2255-c10k-load-testing/doc/development/c10k-benchmark.es.md)
- Reusable k6 scenarios and orchestration scripts: `test/load/` (see its README for exact reproduction commands and RUN IDs).
- `doc/high-availability.md` updated to link to both reports and the tooling.

Both reports are also attached below for convenience.

**Attachments:**
[c10k-benchmark.md](https://github.com/user-attachments/files/31371685/c10k-benchmark.md)
[c10k-benchmark.es.md](https://github.com/user-attachments/files/31371601/c10k-benchmark.es.md)

## Architecture decision records

Two ADRs record the durable decisions this benchmark produced:

- [`ADR-2255-01`](https://github.com/exelearning/exelearning/blob/2255-c10k-load-testing/doc/architecture/adr/ADR-2255-01-balance-websocket-upstream-with-least-conn.md) — balance the WebSocket upstream with `least_conn`, not `ip_hash`.
- [`ADR-2255-02`](https://github.com/exelearning/exelearning/blob/2255-c10k-load-testing/doc/architecture/adr/ADR-2255-02-verify-passwords-with-bun-native-bcrypt.md) — verify passwords with Bun's native bcrypt, never with pure-JS `bcryptjs`.

Both reports are also registered in the `mkdocs.yml` nav, so they ship to the docs site.

## Test plan

- [x] `make fix`
- [x] `make test-unit`
- [x] New regression tests for both bug fixes (fail without the fix, pass with it)
- [x] Full load-test progression documented in the reports above (not part of CI — a manual benchmark exercise)


ignaciogros (approved):
Thanks, @erseco! 

The solution looks right to me and the results speak for themselves.

Two small things that might be good to do before merging:

1. **An ADR for #2255.** Per AGENTS.md §7.11, `ip_hash` → `least_conn` on the WebSocket upstream and native bcrypt in the auth path are durable decisions. The report has the evidence; the ADR should record the decision.
2. **`mkdocs.yml` nav entries** for `doc/development/c10k-benchmark.md` and `.es.md` — the nav is an explicit list, so as it stands both ship to the site unreachable.

erseco:
Thanks for the review, @ignaciogros — both done in 725e7485.

1. **ADRs.** Two rather than one, since they are unrelated decisions in different layers and the ADR guide asks the slug to name the decision, not the topic:
   - [`ADR-2255-01`](https://github.com/exelearning/exelearning/blob/2255-c10k-load-testing/doc/architecture/adr/ADR-2255-01-balance-websocket-upstream-with-least-conn.md) — `least_conn` instead of `ip_hash` for the WebSocket upstream (the ~100%-to-one-instance skew, why Redis makes affinity unnecessary, rejected alternatives; the `worker_connections` tuning is recorded there too, as tuning shipped in the same file).
   - [`ADR-2255-02`](https://github.com/exelearning/exelearning/blob/2255-c10k-load-testing/doc/architecture/adr/ADR-2255-02-verify-passwords-with-bun-native-bcrypt.md) — native `Bun.password.verify` instead of pure-JS `bcryptjs` (the 138s `/healthcheck` stall, no hash migration needed). Writing it surfaced four `bcryptjs.hash()` call sites still on request paths — `auth.ts:674`, `:1011`, `:1129` and `admin.ts:1069` — much colder paths, so they are recorded as follow-up work rather than changed here.

   Both are `status: Proposed`; `make architecture-check` passes.

2. **Nav.** Both reports added under Development in `mkdocs.yml`.

Adding them to the nav made a few things user-facing that were fine while the pages were unreachable, so I also fixed: a stale "Draft" banner and a "once that phase runs" placeholder (the HA phase has since run), a broken `#fix-verification-live-check` anchor, and the repo-relative `test/load/` links, which work on GitHub but 404 on the site — those are now absolute. `mkdocs build` is clean for both pages.

One thing I did **not** change: a handful of intra-page section links use GitHub's anchor slugs (double hyphens around em dashes, accents preserved in the Spanish report), which MkDocs slugifies differently, so those jumps work on GitHub but not on the site. Making them agree means either editing the headings or switching the site-wide `toc` slugify — a separate decision, happy to open an issue if you want it fixed.

ignaciogros:
Thanks, @erseco!

I don't think it's necessary to fix the GitHub anchor slug issue now. In my opinion, we should prioritize making sure that the links work correctly on the site, because that's the link we're sharing.

PR #2322, opened 2026-08-26
 SCORM: the SCO's status and score are decided by the learner's interaction

### LINKED ISSUE #1831: SCORM 1.2 export uses unloadPage() on onunload, causing Chrome Permissions Policy violation and lost scores in Moodle

mnarvaezm:
Closes #1831.

Besides fixing the set of bugs and adding some improvements detailed below, the main goal of this PR is to **define coherently when the status of a SCORM page is written, and with what value**. Until now that decision was spread across the page lifecycle, the pipwerks wrapper and each iDevice on its own, and none of the three agreed with the others. This PR reduces it to a single rule: **the status of a SCO holding graded iDevices is decided by the learner's interaction with them.**

## Status behaviour, according to what the page holds

What gets written to the LMS, and **when**, depending on the iDevices the SCO carries.

### Page with no SCORM iDevices

A content-only page — text, images, video — or one whose game iDevices have the **SCORM option turned off**, is marked **`completed` on entry**, as soon as the page finishes loading.

### Page with one SCORM iDevice

| Moment | Status | Score |
|---|---|---|
| Opens it and does not touch it | nothing is written while on the page; on leaving, `incomplete` | — |
| Starts the activity | `failed` | 0 |
| Answers, currently on 20 | `failed` | 20 |
| Answers, reaches 50 | **`passed`** | 50 |
| Answers, reaches 80 | `passed` | 80 |
| Restarts the activity | `failed` | 0 |
| Leaves the page | whatever it had, **untouched** | — |

The status follows the **score**, not whether the activity is finished. `incomplete` means one thing only: the learner has not started the activity.

### Page with several SCORM iDevices

It behaves just like with one:

- **While none is touched**: nothing is written while on the page and, on leaving, it stays `incomplete`.
- **As soon as any one is used**: the page goes to `failed`.
- **When the score reaches 50 or more**: it goes to `passed`.

Touching **one** is enough for the status to change; the rest of the iDevices have no say in that decision. Where they do have a say is in the score: it is the weighted average of **all** of them on the page, and those not yet attempted enter it with a 0 and their full weight. That is why finishing one of three perfectly leaves the page at 33.33 and `failed`: work has to accumulate in the others to cross 50.

## The score and the index icon update on interaction, without leaving the page

**Every time the learner answers, the score is written to the LMS and `LMSCommit` runs at that very moment**, plus a deferred retry. Moodle refreshes its side index on every commit, so teacher and learner watch the score move **while playing**.

## Leaving the page does not change the status

Neither `pagehide`, nor `visibilitychange`, nor `freeze`, nor the load, nor `LMSInitialize` write status. The only lifecycle write left is the one for pages with no activities, described above. On leaving, all that is written is the resume mode, the session time, the commit and the session close.

## Bugs fixed

**Chrome blocks `unload`** (the original reason for the issue): finalization now happens through `pagehide`, plus `freeze` and `visibilitychange` so progress is not lost if the browser discards the tab, and honouring the bfcache.

**Opening a page marked it as begun.** The pipwerks wrapper ships `handleCompletionStatus` on and rewrote `not attempted` → `incomplete` right after `LMSInitialize`. The switch is turned off in the wrapper rather than in each caller because seven places open the session.

**A single page opened and left untouched stopped the whole package from saving.** With the session left open, the next SCO's `LMSInitialize` was refused with error 101; from then on `connection.isActive` was false and every `LMSSetValue` and `LMSCommit` was dropped **silently**. The session is now always closed, and on pages that were never started `incomplete` is written just before, so that Moodle does not promote them to `completed` inside `LMSFinish`.

**The result did not reach the index until the whole page was finished.** Moodle draws its index icon from `cmi.core.exit`, not from the status, so a page closed as `suspend` was shown as suspended even when the stored status was already `passed`. The exit now follows the verdict.

**`not attempted` was being written**, which Moodle rejects with error 405. Removed from every path, and `SetCompletionStatus` no longer accepts it on 1.2 nor converts `unknown` into it.

**The order of the iDevices decided the pass.** With three equal activities, scores of 100/50/0 gave 50.5 — a pass — and the same 0/50/100 gave 49.5 — a fail — because of how the leftover point was distributed when normalising the weights. It now divides by the real sum of the weights, and the result no longer depends on the order.

**`interactive-video` registered late.** It did so from `cover.hide()`, that is, only once the learner pressed "start". It was the only one of the 35 scored iDevices missing from `cmi.suspend_data` on page load, so the weighted average was computed over the wrong denominator, and a page whose only activity was a video looked like a content page on entry and was marked completed with no score. It now registers from `initSCORM` like the other 34.

**iDevices that scored by themselves on page load:** `3dmol` and `electrical-circuits` in show mode, `select-media-files` in activities with no timer and no access code, and `scrambled-list`. Also fixed were those that lost the score on finishing (`padlock`, `trueorfalse`, `interactive-video`, `geogebra-activity`, `hidden-image`), those that saved it too early because of a visual delay, and the phantom entries in `suspend_data` that stopped a page from ever completing.

**`pipwerks.nav.goBack`/`goForward`** marked the page as completed on every in-SCO navigation.

## Improvements along the way

`trueorfalse` takes a configurable number of attempts; `word-search` can hide the time icon on timed activities; the digit limits on the time and percentage fields of eight iDevices' edition are fixed, along with `sort`'s height and its correct-position count.

## Scope and verification

48 commits, 100 files, 35 iDevices. `ADR-1831-01` is added, with the decision and the consequences accepted with it. **`translations/` is not touched.**

Two tests are worth mentioning. `public/app/common/scorm/scorm-lms-flow.test.js` wires the real runtime on top of an LMS that enforces the Moodle rules verified against its source — 405 when writing `not attempted`, 101 on a second `LMSInitialize`, promotion to `completed` inside `LMSFinish` — and asserts **what was persisted**, not what was called; that is the level at which a "nothing reaches the LMS" bug becomes visible, which per-call mocks do not catch. And `public/files/perm/idevices/scorm-registration.test.js` walks the 35 runtimes and fails if any of them registers its activity outside an initialisation entry point, which is exactly what slipped past with `interactive-video`.

`make fix` clean, 14,765 frontend tests and 2,061 exporter tests green.


erseco (changes requested):
I'm requesting changes for now, mainly because I don't think the SCORM part of this PR should move forward until the team closes the coordination, scope and product/architecture decisions that are still pending.

As discussed in the email thread, #2209 and this PR are alternative implementations of the same underlying #1831 work and overlap substantially. @ignaciogros explicitly proposed that we compare both approaches and decide which line we are going to integrate before continuing two parallel implementations.

I'm still waiting for that decision, and for the requested feedback on that discussion. Since then this branch has continued to receive substantial SCORM changes, so I think it is important that implementation momentum does not end up making the architectural decision by default.

### 1. Please separate the independent iDevice fixes

This PR contains a significant number of useful iDevice fixes that are valuable regardless of which SCORM runtime we finally select.

I would strongly prefer the fixes that are independent from the SCORM runtime/status policy to be moved into a separate PR, or into a small set of focused PRs. If some of them genuinely depend on the new SCORM semantics, those can remain identified as such.

That would allow us to review and merge useful fixes such as the iDevice lifecycle/registration/scoring corrections without coupling them to the unresolved SCORM architecture decision.

@mnarvaezm, given your long-standing work on many of these iDevices, I think this part of the work is especially valuable and should not be blocked by the runtime discussion.

### 2. We still need a decision on the SCORM runtime

There is an unresolved product-policy difference between the two approaches that should be decided explicitly by the team rather than implicitly by whichever implementation merges first.

In particular, we need to decide what `incomplete`, `passed` and `failed` mean while the learner is still working on a page.

#2209 keeps a required activity/page `incomplete` until completion and only then applies the success threshold. This PR deliberately uses `failed`/`passed` as a live verdict from the first learner interaction, reserving `incomplete` for a page that was never started.

Both behaviours can be implemented. They are not the same product contract.

I think we need an explicit decision on that contract before approving either implementation.

### 3. Runtime provenance and licensing still need clarification

The provenance/licensing of the inherited SCORM runtime was raised during the `3ipunt` Moodle plugin validation and is one of the reasons behind #2209 and the related [`moodle-mod_exelearning#105`](https://github.com/exelearning/moodle-mod_exelearning/pull/105) work.

I raised this previously in the report/email thread and we still don't have the historical information needed to close it.

@mnarvaezm, since you have been involved with eXeLearning and these runtimes for much longer than I have, could you please clarify anything you know about the original provenance of `SCOFunctions.js` and the modified `SCORM_API_wrapper.js`?

In particular, it would be useful to know whether the original ADL/CTC code had a licence grant/header before the older migrations, and where the local modifications to pipwerks came from.

If we decide to keep extending the inherited runtime, I think we need a defensible answer to that question. If that provenance cannot be established, the clean-provenance rewrite in #2209 addresses the problem by construction.

### 4. We need to clarify the SCORM 2004 support policy

This PR also makes non-trivial changes to the SCORM 2004 path.

I had understood that the project was moving away from SCORM 2004, but `main` still exposes and implements the SCORM 2004 exporter, so I don't want to assume that it is unsupported.

@ignaciogros, could you please confirm the current project decision here?

If SCORM 2004 is still a supported export target, then I think changes to its runtime semantics need an explicit contract and corresponding real-LMS validation.

If it is no longer a supported target, that should be documented as a project decision rather than continuing to expand that path as part of #1831.

There is also a concrete inconsistency in the current SCORM 2004 contract that should be resolved before merge: the ADR says that an unfinished page remains resumable (`cmi.exit = suspend`) even when a verdict has already been reported, while the added test expects `cmi.exit = normal` for a page whose activity is still in state `1` once `completion_status` is `completed`.

Whichever behaviour we want is fine, but the ADR, implementation and tests should describe the same contract.

---

I want to be clear that I'm not rejecting the work in this PR. The iDevice audit has found useful issues, and several of those fixes should land regardless of which runtime architecture we choose.

My request is to separate that valuable work from the unresolved SCORM policy/runtime decision, clarify the licensing and SCORM 2004 questions, and finish the team decision that was already requested before we continue growing two overlapping implementations.

Once that direction is agreed, I'm happy to review the selected implementation against the common test bank.

ignaciogros:
I'm attaching two SCORM packages for testing purposes, generated with these branches: 1831, 2299.

- [2299.zip](https://github.com/user-attachments/files/31512205/2299.zip)
- [1831.zip](https://github.com/user-attachments/files/31512206/1831.zip)

cristinavaldera:
> I'm attaching two SCORM packages for testing purposes, generated with these branches: 1831, 2299.
> 
> * [2299.zip](https://github.com/user-attachments/files/31512205/2299.zip)
> * [1831.zip](https://github.com/user-attachments/files/31512206/1831.zip)

The 2299.zip does not save califications nor tracking, maybe it is not a SCORM exportation? I´m uploading this new one: 
[2299b.zip](https://github.com/user-attachments/files/31524567/2299b.zip)

PR #2329, opened 2026-08-26
Style fixes: third-party credits and `nav` parameter handling

### LINKED ISSUE #2324: Review JavaScript an licenses in all styles
* Review how styles handle URL parameters when showing/hiding the navigation menu.
* Review if all styles properly acknowledge third-party materials in their `config.xml` file.
* Consider extracting common functions, such as parameter handling, into a shared file included in all exports.

ignaciogros:
## Third-party credits

- **Creative Commons license badges** (`img/licenses.gif`) are a Creative Commons trademark and are not covered by the styles' CC BY-SA license. They are now credited in **flux**, **neo**, **nova** and **zen** — `base`, `educablue` and `universal` already credited them.
- **License texts are now bundled**, as the licenses require. `educablue`, `flux`, `neo`, `universal` and `zen` ship `fonts/OFL.txt`; `nova` ships `fonts/LICENSE.txt`; `flux`, `nova` and `zen` ship `icons/LICENSE.txt`.
- **Credits now match the files actually shipped.** They were checked against each file's own metadata.

## JavaScript fixes and shared code

`nav=false`, the parameter that folds the side menu in the web export, was handled incorrectly:

- Folding or unfolding the menu wiped the whole query string, losing teacher mode (`exe-teacher=1`) and the xAPI credentials, and losing the fragment too.
- Adding the parameter could drop it inside the fragment or produce a second `?`. In the search results this broke deep links.
- The initial check matched `nav=false` anywhere in the URL, so there were false positives like `#nav=false` or `xnav=false`.
- Clicking a search result dropped the teacher mode: hits never carried the parameters the page was opened with.

Now:

- A single `$exeExport.setUrlParam(href, name, value)` in `public/app/common/exe_export.js` sets or removes one parameter, keeping every other parameter and the fragment. `teacherMode.withTeacherParams()` and the search results use it too.
- The seven styles call it directly; their own URL-rewriting code is gone.
- The initial check uses `URLSearchParams`.

New unit tests in `public/app/common/exe_export.test.js` cover the helper.

Verified manually in a browser on an exported site, with each of the styles in turn.


public/app/common/exe_export.js
```diff
@@ -152,11 +178,10 @@ window.$exeExport = {
             if (!href || !this._navParams) return href;
             if (/^(?:[a-z][a-z0-9+.-]*:|\/\/|#)/i.test(href)) return href;
             if (href.indexOf('exe-teacher') !== -1) return href;
```
erseco: Non-blocking: now that `setUrlParam()` handles exact keys and de-duplicates them, I think this guard can be removed. It can produce false positives for URLs such as `page.html?q=exe-teacher` or `exe-teacher-guide.html`, preventing the active Teacher Mode parameter from being propagated. Letting `setUrlParam()` handle the existing-key case should be sufficient.

erseco (approved):
Looks good to me. I checked this against the Teacher Mode contract merged in #1972 and the current consumers, and the canonical `?exe-teacher=1` propagation is consistent with them.

ignaciogros:
Thank you for your review, @erseco.

Your suggestion is already applied. I also improved the code comments and updated the `version` and `compatibility` values in the `config.xml` files of the styles.

PR #2339, opened 2026-08-30
Fix timeline effect

### LINKED ISSUE #2338: Timeline effect opens and closes again on a single click
In an exported page, clicking an `h2` (`.fx-timeline-marker`) or an `h3` (`.fx-timeline-minor h3`) of a timeline expands it and immediately collapses it again. The **Show** button (`.fx-timeline-expand`) behaves the same way. The timeline ends up closed but visually inconsistent: the anchor keeps the `closed` class while the animation is still running.

Reproduced on a page whose timeline sits inside a JSON iDevice (`div.idevice_node`).

[Example](https://github.com/user-attachments/files/31616069/effects-posuere-laoreet-22.zip)

## Cause

Every timeline click handler is bound **twice**. Measured on the affected page:

```js
jQuery._data(document.querySelector('.fx-timeline-container'), 'events').click
    .map(h => h.selector)
// [".fx-timeline-minor h3", ".fx-timeline-marker",
//  ".fx-timeline-minor h3", ".fx-timeline-marker"]

document.querySelectorAll('.exe-fx').length   // 5
$exeFX.counter                                // 10  <- init ran twice
```

`$exeFX.init()` runs twice on that page, and that part is by design: `exe_effects.js:794` initialises the whole page on `ready`, and `exe_export.js:512` (`afterIdeviceRendered`) calls `$exeFX.init(ideviceNode)` again, scoped to the iDevice, so that effects inside late-rendered JSON iDevices are processed too.

The scope argument is enough for the other four effects — accordion, tabs, pagination and carousel end up with exactly **one** handler each, because their `init` unbinds before rebinding. It is not enough for the timeline:

```js
// exe_effects.js:62
if (hasTimeLines) {
    setTimeout(function () { $.timeliner({}); }, 500);
}
```

`$.timeliner()` takes no root. It binds by delegation on the **global** `.fx-timeline-container` selector and never unbinds, so each call adds another layer of handlers. The `off("click", ...)` calls in `$exeFX.timeline` do not help either: they run synchronously during `init`, while both `on()` calls are deferred 500 ms, so the unbinding always happens before the binding.

## Fix

Make `$.timeliner()` idempotent, in the Timeliner block of `exe_effects.js` (around line 803). Namespace the three bindings and drop any previous ones, so the function is safe to call any number of times:

```js
// was: e(t.timelineContainer).on("click", ".fx-timeline-minor h3", function () {
e(t.timelineContainer)
    .off("click.timeliner", ".fx-timeline-minor h3")
    .on("click.timeliner", ".fx-timeline-minor h3", function () {

// was: e(t.timelineContainer).on("click", ".fx-timeline-marker", function () {
e(t.timelineContainer)
    .off("click.timeliner", ".fx-timeline-marker")
    .on("click.timeliner", ".fx-timeline-marker", function () {

// was: e(".fx-timeline-expand").click(function () {
e(".fx-timeline-expand")
    .off("click.timeliner")
    .on("click.timeliner", function () {
```

This fixes the root cause: it holds no matter how many times, or from where, `$.timeliner()` is called.

A one-line alternative is to guard the call site in `exe_effects.js:62`:

```js
if (hasTimeLines && !$exeFX.timelinerCalled) {
    $exeFX.timelinerCalled = true;
    setTimeout(function () { $.timeliner({}); }, 500);
}
```

It is safe — a timeline that only appears in a late-rendered iDevice still gets its single call — but it leaves `$.timeliner()` itself non-idempotent for any other caller.

Note that `$exeFX.counter` reaching twice the number of effects is expected here and is not itself the bug: re-running `init` over an already-initialised effect is what `afterIdeviceRendered` is for, and the other four effects handle it correctly.

ignaciogros:
Timeline effect opened and closed again on a single click.

**Solution:** 

Click bindings namespaced so repeated calls do not stack handlers.


erseco (approved):
Thanks for the fix. The implementation looks correct to me. Namespacing the handlers and removing the previous `timeliner` bindings makes `$.timeliner()` idempotent and addresses the root cause described in #2338 without interfering with unrelated click handlers.

The Codecov warning is expected because the modified Timeliner implementation is effectively a single executable line, but it also highlights that we currently have no regression test for this behavior.

If feasible with the current frontend test infrastructure, I would add a regression test that initializes `$.timeliner()` twice and verifies that a single click only toggles the timeline once. That would protect the actual requirement here: repeated initialization must not stack handlers.

I don't consider the missing regression test a blocker for this PR.

erseco (approved):
Thanks for the fix. The implementation looks correct to me. Namespacing the handlers and removing the previous `timeliner` bindings makes `$.timeliner()` idempotent and addresses the root cause described in #2338 without interfering with unrelated click handlers.

I took the liberty of adding a small regression test to the PR that initializes `$.timeliner()` twice and verifies the three affected interactions: clicking a minor heading, clicking a timeline marker, and using the Show/Hide button. This should also cover the modified line reported by Codecov and protect against handlers being stacked again in the future.

Approved from my side.

ignaciogros:
Thanks for the review and the new test.

PR #2343, opened 2026-08-31
Fixed style order problem in single-page exports

### LINKED ISSUE #2282: Theme stylesheet loads before exe_effects.css in single page exports
In web and SCORM exports the theme stylesheet is the last one in `<head>`, after `exe_effects.css`, `exe_atools.css` and `base.css`. In the single page export the order is reversed: `theme/style.css` comes right after `base.css`, and `exe_effects.css`, `exe_highlighter.css` and `exe_atools.css` are loaded after it.

Because `exe_effects.css` styles components with the same specificity the theme uses (e.g. `.exe-content .fx-accordion-title`), the tie is broken by sheet order: the theme wins in web/scorm and loses in single page. The same theme therefore renders differently depending on the export format — accordion, pagination and timeline fall back to the effects' own defaults (dark `#333333` bar with white text instead of the theme's bar).

Reproduced with an unmodified export from eXeLearning 4.0.3: 13 theme selector groups lose in single page, all of them against `exe_effects.css`.

Expected: the theme stylesheet should be loaded last in every export format, as it already is in web and scorm.

ignaciogros:
Make theme stylesheet load last in single-page exports.

### DIFF
--- a/src/shared/export/renderers/PageRenderer.spec.ts
+++ b/src/shared/export/renderers/PageRenderer.spec.ts
@@ -1002,6 +1002,57 @@ describe('PageRenderer', () => {
             expect(html).toContain('<span class="license">');
             expect(html).not.toContain('href="https://creativecommons.org/licenses/by/4.0/"');
         });
+
+        it('should load the theme stylesheet last, after libraries and base.css (#2282)', () => {
+            const pages = [createTestPage()];
+            const html = renderer.renderSinglePage(pages, {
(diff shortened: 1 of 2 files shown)


erseco (approved):
Looks good to me — nice, focused fix. The stylesheet order now matches the multi-page export, and the tests cover the intended behavior well. 👍

ignaciogros:
Thanks for the review!

PR #2350, opened 2026-09-02
Mermaid: preload the library while the user writes the diagram.

### LINKED ISSUE #2349: Improve Mermaid behavior in the Electron version
See https://github.com/exelearning/exelearning/pull/545#issuecomment-5509111057

ignaciogros:
Mermaid (2.6 MB) was loaded on demand when the editor was deactivated, exactly when the user closed or saved the iDevice. This meant the download, library parsing, `mermaid.initialize()`, and first render all happened during project saving, competing for the same main thread and causing the diagram to take seconds to appear.

## Change

* **`plugin.min.js`**: `$exe.mermaid.loadMermaid()` is now called when the Mermaid dialog opens instead of when the editor is deactivated. The `deactivate` handler is kept because it is still responsible for rendering the diagram after editing.
* **`common.js`**: a new `loading` flag is added to `$exe.mermaid` to prevent the script from being injected twice. It is cleared on `onload` and `onerror`, allowing a failed download to be retried.

## Benefit

The download now happens while the user is writing the diagram code instead of being added to the save operation. By the time the editor closes, the library is already downloaded and compiled, so rendering no longer competes with saving.

There is also **zero cost for users who do not use Mermaid**: the library is still not loaded unless the page contains diagrams or the dialog is opened.


erseco (approved):
Tested in the macOS desktop app and everything works as expected. Mermaid now renders promptly after saving, including when saving quickly after opening the dialog. I also checked editing existing diagrams, cancelling/reopening the dialog, preview, and that Mermaid is not loaded when unused. Looks good to me.

Non-blocking: the new tests in `exemermaid-plugin.test.js` exercise a local `preloadMermaid()` helper that mirrors the production logic rather than the actual `openHTMLDialog()` implementation. This means they would still pass if the preload call were accidentally removed from the plugin.

It would be better to cover the real plugin behavior, ideally verifying that Mermaid starts loading when the dialog opens and that a quick save does not trigger a duplicate load. This does not block the PR.

ignaciogros:
You're right. Thanks for the review and the suggestion.

Issue #674, opened 2025-11-17
Latex code


luquichurri:
When generating LaTeX code, I've encountered some problems.

1. In the "text" iDevice, when I enter color in Latex cede, it appears to work correctly; it displays properly, but it's not correct in the preview mode. I've tried exporting it, and it's still not correct (see images 1 and 2).

![Image](https://github.com/user-attachments/assets/df7db758-0e6f-4f7d-8ef4-f39c8e608c88)
![Image](https://github.com/user-attachments/assets/01e017fb-ac00-4c9e-a339-1d9c9e1e73ac)

2. In the "test" iDevice, LaTeX code appears to display correctly in the answers, but it doesn't work correctly in preview mode. (see images 3 and 4)

![Image](https://github.com/user-attachments/assets/4dc20367-8b5e-4a7c-bb31-f81171f70d83)
![Image](https://github.com/user-attachments/assets/de73562b-dee8-474b-98f0-883c54b7ebc9)

I can send the .elpx files if needed.

PR #2351, opened 2026-09-02
fix(math): upgrade MathJax to 4.1.3 and generate the vendored tree

### LINKED ISSUE #2259: Interactive math speech/explorer is broken by the MathJax 3.2.2/4.0 version mix (restore with matched components)
## What happens

Activating the assistive explorer from the MathJax context menu ("Explorer" / speech features) crashes before any speech data is fetched:

```
Component a11y/sre uses 4.0.0 of MathJax; version in use is 3.2.2
Uncaught TypeError: Cannot read properties of undefined (reading 'SpeechMathItemMixin')
```

Reproduced in Chrome against the static build (main @ 3c7c7e821): the loader fetches `a11y/sre.js`, `a11y/semantic-enrich.js`, `a11y/explorer.js` and the explorer never registers. Zero `sre/mathmaps/*.json` requests are ever issued.

## Why

`public/app/common/exe_math/` mixed two MathJax generations: the active **3.2.2 combined component** (`tex-mml-svg.js`, restored by the "Major rewrite" in #869) and leftover **4.0.0 standalone components** from the earlier v4 upgrade attempt (#752): `a11y/*`, `sre/speech-worker.js`, the 16 `sre/mathmaps/*.json` speech locales (~4.2 MB), `loader/startup/core`, standalone input/output jax, Node-only adaptors. The 3.2.2 loader refuses the 4.0.0 components, so everything behind the a11y menu entry has been unreachable dead weight.

The dead v4 files were removed in the static-size-reduction work (see the PR referencing this issue). That removal did **not** change current behavior (broken → absent) and the baseline screen-reader support is intact: `a11y/assistive-mml` is embedded inside `tex-mml-svg.js` and still emits hidden MathML for every equation.

## What restoring speech takes

Vendor the **version-matched 3.2.2** files from the `mathjax@3.2.2` npm tarball:

- `es5/a11y/{sre,semantic-enrich,complexity,explorer,assistive-mml}.js`
- `es5/sre/mathmaps/<locale>.json` for the locales we choose to support (each ≈ 30–44 KB compressed; `en` + `nemeth` braille ≈ 60 KB, full 13-locale set ≈ 490 KB in the release zip)

and QA the explorer flow (keyboard navigation, speech output with a screen reader, braille). Locale selection is the size/product decision: per-locale zip cost measured at ~30–44 KB each (en 41.7, es 32.6, ko 43.6, hi 41.3, de 35.6, nemeth braille 17.7, …).

Note: exports copy the whole `exe_math` directory into any package containing math, so whatever is vendored ships in every math export too.

## Acceptance

- Context-menu explorer activates without console errors.
- Speech strings are generated for a sample equation in each shipped locale.
- Braille (nemeth) path exercised if shipped.
- An E2E spec covers explorer activation so a future version mix cannot regress silently.

erseco:
> **Stacked on #2260.** The base of this PR is `feature/reduce-static-bundle-size`, so the diff shown here is only the MathJax work. GitHub will retarget it to `main` automatically when #2260 merges. #2260 originally contained the MathJax part; that commit was reverted there and the work moved here — the reasoning is in that PR's description.

## What this does

Finishes the MathJax 4 upgrade that was started in #752 and half-reverted in #869, makes the half-reverted state impossible to reach again, and makes the vendored tree self-contained — which it turned out not to be.

Fixes #2259.

> **Decided: the Speech Rule Engine is not shipped.** Ruled on a review that agreed with removing braille; braille alone turned out to be 19.5 KB of the release ZIP out of the engine's 469 KB, so the two were decided together. See *The decision on speech and braille* below and ADR-2259-03, now `Accepted`.

## The bug

`public/app/common/exe_math/` has shipped **two MathJax generations at once** since January. #752 (`cb778b969`) upgraded it to 4.0.0. #869 (`9272d5c67`) rolled back exactly the 35 files the two versions share and left the other 50 in place. Nothing noticed.

| Part of the tree | Version on `main` |
|---|---|
| `tex-mml-svg.js` (the only file actually loaded) | **3.2.2** |
| 34 of 40 `input/tex/extensions/*.js` | **3.2.2** |
| `core.js`, `loader.js`, `startup.js`, `output/`, `ui/`, `a11y/`, `sre/` | **4.0.0** |

The 3.2.2 loader refuses the 4.0.0 components, so opening the accessibility menu gives:

```
Component a11y/sre uses 4.0.0 of MathJax; version in use is 3.2.2
Uncaught TypeError: Cannot read properties of undefined (reading 'SpeechMathItemMixin')
```

About **5.8 MB of the 8.6 MB directory was unreachable code** that every math export still carried. Broken since 2026-01-03; issue #2259 was opened on 2026-08-14 from a code audit, not from a user report.

## Was there ever a MathJax 4 bug?

No. There is no issue, PR comment or review anywhere stating a reason for the rollback. MathJax 4 shipped fine in v3.0.2 and *closed* two open LaTeX bugs (#674, #648). The rollback happened on the `release/3.1` Bun-rewrite branch as merge hygiene — `"Excluded MathJax v4 upgrade (keeping MathJax v3)"` — to keep the browser bundle matching the new server-side pre-render pipeline, which was built against `mathjax-full@3.2.2`.

The one written technical objection is the TODO at `common.js:89`, claiming seven TeX extensions had *"dependency issues with bundled tex-mml-svg.js (MathJax 3.x)"*. That is a symptom of the downgrade, not of v4: five of the seven render correctly on a consistent v4 tree (verified in a browser).

## Approach

`scripts/vendor-mathjax.ts` regenerates the tree from the pinned packages (`make vendor-mathjax`), and `vendor-mathjax.spec.ts` compares the committed tree against them **by SHA-256**, failing on any missing, extra or changed file. A partial revert like #869 can no longer be committed silently. A second test fails if any vendored `.js` still contains `"3.2.2"`.

`writeVendoredTree()` is `copyFileSync` and nothing else: no patching, minification, rewriting or concatenation. Every behaviour of our own lives outside the vendored tree (`common.js`, `mathjax-packages.ts`), so "we distribute MathJax as-is" is a CI invariant rather than a convention.

## MathJax 4 does not bundle its font, and that was leaking to a CDN

Found while reviewing the branch, and it is why the size numbers moved.

MathJax 4 keeps most of the font **outside** the combined component. `tex-mml-svg.js` carries a base set only — verified by bytes: it is 1,849,738 B against 874,032 B for the `-nofont` variant, and no fragment of any `svg/dynamic/*.js` appears inside it — and fetches the rest on first use through `loader.paths.fonts`, whose stock value in a browser is `https://cdn.jsdelivr.net/npm/@mathjax`.

Nothing configured that path, so these resolved to jsdelivr:

- the glyph ranges behind `\mathbb`, `\mathcal`, `\mathfrak`, `\mathscr`, `\mathsf`, `\mathtt`, the stretchy `\uparrow`/`\downarrow` delimiters, and a long tail of operators and shapes;
- the **`mhchem` and `dsfont` font extensions**, on every page using them — and `dsfont` is one of the five extensions this PR newly enables, so the branch was introducing part of the leak it now closes.

It is invisible in a browser with a network, which is why the suite passed: the formula renders, and only the wire shows the dependency. In practice it means a third-party request inside every exported package, and a missing glyph wherever there is no network — an extracted HTML5 export, an EPUB, an unpacked SCORM package on an isolated LMS, the Electron app. `exe_math/` is copied into **every** export whose content matches `/\\\(|\\\[/` (`LIBRARY_PATTERNS`), not only those that enabled runtime MathJax, so this is per-package weight and per-package behaviour.

### What is vendored, and why exactly that

`paths.fonts` now points at `exe_math/fonts`, so a font asset can never leave the origin, and the assets live next to the bundle:

- **13 of the 40 glyph ranges.** The line is measured, not chosen: exactly the set whose code points MathJax 3.2.2's bundled TeX font could already render, computed by crossing the bundle's own dynamic-range table against `TeXFont.defaultChars` from `mathjax-full@3.2.2`. Vendoring them means the upgrade loses no glyph anyone could previously see.
- **the `mhchem` and `dsfont` font extensions**, now direct dependencies pinned to 4.1.3 like everything else.

Two results of that measurement were counter-intuitive:

| Excluded range | v3 code points it would restore | cost |
|---|---:|---:|
| `latin`, `latin-b`, `latin-i`, `latin-bi` | **0** | 1.5 MB |
| `greek` + `greek-ss` | 1 (ϝ) | 1.6 MB |
| `PUA`, `monospace-ex`, `sans-serif-ex` | 47 | 1.4 MB |
| `cyrillic*`, `hebrew`, `arabic`, `devanagari`, `cherokee`, `braille*`, `phonetics*`, `marrows`, `mshapes`, `sans-serif-[rbi]*`, `monospace-l` | 0 | 2.0 MB |

`latin` is the surprise: accented characters (á, ñ, ü) were **never** in MathJax 3.2.2's font either, so `\text{función}` degrades exactly as it always did. Not a regression, and 1.5 MB not spent.

`EXCLUDED_FONT_RANGES` records every excluded range with its reason, and a test asserts vendored ∪ excluded equals what the font package publishes — so a MathJax upgrade that adds a range fails the build instead of shipping unweighed.

### CHTML, and the menu entries that pointed at nothing

- **Math Renderer → CHTML.** `output/chtml.js` was vendored but its CHTML font is a separate 2.4 MB package that is not. Choosing it fetched two files that were not there. `output/` is now dropped entirely (`output/svg` is inside the combined component, declared via `checkVersion`, so the loader never fetches it) and `common.js` hides the Math Renderer submenu.
- **A renderer persisted in `localStorage`.** The menu stores its settings per origin and `applySettings()` acts on them while the document is being built. One CHTML click on any MathJax page of the same site made every later page request the missing component at startup, without anyone opening the menu. `forgetUnavailableMenuSettings()` drops that key before `defaultReady()`.

Same family as the Speech, Braille and Explorer sections below: an entry MathJax offers because its bundle knows about it, pointing at a file we do not ship.

## Accessibility

MathJax 4 flips v3's defaults: the combined component enables speech and turns hidden MathML **off**. That is a regression for us, because speech runs in a web worker built from a `blob:` URL that calls `importScripts()` and then fetches a locale map — none of which a `file://` document can do. An export opened from disk would announce **nothing at all**.

So:

- **`a11y/assistive-mml` is enabled unconditionally.** Hidden MathML is the only path with no worker and no fetch; NVDA, JAWS and VoiceOver convert it to speech themselves. It is the accessibility floor and it works everywhere, including in an export opened from the filesystem. ADR-2259-02.
- **Speech, braille and the explorer are off, and their menu sections hidden.** See the decision below. Deleting the engine's files is not enough, because speech is compiled into the combined component: the menu would keep offering the toggles and each one would request a file that is not there, leaving `MathJax.typesetPromise()` **unsettled** — `speechError` is never called and the typeset queue wedges, measured.

### The floor is restored per load, not enforced — measured

Review of this branch asked whether the floor ADR-2259-02 promises actually holds. It does not, structurally, and the ADR now says so.

MathJax models assistive MathML and speech as **alternatives**: `setSpeech(true)` calls `setValue(false)` on the `assistiveMml` variable, and `applySettings()` copies the menu's value over `document.options.enableAssistiveMml` on every load. Reproduced in Chrome on this branch:

- clean profile → `enableAssistiveMml: true`, one `mjx-assistive-mml` per formula. The floor holds.
- toggle Speech off and back on in the contextual menu → `{"assistiveMml":false}` in `localStorage` and **zero** `mjx-assistive-mml` nodes for two formulas.

The key is scoped to the origin, so it follows a reader between packages on the same LMS, and in an offline export speech cannot start either — no accessible maths at all, silently. `forgetUnavailableMenuSettings()` now drops a stored `assistiveMml: false` as well, restoring the floor on every load. That is a mitigation, not an invariant; dropping SRE would make it one, since the menu would no longer offer the toggle that trades one for the other.

ADR-2259-02 also now states, for a reviewer who reads the MathJax 4 documentation and finds the opposite advice, why a decision that is right for served pages inverts for a tool that ships offline packages.

## The decision on speech and braille

The review agreed with removing braille. Measuring it first changed the shape of the answer: `nemeth.json` deflates well, so **braille alone is 19,550 B of the release ZIP** out of the engine's 469,461 B, and removing it leaves every structural cost in place — the worker, the seven remaining locale maps, the reach into MathJax's internal `_` namespace, and the toggle that could switch the accessibility floor off. It is not a lever, it is rounding.

The argument accepted for braille — the assistive technology already turns MathML into braille, MathJax only adds the last translation step — applies unchanged to speech: the same pipeline, one step shorter. So the two were decided together and **the Speech Rule Engine is not shipped**.

| | before | after |
|---|---:|---:|
| `exe_math/` raw | 7,158,088 | **4,490,045** |
| `exe_math/` in the release ZIP | 1,999,957 | **1,530,496** |
| vs. `main` (1,854,484 zipped) | +145,473 | **−323,988** |
| files | 87 | **73** |

What is gone: the expression explorer (arrow-key navigation), Nemeth braille, and Auto Voicing. What is not affected: **screen readers**, which get the maths from the hidden MathML and convert it themselves — the path that actually serves blind learners, and the only one that works offline.

Four things carried the decision:

- **Cost per package.** 2,668,043 B raw in **every** export containing LaTeX, not only those that enabled runtime MathJax. For scale, eXeLearning's accessibility toolbar — larger and smaller text, uppercase mode, dyslexia-friendly font, contrast, translate, read-aloud over the whole page — is **649,844 B** and covers every page of every project. Formula-only speech cost 4.1× that.
- **It could not run where our content is read.** Speech and braille come out of the same `blob:` worker that calls `importScripts()` and then fetches a locale map. A `file://` document can do none of it, so in an extracted HTML5 export, an EPUB, an unpacked SCORM package or the Electron app the feature was never available at all.
- **Uneven by construction.** SRE has rules for `af ca da de en es fr hi it ko nb nn sv`; of eXeLearning's interface languages `eo`, `eu`, `gl`, `pt`, `ro` and `va` have none.
- **It hardens the floor.** MathJax models assistive MathML as an *alternative* to speech rather than a floor beneath it: toggling Speech in the menu called `setValue(false)` on it and persisted that per origin, so a reader could end up with no speech (offline) and no MathML either. With speech out of the menu there is nothing left to trade it against. This is what turns ADR-2259-02's promise from "restored on every load" into an invariant.

It also removes `trimSpeechLocaleMenu()` and its reach into `MathJax._.a11y.sre_ts.locales`, which was the one piece of this PR with no compatibility promise behind it.

Reinstating any of it is a small change, and ADR-2259-03 records exactly which four settings and which files.

### On translating the MathJax menu

Asked and answered in the negative, recorded so it is not re-litigated. The menu renders in English while the interface is in Spanish. **This is not new and not caused by this PR:** MathJax dropped menu localisation in v3 and never restored it in v4 — the labels are hardcoded literals, identical in the 3.2.2 build on `main`. eXeLearning has shipped a MathJax 3.x bundle since `0b18ec304` (2025-06-02), so the menu has been English for about fifteen months with no complaint. It *can* be translated from our side — walking `menu.items` and rewriting `.content` in `startup.ready()` renders it in Spanish, verified end to end — at a cost of 133 labels to wrap in `_()`, three submenus built on demand that need a second pass, and `translations/**` being managed by a separate process, so the menu would still render in English on merge. Its own PR if we want it.

## Browser / server parity

The server-side pre-renderer moves from `mathjax-full@3.2.2` to `@mathjax/src@4.1.3`, so a document renders identically whether the export ran in the browser (primary path) or on the server (CLI and external API). The browser pre-renderer (`LatexPreRenderer.js`) uses the workarea's own MathJax, so the vendored font ranges are what keep the two sides equal — before this fix the server resolved ranges from `node_modules` while the browser resolved them from jsdelivr.

v4 dropped the `AllPackages` barrel — every TeX extension now registers itself on import, and a package named but never imported is silently omitted with a console warning. The enabled set is therefore an explicit list in `src/shared/export/prerender/mathjax-packages.ts`, and `mathjax-packages.spec.ts` asserts it equals `externalExtensions` in `common.js`.

## edicuatex

The standalone CDN fallback moves to `cdnjs…/mathjax/4.1.3/tex-svg.min.js` (v4 dropped the `es5/` path segment; verified: the new URL returns 200, the old one 404). Inside eXeLearning the editor loads *this* tree, which ships no Speech Rule Engine, so it also carries `a11y/assistive-mml` — the floor, missing here until @jjdeharo spotted it — and the same speech/braille/explorer settings and hidden menu sections, guarded by `isInExe`. Without them the formula preview stalls on the first toggle.

**Most of this is now upstream and released.** @jjdeharo published `edicuatex` to npm (1.5.4 today) and dropped the Speech Rule Engine there for the same reasons, so of the three eXeLearning-specific edits this branch started with, two are gone: `a11y/assistive-mml` is his, and his font-path handling — keyed off which copy actually answered inside `loadMathJax()` — is better than the `isInExe` guess it replaced. This branch has been aligned with what is released ([`21b90814d`](https://github.com/exelearning/exelearning/commit/21b90814d)): @franmate's i18n priority from [edicuatex#3](https://github.com/edicuatex/edicuatex.github.io/pull/3), and `[tex]/cases` + `[tex]/mathtools`, which were configured but never loaded.

One difference is left, and it has a PR open: [edicuatex#7](https://github.com/edicuatex/edicuatex.github.io/pull/7). His speech guard is gated on `url === MATHJAX_LOCAL_URL`, which is false inside eXeLearning, where the host supplies MathJax — so the guard skips the one deployment that needs it. One line: gate on "not the CDN" instead.

**Next step, separately.** Once that lands and a release exists, a stacked PR will add `edicuatex` to `package.json`, delete the 29 hand-copied files under `public/app/common/edicuatex/`, and vendor from the pinned package with the same SHA-256 drift check `scripts/vendor-mathjax.ts` uses here. It is deliberately not in this PR: it touches `tinymce_5_settings.js`, `prune-dist.ts`, `minify-dist.ts` and `translations.ts`, none of which this one goes near.

## Also in this PR

- **Five extensions enabled** that the 3.2.2 bundle could not load: `begingroup`, `colorv2`, `dsfont`, `texhtml`, `units`. The TODO in `common.js` is retired. `bbm` and `bboldx` stay out **on weight, not availability**: `@mathjax/mathjax-bbm-font-extension` and `@mathjax/mathjax-bboldx-font-extension` both ship an `svg.js` (206 KB and 141 KB), the same shape as the two extensions vendored here. They are out because nobody has asked for the macros and each is heavier than both vendored extensions combined.
- **Duplicated config removed** from `scripts/static-bundle/static-index.html` (28 lines). It now sets `window.MATHJAX_BASE_PATH` and lets `common.js` build the rest, so the extension list, a11y options, font paths and locales are defined once.

## Two findings worth flagging

**1. A test was passing on an artefact.** `latex-rendering.spec.ts` asserted "at least 2 `mjx-container` (one inline, one display)". MathJax 3 nested a *duplicate* container inside `mjx-assistive-mml`, so a single formula counted twice — the assertion was satisfied by one formula, and the display half was never actually verified. v4 emits clean `<math>` there. The test now counts rendered formulas, asserts every one carries hidden MathML, and genuinely checks display math.

**2. The test now uses the supported display delimiter.** It previously wrote display math as `$$…$$`, which does not survive the editor round-trip (a pair collapses to a single `$`, so nothing renders) — reproduced identically on `main`. Not a bug to file: dollar delimiters were deliberately dropped, and only `\(…\)` and `\[…\]` are supported, per the decision on #2269 closing #1990. The test now uses `\[…\]`.

Worth noting separately: `tex.displayMath` still lists `$$…$$` even though the delimiter is unsupported, so `$$` renders in exports but breaks through the editor. Aligning the configuration with that decision is left out of this PR — removing it outright would stop existing `$$` content rendering in exports, where it currently does — and tracked in #2352, which records what a migration would need.

## The regression guard the suite was missing

The font leak survived a green 543-test E2E run because every assertion was structural — container counts and `data-latex` attributes — and a glyph whose range is missing does not blank the formula: MathJax falls back to drawing the character with a `<text>` element in the CSS `unknownFamily` (serif), so the container, the assistive MathML and the text content all still look right.

`latex-rendering.spec.ts` now renders `\mathbb{R} \subset \mathcal{L}` in the runtime-MathJax test and asserts:

- the glyphs drew real `<path d>` outlines and **no `<text>` fallback exists**;
- at least one `/svg/dynamic/` range was requested, and every one that was came back without a 4xx;
- **no script, font, stylesheet or fetch left the origin while typesetting** (images are excluded on purpose — the workarea chrome pulls a Gravatar avatar, and a guard that trips on page furniture is a guard people delete).

Verified in both directions: removing `exe_math/fonts` and re-running turns the whole page's maths off, which is what the guard reports. That is itself worth knowing — a 404 under `fonts/` does not lose one glyph, it wedges the typeset queue for the page.

The same gap existed on the server side and is now closed: `ServerLatexPreRenderer.spec.ts` checked `\mathbb` and `\mathcal` **by `data-latex` only**, which is the input echoed back. It now pre-renders `\mathbb{R} \subset \mathcal{L}` and requires `<path d>` with no `<text>` — the Node counterpart of the browser guard, and the only coverage of `mathjax.asyncLoad`. Falsified by removing that line: the render fails with `dynamic file 'double-struck' failed to load`.

## Verification

| Gate | Result |
|---|---|
| `make fix` | clean |
| Frontend (vitest) | **15,368 pass**, 289 files |
| Backend (`bun test ./src ./scripts`) | 8,343 pass; 2 fail in `Assets API v1`, which pass 31/31 when that spec runs alone — pre-existing cross-spec pollution, unrelated |
| `make test-integration` | **772 pass**, 0 fail |
| `make test-e2e` (chromium, full) | **543 passed**, 11 skipped, 0 failed (17.9 min) |
| `make architecture-check` | OK |

Patch coverage: `vendor-mathjax.ts` 99% lines, `mathjax-packages.ts` 100%, `ServerLatexPreRenderer.ts` 100%; `hideUnavailableMenuEntries()` and `forgetUnavailableMenuSettings()` covered on every branch by `common_mathjax.test.js` (22 tests).

## Architecture records

- **ADR-2259-01** — Generate the vendored MathJax tree from one pinned package *(manifest updated)*
- **ADR-2259-02** — Make hidden MathML the math accessibility floor and bound the speech locales *(updated: what the floor does and does not guarantee, and why this goes against MathJax's own recommendation)*
- **ADR-2259-03** — Weigh the Speech Rule Engine against the accessibility surface it does not cover *(`Accepted`: not shipped)*


jjdeharo:
Hi @erseco — I merged the EdiCuaTeX side of this today ([#4](https://github.com/edicuatex/edicuatex.github.io/pull/4) and [#5](https://github.com/edicuatex/edicuatex.github.io/pull/5)), so the editor is on MathJax 4.1.3 and aligned with what this PR vendors. Two things worth passing back here.

### One gap I think is real

In `public/app/common/edicuatex/js/edicuatex-tools.js` this PR adds `loader.paths` and the `menuOptions` that hide Speech/Braille/Explorer, but it does not add `a11y/assistive-mml` to `loader.load`. In `common.js` it is there (`.concat(['a11y/assistive-mml'])` plus `enableAssistiveMml: true`).

The editor runs in its own iframe with its own `window.MathJax` config, so as far as I can tell it does not inherit the floor set in `common.js`. If that reading is right, an author using a screen reader gets an unlabelled `role="img"` in the formula preview inside eXeLearning, even though the exported content is fine.

I verified the underlying behaviour in a browser, 3.2.2 vs 4.1.3 from the CDN with the same config: `mjx-assistive-mml` present on 3, absent on 4, and the `MJX-SPEECH` element on 4 is empty. Adding the component to `loader.load` brings the hidden MathML back. I have not tested it inside eXeLearning, so this is a reading of the diff rather than a reproduction — you are better placed to confirm it.

Fixed upstream in [edicuatex@5e64791](https://github.com/edicuatex/edicuatex.github.io/commit/5e64791).

### Thanks for ADR-2259-02 and -03

I spent a while trying to get SRE speech to produce anything and failed from both a vendored copy and the CDN. Your ADR explains why: the `blob:` worker plus `importScripts()` and the locale fetch, none of which survives `file://`. That is the piece I was missing, and dropping SRE is clearly the right call for exports.

### One change to #5 you should know about

I did not keep `js/mathjax/` gitignored. Measured on the SVG-only tree (no CHTML, no duplicate bundles): 18 MB on disk but **3.4 MB in git**, 125 files — the repo went from 1.7 MB to 4.99 MiB. That removes the 404-then-fallback round trip on every load, removes the CDN request entirely, and makes an offline or CSP-restricted deployment work from a plain clone. `vendor-mathjax.mjs` is now the regeneration step rather than a required install step. The npm tarball is unaffected — `files` still excludes it.

### And an argument for #6 that turned up by itself

The `edicuatex-tools.js` in this PR starts from `load: ['[tex]/color', '[tex]/mhchem']`, which is the editor as it was before today. The hand-copy and the source have diverged within a single day, through nobody's fault. With `edicuatex` in `package.json` the accessibility fix above would reach you on the next bump instead of by hand. I still owe you the npm account — that part is on me.

erseco:
Thanks @jjdeharo — the gap is real, I have reproduced the mechanism, and it is fixed in `ce9bb6ab0`.

### The accessibility gap

Your reading of the diff was right, and it was worse than the diff suggests. The editor's iframe has its own `window.MathJax`, so it never had the floor from `common.js`; MathJax 4 dropped `assistive-mml` from the combined component; and this branch then removed the Speech Rule Engine, which took away the only remaining path. So inside eXeLearning the formula preview went from "speech, online only" to nothing at all — a bare `role="img"`. Exported content was never affected, but the author's own preview was, and that contradicts ADR-2259-02, which claims the floor holds in every context. Caught by you, not by me.

I confirmed **why your one-line fix is sufficient on its own**, which is worth recording because it is not obvious. `a11y/assistive-mml.js` ends with:

```js
MathJax.startup && (
  MathJax.config.options && MathJax.config.options.enableAssistiveMml !== false &&
    combineDefaults(MathJax.config, 'options', { menuOptions: { settings: { assistiveMml: true } } }),
  MathJax.startup.extendHandler(t => u(t))
)
```

so loading the component registers **both** the document option and the menu setting. That second half matters: MathJax builds the menu inside the document constructor and then writes the menu's settings back over the document's, so `enableAssistiveMml: true` set on its own gets overwritten a moment later. Two consequences: nothing else needs adding, and `options.enableAssistiveMml = false` would silently defeat the whole thing.

I took your exact change rather than inventing a different one, so the next re-vendoring is a no-op here.

### On not gitignoring `js/mathjax/`

Makes sense, and it does not cost eXeLearning anything — I checked before saying so. `files` carries `"!js/mathjax"`, and `npm pack --dry-run` gives 32 files, 179.6 kB packed. So when eXeLearning consumes the package it gets the editor without a second copy of MathJax, which is what we need: inside eXe the editor loads *our* vendored tree. The 4.99 MiB is only paid by someone cloning the repo, and it buys a working offline clone. Your call, and a reasonable one.

One note for whoever wires it up on our side: `dependencies` pulls `mathjax@4.1.3` and the two font packages, the same versions this PR pins, so there is no duplicate install either. And if we ever vendor from the git tree instead of npm we would need to exclude `js/mathjax/` ourselves — the npm route avoids that entirely.

### On #6

Agreed, and your example is the good one: `load: ['[tex]/color', '[tex]/mhchem']` in this PR is the editor as it was *yesterday*, and the accessibility fix reached me through a code review rather than a version bump. That is the whole argument in one paragraph. Whenever the npm account exists, add `edicuatex` to our `package.json` and I will wire the vendoring the same way `scripts/vendor-mathjax.ts` works here — SHA-256 drift check included, so a hand-edit to the vendored copy fails CI instead of quietly diverging.

### On SRE

Glad the ADR was useful. For the record, the decision landed while you were writing: the engine is out, and the ZIP measurement that settled it is that braille alone was 19,550 B of 469,461 — not a lever, so braille and speech were decided together. One thing your investigation adds that I had not written down: you could not get speech to produce anything *even from the CDN*, which is a stronger statement than mine — I only measured that it cannot start from `file://`.

erseco:
@jjdeharo the package is up, so here is where I have taken it and what I propose for the last step.

### Upstreamed the one thing that was still a local edit

[edicuatex#7](https://github.com/edicuatex/edicuatex.github.io/pull/7). Comparing `edicuatex@1.5.1` against the copy vendored here, two of our three eXeLearning-specific changes are already yours — `a11y/assistive-mml`, and the font path, where your version is better than mine: keying `loader.paths` off which copy actually answered inside `loadMathJax()` beats my `isInExe` guess, so I dropped mine.

What was left is the menu hardening, and it needs to be upstream rather than here: inside eXeLearning the editor uses the host's MathJax, and this branch ships none of the Speech Rule Engine, so the menu keeps offering Speech, Braille and Explorer against files that are not there — toggling one leaves `typesetPromise()` unsettled and the preview stops updating without reporting anything. Two details in that PR that cost me time and may save you some: the settings have to go in `menuOptions.settings` rather than `options`, because the document constructor writes the menu's values back over the document's a moment later; and `enrich` gates all four, so it is the only one that actually has to be false.

Your own copies are unaffected — the vendored tree and the CDN both carry the engine.

### Aligned this branch with 1.5.1 in the meantime

`21b90814d` brings the two fixes that were already released and that we were missing: @franmate's i18n priority with fallback to the local catalogue (ours returned the English string when eXeLearning had no translation), and `[tex]/cases` + `[tex]/mathtools`, which were listed in `tex.packages` and never loaded.

I did **not** bring the local-copy loader with CDN fallback. It resolves `js/mathjax/` next to the file and the tarball excludes that directory, so here it would only add a 404 before the CDN fallback it already uses. That one belongs with the migration, where the file is replaced verbatim instead of patched.

### The migration itself: after this merges

I am deliberately not doing it in this PR. It is already 118 files and about MathJax; swapping how EdiCuaTeX is distributed touches a different set — `tinymce_5_settings.js`, `prune-dist.ts`, `minify-dist.ts`, `translations.ts` — and would put a third level on a stack that already has two. Once this lands I will open a PR that adds `edicuatex` to `package.json`, deletes the 29 hand-copied files under `public/app/common/edicuatex/`, and vendors from the pinned package with the same SHA-256 drift check `scripts/vendor-mathjax.ts` uses here, so a hand-edit to the vendored copy fails CI instead of quietly diverging — which is exactly how this file drifted in the first place.

Thanks for doing the npm work. The measurable result is that of the three local edits this PR started with, two are already gone and the third has a PR open — and the next one arrives as a version bump rather than as a code review.

erseco:
Rebased on `main` now that #2260 is in — 118 files, all MathJax and edicuatex, nothing left over from the base. Vendored tree in sync, `make architecture-check` OK, vitest 15,368 and integration 772 green.

On edicuatex: @jjdeharo published to npm and dropped the Speech Rule Engine upstream too, so two of the three local edits here are now his — including the font-path handling, which he keys off which copy actually answered rather than my `isInExe` guess. This branch is aligned with what is released in `21b90814d` (@franmate's i18n priority, and `[tex]/cases` + `[tex]/mathtools`, configured but never loaded).

One difference left, one line, PR open: [edicuatex#7](https://github.com/edicuatex/edicuatex.github.io/pull/7). His guard is gated on `url === MATHJAX_LOCAL_URL`, false inside eXeLearning where the host supplies MathJax, so it skips the deployment that needs it.

The npm migration goes in its own stacked PR once that lands and a release exists — pinning 1.5.4 today would drop the guard.

ignaciogros (changes requested):
Thanks for this improvement, @erseco.

The tests have passed, but I think there's one thing to fix before approving this. Could you please fix it?

## Blocker: accented letters break maths rendering

MathJax 4 does not carry the whole font. It downloads glyphs in groups, on demand. The PR ships 13 of those 40 groups and points MathJax at the local copy, so a group that was left out now returns 404 instead of coming from a CDN.

One of the groups left out is `latin`, which holds á é í ó ú ñ ü ç. Also missing: `latin-b`, `latin-i`, `latin-bi`, `greek`, `greek-ss`, `PUA`, `monospace-ex`, `sans-serif-ex`, `accents-b-i`, `cyrillic`, `hebrew`.

When MathJax asks for a missing group, the first letter degrades quietly and falls back to a serif glyph, which is what MathJax 3.2.2 did too. But the failure is remembered, so the next accented letter makes the whole rendering call fail. Every formula in that call is left on screen as raw `\(...\)`, including formulas with no accents at all.

Confirmed in the browser. A text iDevice with 7 lines and 11 formulas renders **none of them** — not even `\(a^2+b^2=c^2\)` or `\(\alpha+\beta+\gamma=\pi\)`. On the same page, those same plain formulas render 3 of 3 when they run on their own.

**iDevice HTML:**

```
<p>Control: \(a^2 + b^2 = c^2\) y \(\frac{x}{y}\)</p>
<p>Un acento: \(\text{área} = \pi r^2\)</p>
<p>Dos acentos: \(\text{período mínimo}\)</p>
<p>Colateral: \(\text{ñ}\) y \(\text{ç}\) y \(\frac{a}{b}\)</p>
<p>Negrita: \(\textbf{Ámbito máximo}\)</p>
<p>Griego: \(\alpha + \beta + \gamma = \pi\)</p>
<p>El \(\text{área}\) de un círculo con \(\text{radio máximo}\) es \(A = \pi r^2\).</p>
```

| What is rendered | Result |
|---|---|
| Plain formulas alone | 3 of 3 render |
| One accented letter alone | Renders, letter falls back to serif (same as `main`) |
| Two different accented letters | Fails: `dynamic file 'latin' failed to load` |
| The real 7-line iDevice | 0 of 11 render |

The editor, where a single accented formula blanks every formula on the page while the author works; and exported packages in the maths iDevices that re-render on the fly (`mathproblems`, `mathematicaloperations`, `quick-questions-*`, `sort`, `word-search`, `interactive-video`) when they show feedback or move between questions.

**Two ways out:** ship the `latin*` groups as well (about 1.5 MB), or stop MathJax from ever asking for a group that is not there, by removing the excluded groups from its table in `common.js`.

## Why the tests did not catch it

`latex-rendering.spec.ts` checks `\mathbb{R} \subset \mathcal{L}`, and those groups **are** shipped. No test puts an accent inside `\text{}`, and the tests look at the first render of a page, which is the case that works. The missing guard: render two different accented letters after the page has loaded.

## Minor points

1. `\require{bbm}` and `bboldx`: the TeX extensions are shipped but their font files are not, so they hit the same 404.
2. ADR-2259-01 and -02 are still `Proposed` while being implemented; only -03 is `Accepted`.
3. ADR-2259-02 still describes the 5 speech locales and `trimSpeechLocaleMenu()`, which no longer exist.
4. `public/app/common/edicuatex/edicuatex-base-path.test.js:236` still uses the old 3.2.2 CDN URL as a fixture.
5. #2351 edits `edicuatex-tools.js` even though edicuatex was meant to be handled separately. That is a local change someone will have to carry across the next update.

## Checked and fine

- No overlap with #2260, which touched neither `exe_math/**` nor `package.json`.
- One MathJax tree in the repo, with a SHA-256 drift test and a guard against 3.2.2 leftovers.
- The Speech Rule Engine is fully gone, braille included; `a11y/assistive-mml` is kept.
- `package.json` reindented; glyph guard added to `ServerLatexPreRenderer.spec.ts`.
- `https://cdnjs.cloudflare.com/ajax/libs/mathjax/4.1.3/tex-svg.min.js` returns 200.
- No font request leaves the origin.

Thank you.

erseco:
Thanks @ignaciogros — the blocker is real, I reproduced it with your example, and it is fixed in `6c3a4026a`. All five minor points are done too.

### The blocker

Your diagnosis is exactly right, and confirmed in the bundle: `loadDynamicFile()` catches the first failure and marks the range `failed`; every later request for the same range **rejects**, which fails the whole typeset call.

Your 7-line iDevice, before and after:

```
before  {"containers": 0,  "rawDelimiters": 12}   dynamic file 'latin' failed to load
after   {"containers": 12, "rawDelimiters": 0}
```

Where my reasoning went wrong is worth recording. I picked the 13 ranges by measuring which code points MathJax 3.2.2's font could already render, and `latin` scored **zero overlap** — á, ñ, ü were never in the 3.2.2 font either. I concluded "not a regression". That was right about glyph coverage and wrong about the consequence: 3.2.2 degraded a missing glyph quietly forever, while v4 turns the second one into a hard failure of the entire render call.

### Which of your two ways out, and why

I took the second — stop MathJax asking — because **the first does not fix the class of bug**. Shipping `latin*` costs 1.5 MB and still leaves `greek`, `cyrillic`, `PUA` and the rest able to poison a call; anyone writing Cyrillic in `\text{}` would hit the identical failure.

One dead end worth passing on. My first attempt marked the ranges as already resolved, and **it hung the editor**: `getChar()` deletes one placeholder per call and re-renders the whole document to retry, so a resolved-but-empty range costs a full re-render *per accented letter*. Dropping the placeholders outright — your wording, "removing the excluded groups from its table" — is what works: `getChar()` finds nothing, takes the unknown-character path, and the glyph falls back to the CSS `unknownFamily` serif, which is what 3.2.2 did with those same characters.

### The guards you identified as missing

Both added, and both would have caught this:

- `latex-rendering.spec.ts` now renders your case — accented text, bold accented text and Greek alongside plain formulas — and asserts every formula rendered with no raw delimiters left. You were right about why the old tests missed it: they use `\mathbb` and `\mathcal`, whose ranges *are* vendored, and they only check the first render.
- `mathjax-packages.spec.ts` asserts `MATHJAX_VENDORED_FONT_RANGES` in `common.js` equals `VENDORED_FONT_RANGES` in the vendor script. Drift is a bug in both directions: listed but not vendored poisons the typeset; vendored but not listed drops glyphs that are sitting on disk.

### Minor points

1. **`\require{bbm}` / `bboldx`** — same failure through a different door, so the two TeX extensions are no longer vendored at all. Shipping the TeX half without the font extension is worse than shipping neither: it turns an "extension not found" on the console into a poisoned typeset.
2. **ADR-2259-01 and -02** are now `Accepted`.
3. **ADR-2259-02** — the consequences section is rewritten; it still described the speech locales and `trimSpeechLocaleMenu()` that ADR-2259-03 removed.
4. **`edicuatex-base-path.test.js:236`** — fixture updated to 4.1.3.
5. **`edicuatex-tools.js`** — solved at the source rather than justified. [edicuatex#7](https://github.com/edicuatex/edicuatex.github.io/pull/7) is merged upstream, so the file here is now **byte for byte** what their `main` has: `diff` is empty, and there is nothing left for anyone to carry across. The three things this branch was patching are theirs now, and their font-path handling is better than what it replaced — it keys off which copy of MathJax actually answered rather than guessing from `isInExe`. No release carries it yet (npm is on 1.5.4), so it is still a copy from git, but a copy with no edits, which is what the migration to consuming the package needs.

### Verification

`make fix` clean · vitest **15,373** · `make test-integration` **772** · `make test-e2e` (chromium, full) **544 passed, 11 skipped, 0 failed** · `make architecture-check` OK. `exe_math/` is 71 files, 4,480,505 B raw / 1,526,945 B zipped.

ignaciogros (approved):
Blocker and minor points checked!

Thank you, @erseco.

jjdeharo:
@erseco `edicuatex@1.5.5` is published, so the line about "no release carries it yet" is out of date — it carries #7, and the npm migration is unblocked.

@ignaciogros thank you for the review. It found a real bug in EdiCuaTeX too, in a milder form, and I would not have looked without it.

### The blocker does not reach us, checked rather than assumed

`scripts/vendor-mathjax.mjs` copies the whole font directory, so all 40 glyph ranges are there. I ran your 7-line iDevice against the editor:

| | |
|---|---|
| Plain formulas | render |
| One accented letter | renders |
| Two different accented letters | renders |
| ñ, ç, bold accented, Greek | render |
| Plain formula *after* all of the above | renders |

Zero failed requests, `latin.js` and `latin-b.js` served locally. No poisoning, because nothing fails in the first place.

### `\require{bbm}` was the same mistake here, and we took the other way out

Four extensions in `input/tex/extensions` carry glyphs of their own — `mhchem`, `bbm`, `bboldx`, `dsfont` — and only `mhchem`'s font was vendored. So `\require{bbm}` rendered an empty box and asked for a file that 404s, saying nothing on screen. `dsfont` I had not even noticed.

We vendored the three missing font halves instead of dropping the extensions. 396 KB on a 13 MB tree, and `\mathbbm{ABCxyz123}`, `\mathbb{ABCxyz}` and `\mathds{ABCNZ}` now render 9, 6 and 5 glyphs with no failed requests, and nothing after them is affected.

Your reasoning for the other route holds where you applied it: with 13 of 40 ranges shipped, adding `latin*` leaves `greek`, `cyrillic` and the rest able to poison a call. Here there is no remaining door — all 40 ranges and all four font extensions ship — so the feature working seemed better than the feature being absent, in an editor whose job is writing LaTeX. If a range or an extension is ever added upstream without its font, the same hole reopens, so there is now a note next to the list saying to grep the extensions for `font-extension` when it changes.

### One thing to pick up when you sync

Since 1.5.2 there is a `menus/vendor/` directory. `menus/editor.html` used to load Tailwind and SortableJS from CDNs — the second one unpinned, as `@latest` — and both are vendored now, so the menu editor works offline. If the copy here is taken file by file rather than as a tree, that directory needs adding, or that screen loses its styling and its drag and drop.

erseco:
@jjdeharo thanks — 1.5.5 is in, and your last point turned up a gap on our side that I would not have looked for.

### The migration is done: #2359

Stacked on this PR, pinned to 1.5.5. I changed the shape after opening it: the tree is **not committed** here. It is generated by `vendor:edicuatex` as the first step of `build:all` and gitignored, alongside `app.bundle.js`, the yjs bundles, `public/bundles/` and the compiled CSS — everything this repo already generates into `public/`. Electron packages `dist/static/**` rather than `public/**`, and the Dockerfile already asserts the generated artefacts exist, so no path loses the editor. `+306 / −11,912`.

Your reason for going the other way in your repo holds there and not here: a gitignored copy that no deployment ever has is a 404 on every load. Ours is regenerated by every build that serves or packages the editor, and the thing I care about is that there is no second copy left to disagree with the package.

### `menus/vendor/` — in, and asserted by name

Caught before your note, but only because the script copies `menus/` as a tree rather than file by file, which is exactly the failure you described. It is now explicit rather than incidental: `vendor-edicuatex.spec.ts` asserts `menus/vendor/tailwind.css` and `menus/vendor/Sortable.min.js` are in the plan, with your reason in the comment. Verified through to `dist/static/`.

### On `\require{bbm}`: your route is right for you, and I am keeping ours

You read the asymmetry correctly, and I would put it this way: you ship one deployment of about 13 MB, so 396 KB to make a feature work is obviously worth it. `exe_math/` here is copied into **every exported package that contains LaTeX**, so the same 347 KB for `bbm` and `bboldx` is paid by every teacher's SCORM and every EPUB, for two macro sets nobody has asked for. Dropping the TeX halves turns the failure into an ordinary "extension not found" on the console.

Worth noting we do ship `dsfont`'s font extension — it is in our enabled list, so it had to be there. It was `mhchem` and `dsfont` for us and `mhchem` alone for you; between the two reviews all four are now accounted for on both sides.

Your grep note next to the list is the right guard. I have the equivalent: `vendor-mathjax.spec.ts` fails if a range appears that is neither vendored nor in the excluded list with a reason.

### The gap your check surfaced

You verified the blocker does not reach the standalone editor because your tree has all 40 ranges. That is what made me look at the other case: **inside eXeLearning the editor loads *our* MathJax**, which has 13 of 40, and `edicuatex-tools.js` points `loader.paths.fonts` at it. So `\text{área}` in the formula preview asks for `latin`, gets a 404, and the second accented letter fails the render — `MathJax.typesetPromise([preview])` rejects and `edicuatex.js` reports "Syntax error" for a formula whose syntax is fine.

Our fix for this lives in `common.js`, which drops the placeholders of unvendored ranges so the character falls back to serif instead of failing. The iframe has its own `window.MathJax` and does not load `common.js` — the same shape as the `a11y/assistive-mml` gap you found, and I missed it for the same reason.

It cannot be fixed with a list on your side, because which ranges a host trimmed is the host's business. What would work is the host telling you: a second TinyMCE setting next to `edicuatex_mathjax_url` — say `edicuatex_mathjax_font_ranges` — and, when it is present, dropping the placeholders of everything not in it. Six lines next to the `mathjaxIsVendored` gate, and it would make the editor correct against any trimmed host rather than just ours. If that seems reasonable I will open it as a PR with the eXeLearning half ready to go; if you would rather not carry host-specific plumbing, say so and I will find another way. This is author-facing and self-announcing, so it is not urgent either way.

PR #2359, opened 2026-09-03
build(edicuatex): generate the editor from the pinned npm package

erseco:
> **Stacked on #2351.** The base of this PR is `feature/upgrade-to-mathjax-4`, so the diff here is only the EdiCuaTeX change. GitHub retargets it to `main` when #2351 merges.

## Why

EdiCuaTeX was copied into `public/app/common/edicuatex/` by hand, and that is exactly how it drifted. Within a single day the copy here and the upstream source disagreed about which MathJax components to load, and the fix that closed the gap — a missing `a11y/assistive-mml`, so the formula preview announced an unlabelled graphic to a screen reader — reached us through a code review on #2351 rather than a version bump. @jjdeharo published the editor to npm in response ([edicuatex#6](https://github.com/edicuatex/edicuatex.github.io/pull/6)), which makes the next one arrive the ordinary way.

## What changes

`edicuatex` is a pinned dependency at **1.5.5**, `scripts/vendor-edicuatex.ts` derives the tree from it, and `vendor:edicuatex` runs as the **first step of `build:all`**.

**The tree is no longer committed.** It joins `app.bundle.js`, the yjs bundles, `public/bundles/`, the i18n catalogues and the compiled CSS — everything this repo already generates into `public/` and gitignores. There is nothing left to drift, because there is no second copy to disagree with the package.

That is safe on every path that ships the editor, which is why it is worth doing rather than merely tidy:

- **Electron** packages `dist/static/**`, not `public/**`, so it gets the editor through the static build — no `build.files` entry to forget.
- **The Dockerfile** already runs `build:all` and asserts the generated artefacts exist afterwards; the editor's `index.html` joins that list.
- `make up-local` depends on `bundle`, and the CI and E2E workflows run `build:all`.

Verified from a clean state: deleting the directory and running `build:all` regenerates it, and `make build-static` carries it into `dist/static/`.

`make vendor-edicuatex` and `--check` remain for regenerating and for verifying a build, and `vendor-edicuatex.spec.ts` asserts the two agree — so a stale tree from an interrupted build is visible rather than silent.

The paths do not move, so `tinymce_5_settings.js` is untouched.

## What 1.5.0 → 1.5.5 brings

None of it ours to review, all of it upstream: fullscreen and settings controls, a reworked menu, updated translations for the six languages, and **Tailwind and SortableJS served from the package instead of a CDN** — the same offline argument as the MathJax font ranges in #2351.

The version this branch had already been aligned to by hand, `js/edicuatex-tools.js`, is byte-identical to 1.5.5's, so the MathJax-facing behaviour does not change here at all.

## Details worth a second's attention

- **Not vendored**: `package.json`, `scripts/` and `tailwind.config.js`, which build the package rather than run in it, and the two READMEs, which document the standalone project. Their leaving takes two entries out of the static pruner's list, which is a loud-failure list — every path in it must exist.
- **`LICENSE.txt` is vendored.** An editor shipped without its licence is a distribution problem, and the pruner refuses to remove it anyway.
- **`edicuatex-base-path.test.js` moves up one directory.** It is ours, not upstream's, and inside the vendored tree the drift check would report it as an extra file.

## Verification

| Gate | Result |
|---|---|
| `make fix` | clean |
| Frontend (vitest) | **15,373 pass**, 289 files |
| Backend (`bun test ./src ./scripts`) | 8,356 pass; the 2 known `Assets API v1` failures, which pass in isolation |
| `make test-integration` | **772 pass**, 0 fail |
| `make build-static` | builds, and the editor lands in `dist/static/app/common/edicuatex/` |
| `make test-e2e` (chromium, full) | **544 passed**, 11 skipped, 0 failed (17.7 min) |
| `bun run scripts/vendor-edicuatex.ts --check` | in sync, 30 files |


erseco:
@jjdeharo this is the migration we talked about on #2351, now that `edicuatex@1.5.5` is out — pinned to it, and the vendored copy replaced by a generated one so there is nothing left that can drift from the package.

Two things you flagged, both handled:

- **`menus/vendor/`** is in. The script copies `menus/` as a tree rather than file by file, which is the failure mode you warned about, and `vendor-edicuatex.spec.ts` now asserts `tailwind.css` and `Sortable.min.js` by name so it stays that way. Verified through to `dist/static/`.
- **`js/edicuatex-tools.js`** is byte-identical to 1.5.5's, so nothing about MathJax behaviour changes in this PR — #2351 had already been aligned to your `main` by hand.

One difference from what you chose in your own repo, and I think both are right in their place: the tree is **not committed** here. It is generated by `vendor:edicuatex` as the first step of `build:all` and gitignored, next to `app.bundle.js` and the other generated trees. Your reason for un-gitignoring `js/mathjax/` — a copy no deployment ever has is a 404 on every load — does not apply, because every path that serves or packages the editor here runs a build: Electron packages `dist/static/**`, and the Dockerfile already asserts the generated artefacts exist.

Also mentioning the open question from #2351 here so it does not get lost: inside eXeLearning the editor loads *our* MathJax, which ships 13 of 40 glyph ranges, so an accented letter in the formula preview still fails there. Not something this PR changes, and not fixable from a list on your side — the proposal is a second TinyMCE setting telling the editor which ranges the host actually has. Happy to open it with our half ready.

@ignaciogros no rush on this one — it sits on top of #2351 and only makes sense after it.

erseco:
Full chromium E2E on `0025bb981`: **544 passed, 11 skipped, 0 failed** (17.7 min). Verified from a clean state as well — deleting `public/app/common/edicuatex/` and running `build:all` regenerates it, and `make build-static` carries it into `dist/static/`, `menus/vendor/` included.

ignaciogros (approved):
Thanks for this improvement. Pulling external libraries in this way is important for maintaining the project.

The code looks good, but I think there was some inconsistency between the statements and what the PR actually does. Could you please review the changes I just pushed? If everything looks good to you, I think the changes can be merged into `main`. Please check the `Dockerfile` too.

Thank you.

PR #2362, opened 2026-09-05
Prepare v4.0.4 release (and enable i18n for the equation editor)

### LINKED ISSUE #1922: Typo errors in EU translation
See @josemiguelandonegi-max's comment and the attached `.zip` file in #1911:

> The literal text was detected reviewing some translations. There are some related texts with little typo errors

### LINKED ISSUE #2284: Prepare v4.0.4
- See https://github.com/exelearning/exelearning/pull/545#issuecomment-5509111057
- Update style documentation (use of `setUrlParam` in `public/app/common/exe_export.js`).
- Update style designer (separate tool), including the default styles and contents.
- Update changelog.
- Update third-party packages list. 
- Review and complete Spanish (ES) translation.
- Provide automated placeholder translations for new strings in incomplete translations.

ignaciogros:
**This PR:**

- Closes #1922 - Typo errors in EU translation.
- Restored the French (FR) translation, recovering the strings lost during the migration from version 3 and completing the rest with automated translations pending review
- German (DE): fixed misaligned translations that made True and False display unrelated texts and mismatched the delimiter labels in the math editor.
- Fixed wrong icon names. Every style (except Universal) should ship the same 50 icons under the same names.
- Updated styles documentation.
- `make update-licenses` now retrieves copyright holders that cannot be derived from package metadata.
- Updated THIRD PARTY CODE attribution list.
- Updated CHANGELOG.
- Guard translation extraction against missing generated source trees, and update development documentation.
- Translation maintenance cycle: remove obsolete keys.
- Translation maintenance cycle: extract new keys from source code into all XLF files.
- Reviewed and completed the Spanish (ES) translation.
- Automatic placeholder translations for new strings in incomplete translations.
- Improve the skill to fill empty `<target>` elements in translations/messages.*.xlf files.

## Enabling i18n for the equation editor

Since PR #2359, the editor is generated into a gitignored tree by the build, and its interface strings are only there. Extraction scans the working tree, so a checkout that had not been built produced a short key set with nothing looking wrong, and `--remove-obsolete` would delete those strings from every locale.

Three changes close that:

- `make translations`, `translations-cleanup` and `translations-sort` rebuild the vendored tree before scanning.
- The CLI warns when a registered generated tree is missing, and refuses `--remove-obsolete` outright. `--allow-missing-generated` is the explicit, destructive override.
- `GENERATED_SOURCE_DIRS` (`src/cli/commands/translations.ts`) is the registry both read; `doc/development/internationalization.md` documents what to do when a new library is vendored.

Now the editor's strings are extracted and translated in the same cycle as the rest.

## Review request

**This part deserves a specific review, because it falls outside the original scope of the issue, which was preparing v4.0.4.**

The reason is that preparing the release means closing the translation work, and that could not be done while extraction was able to silently lose so many strings per locale. It was a prerequisite, not an addition.

A separate PR will follow with two proposed ADRs: how vendored libraries enter the tree, and where their translatable strings live.


public/CHANGELOG.md
```diff
@@ -1,13 +1,101 @@
...
+- Collaboration: closed connections are now released, preventing servers from accumulating them
+- German (DE): fixed misaligned translations that made True and False display unrelated texts and mismatched the delimiter labels in the math editor
+- Basque (EU): fixed typos and wording in the interface, including a misspelling in the "Made with eXeLearning" link label and the wording of the image optimiser columns
+
+### Upgraded
+
+- fast-xml-parser: 5.4.1 → 5.11.0
+- fast-xml-builder: 1.0.0 → 1.3.1
+- mathjax: 3.2.2 → 4.1.3
+- edicuatex: 1.5.0 → 1.5.5
```
erseco: The previous edicuatex version was 1.4.1

erseco (changes requested):
Thanks @ignaciogros for the work here. I went through the translation changes, the generated catalogues, and the release-related tooling. I think this needs another pass before merging because there are several correctness issues that would ship in v4.0.4.

## Blocking issues

### 1. The restored French catalogue contains misaligned translations

A significant group of restored French targets appears to have been matched by position rather than by source string. These targets are especially problematic because they are not marked with `~`, so they look manually reviewed.

Examples include:

* `Add an activity` → `Répéter l'activité`
* `First activity` → `Répéter l'activité`
* `Next activity` → `Répéter l'activité`
* `Last activity` → `Répéter l'activité`
* `Previous activity` → `activité discursive`
* `Map marker` → `Marcador de mapa`
* `Go to the activity` → `Ajouter une autre activité`

This needs to be fixed at the XLF source before regenerating anything derived from the catalogue.

### 2. Several French translations drop required `%s` placeholders

For example:

* `No. of activities: %s` → `Nombre d'essais`
* `Completed: %s` → `compétition`
* `Average score: %s` → `Sauvegarde du résultat`
* `Average score of completed activities: %s` → `Non complété`

There is at least one additional case in the same group.

These are runtime formatting parameters, so dropping them changes the rendered output rather than just producing an inaccurate translation.

### 3. `trueFalseMessages.generated.ts` is out of sync with `translations/`

`build_and_test` correctly catches this with:

```text
(fail) committed catalogue > should still match `translations/`
       (regenerate with `bun run generate:truefalse-messages`)
```

This is not just a test-maintenance issue. The committed generated catalogue is used by the legacy True/False importer.

At the moment:

* the new `fr` locale is missing entirely;
* the generated German `msgTrue` / `msgFalse` values still contain the old misaligned translations even though this PR fixes the XLF.

Please do **not** regenerate this file until the French catalogue is fixed. With the current XLF, regeneration would also bake incorrect values such as:

```text
Incomplete activity → Choisissez une activité
Weight              → Gras
```

into the generated runtime catalogue.

Once the XLF is corrected, regenerate and format it with:

```bash
bun run generate:truefalse-messages
bunx biome check --write src/shared/import/legacy-handlers/trueFalseMessages.generated.ts
```

### 4. The Basque placeholder errors from #1922 are still present

The PR says it closes #1922, but two malformed printf translations remain unchanged:

```xml
<source>Are you sure you want to delete %s projects?</source>
<target>Ziur zaude %a proiektu ezabatu nahi duzula?</target>
```

and:

```xml
<source>You made %s errors. You have %s attempts left. Do you want to try again?</source>
<target>% errore izan dituzu. % saiakera gelditzen zaizkizu. Saiatu nahi duzu berriro?</target>
```

Both should preserve the `%s` placeholders from the source before considering #1922 closed.

### 5. The generated third-party attribution list is incomplete

`public/libs/README.md` drops `@codecov/bundle-analyzer`, although it is still declared as a dev dependency.

Regenerating with a complete dependency tree restores it.

More importantly, `update-licenses` currently warns and skips a package when its metadata cannot be read, but still writes the output and exits successfully. That means a partial `node_modules` can silently produce an incomplete attribution file.

For release/legal metadata I think generation should fail, or there should at least be a CI drift check that prevents an incomplete generated file from being committed.

## Other correctness issues

### Translation tooling

`.agents/skills/xlf-translate/SKILL.md` still defines `all` as the ten locales listed there, and French is not one of them. The restored `messages.fr.xlf` still contains empty targets, so future runs of the documented workflow will skip that locale.

`findMissingGeneratedSources()` also only checks whether each generated directory exists. A partially generated `public/app/common/edicuatex/` therefore passes the guard. With `--remove-obsolete`, an interrupted vendor operation can make the missing EdiCuaTeX strings look genuinely obsolete. `vendor-edicuatex.ts --check` already has stronger knowledge of what a complete generated tree looks like.

### `update-licenses`

There are two smaller correctness issues worth fixing while this code is being changed:

1. When a `BOILERPLATE_COPYRIGHT` match is rejected, the code gives up on that regex instead of continuing to look for a later valid match.
2. `COPYRIGHT_OVERRIDES[packageName]` is read from a normal object with an arbitrary package name. Names such as `constructor` or `toString` can resolve inherited `Object.prototype` members. `Object.hasOwn()`, a null-prototype object, or a `Map` would avoid this.

### Documentation

The icon-color fallback described in `doc/development/styles.md` omits `THEME_ICON_COLOR_MAP`, so the documented chain to the final `#6E9F41` fallback is not the chain used by the included themes.

The CHANGELOG also says:

```text
edicuatex: 1.5.0 → 1.5.5
```

but v4.0.3 did not ship EdiCuaTeX 1.5.0. Its vendored editor came from the earlier 1.4.x/v1.4.1 sync; 1.5.0 was packaged later. The release notes should use the version/snapshot actually shipped by v4.0.3 as the baseline.

## Minor test-quality issues

These are not blockers by themselves:

* `update-licenses.spec.ts` duplicates the complete `COPYRIGHT_OVERRIDES` constant in an equality assertion, making the test primarily a change detector for the implementation rather than behavior.
* `translations-sort.spec.ts` restores `console.warn` / `console.log` after awaited calls without `try/finally`, so a rejection can leak the stubs into subsequent tests.

---

The main reasons for requesting changes are the corrupted French catalogue, lost placeholders, stale generated True/False catalogue, unresolved Basque placeholder errors, and incomplete third-party attribution generation. The smaller items can be handled separately if needed, but I don't think the current state is safe for a release-preparation merge.

ignaciogros:
Thanks a lot for the review and for reporting these issues. Many of the French translation errors came from v3.0.2. Without your report, the translation would have been in a bad state.

I think all of them have been fixed. 

**Changes:**

- Italian (IT) translation: restore missing `%s` placeholders.
- Portuguese (PT) translation: fix a misaligned translation.
- Basque (EU) translation: restore the `%s` placeholders reported in #1922.
- Translation skill: add French to the available languages list.
- Improve i18n documentation.
- Improve styles documentation: fallback tint for Material ("General") icons.
- Improve the Update Licenses Command, and add `make update-licenses-check` (it fails when the attribution list has drifted).
- Complete the generated third-party attribution list.
- Improve translation-related scripts.
- Fully review the French (FR) translation; every correction carries the `~` marker.
- Sync `trueFalseMessages.generated.ts` with `translations/`.
- Update the v4.0.4 CHANGELOG.

**Additional comments:**

- **The CI drift check for `update-licenses` is the one thing left out.** The command now refuses to write an incomplete attribution file, and `--check` / `make update-licenses-check` report drift without writing, so the check itself exists. Wiring it into the workflow is not in this PR: `ci.yml` fails any pull request that modifies `ci.yml` or `app.yml`, so that step has to land as a separate change.
- **Why an unreadable package keeps its recorded entry instead of being re-derived:** neither `package.json` nor `bun.lock` carries a license or a copyright holder (that data exists only inside the installed package). A package whose `os` excludes the platform (`@codecov/bundle-analyzer` on Windows) is never installed there, so the entry already in `public/libs/README.md` is the only source on that machine. CI installs it and regenerates from real metadata, so `--check` still verifies the value.
- Auditing the French catalogue for the problems you found turned up the same class of defect in two locales you did not mention: four dropped placeholders across `it` and `pt`. They are fixed here too.
- On `THEME_ICON_COLOR_MAP` itself, beyond documenting it: all seven of its values are already declared by the themes in CSS (`--exe-icon-color`, or `--exe-icon-picker-color` for `educablue`), and the resolution chain reads the CSS first, so the table duplicates data the styles already own. I would rather styles were fully self-describing here, but removing it means superseding ADR-1247-04 and re-checking the seven bundled themes by hand, so it will get its own PR rather than widen this one.

Please could you do a second review?

erseco (changes requested):
Thanks @ignaciogros, this is a big improvement. I went through the new head (`9721db4`) with the same scope as the first pass: the French catalogue, the generated catalogues, and the release tooling.

The five blocking items from the previous review are resolved: the misaligned French strings I listed are corrected and marked with `~`, the `%s` placeholders are back in FR/IT/PT/EU, `trueFalseMessages.generated.ts` matches `translations/` and its test passes, `@codecov/bundle-analyzer` is back in `public/libs/README.md`, and `--check` exists. CLI tests and Biome are green, the XLF files are in sync with the code (extract 0, cleanup 0, sort/format no-op), and `dist/cli.js` bundles correctly.

I still found a few things that I think should not ship in a release-preparation merge, plus some smaller items.

## Blocking issues

### 1. Another block of misaligned French translations

Around `translations/messages.fr.xlf:852-920` there is a second group of about 17 units whose targets belong to neighbouring sources. It is the same class of defect as the first review, this time in the AI-prompt and EdiCuaTeX area. All of them carry `~`, so they are shown to the user as-is:

* `The questions could not be generated. Incorrect format` → `Une erreur est survenue lors de la récupération des questions…`
* `Generate 10 words followed by their definitions…` → `Générez 30 mots et leurs définitions…`
* `Provide 4 elements for each of these groups…` → `Carnivores#Lion#Tiger#Eagle#Shark`
* `Generate 10 true or false questions…` → `0#0#0#0`
* `EdiCuaTeX - LaTeX Equation Editor` → `0. False#Water boils at 90°C at standard pressure…`
* `Application created by Juan José de Haro` → `0. False#Mount Everest is the tallest mountain in Africa…`
* `Recent` → `0. False#The Earth is the third planet from Mars…`

Each target is the translation of the *next* source in the file, so it looks like the same off-by-one that hit the earlier block. The same region is shifted in `de` and `pt` on `main` already (not introduced here, I can open a separate issue), but `fr` is new in this PR.

### 2. Renaming theme icons without an alias breaks projects that already use them

The PR renames `neo/icons/objetives.png` → `objectives.png` and `educablue/icons/think-alt.svg` → `think_alt.svg`. Both names shipped in v4.0.0 through v4.0.3, so any saved project that picked that icon stores `objetives` / `think-alt` as its icon name.

There is no alias anywhere in the resolution chain: `deriveBlockIcon` (`src/shared/block-icon.ts` and its JS twin `blockIconRuntime.js`), `LEGACY_ICON_MAP` in `blockNode.js`, and `setThemeIconFiles()` in `IdeviceRenderer.ts` all pass the raw name through. `resolveIconName()` then misses the map and falls back to `${baseName}.png`, so the exporter emits `theme/icons/objetives.png`, which 404s in preview, HTML5/SCORM/EPUB exports and the workarea.

A small alias table (`{ objetives: 'objectives', 'think-alt': 'think_alt' }`) in the shared derivation point (TS + JS twin) or in `setThemeIconFiles()` keeps old projects working.

### 3. `inspectEdicuatexTree` only looks at `drift.missing`

`src/cli/commands/translations.ts:81` treats the tree as complete when nothing is *missing*, and ignores `drift.changed` / `drift.extra`. A stale vendored tree that still has every file therefore passes as "in sync with the pinned edicuatex package", and `--remove-obsolete` proceeds. The `--help` text, the doc and the registry comment all say the guard protects against exactly this case.

Reproduced: remove `'Huger text'` from `public/app/common/edicuatex/lang/en.js`, `vendor-edicuatex.ts --check` reports `changed lang/en.js`, `findUntrustedGeneratedSources()` returns `[]`, and `bun cli translations --clean-only --remove-obsolete --locale=fr` deletes the `Huger text` trans-unit without any warning. The realistic path is an edicuatex bump plus `bun install` without `make vendor-edicuatex`, then a direct CLI call.

`drift.missing.length + drift.changed.length === 0` (or simply reusing the exit status of `--check`) closes it.

### 4. A partial `node_modules/edicuatex` aborts every translations command

`inspectEdicuatexTree` only guards `existsSync(packageRoot)`; `buildVendorPlan` then runs `readdirSync` on `css/icons/js/lang/menus` unconditionally. With a package directory that only has `index.html`, `LICENSE.txt` and `lang/` (an interrupted install), `findUntrustedGeneratedSources()` throws `ENOENT: scandir …/node_modules/edicuatex/css`, and `bun cli translations --extract-only` and `translations:sort`, which are non-destructive and did not depend on `node_modules` before this PR, exit 1 from the generic catch in `src/cli/index.ts` without mentioning `make deps` / `make vendor-edicuatex`. The spec avoids it by creating all five directories.

The `inspect` contract already documents "returns null when the check cannot be made", so wrapping `buildVendorPlan` in a try/catch and returning `null` (or an `incomplete` problem) is enough.

## Other correctness issues

### `update-licenses`

**The boilerplate guard misses the Apache-2.0 and MIT templates.** `BOILERPLATE_COPYRIGHT` (line 138) only rejects captures that start with `owner`/`holder`/`notice`/`[`. Running the extractor over the real LICENSE files in `node_modules`:

```text
pdfjs-dist, @material-symbols/svg-400, @mathjax/src
  → "License. Subject to the terms and conditions of"   (from "2. Grant of Copyright License. Subject to…")
minimist
  → "HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER"   (holder\b does not match HOLDERS)
```

Today this is masked because those packages are exactly the `COPYRIGHT_OVERRIDES` set. The next Apache-2.0 dependency without an `author` field (common for Google/Mozilla/Apache packages) will get that text as its copyright holder, and `--check` will pass because generation and check agree. The new tests only use 2–5 line excerpts that omit section 2. Adding `license` to the guard, or requiring the capture to start with `Copyright`/`(c)`/`©`, fixes it.

**The year pattern requires whitespace right after the year**, so the most common BSD/MIT form `Copyright (c) YYYY, Holder` falls through to the generic `/Copyright\s+(.+)/` and keeps the year. This is already in the generated output:

```text
public/libs/README.md:90    Copyright: 2015, Scott Motte     (dotenv)
```

and `--check` now freezes it as correct. Same root cause for `2013, 2014, 2015 Foo`, `2023. Foo`, and `2019 – present Foo` (the en-dash is not in `[,-]`). Allowing `[,.]?` after the year and the en-dash in the range covers these.

**On the carry-over from the recorded README:** I follow the reasoning, a package excluded by `os` is never installed on Windows, so the README is the only local source there. My concern is that it makes `--check` self-referential on that platform (a hand-edited or stale entry passes forever and gets re-emitted), and it costs ~80 lines (`parseRecordedPackages` plus the two branches) for one package. Since the PR already introduces `COPYRIGHT_OVERRIDES`, extending it to `{ copyright, license }` for platform-excluded packages gives identical output on every OS, is reviewable in the diff, and removes the reverse parser of our own output. Not a blocker, but I think it is the simpler design.

### CHANGELOG baseline

The "Upgraded" section uses the last dependabot bump as the "from" version rather than what v4.0.3 shipped. Checking `bun.lock` at the `v4.0.3` tag:

| Package | CHANGELOG says | v4.0.3 actually shipped |
|---|---|---|
| mermaid | 11.17.0 | 11.12.3 |
| jose | 6.2.9 | 6.1.3 |
| sass | 1.103.0 | 1.97.3 |
| electron | 43.4.1 | 43.2.0 |
| @biomejs/biome | 2.5.9 | 2.4.5 |
| @xmldom/xmldom | 0.9.11 | 0.9.10 |
| @material-symbols/svg-400 | 0.40.2 → 0.47.0 | not present (new dependency, added after the tag) |
| fast-xml-builder | 1.0.0 → 1.3.1 | transitive, not in `package.json` |

It is the same issue that was corrected for edicuatex (1.5.0 → 1.4.1), applied to the rest of the list. `git diff v4.0.3 -- bun.lock` gives the right baselines.

### The drift check can run in CI without touching `ci.yml`

`make update-licenses-check` is currently not executed by anything (no workflow, no Dockerfile). I understand `ci.yml` cannot be modified from a PR, but it does not need to be: `ci.yml:71` already runs `make lint`, and `lint:` already aggregates `architecture-check` for exactly this purpose. Adding the target to that line is enough:

```make
lint: check-bun lint-ts lint-js lint-tests architecture-check update-licenses-check
```

`node_modules` is installed in that job, and `Dockerfile:35` already chains `vendor-edicuatex.ts --check` the same way. Without it, the Makefile comment ("drift is a build failure, not a warning") describes something that does not happen yet.

### One empty French target reaches the static build as `""`

`translations/messages.fr.xlf:898` (`Provide only one ordered list of steps or items, each separated by '#'.`) has a target of exactly `~`. The server path (`translation.ts`) falls back to the source when the stripped target is empty, but `parseXlfContent` in `build-static-bundle.ts` keeps the key with `""`, and `getContentTranslation()` in `locale.js` returns whatever is in the catalogue, so the AI prompt example shows up empty in the static build in French. It is the only new empty `~` target in the PR (de and pt each have one on `main`).

## Minor

* **Coverage:** the single uncovered line codecov reports is `translations.ts:82` (`return { complete: true, … }`). It is testable: the "incomplete" fixture already sets up `node_modules/edicuatex`, writing the same files into the generated tree exercises the complete branch. Per the Definition of Done, either cover it or justify it in the description.
* **`iDevice` translated as a common noun in FR:** `Importation d'outils pédagogiques` (4206), `À propos de cet outil pédagogique` (9378), `…premier outil pédagogique de texte libre…` (11842), and `appareils iDevices` (4214), while the other 58 occurrences keep `iDevice`. The skill this PR extends says not to translate proper nouns, and two of these do not carry `~`, so they read as reviewed. The menu ends up showing `Importation d'outils pédagogiques` next to `Importer l'iDevice au format ZIP`.
* **`SKILL.md` Step 1.5:** the terminology-anchoring loop at line 81 iterates `ca va de eo eu gl it pt ro` (no `fr`, no `es`) while line 47 now says "all eleven languages" and `$langs` at line 272 includes `fr`; line 206 still says "27 strings × 10 languages". An agent following the snippet literally never consults the existing FR targets and re-translates French from scratch against a catalogue with 3,002 `~` entries pending review, which is exactly the synonym drift the new section is meant to prevent.
* **`translations.ts` header comment** describes `--allow-missing-generated` as "Extract even though a generated source tree is missing", but extraction is never blocked; the flag only unlocks `--remove-obsolete` (the `printHelp` text is correct).
* **Duplicated warning block:** `const headline = kind === 'missing' ? 'missing' : 'incomplete'` (line 541) is an identity ternary over `kind`, and the three-`warning()` block is hand-copied in `translations-sort.ts:186` with a clause that already differs. A shared `warnUntrustedGeneratedSources(problems, suffix?)` exported next to `findUntrustedGeneratedSources` keeps both commands in step.

## On `THEME_ICON_COLOR_MAP`

Agreed, superseding ADR-1247-04 and re-checking the seven bundled themes belongs in its own PR. Documenting the current chain here is fine.

---

The reasons for requesting changes again are the second misaligned French block, the icon rename without an alias (it breaks existing projects), and the two `inspectEdicuatexTree` issues. The `update-licenses` regex fixes and the CHANGELOG baselines are quick to address in the same PR; the rest can be split off if you prefer.

ignaciogros:
Thanks again for the review. I’ve made the following changes. **Could you please take another look?**

**Changes:**

- Rename an icon a release already shipped: old name → current name.
- Third party code list format revision.
- CHANGELOG: "Upgraded" and "Fixed" sections revision.
- Fixed misaligned translations that showed unrelated texts in Catalan (CA), German (DE), French (FR), Esperanto (EO), Galician (GL), Italian (IT), Portuguese (PT), Romanian (RO) and Valencian (VA).
- Review of the `xlf-translate` skill. Generalized text for easier maintenance.
- Update Licenses Command: fix copyright extraction and run the consistency check in `make lint`.
- Translations Command: detect stale vendored trees and share the warning.

erseco (changes requested):
Thanks @ignaciogros, this round closes almost everything. I went through the 18 commits since `9721db4` (new head `e5358c6`) and re-ran the same checks as before.

## What is fixed

All 14 points from the previous review are addressed, and the fixes are the right ones:

* **French catalogue:** the second shifted block (852–920) is restored by source string, the bare `~` target is filled, and `iDevice` is a proper noun again in all six units. The same block is also restored in `de` and `pt`, which covers most of #2366.
* **Icon rename:** `RENAMED_THEME_ICONS` / `resolveRenamedThemeIcon` is in the shared derivation point (`block-icon.ts` + `blockIconRuntime.js`) and is applied on the picker, bridge and export paths, each with its own test.
* **`inspectEdicuatexTree`:** stale files (`drift.changed`) are now detected, the "refuse to remove obsolete when stale" test pins it, and the previously uncovered line is covered.
* **`update-licenses`:** the Apache section 2/4d and MIT `HOLDERS` prose are rejected, `Copyright (c) 2015, Scott Motte` now yields `Scott Motte`, the README carry-over is gone in favour of `COPYRIGHT_OVERRIDES` with `{ copyright, license }`, and `make lint` runs `update-licenses-check`.
* **CHANGELOG:** all six "Upgraded" baselines now match `bun.lock` at `v4.0.3`, and the two entries that were not upgrades are gone.
* **Skill and CLI docs:** `$langs` is derived from disk, the "ten languages" remnants are gone, the header comment is right, and the warning block is shared via `warnUntrustedGeneratedSources()`.

Verification on `e5358c6`: `bun test` on `src/cli`, `block-icon`, `IdeviceRenderer` and the True/False generator: 742 pass / 0 fail. Vitest on the three touched icon files: 782 pass / 0 fail. `update-licenses --check`, `vendor-edicuatex --check`, XLF extract/cleanup/sort/format and `generate:truefalse-messages` + Biome all leave the tree clean. The 73 changed XLF targets all translate their own source, no `id`/`resname`/`source` changed, no placeholder lost.

## Blocking

Three of the fixes leave a gap in their own path, and one is a regression. All four are small.

### 1. The `catch {}` in `inspectEdicuatexTree` disables the guard on the destructive path

`translations.ts:90` returns `null` on *any* exception from `detectDrift(buildVendorPlan())`, and `findUntrustedGeneratedSources()` treats `null` as "nothing to report". The comment justifies it to keep `--extract-only` and `translations:sort` working, but a problem only *warns* on those two commands (`translations.ts:576-590`, `translations-sort.ts:190`); the only thing a problem blocks is `--remove-obsolete`.

So the case blocking issue 4 was about is now silent: an interrupted `bun install` leaves `node_modules/edicuatex` without `css/`, the vendored tree is the previous version, and `translations --clean-only --remove-obsolete` no longer warns or refuses. It scans the old `lang/en.js` and deletes every trans-unit that only the newer package carries from all 12 catalogues. At `9721db4` the exception aborted before touching any XLF.

Returning `{ complete: false, detail: 'package tree unreadable' }` from the catch keeps extract/sort working (warning only) and still blocks the deletion. Narrowing the catch to `ENOENT` under `packageRoot` would also be fine.

### 2. The workarea legacy fallback still indexes with the raw name

`blockNode.js:261`: the theme lookup uses `derived.value` (mapped), but `LEGACY_ICON_MAP[legacy]` still uses the raw name. A neo project saved in v4.0.3 with `<iconName>objetives</iconName>`, opened and switched to a style that does not ship `objectives.*` (`universal`), ends up with no icon, while the same block stored as `objectives` falls back to the Material `target`. `LEGACY_ICON_MAP[derived.value]` fixes it.

### 3. The exporter `.png` fallback still uses the old name

`IdeviceRenderer.ts:91` consults the rename table only against `iconResolutionMap`; the `${baseName}.png` fallback keeps the old spelling. `PrintPreviewExporter` and `Html5Exporter` swallow a failed `fetchTheme` and continue with server-absolute paths, so with an empty map a block with `objetives` still emits `theme/icons/objetives.png` (404) although `objectives.png` exists. The spec only exercises the `iconName` route, which `deriveBlockIcon` has already mapped.

```ts
const name = resolveRenamedThemeIcon(baseName);
return this.iconResolutionMap.get(name) ?? `${name}.png`;
```

### 4. Regression in the copyright extractor for multi-line notices

`yearSeparator` went from `\s+` to `[^\S\n]+`. I follow the reason (the unfilled MIT template `Copyright (c) 2020 ` followed by a blank line), but a notice whose holder sits on the next line no longer matches any year pattern; it then falls to the year-less pattern, the year gets stripped and `\p{L}` rejects it. Probed against `node_modules` at this head:

```text
yjs/LICENSE                      "Copyright (c) 2023\n  - Kevin Jahns"   → null   (was "- Kevin Jahns" at 9721db4)
@isaacs/balanced-match           "Original code Copyright Julian Gruber"  → null   (was "Julian Gruber")
synthetic                        "Copyright (c) 2015-present\nAcme Corp" → "2015-present"
```

Both real cases are masked today because the packages have `author` in `package.json`; the next dependency without it gets `Unknown` or a bare year range in a legal file, and `--check` freezes it. Allowing at most one line break after the year (`(?:[^\S\n]*\n)?` followed by a non-blank line) and rejecting captures that are only a year/range after cleanup keeps the MIT-template fix and restores these.

## Non-blocking

* **Remaining extractor heuristics.** Only the year-less pattern is line-anchored; the three year patterns match mid-line, so `   Portions Copyright 2017 Google Inc.` (`@playwright/test`) and quoted third-party notices in blockquotes (`> Copyright (c) 2011 Einar Otto Stangvik`, vite/vitest) are captured; `Holder until <url>` (`fast-uri`) keeps `until`; `{yyyy} {name of copyright owner}` (`mhchemparser`) passes the guard because it only knows `[yyyy]`. Anchoring all four patterns with `^[^\S\n]*` (flag `m`) unifies the two guards. All masked by `author` today.
* **Override licence is not cross-checked when the package is installed.** `AttributionOverride.license` only applies when the package is absent. If dependabot ever bumps `@codecov/bundle-analyzer` to a relicensed version, CI (ubuntu) regenerates with the new licence while `make lint` on Windows generates from the override and fails on a drift CI cannot reproduce. A one-line assert that `pkg.license === override.license` when both exist closes it.
* **Unbounded re-render in collaboration.** `YjsProjectBridge.js:1540` compares the raw `blockData.iconName` (`objetives`) with the already-normalised `blockNode.iconName` (`objectives`), so every remote update of that block re-runs `makeIconNameElement()` until someone re-saves it. Idempotent, but unbounded. Compare `normalizeBlockIcon(undefined, blockData.iconName).value` or drop the raw-name clause, since both sides are normalised.
* **Third live copy of `RENAMED_THEME_ICONS`.** `blockNode.js:106` calls it the "degraded path", but `window.eXeBlockIconRuntime` does not exist when `app.bundle.js` is evaluated (the runtime lands in group 4 of `yjs-loader.js`), so the local copy is the production path in the workarea. Nothing asserts the three tables are equal. Either `require('../../../../common/blockIconRuntime.js')` as `YjsProjectBridge.js:3` does (plus `KEEP_LOOSE_APP_FILES` in `prune-dist.ts`), or a spec in the style of `mathjax-packages.spec.ts:55-63`.
* **`think-alt` notes contradict each other after `60ab111`.** `styles.md:108`, `blockIconRuntime.test.js:82` and `block-icon.spec.ts:40` say educablue shipped the hyphenated name in v4.0.0–v4.0.3; `block-icon.ts:46-48` and `blockIconRuntime.js:129` say educablue arrived after v4.0.3. `git ls-tree v4.0.3` confirms the latter (only `base/` and `users/`). Worth aligning the three sentences with `block-icon.ts`.
* **Strict compare with a missing `value`.** `blockNode.js:237` and `YjsProjectBridge.js:11`: a descriptor without `value` normalises to `name: ''`. I found no current producer that omits `value`, so low confidence; `(iconData.value || '')` makes the intent explicit.
* **Nits:** `carriedOver` and the "(N carried over)" message (`update-licenses.ts:471/517`) still name the removed mechanism; Biome reports two `noUselessStringRaw` infos at `update-licenses.ts:196/204`; the *apply* blocks in `SKILL.md` still show `C:\...\` paths (pre-existing).

Out of scope for this PR but adjacent to the corrected `eo` hunk: `No decimals` → `~Respondo (opcia)` at `messages.eo.xlf:7546`. I will add it to #2366.

---

Requesting changes only for the four items above; each is a few lines. Happy to approve on the next push.

ignaciogros:
**Done:**

- Block a destructive translation run when the pinned package is unreadable.
- Map the renamed icon on the workarea legacy fallback.
- Map the renamed icon on the exporter's .png fallback
- Restore the copyright extractor for multi-line and qualified notices.
- Cross-check an override licence against the installed package.
- Stop the unbounded re-render on the collaboration path.
- Pin the three copies of the rename table together.
- Align the notes that contradicted each other.
- Make the skill's apply commands runnable anywhere.

**Not done:**

- **Anchoring the three year patterns to the start of a line** (non-blocking). It would reject `Portions Copyright 2017 Google Inc.` in `@playwright/test`, where the extraction is right today. That line is the only attribution the file carries. The concrete wrong outputs the review named (`fast-uri`, `mhchemparser`) are fixed directly in commit 4 ("Restore the copyright extractor for multi-line and qualified notices") instead.
- **`No decimals` → `~Respondo (opcia)` in `messages.eo.xlf`.** I'll do that as part of #2366 when this PR is merged into `main`.

Thanks a lot for those comments. Could you review this PR again once the tests have passed? I hope this is the last one :-)

erseco (approved):
Thanks @ignaciogros. I went through the 8 commits since `e5358c6` and re-ran the same checks as the previous rounds. **All four blocking items are fixed**, each with a test.

While verifying the extractor I found that the fix for it had reopened a hole next to the one it closed, plus three smaller things. They were all a few lines, so rather than send the PR back a fourth time **I pushed the fixes myself** — four commits on top of `95213e7`. Details below so you can check my work.

## The four blocking items, verified

**1. `inspectEdicuatexTree` no longer swallows the exception** (`e6391e5`). I reproduced the scenario the guard exists for — a `node_modules/edicuatex` left with only `index.html`, `LICENSE.txt` and `lang/`, against a vendored tree from the previous version:

```text
kind:   "incomplete"
detail: "the pinned package could not be read (ENOENT: ... scandir .../node_modules/edicuatex/css)"
```

`--remove-obsolete` refuses on that, `--extract-only` and `translations:sort` only warn. Exactly the split the comment claims.

**2. The workarea legacy fallback** (`823a8910`). `LEGACY_ICON_MAP[derived.value]` is right, and mapping first loses nothing: the map is keyed `objectives: 'target'`, with no `objetives` or `think-alt` key.

**3. The exporter `.png` fallback** (`99a6e31c`). `${renamed}.png`, with a spec pinning the empty-resolution-map branch `PrintPreviewExporter` / `Html5Exporter` actually take.

**4. The copyright extractor** (`0b274c85`). The multi-line regression is gone — `yjs` and `@isaacs/balanced-match` are back, `fast-uri` loses its trailing `until`, and the `{yyyy}` placeholder is rejected. See below for what I changed on top.

**Override licence cross-check.** Live, not dead code: `@codecov/bundle-analyzer` is installed on darwin (`os: ["darwin","linux"]`, `license: MIT`) and matches, so a relicensing bump now fails at generation instead of producing a drift only Windows can see.

Everything else from the last round — the re-render, the three-copy spec, the contradicting notes, the `C:\...` paths, the `carriedOver` naming, the two Biome infos — is done. The corrected `think-alt` note checks out: at `v4.0.3` every style already shipped `think_alt`, and the hyphenated `educablue/icons/think-alt.svg` first appears in `64df99f2` (#2258), after the tag. `neo/icons/objetives.png` is in `v4.0.3`.

## What I pushed

**`d496483` — the extractor takes licence prose as the holder again.** `holderOnNextLine` accepts any non-blank next line, and that is exactly where every licence template puts its prose. Measured against the exported function:

| Input | `e5358c6` | `95213e7` | now |
|---|---|---|---|
| `Copyright (c) 2020\nPermission is hereby granted, free of charge…` | `null` | **that whole sentence** | `null` |
| `Copyright (c) 2019\nTHE SOFTWARE IS PROVIDED "AS IS"` | `null` | **that clause** | `null` |
| `Copyright (c) 2015 Bitmovin GmbH, AT` | `Bitmovin GmbH, AT` | **`Bitmovin GmbH`** | `Bitmovin GmbH, AT` |
| `Copyright (c) 2015,Scott Motte` | `2015,Scott Motte` | `2015,Scott Motte` | `Scott Motte` |
| `Copyright (c) 2023\n  - Kevin Jahns` (`yjs`) | `null` | `Kevin Jahns` | `Kevin Jahns` |
| `…Gary Court until <url>` (`fast-uri`) | `Gary Court until` | `Gary Court` | `Gary Court` |

The first row is the MIT template as most packages ship it — without a blank line before the permission paragraph. It is the same bug the `[^\S\n]` change was made to close, just in the variant that has no blank line, and `--check` inside `make lint` would have frozen the result as correct. Capitalisation cannot separate that from a name (`THE SOFTWARE…` is the counter-example), so the continuation now has to carry a **list bullet** — the shape `yjs` actually uses, and one no licence text opens a prose line with. An unbulleted continuation yields `null`, and `Unknown` in a legal file is honest where prose is not.

The `until` strip was running on *every* notice, not only where a URL was removed, so it silently shortened real names — hence `Bitmovin GmbH, AT`. It is now conditional on a URL actually being stripped. A dangling conjunction is the opposite case (no holder ends in a bare "and", which is how `clone` writes it), so that one still goes unconditionally, with `[\s,]+` in front keeping `Switzerland` and `Legrand` whole.

I checked this the same way as the rest: over **every LICENSE and README in `node_modules`**. The generated attribution file is **byte-identical — 0 of 815 packages change** — and the single README result that moves is `ejs`, which had started returning the bare address `mde@fleegix.org` and now returns `null`. Five new specs cover the cases in the table.

**`8c8f865` — CHANGELOG.** Renaming `## Unreleased` to `## v4.0.4` left the earlier `### Removed` behind, so v4.0.4 rendered with that heading twice, split by a `---` that made the xAPI entry read as belonging to no release. Merged, with the rule back where the other releases put it (only before the next `## vX`).

**`ee66c60` — `internationalization.md:120`.** Still said a hook whose package is "absent or half-installed" reports nothing rather than failing. Half-installed is precisely what `e6391e5` turned into a reported problem, so the doc contradicted the fix. The absent case is unchanged.

**`3dd551e` — `SKILL.md:218`.** The new data-file route opens with `ReadAllText($json, $enc)` and never assigns `$json`, so an agent following it verbatim hits an `ArgumentNullException` on the step the section exists to enable. Now names the file and says what shape the JSON has.

## Verification on `3dd551e`

* `make lint` → exit 0, including `update-licenses-check`: `public/libs/README.md is up to date (61 packages)`. `--check` writes nothing.
* `make test-unit` → **8744 pass / 0 fail** (8739 before, +5 new specs). Vitest → **287 files, 15357 pass / 0 fail**.
* No `.skip` / `.only` / `.todo`; no `translations/**` file touched by any of these commits.
* CI on `95213e7` was green throughout — `build`, `build_and_test`, 12 E2E jobs, `codecov/patch`, `codecov/project`, `codecov/bundles`.

(The local `make test-unit` coverage gate flags `test/e2e/playwright/helpers/moodle-serving-model.ts` and `prebuilt-package.ts`; both are untouched by this PR and by my commits, and codecov reports the patch fully covered. A bare `bun test` also fails 42 route/DB tests that need a database — identical on a clean `95213e7`, so unrelated.)

## Non-blocking, for later

* **`LEGACY_ICON_MAP[derived.value]`** (`blockNode.js:264`) is a plain-object lookup without `Object.hasOwn`, so `constructor` / `toString` resolve to inherited members — the hazard the three `resolveRenamedThemeIcon` twins and `COPYRIGHT_OVERRIDES` were deliberately hardened against in this PR, on a line the PR touches. Same for `themeIcons[name]` two lines above.
* **The "degraded path" comments understate themselves.** `blockNode.js:39-40` and `74-76` call the local copy the fallback, but `yjs-loader.js` loads `blockIconRuntime.js` in group 4 and `app.bundle.js` is evaluated before that, so `window.eXeBlockIconRuntime || localBlockIconRuntime` binds the local one in the workarea. The new `block-icon.spec.ts` comment states it correctly; these two are the same class of contradiction `85de82bc` cleaned up.
* **`updateRemoteBlock` and `_applyBlockUpdate` now disagree** on what "the icon changed" means — the first compares `source`/`value`, the second still compares `blockNode.iconName` against the normalised name and repairs it. Both render correctly; only the stale `iconName` on the first path differs, and it is what a later local save writes back.
* **`readTable()`** (`block-icon.spec.ts`) scrapes the literal with a regex, so a `//` comment inside the braces containing `foo: 'bar'` would read as an entry. Fine today.
* **`COPYRIGHT_OVERRIDES[name].copyright`** wins over real metadata unconditionally, with none of the staleness cross-check you added for `license`. If `pdfjs-dist` or the `@mathjax/*` packages ever start shipping an `author` — the exact condition the docblock names as the entry's justification — the recorded holder keeps winning and nothing reports the entry as obsolete.
* **Pre-existing:** `caniuse-lite` → `and certain other rights. Our licenses are` and `mdn-data` → `and Related Rights in the Work…`, both CC prose. Neither is a direct dependency, so neither reaches `public/libs/README.md`.
* Your `@playwright/test` reasoning is right — `   Portions Copyright 2017 Google Inc.` is mid-line and anchoring would drop it. One correction with no consequence for the code: the file carries *two* `Portions Copyright` lines (189 is Microsoft), and the README says `Microsoft Corporation` because `package.json` has an `author`, so its LICENSE is never read.

---

Approving. Thanks for the very thorough turnaround on this one — the icon rename and the edicuatex guard in particular came out well, and the extractor is in far better shape than when we started, with the whole dependency tree as evidence rather than a handful of samples.

PR #2365, opened 2026-09-06
Icon tinting is driven entirely by theme CSS

### LINKED ISSUE #2363: Remove `THEME_ICON_COLOR_MAP` so icon tinting is driven entirely by theme CSS
`getCurrentThemeIconColor()` in `public/app/workarea/project/idevices/content/blockNode.js`
holds `THEME_ICON_COLOR_MAP`, a hardcoded table of one tint colour per bundled theme. Every
value in it is **already declared by the theme itself in CSS**, and the resolution chain reads
the CSS first. The table is duplicated data: a second place where a theme's identity lives,
inside the application's JavaScript rather than in the theme.

Styles should be fully self-describing. A theme should be able to state its picker tint in its
own stylesheet and be treated exactly like the seven that ship with the application.

### 1. All seven values are duplicated

| Theme | `THEME_ICON_COLOR_MAP` (`blockNode.js`) | Theme CSS (`public/files/perm/themes/base/<id>/style.css`) |
|---|---|---|
| `base` | `#d86e41` | `--exe-icon-color: #d86e41` |
| `flux` | `#eda900` | `--exe-icon-color: #eda900` |
| `nova` | `#f5c200` | `--exe-icon-color: #f5c200` |
| `neo` | `#e3ac3b` | `--exe-icon-color: #e3ac3b` |
| `zen` | `#d40055` | `--exe-icon-color: #d40055` |
| `universal` | `#0d2953` | `--exe-icon-color: #0d2953` |
| `educablue` | `#0d77d1` | `--exe-icon-picker-color: #0d77d1` |

`educablue` is the proof that the CSS mechanism already covers the hard case — a light
`--exe-icon-color: #fff` for the coloured block header plus a separate, readable
`--exe-icon-picker-color` for the picker chips — with no help from the JavaScript table. It is
nonetheless listed in the table as well.

### 2. The table is already close to unreachable

The resolution loop returns on the first candidate element that yields anything:

```js
for (const colorSource of colorCandidates) {           // headElement, blockNameElementText, iconElement
    const styles = window.getComputedStyle(colorSource);
    const customColor = --exe-icon-picker-color || --exe-icon-color || --icon-primary;
    if (customColor) return customColor;
    const color = styles.color;
    if (color && color !== 'rgba(0, 0, 0, 0)') return color;   // ← almost always returns here
}
// THEME_ICON_COLOR_MAP is only consulted past this point
```

`styles.color` is inherited and virtually never transparent, so the table is reached only when
the block has **no header, no title element and no icon element**. The existing test confirms
this: it has to null all three out to exercise the fallback
(`blockNode.test.js`, "falls back to each theme style-icon color so General icons match Style
icons").

This makes the removal low-risk, and it also corrects a claim worth not repeating: a
third-party theme that declares none of the variables does **not** fall to `#6E9F41`; it falls
to the computed `color` of its own block header. `ADR-1247-04` states otherwise in its
Consequences section.

### 3. It was convenience, not a workaround

Introduced in c169c6cef ("Material Icons as the default icon set for block/iDevice icons",
#1497). The themes already declared `--exe-icon-color` at that point, and neither the commit
message nor the code comment records a timing or scoping problem that the table was working
around. There is no hidden reason to keep it.

## Proposed change

1. Delete `THEME_ICON_COLOR_MAP` and its branch in `getCurrentThemeIconColor()`, leaving the
   chain as `--exe-icon-picker-color` → `--exe-icon-color` → `--icon-primary` → computed
   `color` → `#6E9F41`.
2. Confirm each bundled theme declares what it needs. All seven already do; check whether the
   multi-hued themes (`neo`, `universal`) want a distinct `--exe-icon-picker-color`, since the
   code comment notes they deliberately keep the theme accent rather than matching their
   multi-coloured Style icons.
3. Replace the table-shaped test with one that asserts the *chain*, not a copy of the constant.
   The current test enumerates all seven values and is a change detector for the
   implementation rather than a test of behaviour.

## Files to touch

- `public/app/workarea/project/idevices/content/blockNode.js` — the table and the fallback branch.
- `public/app/workarea/project/idevices/content/blockNode.test.js` — the three tint tests around
  line 1712.
- `doc/development/styles.md` — the "Which colour the icon picker actually uses" section
  (the chain diagram and the paragraph naming the table).
- `doc/architecture/adr/ADR-1247-04-dual-icon-rendering-and-theme-tint.md` — the Decision text
  ("with per-theme fallbacks"), the Evidence bullet on per-theme tint, and the Consequences
  bullet on custom themes. **The ADR is `status: Proposed`, so it can be amended in place**
  rather than superseded; only promote it to a new ADR if the decision itself is judged to have
  changed.
- `doc/architecture/changes/1247-material-icons-default-icon-system/design.md` — line 100
  mentions the fallback. Implemented designs are not rewritten, so confirm whether this counts
  as a link/typo fix or needs a note instead.
- Possibly `public/files/perm/themes/base/*/style.css`, if step 2 finds a theme that wants an
  explicit picker accent.

## Verification

Automated tests cannot confirm this one. After the change, open the icon picker in the
application and compare the "General icons" tint against the "Style icons" artwork in **each of
the seven bundled themes**, plus at least one theme that declares none of the variables, to
confirm the computed-colour step behaves sensibly.

## Out of scope

- The `#6E9F41` default itself (`--icon-primary` in `assets/styles/abstracts/_variables.scss`).
- The dual rendering technique (mask + `currentColor` vs ``), which is the substance of
  ADR-1247-04 and is not in question here.

## Context

Raised during the review of #2362, where the fallback chain was documented in
`doc/development/styles.md` but deliberately left in place: correcting the documentation was in
scope for a release-preparation PR, changing the behaviour of the icon picker was not.

ignaciogros:
Merge this PR after #2362 

- Remove THEME_ICON_COLOR_MAP so icon tinting is driven entirely by theme CSS. No style is named anywhere in the application.
- Neo style: use the style's own forest green for Material icons (General icons).

### DIFF
--- a/.github/workflows/build-electron-installers.yml
+++ b/.github/workflows/build-electron-installers.yml
@@ -40,7 +40,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest, macos-15, windows-latest]
     runs-on: ${{ matrix.os }}
 
     steps:
@@ -184,7 +184,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest, macos-15, windows-latest]
     steps:
       # Needed by the boot assertion below, which lives in scripts/.
       - name: Checkout code
@@ -259,7 +259,7 @@ jobs:
 
       # ---------- macOS test ----------
       - name: Test install on macOS
-        if: matrix.os == 'macos-latest'
+        if: matrix.os == 'macos-15'
         run: |
           set -euo pipefail
 
@@ -566,7 +566,7 @@ jobs:
           set -euo pipefail
           case "${{ matrix.os }}" in
             ubuntu-latest)  log=ubuntu-app.log ;;
-            macos-latest)   log=mac-app.log ;;
+            macos-15)       log=mac-app.log ;;
             windows-latest) log=windows-app.log ;;
             *) echo "::error::Unknown runner '${{ matrix.os }}'"; exit 1 ;;
           esac


erseco (changes requested):
Reviewed with #2362 treated as the base, as the PR body asks. Findings are split into what this PR's own delta introduces and what it inherits byte-for-byte from #2362.

## Changes requested (this PR's own delta)

1. **`doc/development/styles.md:150` – the new fallback sentence is wrong outside the picker.** "without it the tint falls back to the application's own `--icon-primary` green" only holds in the icon picker modal, where `_modals.scss` resolves `var(--modal-icon-color, var(--icon-primary))`. In the editor content area and in every export the glyph is `background-color: currentColor` with no `color` set (`_execontent.scss:240-244`, `base/style.css:436-441`), so it takes the block header's text colour, which is what the replaced sentence said. `--icon-primary` is not referenced by any theme `style.css`, by `public/style/workarea/base.css`, or by anything under `src/shared/export/`. Please restore the previous wording here and keep the `--icon-primary` explanation in the picker section (lines 218-243), where it is correct.

2. **`public/app/workarea/project/idevices/content/blockNode.js:414-426` – with the map gone, the JS resolver is a copy of the CSS fallback chain.** The `--icon-primary` read, the computed-`color` branch and the `'#6E9F41'` literal re-implement what `_modals.scss` already does in all nine picker rules. The modal is mounted in `.modals-container`, outside `.exe-content`, and `--icon-primary` is on `:root` (also in the static bundle), so the CSS fallback always applies. Reading only `--exe-icon-picker-color` / `--exe-icon-color` and calling `setProperty('--modal-icon-color', …)` only when one is found gives byte-identical output for all seven bundled themes, removes the duplicated `#6E9F41` (pinned in three tests against `_variables.scss:110`), and lets the docs drop the "safety net the editor never reaches" paragraphs.

3. **`public/app/workarea/project/idevices/content/blockNode.test.js:1724-1783` – the replacement tests no longer catch mutations the deleted ones caught.** Removing `color &&`, removing the `rgba(0, 0, 0, 0)` check, or removing the whole computed-colour guard all pass the new suite; `origin/main`'s suite failed the first one. Two causes: "walks the candidate elements in order" nulls `headElement`, so the loop never has to skip an attached element that yields nothing; "falls back down the variable chain to --icon-primary" sets the variable to `#6E9F41`, the same value as the literal, so it cannot tell the variable read from the fallthrough. Keep `headElement` attached with no variable in the walk test, and use an arbitrary hex for `--icon-primary`. If item 2 is taken, the tests shrink with it.

4. **`public/files/perm/themes/base/neo/style.css:291` – dead fallback literal.** `color: var(--exe-icon-color, #1a7d1e)` restates the value that `.exe-content` declares four lines above on an ancestor, so the `var()` fallback can never fire. This is why the tint change needed two lines that must agree, while the ADR and styles.md now promise "a one-line edit in the theme's own stylesheet". `color: var(--exe-icon-color);` is enough.

5. **`doc/architecture/changes/1247-material-icons-default-icon-system/design.md:101` and `ADR-1247-04` – record the amendment where the tooling reads it.** `design.md` is `status: implemented`, and AGENTS.md §7.11 / `doc/architecture/changes/README.md` say implemented designs are not rewritten beyond typo/link fixes; the new "Note (issue #2363)" is substantive. The ADR is still `Proposed`, so it can carry the note. Also `related.prs` (ADR line 14) and `implementation_prs` (design.md line 12) still read `[1497]`; `make architecture-records` reads frontmatter, not prose, so #2363/#2365 are invisible to the generated index. Please move the note to the ADR and add 2365 to both lists.

Everything else in the PR's own delta checks out: every bundled theme declares `--exe-icon-color` on `.exe-content`, educablue declares `--exe-icon-picker-color`, no `#e3ac3b` leftovers remain, and `make architecture-check` passes.

## Inherited from #2362 (not this PR's delta)

This branch currently carries all of #2362, so the GitHub diff shows findings that belong there. They drop out of this PR once #2362 merges first, as the PR body intends, but they ship with this PR if it merges first. Listed for visibility; they should be addressed on #2362:

- `neo/icons/objetives.png` → `objectives.png` and `educablue/icons/think-alt.svg` → `think_alt.svg` are renamed with no alias. Theme icon ids are the file base names, so projects that saved `objetives` / `think-alt` lose the icon in the editor and get a broken `` in every export (`IdeviceRenderer.ts:80`, `blockNode.js:298`).
- `src/cli/commands/translations.ts:81` – the new edicuatex drift guard ignores `drift.changed`, so a stale but complete vendored tree passes and `--remove-obsolete` deletes keys from every locale. `:80` – `readdirSync` inside the guard is unguarded and aborts every translations command on a partial `node_modules/edicuatex`.
- `src/cli/commands/update-licenses.ts:138` – copyright extraction writes disclaimer fragments as attribution for placeholder MIT/BSD/Apache texts. `:434` – `--check` carries README values over for packages missing from `node_modules`, so it cannot detect drift there. `Makefile:357` – `update-licenses-check` is not invoked by CI or `make lint`.
- `translations/messages.gl.xlf:2058` – a reviewed target ("Texto moi grande") was overwritten with a `~` placeholder. `translations/messages.eu.xlf` – seven reviewed Basque corrections from #1922 carry the `~` prefix.
- Minor: `translations.ts:11` header misdescribes `--allow-missing-generated` and the `src/cli/index.ts:200/205` help is stale; `Makefile:477` `vendor-edicuatex` rm-rf's and rewrites the tree on every run instead of trying `--check` first.

PR #2370, opened 2026-09-07
Remove `THEME_ICON_COLOR_MAP`. Icon tinting driven by theme CSS.

ignaciogros:
- Remove `THEME_ICON_COLOR_MAP` so icon tinting is driven entirely by theme CSS.
- Use Neo's own forest green for Material icons (General icons).
- Document the tint as a theme-CSS contract and record the amendment.
- Remove dead code from the CSS style files.
- Point icon-tint records at PR #2370 and note the change for style authors.


erseco (changes requested):
@ignaciogros Thanks for this — moving the tint into theme CSS is the right direction, and removing the JS colour map is a real simplification. I ran the review against the branch tip: `npx vitest run public/app/workarea/project/idevices/content/blockNode.test.js` is 311/311 green and `make architecture-check` is OK (24 records, 6 changes). I also confirmed that all seven bundled themes currently declare `--exe-icon-color` on `.exe-content`, so nothing is visibly broken today.

That last point is exactly why I'm requesting changes: the PR removes **both** safety nets at once (the JS `THEME_ICON_COLOR_MAP` *and* the literal `var(--exe-icon-color, #hex)` fallbacks in all 7 stylesheets) and adds nothing that fails when the remaining invariant is violated. Plus a few documents now contradict each other.

---

## Blocking

### 1. The ADR states two incompatible rules for the same case

`doc/architecture/adr/ADR-1247-04-dual-icon-rendering-and-theme-tint.md`

Line 71 (edited in this PR): *"Multicolor themes (neo, universal) cannot be matched by a single tint, so each simply declares the one it wants — `universal` its accent, `neo` its own forest green"*.

Line 80, in the Decision paragraph (also edited in this PR, but only in its middle clause), still ends with *"multi-hued themes keep the theme accent"* — which is precisely the rule this PR supersedes when it moves neo from its accent `#e3ac3b` to `#1a7d1e`. The ADR is the record of the decision, so this needs to say one thing.

### 2. `design.md` documents an implementation this PR deletes

`doc/architecture/changes/1247-material-icons-default-icon-system/design.md:100`

> Per-theme tint via `--exe-icon-color` / `--icon-primary` / `--modal-icon-color` with `THEME_ICON_COLOR_MAP` fallback and default `#6E9F41` (`blockNode.js` `getCurrentThemeIconColor`)

Every element of that sentence is removed by this PR, and the frontmatter now adds #2370 to `implementation_prs` — so a reader trusting the traceability field is sent to code that no longer exists, and the design directly contradicts ADR-1247-04 line 71 in the same commit. AGENTS.md forbids rewriting implemented designs, so either add a pointer to the ADR amendment, or drop the `implementation_prs` bump here.

### 3. Removing the literal fallback makes the CSS degrade *worse*, not just differently

`public/files/perm/themes/base/neo/style.css:291` (same shape in `nova/style.css:289` and `zen/style.css:327`)

An undefined custom property makes the whole declaration invalid at computed-value time, and for an inherited property like `color` that computes to `inherit` — it does **not** fall back to the next declaration in the cascade. All three themes have an earlier rule of *equal* specificity:

```css
.exe-content .box-head .box-title,
.exe-content .box-head .box-icon { color: #002702; }   /* neo */
```

With `var(--exe-icon-color, #e3ac3b)` an undeclared variable still produced a definite colour. With bare `var(--exe-icon-color)` it wipes the theme's own explicit icon colour and inherits from `.box-head`. And since `.box-icon` is a `<button>` in the editor (UA `color: buttontext`) but not in the export, the visible result diverges between the two. ADR-1247-04 line 106 documents the *picker* fallback, not this one.

### 4. Nothing fails when a theme stops declaring `--exe-icon-color`

The whole change is safe only because the seven bundled themes happen to declare it. Add an 8th theme, or drop that declaration during a refactor, and you get a silent double regression — the header icon falls back to `inherit` in content and exports, and the picker chips go to the app's `#6E9F41` green — with `make test-unit`, `make test-e2e` and `make architecture-check` all still green.

A unit test that reads `public/files/perm/themes/base/*/style.css` and asserts each declares `--exe-icon-color` pins the invariant the ADR now calls a contract. ~10 lines.

---

## Contract questions — your call, but please record the answer

### 5. `--icon-primary` is dropped from the resolver without a note

`public/app/workarea/project/idevices/content/blockNode.js:439`

`--icon-primary` is a first-class application variable (`assets/styles/components/_menu.scss:211`, `_menu_nav.scss:76,301`, `_variables.scss:110`). A theme author who overrode it on `.exe-content` to recolour the app chrome previously also got that colour in the icon picker, because the closer `.exe-content` declaration beat the `:root` one when read off the block header. Now the resolver returns `''`, the modal is left untouched, and `_modals.scss` resolves `var(--modal-icon-color, var(--icon-primary))` against `:root` — the app's green. ADR-1247-04 lists only `--exe-icon-picker-color` and `--exe-icon-color` as removed/kept, and never records that `--icon-primary` was an accepted theme input.

### 6. The resolved value is copied verbatim onto a node outside `.exe-content`

`public/app/workarea/project/idevices/content/blockNode.js:1787`

A theme declaring `--exe-icon-color: currentColor` — a natural way to make the General icons follow the header text, and something the new docs' "General icons are painted with `currentColor`" framing invites — yields the literal token `currentColor` from `getPropertyValue`. `setProperty('--modal-icon-color', 'currentColor')` on the modal body then resolves against the *modal's* text colour, giving near-black chips instead of the theme accent. The new `if (themeIconColor)` guard only covers the empty case, not the context-sensitive one.

---

## Cleanups

### 7. The three-candidate walk is now dead logic

`public/app/workarea/project/idevices/content/blockNode.js:432`

Both remaining lookups are custom properties, which inherit. In the real workarea `this.headElement` (`<header>`, line 755) is the ancestor of `this.blockNameElementText` (`<h1>`, line 860) and `this.iconElement`, so `getComputedStyle(headElement).getPropertyValue(...)` already returns whatever the descendants would. The walk was meaningful before this PR because `styles.color` is a per-element computed value that genuinely differed; with that branch deleted, the loop, the `filter(Boolean)` and the 3-element array are dead flexibility:

```js
const styles = getComputedStyle(this.headElement || this.blockNameElementText || this.iconElement);
```

Related: `if (!window.getComputedStyle) break;` (line 434) is loop-invariant but re-evaluated per candidate — it belongs above the loop, or can go entirely since `getComputedStyle` exists in every supported target and in both test environments.

### 8. The new test locks in a DOM shape that cannot occur

`public/app/workarea/project/idevices/content/blockNode.test.js:1772`

*"keeps looking when an attached element declares no theme variable"* appends `block.headElement` and `block.blockNameElementText` as two independent children of `document.body` and declares `--exe-icon-color` only on the title. In the real block the title is *inside* the header, so the property inherits downward and the head already resolves it — the "walk must reach the title" branch the comment describes is unreachable. The test asserts an impossible arrangement and will keep passing even if the resolver breaks for the real nested DOM.

### 9. `styles.md` overstates the modal contract

`doc/development/styles.md:262` claims *"every rule in `assets/styles/components/_modals.scss` reads `var(--modal-icon-color, var(--icon-primary))`"*. Within `#change-block-icon-modal-content`, `.icon-options-section-title` uses `color: var(--icon-gray)` (line 63), and `.option-block-icon.exe-icon` uses `background-color: var(--main-background-color)` and `border: 1px solid var(--gray-border-light)` (lines 34-35). A theme author following the doc sets `--exe-icon-color`, expects the whole picker to follow, and finds the section headings and chip backgrounds unchanged.

### 10. No E2E for a user-visible colour change

AGENTS.md: *"User-visible behavior changes (workarea, preview, export, embedding, collaboration) must include or update a Playwright spec"*. `test/e2e/playwright/specs/block-icons.spec.ts` asserts only the `--exe-material-icon-url` data: URI, never a colour, so nothing in E2E would notice a tint regression in a real browser. The unit tests structurally cannot cover it either — they set custom properties inline on the element itself and never exercise real CSS inheritance from `.exe-content`.

### 11. Test setup without teardown

`public/app/workarea/project/idevices/content/blockNode.test.js:1774,1785,1797` — three `document.body.appendChild(...)` calls; the file's only `afterEach` (line 122) just nulls `block`. Harmless today since each test reassigns `block.headElement`, but it is setup/teardown asymmetry in exactly the tests that depend on document attachment for `getComputedStyle`, and a future test doing `document.querySelector('header')` or looking for `#change-block-icon-modal-content` will pick up a stale node.

---

For what it's worth, the JS refactor is behaviour-preserving in practice: `--icon-primary` is declared on `:root` (`_variables.scss:110`), so the old `styles.color` and `THEME_ICON_COLOR_MAP` branches were already unreachable in the real app. And neo's `#1a7d1e` does exist in its own palette (lines 472/476), so that change looks deliberate and fine — it just needs the ADR to agree with it.

Items 1-4 are what I'd like to see before merge; the rest can be split off if you prefer.

ignaciogros:
Thanks for the review. I think every point is addressed.

**Commits since your review:**

- `1f84f36df` Read the icon tint off the block header, and resolve `currentColor` there.
- `9c17ae9d3` Assert in a real browser that the applied glyph takes the style's tint.
- `a151ab17c` Fail when a bundled style stops declaring `--exe-icon-color`.
- `1ec419cb2` Fall back to a neutral tint in the icon picker, not the application's green.
- `04883ac48` Make the icon-tint records state one rule, and record what is not a theme input.

**From a new full review after those commits:**

- `3be28a61a` Keep the icon picker open when `getComputedStyle` is missing.
- `40bf48824` Check where a style declares `--exe-icon-color`, not just that it does.

Could you please take a look at these changes? I hope everything's OK now.

Thanks!

erseco (approved):
@ignaciogros All four blocking items are closed. I checked each one against the branch tip rather than against the commit messages. Approving.

## Verified

| # | Item | Where |
|---|------|-------|
| 1 | ADR stated two incompatible rules | Decision paragraph now reads *"a multi-hued theme, which no single tint can match, simply declares whichever one it wants"*, agreeing with line 71 and with Consequences |
| 2 | `design.md` documented deleted code | `implementation_prs` back to `[1497]`; the amendment lives in the ADR's References |
| 3 | Bare `var(--exe-icon-color)` degrades to `inherit` | Still bare, but item 4 removes the silence — see below |
| 4 | Nothing failed when a style dropped the tint | `src/shared/block-icon.spec.ts`, and `40bf848` checks *where* it is declared, not just that the text appears |
| 5 | `--icon-primary` dropped without a note | New `--modal-icon-default: var(--text)` across `_modals.scss`; the ADR now records explicitly that `--icon-primary` is not a theme input |
| 6 | `currentColor` copied verbatim onto the modal | Resolved against the header before copying, with a test |
| 7 | Three-candidate walk was dead logic | One `a \|\| b \|\| c`, `getComputedStyle` guard hoisted out |
| 8 | Test locked in an impossible DOM | Rewritten as the headerless case, with a comment pointing at the E2E for the real inheritance |
| 9 | `styles.md` overstated the modal contract | New warning block naming `--icon-gray`, `--main-background-color`, `--gray-border-light` |
| 10 | No E2E for a user-visible colour change | `block-icons.spec.ts` compares the applied glyph's ink against the style's declared tint |
| 11 | Setup without teardown | `attach()` helper plus `afterEach` |

Ran locally at `40bf848`:

- `npx vitest run public/app/workarea/project/idevices/content/blockNode.test.js` — 313/313
- `bun test src/shared/block-icon.spec.ts` — 21/21
- `make architecture-check` — OK (24 records, 6 changes)

I also mutation-tested the new guard: flipping neo's declaring selector from `.exe-content` to `.exe-export` makes `neo declares --exe-icon-color on .exe-content` fail. It pins what it claims to pin, which is the whole point of item 4.

(`make lint` fails here on `update-licenses-check` because `@mathjax/src` is not installed in my worktree. Local environment, not this PR.)

---

## Non-blocking

### 1. The `currentColor` path dropped the transparent-colour guard

`public/app/workarea/project/idevices/content/blockNode.js:452`

Before this PR the resolver refused a computed `rgba(0, 0, 0, 0)` and moved on. That guard covered the automatic fall-through to `styles.color`, which this PR deliberately removes, so losing it is not a regression. The concern survives on its own though: the `currentColor` branch is new here, and `doc/development/styles.md:261` now advertises it. A style declaring `--exe-icon-color: currentColor` on a header whose computed colour is transparent — a `background-clip: text` gradient heading is the realistic one — writes `--modal-icon-color: rgba(0, 0, 0, 0)` onto the modal body, and every picker chip renders invisible. That is strictly worse than the neutral grey the empty-string path exists to give.

```js
if (customColor.toLowerCase() === 'currentcolor') {
    return styles.color && styles.color !== 'rgba(0, 0, 0, 0)' ? styles.color : '';
}
```

### 2. The contract guard accepts declarations that produce no tint

`src/shared/block-icon.spec.ts:161`

`/--exe-icon-color\s*:\s*\S/` matches `--exe-icon-color: initial` and `--exe-icon-color: var(--typo)`. Both make `color: var(--exe-icon-color)` invalid at computed-value time, so the header glyph falls back to the inherited text colour and `getCurrentThemeIconColor()` returns `''` — precisely the colour-only regression this spec exists to catch, with CI green. The realistic way in is a palette refactor that typos a variable name, not someone writing `initial`.

### 3. `readdirSync` without a `style.css` filter

`src/shared/block-icon.spec.ts:152`

`styles` is every directory under `themes/base`. Add a shared `img/` or `fonts/` folder there — a natural refactor, since seven styles duplicate artwork — and the suite throws `ENOENT: .../img/style.css` instead of reporting a missing tint.

### 4. Two small things in the E2E probe

`test/e2e/playwright/specs/block-icons.spec.ts`

- `probe.style.color = declared` resolves against `document.body`, so a style declaring `currentColor` — the case the resolver special-cases and the docs now advertise — makes `expected` the workarea body's colour while `headerColor` is the block header's, and the test fails on a correct style. Appending the probe inside `.box-head` makes the two agree.
- The probe is a single `page.evaluate`. The preceding `waitForFunction` only waits for the JS-applied `--exe-material-icon-url`, which does not imply the style's `style.css` has finished applying. Moving the probe inside a `page.waitForFunction` costs nothing and matches AGENTS.md §10.

---

## Looked at, not asking you to change

- The bare `var(--exe-icon-color)` in the seven stylesheets still computes to `inherit` rather than falling through to the next declaration — my original item 3. With `src/shared/block-icon.spec.ts` in place it can no longer happen silently, which was the actual objection. Fine as it stands.
- `--icon-primary` has no migration note for third-party styles. `themes/users/` is empty and nothing in the repo overrides it, and the ADR and `styles.md` both record the decision. Enough.
- Requiring the selector to be exactly `.exe-content` rejects a style that declares the tint on `:root` or `body`, which would work. That strictness is the convention for bundled styles and the test comment says so — deliberate, not a bug.
- Neo's `#e3ac3b` to `#1a7d1e` has no test pinning the hue. Pinning it would re-introduce the per-theme colour table this PR exists to remove. Leave it.
- The JSDoc calls the `blockNameElementText` / `iconElement` fallbacks unreachable in a browser while a new test pins them. Cosmetic, but pick one: either they do something, or it is `const colorSource = this.headElement;` and the test goes with them.

The docs came out well — the "Not everything in the picker follows your tint" block in `styles.md` is the part a style author actually needed.

ignaciogros:
Closes #2363

Issue #2263, opened 2026-08-14
mixed and missing language - corrected


jdziew63-pl:
In Project proprties the language used is polish - see attachment
Despite this, when editing a project, when asked about Fill in (Cloze), the check function key under the text seems to be in Spanish (Comprobar). Pressing it shows the text in Spanish (Tu puntuacion es 0 (0/1) Lo siento. No has superado la prueba
see attachment

<img width="1192" height="596" alt="Image" src="https://github.com/user-attachments/assets/a81dbbd3-90ed-449c-b462-e4fc725ae026" />
<img width="1089" height="547" alt="Image" src="https://github.com/user-attachments/assets/7a66913f-73be-4043-be44-5fc14f05714c" />

2. Is there a Manual for help for ver. 4.0.3 available in english - again trying to reach it reveals only spanish version ?
3. In Interactive activities the question type Fill in (Cloze) can not be found. How can I reach it

jdziew63-pl:
Please - help in solving problems. My used previous version of eXe was 2.9

ignaciogros:
Thanks for reporting this, @jdziew63-pl.

We'll try to fix it in v4.0.4, which will be released soon.

ignaciogros:
@mnarvaezm:

**There are Spanish strings in Fill in (Cloze) / Form iDevice.**

The runtime (export-side) default messages of several iDevices are hardcoded in **Spanish** and are used as a fallback whenever a key is missing from the saved `msgs` object, regardless of the project language:

- `public/files/perm/idevices/base/form/export/form.js` → `$form.msgs` (`msgCheck: 'Comprobar'`,
  `msgTestResultNotPass: 'Lo siento. No has superado la prueba'`, ...)
- `public/files/perm/idevices/base/scrambled-list/export/scrambled-list.js` → `getMessages()`
- `public/files/perm/idevices/base/trueorfalse/export/trueorfalse.js` → `msgsdefault`

In addition, some keys have no `c_()` counterpart in the edition side, so they are **always** rendered in Spanish even after saving the iDevice:

- `form/edition/form.js` `refreshTranslations()` is missing `msgOk` and `msgKO`
- `trueorfalse` is missing `textButtonScorm`

**Fix:** 

make all export-side defaults English (source language), as
`interactive-video/export/interactive-video.js` already does, and add the missing keys to
`refreshTranslations()` so they go through `c_()`.

ignaciogros:
Hi @jdziew63-pl,

[...]

PR #2371, opened 2026-09-07
fix(scorm): homogenise SCORM score reporting across the 35 iDevices

mnarvaezm:
## Context

eXeLearning's 35 gradable iDevices behaved inconsistently when saving a score to the LMS. Each one
decided in its own way when to register with the activity registry, what counts as having started,
when to publish a score and when to declare itself finished. None of that was unified, so two
activities an author perceives as equivalent behaved differently in Moodle, and neither the learner,
nor the teacher, nor the author could predict which was which.

This PR homogenises how the score is saved. All 35 iDevices now share the same contract with the
LMS: they publish a zero as soon as the learner explicitly starts, they report progress as questions
are answered, they declare the activity finished at the same point, and they return the score to
zero on a restart. A page left without having interacted with any activity is recorded incomplete
and **with no score**, rather than as a zero. The access code stops being a purely visual unlock and
starts the activity, exactly as the start button does. Underneath, the SCORM 1.2 runtime is adjusted
so that data reaches the LMS at the right moment instead of one page later.

## What each role gains

**The learner** sees a course menu that matches what they are doing: an activity they started shows
as started, one they finished shows as finished, and restarting clears the previous result instead
of leaving it stranded.

**The teacher** gets a trustworthy record. A page visited but not worked on is stored incomplete and
unscored, rather than as a real zero that drags the learner's average down, and a resumed attempt
that is finished gets closed instead of staying suspended forever.

**The author** can choose one iDevice over another on pedagogical grounds, without the choice
changing how their page is graded. And the weight they assign to each activity is honoured: it was
being stored incorrectly, and the page's mark could depend on the order the iDevices were placed in.

### A page or SCO with no SCORM iDevices

This covers pages holding only text, images or video, and also pages whose iDevices have the SCORM
option **disabled**. With no gradable iDevice present, nothing holds the page back from completing
when it is left.

| Action | Status | Score |
|---|---|---|
| First entry | **Not attempted** | none |
| On leaving | **Completed** | none |
| Entering again | **Completed** | none |

There is nothing to answer and nothing to restart on these pages, and **a score must never appear**.

Once the page has been visited and left, it stays **Completed**. Entering it again later must keep
that same status.

### A page with one SCORM iDevice enabled

The page's score is the one obtained in the iDevice itself, expressed out of 100.

| Learner action | Status | Score |
|---|---|---|
| Enters | **Not attempted** | none |
| Leaves without doing anything | Incomplete | none |
| **First interaction** | Incomplete | **0** |
| Answers question by question | Incomplete | the partial one |
| Answers **every** question | **Passed** or **Failed** | the final one |
| Leaves halfway and comes back | Incomplete, progress restored | the one they left |
| Restarts the activity | Incomplete | **0** |

**Why is there no 0 on entry?** It is deliberate. A page that has merely been visited, without the
activity actually being started, must not receive a score.

**What counts as the first interaction** depends on how the iDevice is configured:

- with **Time**, pressing the start button;
- with **Access**, entering the password correctly;
- with neither time nor access, **answering for the first time**.

At that moment the score becomes **0**. If the learner is coming from an earlier attempt that had a
different score, it must **drop to 0** rather than linger until they answer something.

**On answering every question**, the status becomes **Passed** when the score is 50 or above, and
**Failed** when it is below.

**On restarting**, the activity returns to **Incomplete** with a score of **0**, both in the iDevice
itself and in the SCORM menu. The previous attempt's result must not remain visible in either place.

### A page with more than one SCORM iDevice enabled

Each iDevice contributes its share to the page's total score, calculated out of 100.

The status stays **Incomplete** until the learner has answered **every question of every gradable
iDevice**. Once they are all complete it becomes:

- **Passed**, when the score is 50 or above;
- **Failed**, when the score is below 50.

Two situations are worth keeping in mind while testing:

- An iDevice **that was never started contributes 0 at its full weight** to the page's score. With
  three equally weighted iDevices, scoring 100 % on one without starting the other two gives the
  page roughly **33/100**.
- While any question is still pending, the status stays **Incomplete** regardless of the score shown
  at that moment. Seeing **100 points on an Incomplete page** can therefore be perfectly coherent.

To diagnose one iDevice's behaviour, test it **alone on its own page**. With several of them, the
status and the score belong to the page as a whole and it is harder to tell which iDevice is causing
a failure.

## SCORM 1.2 runtime

- **Commit after every report.** The score only reached the LMS when the page was left.
- **Anything reported before the session opened is now published.** iDevices register on
  `jQuery ready` while `loadPage()` runs on `body onload`, after every image and video has loaded;
  whatever landed in that window stayed invisible to the LMS until the next flush.
- **A page's mark no longer depends on the order of its iDevices.** The historical algorithm scaled
  the weights by largest-remainder rounding; with equal weights every fraction ties, so the spare
  point went to whichever activity registered first. Three activities scoring 100/50/0 aggregated to
  50.5 and the same three as 0/50/100 to 49.5 — the same work by the learner, opposite verdicts
  against a threshold of 50. It is now the exact weighted mean, which is symmetric.
- **A default weight of 100** when the activity has no stored weight, instead of 1.
- **`cmi.core.exit` is cleared as soon as the attempt turns terminal**, not only at page unload. On a
  resumed attempt the previous visit's `suspend` stayed stored while `lesson_status` was already
  `passed`, and Moodle drew the menu icon from that stale value.
- **A restart may downgrade a terminal status restored on page load.**
- **A second retry of the commit**, so Moodle redraws the menu icon — it loses the race against its
  own `sendBeacon`.

## iDevices

**Opening zero and access code.** When the learner starts explicitly — start button, link, or the
correct password — the activity publishes 0 and incomplete. The access code additionally starts the
game where that is appropriate, with each iDevice's own particulars: `discover` with a single level
starts, and with two or three shows the level selector, because the learner has to choose; the
memory mode of `flipcards` starts while its other modes are already live; a timed `word-search`
starts its clock.

**Restart.** The score returns to 0 and the status to incomplete, in the game and in the menu. The
"score only once" lock is removed from the four iDevices that carried it, and `gameOver` is lowered
on a replay so the activity is playable again.

**Completion.** The activity is marked finished on the last answer. The presentation modes of
`3dmol` and `electrical-circuits` finish on reaching the last item. `interactive-video` now
registers, scores and finishes, which it did not do before.

**SCORM weight.** The weight the author chose is honoured, where it used to be lost; `rubric` accepts
only values from 1 to 100; `quick-questions-video` and `interactive-video` stored and restored it
incorrectly.

**Assorted.** One card sound per click, the countdown indicator only on timed activities, `sort` was
undercounting correct positions, `mathematicaloperations` ignored an option stored as a string, and
the removed lives panel crashed every game over.

## Translations — refs #2263

Three iDevices carried hardcoded Spanish in their export defaults, so a user running the interface
in a language with no catalogue saw Spanish instead of English. They now fall back to English, and
the missing keys are added to `refreshTranslations`.

**Nothing under `translations/` is touched.**

## Documentation

- **ADR-2209-02** — decision point 9 is corrected: the aggregation algorithm is no longer the
  historical one, and the reason it changed is recorded.
- **`doc/development/scorm12-runtime-contract.md`** — Moodle's beacon race and its retry, clearing
  the `exit` when the attempt turns terminal, the single aggregation algorithm, and a **new §6.1**
  separating writing the status (`LMSSetValue`) from storing it (`LMSCommit`): until the first
  commit, the LMS's reports still show the state from before the visit, which is why a page open and
  untouched reads "not attempted".
- **`doc/elpx-format/tracking-emission.md`** — its SCORM section described the pre-registry model
  (`passed` at 50, else `failed`, and the legacy `suspend_data` line format) and applied it to
  SCORM 1.2. It now separates the two formats and describes 1.2 as the registry implements it.

## Tests

9 new test files and 43 extended, colocated with the code. Every fix landed with its own test and was
verified by mutation — reverting the fix and checking that the test fails — because several of these
paths are easy to cover with a test that passes just as happily with the bug present.

## How to review it

Check that, after each action performed in the package, **the score and the status (icon) of every
entry in the SCORM menu update correctly** and match the logic described under "Expected behaviour".

The review must be **detailed, iDevice by iDevice**. A general pass over the package is not enough:
each iDevice has its own behaviour and all of them were modified to fit this new logic.

### Moodle updates asynchronously

The package sends the score and the status to Moodle, but those changes **take a moment to show up**
in the SCORM menu — around a second.

For a learner that delay is irrelevant, but while testing, watching for every change, it can feel
much longer. **Wait a moment** before judging what the menu shows as right or wrong.

### Sample packages

**Four SCORM packages** are available to make this easier, covering every gradable iDevice in
different configurations: game mode, time limit, access form, and so on. The point is to be able to
check each one individually and confirm that the status and the score update correctly in every
situation.

[idevices-scorm-1d_scorm.zip](https://github.com/user-attachments/files/32037640/idevices-scorm-1d_scorm.zip)
[idevices-scorm-2d_scorm.zip](https://github.com/user-attachments/files/32037644/idevices-scorm-2d_scorm.zip)
[idevices-scorm-3d_scorm.zip](https://github.com/user-attachments/files/32037645/idevices-scorm-3d_scorm.zip)
[idevices-scorm-4d_scorm.zip](https://github.com/user-attachments/files/32037649/idevices-scorm-4d_scorm.zip)
[idevices-scorm-1bg_scorm.zip](https://github.com/user-attachments/files/32084619/idevices-scorm-1bg_scorm.zip)
[idevices-scorm-2bg_scorm.zip](https://github.com/user-attachments/files/32084621/idevices-scorm-2bg_scorm.zip)
[idevices-scorm-3bg_scorm.zip](https://github.com/user-attachments/files/32084628/idevices-scorm-3bg_scorm.zip)
[idevices-scorm-4bg_scorm.zip](https://github.com/user-attachments/files/32084633/idevices-scorm-4bg_scorm.zip)

Inside these packages there are two kinds of page whose title carries a word saying **how the test
must be run**: **Access** and **Time**.

### Pages whose title ends in "Access"

These pages show an access form before the activity can be reached. The password is **1234**.

On entering it correctly:

- in the SCORM menu the status becomes **Incomplete** and the score **0**, signalling that the
  learner has started the activity;
- the activity **starts automatically**, with no need to press the start button as well. The one
  exception is the iDevices that require a choice before playing: `discover` configured with **2 or 3
  levels** changes the status and the score but shows the level selector instead of starting the
  game, because the learner has to choose which level to play. With **a single level** there is
  nothing to choose and it starts directly.

An **incorrect** password must change nothing — no status, no score — and the password field must be
cleared.

### Pages whose title contains "Time"

On these pages you must **wait for the time to run out** to check that the behaviour is correct.

When the time ends the activity is treated as finished **whether or not the learner answered every
question**, and its status becomes:

- **Passed**, when the score is 50 or above;
- **Failed**, when the score is below 50.

### Pages with several SCORM-enabled iDevices

On pages holding several gradable iDevices with SCORM enabled, two things need particular attention:

- **The page status.** It must not turn Passed or Failed until every question of every SCORM iDevice
  on the page has been answered. While any question is still pending, the page must stay Incomplete.
- **The score weighting.** Check that the weight assigned to each iDevice is correctly reflected in
  the page's final score. To make this easier, each iDevice's title states the weight it contributes
  to the final mark.

So during testing it is not enough to check the mark obtained in each activity on its own: you also
have to verify that the weighted sum of all of them matches the score Moodle shows for the SCO.

### Updated example SCORM packages

The example SCORM packages have been updated. They now include new iDevices with the **"Show a button to save the score"** option enabled, so this manual mode can also be tested alongside the automatic mode and the activities with SCORM score saving disabled.

### Pages whose title contains "Save"

On these pages, the iDevices have the SCORM option **"Show a button to save the score"** enabled.

In this mode, the activity **does not automatically send the score to Moodle** while the learner is playing. The learner can answer questions, see their results and complete the activity, but the result is only recorded when they press the **Save score** button.

During testing, check the following:

- The **Save score** button is visible from the beginning.
- If the learner presses it **before starting the activity**, no score should be sent and they should be told to start the activity first.
- Once the activity has started, pressing the button saves **the score obtained at that moment**.
- The button can be pressed **as many times as required**. Each save replaces the previous score, whether the new score is higher or lower.
- If the learner saves while there are still unanswered questions, the score is recorded, but the status must remain **Incomplete**.
- If the learner completes the whole activity but **does not press Save score**, the result is not sent to Moodle. For the purposes of the page calculation, that iDevice still has no saved score.
- **Restarting** the activity or playing again **must not send any score** in this mode.
- If there is already a saved score from a previous attempt, entering the activity again, starting a new game or restarting it **must not change it**. It is only replaced when the learner presses **Save score** again.
- If the learner leaves without saving, the score shown only inside the iDevice **must not replace the score stored in Moodle**.

When the learner has completed the activity and presses **Save score**, the status becomes:

- **Passed**, if the saved score is 50 or above;
- **Failed**, if the saved score is below 50.

This mode therefore allows **the learner to decide when to submit their result to Moodle**.


erseco (changes requested):
Thank you, Manuel, for this work. Homogenising score reporting across 35 iDevices is one of those jobs nobody notices when it goes right and that drags defects along for years when it is left half done, and this is done with care: every hard decision is argued in the code itself, and the reasoning sits exactly where someone will need to read it two years from now.

I ran the branch through the two test batteries that landed with #2299 and #2310, and through a comparison against `main` so that any difference could be attributed. The full report is attached as a PDF; the summary follows.

📄 **Full report (PDF)**: [Informe-PR2371-pruebas-2026-09-08.pdf](https://github.com/user-attachments/files/31930871/Informe-PR2371-pruebas-2026-09-08.pdf)

## What is good

**There is no grading regression.** The four scenarios of the #2310 trace recorder, played in a real browser under the mod_exelearning serving model with the #105 plugin runtime, produce exactly the same mark as `main` and match the hand-computed oracle. The writes that reach the LMS are identical: same values of `cmi.suspend_data`, `cmi.core.score.raw` and `cmi.core.lesson_status`, in the same order.

| Suite | Result |
|---|---|
| SCORM 1.2 contract (#2299), both runtimes | 120 / 120 |
| Grading fixtures (#2310) | 29 / 29 |
| Trace recorder (#2310), real browser | 4 / 4 |
| scorm-again smoke (#2310) | 2 / 2 |
| Unit tests for `public/app/common/scorm` + the 35 iDevices | 4,912 / 4,912 |
| `make lint` | clean |

Beyond that:

- **The #2209 contract still holds.** No new deviation against the independent oracle.
- **The work ships with tests**: 52 test files touched, 9 new, for 35 iDevice runtime files and 3 core ones.
- **The weighted mean in place of largest-remainder is the right call**, and the comment makes the right argument: the historical distribution made the mark depend on the order the author placed the iDevices in. That settles decision 3, left open by the 27-08 audit.
- **Committing after publishing a mark is well spotted.** Moodle only refreshes its course-structure menu on `LMSCommit`, and that was unresolved.

## What I am asking to change before merging

### Blocking

**1. `exe-scorm12-policy.js:362` — the entry flush commits stale state.**
The flush runs after `activities.load()`, and `load()`'s merge passes the stored score and `completed` as the descriptor over the live ones: `normalize()` only falls back to `previous` when the value is `undefined`, and `decodeRecord()` returns `score: null` and `completed: false`, which are not. An answer given before `loadPage()` finishes is overwritten by the restored record, and the flush commits what was overwritten. The comment already anticipates the problem ("before load() merges the stored payload over them"), but `reportedBeforeEntry` only remembers *that* there was work, not *which*. This path also never calls `persistActivities()`, so the committed `score.raw` and status disagree with `cmi.suspend_data`.

**2. `adaptative-quiz.js:1034` — opening the page publishes a zero over a stored mark.**
`startGame()` always ends in `saveScormScore()`, and inside a SCORM package the path `initScormData` → `maybeStartAfterScorm` → `beginActivity` → `startGame` reaches it with no learner input at all, when there is neither a timer nor an access code. `crossword` and `flipcards` solve exactly this with the `reportScorm` flag, which does not exist here. `complete.js:305` has the same problem via the maximize handler. Suggestion: give `startGame` the same `reportScorm = false` parameter and pass `true` only from the play button and the access code, the way its siblings already do.

**3. `trueorfalse.js:178` — already published packages lose "Try again".**
A missing `attemptsNumber` resolves to 1, `pendingAttempts` is decremented on the first Comprobar, and `gameOver` hides `#tofRebootTest` once it reaches 0. On `main` the button was always shown. The result is that every `.elpx` authored before this branch goes from unlimited retries to a single one, in HTML5 exports too, until the author re-saves each iDevice. AGENTS.md asks that existing features not be removed. Suggestion: let the absence of the field mean "unlimited", and make 1 the default only for new iDevices created from the editor.

**4. `az-quiz-game.js:1293 and 1376` — the last-word condition can never be true.**
The predicate compares against `mOptions.numberWords`, which is assigned nowhere in the repository: `grep -rn numberWords public src` returns only these two lines and the test. It is always false, so the early report the change exists for is never emitted and the window the commit says it closes is still open. The condition the file already uses is `answeredWords >= validWords` (line 1154), and it is the correct one because the rosco revisits skipped words. The test passes because it injects `numberWords: 4` by hand into the options object, so right now it pins a fiction; it would be better for it to build the state through the normal path.

**5. `interactive-video.js:1621` — the mark is committed before the last answer is added.**
`updateResult` calls `resultsViewer.getFinalResult()`, which now drags `finalizeScorm` → `gameOver = true` → `sendScore(true)` → commit, *before* `updateScore()`. The first committed completed report goes out without the last answer's point, and only a second send corrects it; if the page unloads between the two, the LMS keeps the first. Swapping the two lines is enough.

**6. `quick-questions-multiple-choice.js:2003 and 2084` — the predicate ignores itinerary mode and lives.**
`activeQuestion + 1 >= numberQuestions` does not hold in itinerary mode (`order === 2`), where `updateNumberQuestion` jumps to hit or error targets and ends on −2, and it does not cover the lives-exhausted case that `quick-questions.js:1728` and this file's own `newQuestion` do handle. It reports completed too early in itinerary mode and too late with lives.

**7. `padlock.js:308` — the report goes out before the activity is registered.**
In `enable()`, `sendScore(true, instance)` and `startGame(instance)` run before `registerActivity(mOptions)`, so a padlock restored as solved reports while `mOptions` still has no `ideviceId`: the send is dropped on the registry path and written to an undefined key on the legacy one. That is precisely the value the change means to publish. Registering first is enough.

**8. `periodic-table.js:999 and 1036` — the early report ended up inside the timer.**
On both mobile paths the "report now" block sits inside the 5 s reveal `setTimeout`, so it arrives late and merely duplicates `gameMobileOver`'s own report. The desktop wrong-answer terminal branch (1188) got no early report at all. And when the clock expires on mobile, the desktop `gameOver()` is called (1616), which never closes `#ptlLightboxMobile` nor shows `#ptStartGameMobileDiv`, leaving the lightbox open with no play-again control.

### Non-blocking

**9. `3dmol.js:1583` — `startGame` never lowers `gameOver`.**
It resets hits, errors and `gameStarted`, but not `gameOver`. Its twin `electrical-circuits` did get that line in this same branch (1016). On "New game", every intermediate 3dmol report travels with `activityCompleted = true`.

**10. `common.js:1923` — the commits are not coalesced.**
`triggerMoodleDetection` schedules one retry `setTimeout` per report without cancelling the previous one, plus another when the status changed. Measured with the recorder, a page with four activities goes from 0 commits on `main` to 6. The intent is right and well argued; what is missing is coalescing: a single per-page timer, rescheduled.

## Two things that are not defects, but worth confirming

- **The padlock scores 10 through any of the three ways in**, including the clock running out with the learner having typed nothing. The comment defends it well (a padlock is a gate, not a question) and nothing was reported at all in that case before. I only want it said out loud, because it changes marks.
- **The default weight moves from 1 to 100.** Together with the weighted mean, that re-scores attempts stored under the old rules. `doc/development/scorm12-runtime-contract.md` and ADR-2209-02 still say the move is "at most one weight-point"; that sentence needs updating, or the change needs bounding.

## A note on method

A warning for anyone reproducing this. The scorm-again smoke initially failed on the scenario of a scored page left unanswered, writing `cmi.core.score.raw = 0` where the contract requires no write at all, and it passed on `main`. It looked like the most serious regression of the batch, and **it was not**: reverting the iDevice, then `common.js`, then the SCORM runtime, and finally the whole of `public/` did not fix it, and this branch does not touch `src/`. The cause was the worktree's stale bundles. With `make bundle` both scenarios go green. The E2E tests run against what is built, not against the source.

erseco:
## Code review — SCORM score reporting homogenisation

Automated review of the branch diff. 14 findings, most severe first. Line numbers refer to the branch head.

### Correctness

**1. `trueorfalse` loses unlimited retries on already-published content** — `public/files/perm/idevices/base/trueorfalse/export/trueorfalse.js:917`

The new `attemptsNumber` field defaults to `1` (lines 176-179) and is decremented on every check. Any `.elp` saved before this branch has no `attemptsNumber`, so `addEvents` (line 696) sets `pendingAttempts = 1`; the learner presses *Comprobar* once, `gameOver` decrements to 0 (lines 849-851) and line 920 hides `#tofRebootTest`. Quizzes that were retryable before the upgrade no longer are, and no editor setting on the stored content can restore it. This contradicts this same branch's `repeatActivity = true` / "let every activity be repeated" direction and AGENTS.md's "Do not remove existing features". Suggest defaulting a missing `attemptsNumber` to unlimited.

**2. `exit=""` survives into a reopened attempt closed abruptly** — `public/app/common/scorm/scorm12/exe-scorm12-policy.js:194`

`clearExitWhenTerminal` writes `cmi.core.exit = ""` mid-session as soon as the attempt turns terminal, but nothing rewrites `"suspend"` when the attempt is later reopened except the unload-time exit policy. Sequence: learner finishes a page → `applyDecidedStatus` writes `passed` + `writeExit('')`; learner restarts an activity → `reconcilePendingActivities` downgrades to `incomplete`, and `clearExitWhenTerminal('incomplete')` writes nothing, so the LMS still holds `exit=""`. If the tab is then closed, the mobile browser kills it, or Moodle navigates the iframe without firing `pagehide`, `applyExitPolicy` never runs: the LMS ends the attempt as a normal completion and `cmi.suspend_data` is not restored on re-entry. Before this change no exit was written in that window, so the previous `"suspend"` survived.

**3. Activity-weight default diverges: registry 1 vs `common.js` 100** — `public/app/common/scorm/scorm12/exe-scorm12-activities.js:190`

The "no usable weight means 100" change landed only in `common.js` (`reportActivity`, `getFinalScore`). The registry still defaults a missing weight to `1` in `normalize()` (190/203), `decodeRecord()` (335), `decodeLegacyPool()` (368-371) and the legacy migration (683). A page played under the old build (or a pre-rewrite package) stores weight 1 in `cmi.suspend_data`; on resume `activities.load()` merges the stored payload *over* the live records, so both go back to 1. Replaying one activity re-registers it at weight 100 while the untouched one stays at 1, so `aggregateScore()` returns essentially the replayed activity's score alone (100×1 vs 1×1) instead of the mean — and `cmi.core.score.raw` plus the mastery comparison follow it. The `pendingReports` re-apply loop in `applyEntryPolicy` does not carry `weight` either.

**4. `form`: Reset/Start overwrite the manually saved score with 0** — `public/files/perm/idevices/base/form/export/form.js:878`

`$form.saveScormScore` gates on `data.isScorm > 0`, whereas every sibling iDevice in this PR gates on `isScorm !== 1` precisely so manual mode is never auto-submitted. In manual SCORM (isScorm 2): the learner presses *Save score* and the LMS stores 90; they then press Reset → `rebootGame` → `saveScormScore(data)` passes the `> 0` gate → `sendScore` → `sendScoreNew(true, data)` → `updateActivity` writes `score.raw = 0` and commits. `startGame(data, true)` from `#frmStartGame` does the same.

**5. `beforeafter`: entering the access code auto-submits a 0 in manual SCORM** — `public/files/perm/idevices/base/beforeafter/export/beforeafter.js:705`

The new `enterCodeAccess` → `startGame(instance)` path routes through a `startGame` that reports on `isScorm > 0` (line 636 region). With manual SCORM (isScorm 2) behind an access code, typing the code fires `sendScore(true, instance)` → `score.raw = 0` + commit, replacing whatever the learner had previously saved, before they have interacted with the activity at all.

**6. `interactive-video`: NaN score shown when no scoring slides** — `public/files/perm/idevices/base/interactive-video/export/interactive-video.js:631`

`reportScore()` is now called at activity start and divides by `numSlides`, which is 0 for a SCORM-enabled video with no scored questions. In `cover.hide()`, `activeSlide` stays 0 and `scoreNIA` is false → `(0 * 10) / 0` → `NaN` → `.toFixed(2)` → `"NaN"` rendered in `.Games-RepeatActivity`. `sendScoreNew`'s `Number.isFinite` guard keeps the bad value out of the LMS, but the learner sees NaN for the whole visit.

**7. `geogebra-activity`: pass/fail message fallbacks are swapped** — `public/files/perm/idevices/base/geogebra-activity/export/geogebra-activity.js:522`

In the `getOptions` block this diff cleaned up, `msgUnsuccessfulActivity` falls back to `'Activity: Passed. Score: %s'` and `msgSuccessfulActivity` to `'Activity: Not passed. Score: %s'`. An activity missing `messagesEval[1]`/`[2]` tells a failing learner they passed, and vice versa.

**8. `discover`: access-code single-level game labels itself "Master"** — `public/files/perm/idevices/base/discover/export/discover.js:1700`

The new access-code path calls `startGame(instance, 2)`, but level index 2 is the *Master* label; every other single-level entry point uses 0 (*Rookie*). With `gameLevels == 1`, `getCardsLevels` ignores `nivel` when picking cards but still sets the on-screen label from it, so `#descubreInfo` reads "Level: Master" — while the play-again handler for the same game (lines 1039-1041) uses `startGame(instance, 0)` → "Level: Rookie".

### Robustness

**9. `writeExit` lacks the capability guards its siblings have** — `public/app/common/scorm/scorm12/exe-scorm12-policy.js:167`

It calls `client.hasWrittenValue(EXIT)` and `client.getCachedValue(EXIT)` unguarded, while `commitSession` in `common.js` — added in this same PR — wraps each client method in `typeof … === 'function'` because the Moodle plugin injects its own vendored copy of the runtime, possibly from a different release. Against such a client `writeExit` throws a TypeError, and it is reached from `applyDecidedStatus` (every activity report) and from `applyExitPolicy` at unload — the one path with no retry.

**10. `activities.update(...)` sits outside the `if (activities)` guard** — `public/app/common/scorm/scorm12/exe-scorm12-policy.js:436`

Currently safe only because `reportedThisSession(null)` returns `[]`, so the loop body never runs without a registry. Any change that lets `reportedThisSession` return records from another source, or that recomputes `pendingReports` after a re-read of `deps.getActivities()`, dereferences null on pages with no registry (SCORM 2004, pre-rewrite packages) and throws out of `applyEntryPolicy`, leaving `state.entryApplied` false with the session half-initialised.

### Efficiency

**11. `electrical-circuits`: every answer reports twice** — `public/files/perm/idevices/base/electrical-circuits/export/electrical-circuits.js:1436`

`answerQuestion`/`answerQuestionThree` now call `saveScormScore` on every answer while `showQuestion` (line 1311) still reports after the reveal delay. A 10-question activity issues ~20 `sendScoreNew` cycles instead of 10; each cycle does two `LMSGetValue('cmi.core.lesson_status')` reads (`common.js:1748`, `1778`), one synchronous `LMSCommit` from `updateActivity`'s `finally`, and one deferred 1200 ms `LMSCommit` timer (`common.js:1883`) — ~40 status reads, ~20 synchronous commits and ~20 pending timers for one activity. The paired reports carry identical data; nothing changes between them but the reveal animation. The same duplicate-report pattern was added to `3dmol`.

### Maintainability

**12. `discover`: `gameStarted` is toggled three times to bypass a gate** — `public/files/perm/idevices/base/discover/export/discover.js:1966`

`mOptions.gameStarted = true; saveScormScore(); mOptions.gameStarted = false;` appears at 1063, 1714 and 1966, purely to smuggle a report past `sendScoreNew`'s `if (game.gameStarted || game.gameOver)` gate. This patches the caller instead of giving `sendScoreNew` an explicit way to report an abandoned attempt: the invariant "gameStarted reflects whether a round is running" is deliberately violated for the duration of a synchronous call, and anything `sendScoreNew` reaches that re-enters the iDevice — or any future async step inside it — observes a lie. Each of the three sites re-derives the reasoning in its own long comment rather than sharing one helper (e.g. `reportAbandonedAttempt(instance)`), and the shape has already spread to `mathematicaloperations`' `reloadGame`.

**13. Dead `initialScore` write kept in 4 iDevices, deleted in 4** — `public/files/perm/idevices/base/electrical-circuits/export/electrical-circuits.js:1137`

The "score only once" lock was removed but the now-dead `initialScore` write was kept and merely relocated from `$eXeEC.initialScore` to `mOptions.initialScore`; same in `mathematicaloperations:1823`, `quick-questions:1264` and `quick-questions-multiple-choice:1561`, while `3dmol`, `mathproblems`, `quick-questions-video` and `select-media-files` deleted theirs. Nothing reads `initialScore` any more (its only reader was the removed `|| $X.initialScore === ''` condition), yet the module-level `initialScore: ''` field is still declared — so the eight files now disagree on whether an iDevice tracks an initial score.

### Test coverage

**14. `aggregateScore()` JSDoc cites a spec file that does not exist** — `public/app/common/scorm/scorm12/exe-scorm12-activities.js:236`

The doc comment asserts that `exe-scorm12-activities.spec.js` "pins them against each other". That file does not exist — frontend tests are `*.test.js` per AGENTS.md §5.1 — and no test cross-checks `aggregateScore()` against `common.js`'s `getFinalScore()`. `exe-scorm12-activities.test.js` only tests the registry's own aggregate and `common.test.js` only tests `getFinalScore` in isolation, so the invariant the comment claims is enforced is exactly the one finding 3 shows is already broken.

mnarvaezm:
Ernesto, thank you very much for your reports. They are extremely useful. I’m reviewing them one by one, incorporating the changes where appropriate or explaining the current behavior when necessary. I’ll start with the True/False iDevice.

Regarding attempts, I think the main issue here is consistency with the rest of eXe.

In 2.9, attempts did not exist in this iDevice: neither the concept nor the button.

In `main`, there is no setting for them, and the `#tofRebootTest` button is displayed unconditionally after every check. In practice, this means attempts are unlimited and the author has no way to restrict them.

The problem is that this makes True/False an exception compared with the rest of the product. In `main`, there are already eight iDevices that keep track of attempts — `complete`, `discover`, `flipcards`, `padlock`, `scrambled-list`, `select-media-files`, `sort`, and `word-search` — and in six of them the author can configure the number of attempts from the editor.

In addition, `scrambled-list` already uses exactly the same approach in `main` as the one applied here: from 1 to 9 attempts, with 1 as the default value.

That is why the same mechanism has been added to True/False, using the same range and the same default value. This is not about removing an existing feature, but about making this iDevice behave consistently with the others.

It also does not contradict the `feat` in this branch related to repeating activities: `repeatActivity` determines whether the score can be sent to the LMS again, while `attemptsNumber` determines how many checks the iDevice allows within an attempt. They are separate concerns.

As for the “Try again” button, it has not been removed. It is simply shown or hidden depending on the iDevice configuration, just as many other elements in eXe are. If the author configures only one attempt, the button would have no possible function, so it is not displayed.

I’ll continue reviewing the rest and will keep you posted.
@erseco  @ignaciogros @cristinavaldera

cristinavaldera (approved):
We have reviewed the behaviour of all the idevices from the 4 zips attached regading to the change of state and the calification.

Everything is working as expected. Thank you!!!

ignaciogros:
The manual tests have produced good results in all cases. I think that, after reviewing @erseco’s and @cristinavaldera’s latest comments, the proposal is ready to be merged into `main`.

In fact, I think it would be best to do it soon and handle the problems of each iDevice (which are already quite detailed) in separate issues. If we haven’t made any mistakes in our testing, any problems that arise would be minor, and the current proposal represents a significant improvement.

I think it makes sense to merge this as soon as possible.

erseco:
Re-reviewed at `156bae5a0`, against the ten points and the two confirmations I left on 07-09. Almost all of it is done, and done properly — several of the fixes are better than what I asked for, because they moved the rule into the runtime instead of patching the iDevice that happened to show it.

## The previous review, point by point

| # | Point | Status |
|---|---|---|
| 1 | `exe-scorm12-policy.js` — the entry flush commits stale state | **Fixed** |
| 2 | `adaptative-quiz` / `complete` — opening the page publishes a zero | **Fixed** |
| 3 | `trueorfalse` — already published packages lose "Try again" | **Fixed** (as a product decision, argued in code) |
| 4 | `az-quiz-game` — the last-word condition can never be true | **Fixed** |
| 5 | `interactive-video` — the mark is committed before the last answer | **Fixed** |
| 6 | `quick-questions-multiple-choice` — predicate ignores itinerary and lives | **Fixed** |
| 7 | `padlock` — the report goes out before the activity is registered | **Fixed** |
| 8 | `periodic-table` — the early report ended up inside the timer | **Fixed** (all three sub-points) |
| 9 | `3dmol` — `startGame` never lowers `gameOver` | **Fixed** |
| 10 | `common.js` — the commits are not coalesced | **Still open** |
| C1 | The padlock scores 10 through any of the three ways in | **Resolved** — the mark now follows the outcome |
| C2 | The default weight moves from 1 to 100 | **Resolved** — documented in the runtime contract |

What I checked, rather than took on trust:

- **1** — `applyEntryPolicy()` now captures the pre-session reports as records (`reportedThisSession()`), re-applies them after `activities.load()`, and calls `persistActivities()` before `applyDecidedStatus()` and the commit. Both halves of what was wrong are gone.
- **2** — `startGame(id, reportScorm = false)` in `adaptative-quiz`, with `true` only from the play button and the access code; `complete`'s maximize handler is now gated on `!gameStarted && !gameOver`.
- **4** — `numberWords` no longer exists anywhere in the repository; both branches use `answeredWords >= validWords`.
- **5** — `updateResult()` now runs `updateScore()` before `getFinalResult()`.
- **7** — `registerActivity()` at `padlock.js:327` precedes both `sendScore()` and `startGame()`.
- **8** — the report sits outside the 5 s reveal on both mobile branches, the desktop out-of-attempts branch has one at last, and the mobile clock expiry calls `gameMobileOver()` instead of the desktop `gameOver()`.
- **C1/C2** — `showFeedback()` leaves the mark alone (10 opened, 0 timed out) and reports nothing on page load; `doc/development/scorm12-runtime-contract.md` now says outright that the default weight moved from 1 to 100 and works the arithmetic through.

## What is still blocking

### 1. `relate.js:920` — opening the page publishes a zero over a stored mark

This is exactly the defect of my previous point 2, in an iDevice that did not get the treatment. `init()` → `loadGame()` → `addEvents(i)` ends with:

```js
if (mOptions.type < 2 && !mOptions.itinerary.showCodeAccess) {
    $eXeRelaciona.startGame(instance);
}
```

and `startGame()` now ends in `saveScormScore()` (`relate.js:391`) with no `reportScorm` parameter to hold it back. So a relate iDevice that is untimed and has no access code publishes a 0 with no learner input at all. `crossword`, `flipcards`, `select-media-files`, `adaptative-quiz` and `dragdrop` all gate that call behind the flag; `relate` is the one that does not. Same fix as its siblings.

I swept the other 34 for the same shape. The only other synchronous start on a load path is `identify.js:100`, and `identify`'s `startGame()` does not report — so `relate` is the single instance.

### 2. `exe-scorm12-policy.js:487` — the replay is what turns that into data loss

Worth separating from the iDevice, because the iDevice is only the trigger. The replay loop runs over anything `reportedThisSession()` returns, and that predicate accepts any record with `score !== null` — including the 0 an activity publishes when the page starts it for the learner. The loop applies it over the freshly restored attempt, and `persistActivities(); applyDecidedStatus(); client.commit();` makes it permanent in one visit. A learner who scored 80, left, and comes back to the page loses it.

The comment above the loop argues the case honestly — an iDevice cannot resume, so interacting again really is a new attempt — and I agree with it *for a start the learner asked for*. What it cannot see is the difference between that and a start the page performed on its own. Right now the whole runtime's correctness rests on all 35 iDevices never reporting on their load path, enforced by hand, one file at a time; `relate` is what that costs. Carrying a `startedByLearner` flag on the report (or refusing a pre-session report whose score is 0 and whose `answered` is 0) puts the rule in the one place that can hold it.

### 3. The terminal report goes out twice in eight iDevices

`3dmol` and `electrical-circuits` got the right shape:

```js
const alreadyReportedFinished = mOptions.gameOver === true;
```

Nothing else did — `alreadyReportedFinished` appears in those two files and nowhere else. Everywhere else the new "raise the flag and report now" block runs, and then the `gameOver()` the reveal delay calls reports the same numbers again: `az-quiz-game.js:900`, `periodic-table.js:1706` and `:1753`, `guess.js:982`, `quick-questions.js:1260`, `quick-questions-multiple-choice.js:1557`, `mathproblems.js:1093`, `sort.js:2103`. `hidden-image` is the same defect one level down — the report was added to `answerQuestion()` (`:1152`) but the one in `showQuestion()` (`:780`) was not removed, so it reports twice per answer and three times on the last one.

Each duplicate is a full `sendScoreNew` → `updateActivity` → `commitSession()` plus a `triggerMoodleDetection` pair, so on Moodle it is a duplicated HTTP round-trip per answer. Nothing is graded wrong, but it doubles the traffic the PR set out to make deliberate.

### 4. Point 10 is still open, and these three make it worse

`triggerMoodleDetection` (`common.js:1961`) still schedules a fresh `setTimeout` per report with no `clearTimeout` of the previous one. With the duplicate reports above, a twenty-question quiz queues forty-odd deferred commits, all firing after the learner has moved on. A single timer id on the scorm object, rescheduled, gives the same "carry whatever settled after the last write" guarantee at one commit per burst. The timers also outlive `lifecycle.finish()`; that is harmless today only because the vendored pipwerks wrapper guards on `connection.isActive`.

## Minor, not blocking

- `interactive-video.js:164` — `getIdeviceRoot()` returns `$('.exe-interactive-video').eq(0)`. Two interactive videos on one page and the second one's save button gets no handler at all. Resolving the root from the button's own `.idevice_node`, the way the rest of the PR does, costs nothing.
- `rubric.js:1618` — `showScormMessage()` is new, and unreferenced: the same commit removed the refusal path that was its only caller.
- `mathematicaloperations.js:1827`, `quick-questions.js:1264`, `quick-questions-multiple-choice.js:1561` — `initialScore` is assigned and read by nobody. `7dd655c0d` documents it instead of removing it; deleting the property and its writers is the honest version.
- Eight comments still state the completion rule as `gameOver === true || auto !== true`, which `397ba2fa5` replaced with `gameOver === true` in `sendScoreNew`: `3dmol.js:913`, `electrical-circuits.js:689`, `identify.js:952`, `mathproblems.js:1190`, `puzzle.js:661`, `padlock.js:412`, `challenge.js:1043`, `interactive-video.js:577`. Some were written after the rule changed.
- The silent-video time fields widened from one digit to three without the `focusout` clamp their siblings got, so an author can store more than the `max="120"` the input declares: `quick-questions-multiple-choice.js:2750`, `quick-questions.js:2499`, `trivial.js:2705`, and the new handler at `guess.js:1754`.
- `periodic-table.js:999` and `:1032` — the new blocks are indented as if the inner `if` belonged to the following `setTimeout`. It parses fine; it reads wrong. Nothing will catch it, since `lint:public` only covers `public/app/`.

## Where that leaves it

Ten of the twelve points are closed, and the two that matter most — the entry flush and the periodic-table timing — are closed well. I am not approving yet, on one item: `relate` plus the replay loop is a silent, permanent loss of a learner's stored score, which is the failure this PR exists to prevent. Fix `relate` the way its five siblings are fixed and I will re-run the batteries; moving the rule into the policy would close the class rather than this instance, and I would rather see that. The duplicate reports and the commit coalescing I would take in a follow-up if you prefer, as long as they are tracked.

Thanks again for the care in this — the reasoning left in the code is what made this second pass quick.

erseco:
Pulled and re-reviewed at `323dca98a` — three new commits since my last pass.

## The three new commits

All three are correct, and the two puzzle bugs are good finds. What I checked rather than took on trust:

**`6fee3a8be` — record an unusable mark as zero, not as NaN.** The guard belongs in `saveEvaluation()` and `Number.isFinite` is the right predicate: `state` is decided by `parseFloat(score) >= 5`, which a NaN fails, so a passed activity was being iconed as failed; `!isNaN` would have let the `Infinity` from a zero count straight through. `showEvaluationIcon()` re-parses anyway, so narrowing the stored value to a number breaks nothing downstream.

**`b34b2a97e` — require a question when the score is saved.** `hasScorableSlide()` matches the runtime exactly: the same six types the export counts at `interactive-video.js:706-711`, and the `scoreNIA` short-circuit mirrors `numSlides = scoreNIA ? scoreSlides.length : activeSlide`. Placing the refusal above the `slide.url` rewrite loop is the right call and the comment says why. Hoisting `getValues()` and the checkbox read into `scormValues` / `scoreNIA` is a real simplification, not just motion.

**`323dca98a` — report the replay at once and lock the board until it is built.** Three separate defects, all real:
- `PZLP-TileChang` matched nothing — a class selector matches whole tokens, so every swap puzzle kept the previous attempt's tiles through the rebuild. `PZLP-TileChange` is what the other seven references in the file spell.
- The replay published nothing: `sendScoreNew` drops a game that is neither started nor over, and `showPuzzle()` does not raise `gameStarted` — `placePuzzlePieces()` does, off the image's load event. Raising it explicitly before the report is right, and "intermittent because a cached image can fire load first" is the correct explanation.
- `gameActived = true` matches the polarity the click handlers read at `:1270` and `:1306`, and `placePuzzlePieces()` still clears it last (`:613`). `selectedTile = null` is needed for the same reason.

## Blocking: `make test-unit` is red

The three tests `6fee3a8be` adds to `common.test.js` cannot pass in this repo's environment:

```
FAIL public/app/common/common.test.js > gamification.report > saveEvaluation with an unusable mark > records a zero for a division by zero
FAIL public/app/common/common.test.js > gamification.report > saveEvaluation with an unusable mark > records a zero for a count of zero
FAIL public/app/common/common.test.js > gamification.report > saveEvaluation with an unusable mark > leaves a usable mark alone

TypeError: Cannot read properties of undefined (reading 'removeItem')
 ❯ givenActivity public/app/common/common.test.js:3669:22

Test Files  1 failed (1)
Tests       3 failed | 437 passed (440)
```

`localStorage` does not exist under this project's happy-dom setup — not bare, not on `window`, not on `globalThis`; I probed all three. `vitest.setup.js` does not provide one either. The repo's own pattern is to install a stub, as `exe_export.test.js:162` does:

```js
window.localStorage = setupLocalStorageStub();
```

`bun test:frontend` runs the same config, so CI's frontend job fails on this too. The fix is small, but it is Definition-of-Done item 2, so it has to land before this can go in.

## Still open from my previous comment

Verified unchanged at this head, none of them touched by the three new commits:

1. **`relate.js:920`** — still calls `startGame()` from `addEvents()` on the load path, and `startGame()` still ends in `saveScormScore()` (`:391`) with no `reportScorm` parameter. Still the one iDevice of the 35 that publishes a zero with no learner input.
2. **`exe-scorm12-policy.js:487`** — the replay loop still accepts any record with `score !== null`, so that zero still lands over the restored attempt and is persisted and committed. This is the one I would most like to see closed at the policy rather than per-iDevice.
3. **Duplicate terminal reports** — `alreadyReportedFinished` still appears only in `3dmol.js` and `electrical-circuits.js`; the other eight (`az-quiz-game`, `periodic-table`, `guess`, `quick-questions`, `quick-questions-multiple-choice`, `mathproblems`, `sort`, `hidden-image`) still report twice.
4. **`common.js:1961`** — `triggerMoodleDetection` still schedules a `setTimeout` per report with no `clearTimeout`.

## Where that leaves it

The new work is good and I have no changes to ask for in it. The verdict does not move, though: the red unit tests are a hard gate, and `relate` plus the replay loop is still a silent, permanent loss of a learner's stored mark. Fix the `localStorage` stub and `relate`, and I will re-run the full batteries — the other three I am happy to see tracked as follow-ups if you would rather land this.

erseco:
Re-reviewed at `5c2dc113d` — two more commits.

## The two new commits

**`c9194d322` — pin the replay report ahead of the image's load event.** The source change is the comment only; the code is unchanged. The new wording is the more accurate of the two: setting `src` queues the load event as a task, so it cannot overtake the rest of a synchronous handler *even for a cached image* — the previous comment's "it looked intermittent because a cached image can fire load before the thread gets here" was the weaker claim. The tests earn their place: unlike the ones above them they let `showPuzzle` and `showImagePuzzle` run for real and assert the whole tuple (`gameStarted: true, gameOver: false, gameActived: true, pictureArrived: false`), so they pin the ordering rather than the stub. One nit: the comment says "jsdom does not fetch the image" — this harness is happy-dom.

**`5c2dc113d` — give the harness the local progress report.** Right call, and the reason is well stated: a 500 ms timer firing after the test that armed it has cleared its own mocks is exactly the shape that fails a whole run with every test passing. `report` belongs next to the other seven `mockGamification*` objects, and a spec that wants to assert on it still installs its own.

## The blocker from my last comment is not fixed

`localStorage` is still the only thing red in the frontend suite. Full run at this head, with dependencies installed:

```
FAIL public/app/common/common.test.js > gamification.report > saveEvaluation with an unusable mark > records a zero for a division by zero
FAIL public/app/common/common.test.js > gamification.report > saveEvaluation with an unusable mark > records a zero for a count of zero
FAIL public/app/common/common.test.js > gamification.report > saveEvaluation with an unusable mark > leaves a usable mark alone

TypeError: Cannot read properties of undefined (reading 'removeItem')
 ❯ givenActivity public/app/common/common.test.js:3669:22

Test Files  1 failed | 295 passed (296)
Tests       3 failed | 15944 passed (15947)
```

`5c2dc113d` was the natural place to fix it — it is the same harness file, and the same class of problem — but it adds `report` and no `localStorage`. Node says so itself while the suite runs: `ExperimentalWarning: localStorage is not available because --localstorage-file was not provided`. The repo's pattern is `window.localStorage = setupLocalStorageStub()` (`exe_export.test.js:162`); putting one in `vitest.setup.js` would close it for good.

Everything else is green, and I want to be precise about that, because my previous comment was measured on a worktree that turned out to be missing `scorm-again`:

| Suite | Result |
|---|---|
| `public/app/common/scorm` (contract + runtime, 15 files) | 874 / 874 |
| Frontend suite, everything else | 15,944 / 15,947 |
| `make lint` | clean (43 pre-existing warnings, no errors) |

The three contract-suite failures I saw before were my own missing dependency, not the branch — `bun install` clears them. `focusedEditMode.test.js` also passes on the clean run. So `common.test.js` is the whole of it.

## Still open

Unchanged at this head, none of them touched by these two commits:

1. **`relate.js:920`** — still starts from `addEvents()` on the load path into a `saveScormScore()` with no `reportScorm` parameter (`grep -c reportScorm relate.js` → 0).
2. **`exe-scorm12-policy.js:487`** — the replay loop still re-applies any record with `score !== null` over the restored attempt, then persists and commits it.
3. **Duplicate terminal reports** — `alreadyReportedFinished` still in 2 files of the 35.
4. **`common.js:1972`** — `triggerMoodleDetection` still schedules per report with no `clearTimeout`.

## Where that leaves it

The new work is good; the verdict does not move. One stub and `relate` stand between this and an approval — everything else on the branch is green and I have re-run it end to end. Fix those two and I will do the final pass; the duplicate reports and the commit coalescing I am happy to see tracked as follow-ups.

erseco (changes requested):
Superseding my review of 07-09, whose list is now mostly out of date. Re-reviewed at `5c2dc113d`.

**Nine of the ten points from that review are closed**, and the two I only asked to have confirmed are resolved. I checked each one in the code rather than by commit message; the detail is in [this comment](https://github.com/exelearning/exelearning/pull/2371#issuecomment-5617022052), and the two passes over the commits that followed are [here](https://github.com/exelearning/exelearning/pull/2371#issuecomment-5617912750) and [here](https://github.com/exelearning/exelearning/pull/2371#issuecomment-5618929077).

Two things stand between this and an approval:

1. **`make test-unit` is red.** The three tests `6fee3a8be` adds to `common.test.js` call bare `localStorage`, which does not exist in this project's happy-dom harness — not bare, not on `window`, not on `globalThis`. The repo's pattern is `window.localStorage = setupLocalStorageStub()` (`exe_export.test.js:162`); `vitest.setup.js` is the place for it. Everything else is green: 874/874 across `public/app/common/scorm`, 15,944/15,947 over the rest of the frontend suite, `make lint` clean.

2. **`relate.js:920` publishes a zero over a stored mark.** `addEvents()` calls `startGame()` on the load path and `startGame()` ends in `saveScormScore()` (`:391`) with no `reportScorm` parameter — the same defect the branch fixed in `adaptative-quiz`, `crossword`, `flipcards`, `select-media-files` and `dragdrop`. `applyEntryPolicy()` then replays that zero over the restored attempt and commits it (`exe-scorm12-policy.js:487`), so the learner's stored score is gone for good. I swept the other 34 iDevices: `relate` is the only one left. Closing it in the policy rather than per-iDevice would end the class, and I would prefer that, but the `reportScorm` flag alone unblocks the PR.

Two smaller things I am happy to see tracked as follow-ups rather than fixed here: the terminal report going out twice in eight iDevices (`alreadyReportedFinished` is in 2 files of the 35), and `triggerMoodleDetection` scheduling a commit per report with no `clearTimeout` (`common.js:1972`).

Thanks for the pace on this — nine points in three days, each with its reasoning left in the code, is why the second pass took an afternoon instead of a week.

cristinavaldera (changes requested):
All the idevices with "Save button" configured work as expected. Just Drag and Drop idevice seems to be malfunctioning (not possible to drag)

erseco (approved):
Approving at `1325f3379`. Both blockers from my 10-09 review are closed; I checked each in the code and re-ran the batteries rather than reading commit messages.

## Closed

**`make test-unit` is green.** `042fb2bd2` bridges `localStorage` in `public/vitest.setup.js`, guarded on it being missing, preferring `window`'s, with the reason it cannot live in a `beforeEach` written down. Measured at this head:

| Suite | Result |
|---|---|
| Frontend (`vitest run`) | 15,987 / 15,987 across 296 files |
| Backend (`bun test ./src ./test/helpers ./scripts ./app`) | 8,752 / 8,752 |
| `make lint` | clean (43 pre-existing warnings) |

**`relate` no longer publishes a zero over a stored mark.** `bda54ee04` gives `startGame` a `reportScorm` parameter and passes `true` only from the three learner entries (`:687` timed, `:840` start button, `:1561` access code); the load path at `:922` is silent. `sort` had the same defect in a shape my sweep could not catch — `loadGame()` triggered the start button's `.click()` instead of calling `startGame` — and it is fixed too, with the button's hide done by hand. Good catch. The E2E in `scorm12-sco-runtime.spec.ts` pins it: 80 stored, reload, mark intact, learner restart reports.

The class-level fix is not there — `exe-scorm12-policy.js` is untouched, so the replay loop still accepts any record with `score !== null` and the rule still rests on all 35 iDevices staying silent on their load path. I would still prefer that closed in the policy, but not at the cost of holding this.

## The puzzle work (`364c6d9c1`)

Three real finds. The hit, score and completion now settle in `checkIfSolved()` instead of after the (rows x columns + 1) x 300 ms reveal, so a learner who leaves or presses save during it no longer stores a mark one puzzle short. `canSendScore()` asks the two questions `gameStarted` cannot. And `gameOver(1)` -> `gameOver(0)` means whoever finishes is no longer shown the losing image.

Three small things in that block:

- `puzzle.js:1244` — the inline `alert` drops half of the refusal. `sendScoreNew` does `$repeatActivity.text(message).show()` *and* `alert(message)` (`common.js:1827-1829`); this one only alerts, which is exactly the presentation the helper you removed existed to keep identical.
- `puzzle.js:1040-1046` — `saveScormScore`'s doc block is now orphaned above `canSendScore`, which brought its own; `saveScormScore` (`:1082`) has none.
- `puzzle.js:1080-1081` — double blank line.

## Follow-ups, to track rather than fix here

1. Duplicate terminal reports: `alreadyReportedFinished` is in 2 files of the 35, so `az-quiz-game`, `periodic-table`, `guess`, `quick-questions`, `quick-questions-multiple-choice`, `mathproblems`, `sort` and `hidden-image` still report twice.
2. `common.js:1961` — `triggerMoodleDetection` still schedules a commit per report with no `clearTimeout` of the previous one.
3. Move the pre-session guard into `exe-scorm12-policy.js:487`, so a start the page performed cannot be replayed over a restored attempt.

## Minor, unchanged

- `interactive-video.js:164` — `getIdeviceRoot()` returns `.eq(0)`: a second interactive video on the page gets no save handler.
- `rubric.js:1618` — `showScormMessage()` has no callers.
- `initialScore` is assigned and read by nobody: `mathematicaloperations.js:1829`, `quick-questions.js:1264`, `quick-questions-multiple-choice.js:1561`, and now `puzzle.js:1005` and `:1841`.
- Seven comments still state the completion rule as `gameOver === true || auto !== true`: `identify.js:952`, `mathproblems.js:1190`, `padlock.js:412`, `3dmol.js:913`, `interactive-video.js:577`, `challenge.js:1043`, `electrical-circuits.js:689`.
- The silent-video time fields still have only a `keyup` handler and no `focusout` clamp, so a value above the declared `max="120"` can be stored: `quick-questions-multiple-choice/edition:2750`, `quick-questions/edition:2499`, `trivial/edition:2705`, `guess/edition:1754`.
- `periodic-table.js:999` and `:1032` — both new blocks are still indented as if the inner `if` belonged to the following `setTimeout`.

## Not mine to close

@cristinavaldera's report of 11-09 (Drag and Drop not draggable) is the newest thing here and nothing has landed since — please settle it before merging. For what it is worth: the whole `dragdrop` diff against `main` is 35 lines and none of it touches drag initialisation, and the CSS fix is inert (`.DADP-Definition` appears in `dragdrop.css` and in no HTML in the repository). The known way a board comes up undraggable is `initializeDragAndDrop` giving up after 10 x 200 ms when jQuery UI is not ready, which is unchanged here. Worth reproducing against `main` before attributing it to this branch either way.

Thanks for the pace on this — nine points and then two blockers, each with its reasoning left in the code.

cristinavaldera:
> @cristinavaldera's report of 11-09 (Drag and Drop not draggable) is the newest thing here and nothing has landed since — please settle it before merging. For what it is worth: the whole `dragdrop` diff against `main` is 35 lines and none of it touches drag initialisation, and the CSS fix is inert (`.DADP-Definition` appears in `dragdrop.css` and in no HTML in the repository). The known way a board comes up undraggable is `initializeDragAndDrop` giving up after 10 x 200 ms when jQuery UI is not ready, which is unchanged here. Worth reproducing against `main` before attributing it to this branch either way.

It seems to be a problem in the scorm package used for testing, a library was missing.

- In main it works perfectly. 
- When opening the test SCORM package in the PR branch and exporting again as SCORM, it works. 
So, we do not need to look at this anymore.

Thank you!

cristinavaldera (approved):

PR #2374, opened 2026-09-08
Document AI policy for first-time contributors

erseco:
## Summary

- document that first-time contributions must be the contributor's own work
- clarify that generative AI must not substantially produce the implementation, tests, PR description, or review responses for a first contribution
- explain that `good first issue` tasks are intended to help contributors learn the codebase and demonstrate understanding
- state that non-compliant first-time PRs may be closed

This is intentionally scoped to **first-time contributors** rather than establishing a blanket ban on AI-assisted development in the project.

## Architecture decision

Adds `ADR-2374-01` as a proposed decision, as required by the repository's architecture policy for changes to AI-assisted generation policy. The ADR records the rationale, alternatives, consequences, enforcement risks, and a reference to the comparable Strimzi `good-start` policy.

## Testing

Documentation-only change.

### DIFF
--- a/doc/architecture/adr/ADR-2374-01-require-human-authored-first-contributions.md
+++ b/doc/architecture/adr/ADR-2374-01-require-human-authored-first-contributions.md
@@ -1,0 +1,117 @@
+---
+id: ADR-2374-01
+title: "Require human-authored first contributions"
+status: Proposed
+date: 2026-09-08
+tracking_issue: 2374
+deciders:
+  - "@erseco"
+reviewers: []
+related:
+  prs: [2374]
+  changes: []
+  adrs: []
+supersedes: []
+superseded_by: []
+ai_assistance:
+  tool: "ChatGPT"
+  model: "GPT-5.6 Sol"
+---
+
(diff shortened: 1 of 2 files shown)


ignaciogros (commented):
I agree with the proposal because it will limit the number of contribution attempts that do not take into account the application’s architectural details or its possible evolution. Encouraging participation is really important, but I believe we should encourage the involvement of those who understand the project and are able to continue making meaningful contributions.

What is presented as a problem does not seem like one to me: “The boundary of ‘substantially produced’ can require case-by-case interpretation.”

In my opinion, such interpretation is necessary, because we are talking about individual contributions. Someone with programming experience could always contact the project maintainers to propose a larger development, which could then be developed with the necessary resources.

I suggest @cristinavaldera, @mnunezcedec and @eXeLearningProject review this proposal.

cristinavaldera (approved):

mnunezcedec:
Hi @ignaciogros I support this proposal. A good idea can come from anywhere, even an AI, but true quality lies in the implementation—ensuring it fits the architecture, is maintainable, and includes proper tests. (@erseco )

PR #2375, opened 2026-09-08
fix(form): use configured pass rate for score

### LINKED ISSUE #2317: Form iDevice: pass/fail message ignores passRate — threshold hardcoded to 50
## Description
In the Form iDevice, the on-screen pass/fail message ("Congratulations! You 
passed the test" / "Sorry. You failed the test") always uses a hardcoded 
threshold of 50%, ignoring the activity's configured `passRate`.

In `form.js`, `gameOver()` calls:

    $form.showScore(50, data);

while `showScore(passRate, data)` is designed to receive the configured value:

    const isPass =
        (data.rightQuestions / data.totalQuestions) * 100 >=
        parseInt(passRate, 10);

So any score ≥ 50% shows "passed" regardless of configuration.

## Steps to reproduce
1. Create a Form iDevice with 10 questions.
2. Export as SCORM 1.2 and open the package (or preview).
3. Answer 6 of 10 correctly and finish.

## Expected
The message should respect the configured `passRate` (e.g. fail if 
passRate is 100 and score is 60%).

## Actual
"Congratulations! You passed the test" is shown for any score ≥ 50%.

## Suggested fix
    - $form.showScore(50, data);
    + $form.showScore(data.passRate, data);

## Related
- `data.passRate` defaults to 5 (`data.passRate = data.passRate ?? 5`) and 
  there is no field in the Form iDevice editor (General settings or SCORM 
  tab) to change it — see separate issue.

## Environment
- eXeLearning 4.0.3
- Export: SCORM 1.2

ayushdwivedi-stack:
## Summary

Use the configured `passRate` when calculating the Form iDevice score instead of the hardcoded `50%` threshold.

## Changes

- Removed the hardcoded `50` value from the score calculation.
- Passed `data.passRate` to `showScore()`.

## Validation

- Ran `git diff --check`
- Ran `node --check public/files/perm/idevices/base/form/export/form.js`

Fixes #2317

### DIFF
--- a/public/files/perm/idevices/base/form/export/form.js
+++ b/public/files/perm/idevices/base/form/export/form.js
@@ -765,7 +765,7 @@ var $form = {
         if (data.addBtnAnswers & showAnswers.length) showAnswers.show();
         data.gameOver = true;
         $form.checkAllQuestions(data);
-        $form.showScore(50, data);
+        $form.showScore(data.passRate, data);
         if ($('body').hasClass('exe-scorm') && data.isScorm > 0) {
             $form.sendScore(data);
         }


erseco:
Thanks for spotting this — the hardcoded `50` in `gameOver()` is a real bug, and the exported activity does report a pass for anyone over 50% regardless of what the activity asks for.

The swap on its own does not fix it, though: `data.passRate` holds a hardcoded **5**, so `showScore(data.passRate, data)` grades everyone over 5% as passed — a 1-out-of-10 answer sheet included. Three fields claim to hold the threshold and none of them reach the score:

- `dropdownPassRate` is the authored value (the edition dropdown, 10…100, and where `ScormTestHandler` puts the `passRate` of an imported legacy SCORM Test). `save()` collects it, pushes its id onto `dataIds` — which this iDevice never iterates — and `getDataJson()` leaves it out, so it never leaves the editor.
- `passRate` is what the export runtime reads, and `getDataJson()` writes `data.passRate = 5` on every save.
- `gameOver()` ignores both.

#2409 fixes the whole chain and keeps your one-line change at the end of it: edition persists the authored value, and the export resolves it, reading a stored `5` as the old hardcode rather than a 5% threshold so no existing content is regraded. It also adds unit coverage at both ends plus a Playwright spec where an activity asking for 100% must fail a 1-out-of-2 sheet.

Credited you in the PR description. Closing this one is up to @erseco.

PR #2389, opened 2026-09-13
Prepare v4.0.4 release (II)

### LINKED ISSUE #2378: Prepare v4.0.4 (II)
- Update changelog (add the new GL translation revision, and review the SCORM-related sections).
- Typo in the style's CSS and JS files: `searchBarTogger` → `searchBarToggler`.
- Accessibility improvement in the style's JS file: add `aria-expanded` and `aria-controls` to `#siteNavToggler` and `#searchBarToggler`.
- Neo and Flux styles: fixed responsive layout detection.

ignaciogros:
- Fix: rename the search bar toggler from `searchBarTogger` to `searchBarToggler` in the styles (CSS and JS) and in `search-preview-navigation.spec.ts.`
- Styles: announce the navigation and search bar togglers with `aria-expanded` and `aria-controls`, kept in sync on every toggle and when `nav=false` is applied.
- Styles: detect the responsive layout from position instead of float in Neo and Flux, removing the float rules it relied on.
- Changelog revision.


erseco (changes requested):
Reviewed `origin/main...2378-prepare-v404-ii` (7 theme `style.js`/`style.css` pairs, 1 E2E spec, CHANGELOG).

**What I verified as correct**

- The `searchBarTogger` → `searchBarToggler` rename is complete and 1:1 (6 CSS / 3 JS occurrences per theme, matching the counts on `main`). The only leftovers of the old spelling are the recorded exports under `test/fixtures/export/**`, which nothing byte-compares — inert.
- The `isLowRes()` switch from `float` to `position` in Neo and Flux is a real fix, not cosmetics: both themes set `#siteNav { position: fixed }` on desktop, which forces computed `float` to resolve to `none`, so the old check returned `true` at *every* width. Both have `position: static` inside their `max-width: 750px` block, so the new check is accurate. Leaving Zen/Nova/Base on `float` is the right call — those really do use floats with no `position` — and dropping the now-inert `float` declarations is a no-op in both breakpoints.
- All 7 `style.js` files parse (`node --check`), and every version in the CHANGELOG "Upgraded" section matches `package.json` / `app/package.json`, including `electron-context-menu 5.0.0`, which only exists in `app/package.json`.

CI is green, so nothing below comes from a failing job — the two screenshots are a manual repro (the `really-simple` web export fixture, with this branch's `theme/` and the current `public/app/common/exe_export.js`, served locally; CSS transition disabled so the measurements are stable).

---

## Blocking

### 1. `aria-expanded="false"` is not true in Neo, Flux and Educablue

The new state is announced, but in these three themes the closed nav is only moved/shrunk out of sight — it stays in the tab order and in the accessibility tree. So the toggler now tells a screen-reader user the menu is collapsed while its links are still reachable and still read out. Before this PR no state was announced at all, so this is a change from "no information" to "wrong information".

- `public/files/perm/themes/base/neo/style.js:9` — `neo/style.css:405`: `body.siteNav-off #siteNav { left: -330px }`, no `display` / `visibility` / `inert`. `display: none` exists only in the `max-width: 750px` block (`neo/style.css:1076`), so only ≥751px is affected.
- `public/files/perm/themes/base/flux/style.js:23` — same shape: `flux/style.css:504` vs the mobile `display: none` at `flux/style.css:1191`.
- `public/files/perm/themes/base/educablue/style.js:41` — worse, at **every** width: `educablue/style.css:535` and `:941` both close with `width: 0` (the `display: none` at `:941` is commented out), and `#siteNav` has `overflow-y: auto`, which forces `overflow-x: auto` — the links are merely clipped.

Neo, desktop — collapsed, `aria-expanded="false"`, two real <kbd>Tab</kbd> presses land on "Page 1" at x = −314px:



Educablue, desktop — same, one <kbd>Tab</kbd> away, inside a zero-width nav:



The smallest fix is one line in the `navExpanded()` helper each theme already has, so the DOM state and the announced state can never drift apart:

```js
navExpanded: function (visible) {
    $('#siteNavToggler').attr('aria-expanded', visible ? 'true' : 'false');
    $('#siteNav').prop('inert', !visible);
},
```

(`inert` removes the subtree from focus *and* from the accessibility tree; Chrome 102+, Safari 15.5+, Firefox 112+. The themes that already use `display: none` are unaffected by the extra line.)

---

## Non-blocking

### 2. The shared search close path doesn't reset the toggler — `public/app/common/exe_export.js:802`

The new `aria-expanded` on `#searchBarToggler` is written in 7 theme files, but the search-result click handler in the shared runtime does `$('#exe-client-search').hide()` without touching it. Repro: arrive via `?q=term`, open the search bar, search the same term and click the hit for the page you are already on — the href differs only by `#block`, so it is a same-document fragment navigation with no reload. The bar hides; the toggler keeps announcing "expanded" until the next full page load. Per AGENTS.md ("single source of truth"), the reset belongs next to the `.hide()` rather than being duplicated per theme.

### 3. No test covers the new behaviour — `test/e2e/playwright/specs/search-preview-navigation.spec.ts:195`

The only test change in the PR is the selector rename. Nothing asserts the `aria-expanded` sync, and nothing covers the Neo/Flux `isLowRes()` flip, which *is* a user-visible desktop change: the toggler now takes the `else` branch, so `params('remove')` clears `?nav=false` on reopen (it never did on desktop before) and `$('#exe-client-search').hide()` no longer fires on every menu toggle. AGENTS.md §5.3 asks for a Playwright spec on user-visible behaviour changes — ideally landing together with the fix for §1.

### 4. The a11y work is missing from the release notes — `public/CHANGELOG.md:71`

"Neo and Flux styles: fixed responsive layout detection" is there, but the `aria-expanded` / `aria-controls` work listed in the PR body isn't, in a release whose notes claim "meeting WCAG 2.2 level AA".

ignaciogros:
Thanks a lot for the review. I think everything's been fixed. Could you please take another look at it?

erseco (changes requested):
Second pass over `origin/main...2378-prepare-v404-ii`. All four points from my previous review are genuinely addressed — thanks:

- **`aria-expanded="false"` was a lie in Neo/Flux/Educablue** — fixed exactly as suggested. Every theme now has a `navExpanded()` helper that writes `aria-expanded` and `$('#siteNav').prop('inert', !visible)` together, and every state change routes through it (URL `nav=false` at init, both toggler branches, the search toggler's low-res close, and Flux/Nova's `setNavOff()` / sessionStorage restore). I checked all 7 themes: in each one the collapsed nav really is visually hidden at every breakpoint (`display:none` in base/zen/nova/universal and the mobile blocks of neo/flux; `left:-330px` on neo/flux desktop; `width:0` on educablue), so `inert` never lands on a menu the user can still see. The togglers are injected with `$('#siteNav').before(...)`, i.e. outside `#siteNav`, so making the nav inert cannot trap the button that reopens it.
- **Shared search close path** — fixed in the single shared place, `exe_export.js:803`, next to the `.hide()`. I grepped every `$('#exe-client-search').hide()` site: the 6 theme ones all set the toggler already, and the `#exe-client-search-reset` handler never hides the bar, so there is no third path.
- **Tests** — a new E2E spec plus a frontend unit test now exist (see finding 1).
- **CHANGELOG** — entry added under *Changed*. I also re-verified every version in the new "Upgraded" block against `package.json` / `app/package.json` (`fast-xml-parser 5.11.1`, `i18n 0.15.4`, `electron 44.2.0`, `electron-context-menu 5.0.0`, `edicuatex 1.5.5`, `biome 2.5.11`, …) — all match.

Scope of the `isLowRes()` change is still right: nova/zen/base use `float:left` with no `position` on `#siteNav` (so the `float` check is accurate), universal sets `position:fixed` only above 576px (so its existing `position` check is accurate), educablue uses `matchMedia`. Only Neo and Flux were broken, and only they were changed.

One blocker and two minors below.

---

### 1. Blocking — `test/e2e/playwright/specs/theme-nav-toggler-a11y.spec.ts:71` — the new spec fails on Firefox; CI is red

The current run ([34770423667](https://github.com/exelearning/exelearning/actions/runs/34770423667), commit 4d2115e88) has `e2e (firefox, 1)` and `e2e (firefox, 4)` failing. Shard 4 reports `4 failed` and they are all four of the new tests — neo, flux, educablue and base — each failing through both retries at line 71:

```js
expect(await firstNavLink.evaluate(a => (a.focus(), document.activeElement === a))).toBe(true);
```

This is the *initial* state probe, before any click, and line 70 (`el.inert === false`) passes — so this is not the `inert` work misbehaving, it is the focus probe itself: programmatically focusing an element inside the preview iframe does not update that document's `activeElement` in Firefox when the iframe's browsing context isn't the focused one. The previous green E2E run (34763213621) was on 9527d7ef1, *before* the a11y commits, so this spec has never passed on Firefox. Shard 1's single failure is `admin-impersonation`, unrelated.

Two consequences, not one:

1. the branch cannot merge with 2 of 8 e2e shards red;
2. the mirror assertion at line 78, `…toBe(false)` after collapsing, is **vacuous wherever the probe can't take focus** — it would pass even if `navExpanded()` never set `inert`, which is precisely the regression the spec exists to catch.

Please make the reachability check frame-focus-independent (e.g. click something inside the iframe first to focus the browsing context, or replace the probe with an assertion that doesn't depend on it) rather than skipping Firefox.

### 2. Low — `neo/style.js:16` (same line in `base:30`, `flux:30`, `nova:30`, `universal:55`, `zen:16`) — `aria-controls="exe-client-search"` can point at nothing

The search toggler is injected unconditionally with `aria-controls="exe-client-search"`, but `#exe-client-search` is only rendered when the project's search option is on (`src/shared/export/renderers/PageRenderer.ts:194` emits it conditionally). In an export with search disabled the button carries a dangling IDREF. Impact is limited — CSS keeps the button `display:none` without `.exe-search-on`, so assistive tech and most axe rules skip it — but in a release whose notes claim WCAG 2.2 AA it is the kind of thing an automated audit will report. Same shape as `aria-controls="siteNav"`, which is always safe because `#siteNav` always exists in website exports.

### 3. Low — `public/app/common/exe_export.test.js:2168` — the new unit test leaks the jQuery mock if it fails

`window.$ = vi.fn(...)` is restored by a plain `window.$ = originalJQuery;` as the last statement, *after* the `expect`. If that assertion ever fails, the mock stays installed and every later test in the file gets a `$` whose `#siteNav` is hard-coded visible — one real failure turns into a cascade that hides its own cause. `try/finally` (or `afterEach`) fixes it. Noting it because the test is new; the neighbouring tests share the same pattern, so fixing it here only is fine.

---

Nothing else: I traced the remaining close/open paths (`#siteNav a` click handlers in educablue/universal — they trigger a full navigation, so state resets; Flux/Nova's pre-paint `sessionStorage` guard — mirrored to `<body>` and then `navExpanded(!off)`), checked that the removed `float` declarations in neo/flux are genuinely inert in both breakpoints, and confirmed the only remaining `searchBarTogger` spellings are in `test/fixtures/export/**`, which nothing byte-compares.

ignaciogros:
Thanks, @erseco! I hope everything's fixed now.

test/e2e/playwright/specs/theme-nav-toggler-a11y.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+            // Reopen: desktop branch clears nav=false again.
+            await navToggler.click();
+            await expect(navToggler).toHaveAttribute('aria-expanded', 'true');
+            expect(await nav.evaluate(el => el.inert)).toBe(false);
+            expect(await canFocus(firstNavLink)).toBe(true);
+            expect(await nextButton.getAttribute('href')).not.toContain('nav=false');
+
+            // Educablue has no search toggler: the search form is always visible.
+            const searchToggler = iframe.locator('#searchBarToggler');
+            if ((await searchToggler.count()) === 0) return;
```
erseco, line 111: Non-blocking nit: this returns when the toggler is **absent from the DOM**, which is exactly the state the new `if ($('#exe-client-search').length)` guard produces when it misfires. If the search-box markup ever moves from `PageRenderer` to runtime injection, or the id changes, or a theme's toggler injection is reordered before the box exists, neo/flux/base lose their search button from every export — and this spec, the only test covering it, still reports pass, because `count()` is 0 and it returns before the first search assertion.

Gating on the one theme that genuinely has no search toggler keeps the same skip without the blind spot:

```js
if (themeId === 'educablue') return; // educablue has no search toggler; the form is always visible
```

erseco (approved):
Third pass over `origin/main...2378-prepare-v404-ii` (25 files). All three points from my previous review are genuinely fixed — approving.

- **Firefox E2E** — resolved. The probe is now `canFocus()`, which focuses `#siteNavToggler` first and **throws** `'The preview document cannot take focus'` if that doesn't take, then focuses the link. So the collapsed-state assertion can no longer pass vacuously in a frame that can't take focus, and the spec clicks `main h1` first to give the iframe's browsing context focus. All 12 e2e shards (chromium/firefox/static × 4) are green on 326830f, plus `build`, `build_and_test` and codecov patch/project.
- **Dangling `aria-controls`** — resolved in the right place. `#exe-client-search` is static markup emitted by `PageRenderer.ts:192-195` (`searchBoxHtml`), inside `<main>`, so it is in the DOM well before the theme's `$(function(){ myTheme.init(); })` runs — the `$('#exe-client-search').length` guard can't race. The togglers are `position: fixed` in every theme, so omitting one shifts no layout. The new E2E case asserting `[aria-controls="exe-client-search"]` has count 0 when search is off is the right check.
- **Leaking jQuery mock** — resolved with `try/finally`.

Verified independently this pass:

- All 7 `style.js` parse (`node --check`); no `searchBarTogger` spellings remain outside `test/fixtures/export/**`.
- `navExpanded()` never marks a *visible* menu inert: every call site is paired with the `siteNav-off` class, and `#siteNav { position: fixed }` (neo/flux) / `display:none` (base/zen/nova/universal mobile) / `width:0` (educablue) hide it in every breakpoint. `#exe-client-search` is a sibling inside `<main>`, not a descendant of `#siteNav`, so `inert` cannot swallow the search input on the low-res path that does `navExpanded(false); bar.show(); $('#exe-client-search-text').focus()`.
- The `isLowRes()` flip is still correctly scoped: neo and flux both declare `#siteNav { position: fixed }` unconditionally and `position: static` inside `@media (max-width: 750px)`; the removed `float` declarations were inert in both breakpoints. No `float` was removed from base/zen/nova/universal, which still check it. No `.in-iframe` rule changes `#siteNav`'s position, so the preview exercises the same branch a real export does.
- `#exe-client-search { display: none }` is the default in all six themes that inject a search toggler, so the hardcoded `aria-expanded="false"` is accurate at load; `aria-expanded="true"` on the menu toggler is likewise accurate (the menu is open by default at every breakpoint in all 7 themes).
- Passing a boolean to `$(this).attr('aria-expanded', bar.is(':visible'))` (neo/zen/base/universal) is safe — `aria-expanded` is not in jQuery's `expr.match.bool` set, so it stringifies rather than removing the attribute.
- `#exe-client-search-reset` still does not hide the bar (`exe_export.js:688-695`), so `exe_export.js:803` remains the only close path needing the reset.
- Every version in the CHANGELOG "Upgraded" block matches `package.json` / `app/package.json` (`fast-xml-parser 5.11.1`, `i18n 0.15.4`, `electron 44.2.0`, `electron-context-menu 5.0.0`, `edicuatex 1.5.5`, `biome 2.5.11`, `sass 1.103.1`, `jose 6.2.10`, `mermaid 11.17.2`, `pdfjs-dist 6.3.289`, `@xmldom/xmldom 0.9.12`, `mathjax 4.1.3`). The `<version>20260913</version>` bumps are metadata only — nothing keys off the old `20260827`, and theme cache-busting uses a timestamp (`theme.js:340-346`), not this field.

One non-blocking nit inline, on the search half of the new spec. Feel free to merge with or without it.

test/e2e/playwright/specs/theme-nav-toggler-a11y.spec.ts
ignaciogros: I've changed the comment so that it doesn't mention any specific style. I think it's better to use a generic solution. EducaBlue could have that control at some point in the future. Thanks for the reviews!

Issue #2393, opened 2026-09-13
test(static): unsaved-changes dialog blocks Relate SCORM runtime E2E before export


erseco:
## Problem

The static SCORM regression **preserves the stored Relate mark on load and reports a learner restart** fails before it exports a package.

Location: `test/e2e/playwright/specs/scorm12-sco-runtime.spec.ts:683`.

After adding a Relate iDevice and entering `France` / `Paris`, clicking its **Options** link is blocked by `#modalAlert`:

> Unsaved changes detected. Save your iDevice before continuing.

Playwright times out after 20 seconds because the dialog intercepts pointer events. The test never reaches its SCORM export/runtime assertions; this observation does not establish a SCORM score-persistence defect.

## Reproduction

Base commit: `37922ad85`; macOS ARM64; Chromium static project.

```sh
make build-static
bun x playwright test --project=static test/e2e/playwright/specs/scorm12-sco-runtime.spec.ts -g 'preserves the stored Relate mark'
```

Also fails under `make test-e2e-static`.

## Evidence

While validating #2390:
- Full static suite: 440 passed, 134 existing skips, this one failure (16.0 minutes).
- The same test failed again in isolation at the same click.
- Restoring the original ResourceFetcher from `37922ad85` into the static distribution and repeating the isolated test produced the same failure. The original file was restored only for that comparison; the #2390 fix was then reinstated.

The error-context snapshot identifies the unsaved-changes modal and the open Relate editor. The root cause of the unexpected dialog has not yet been determined.

## Expected result

The author can open the Relate iDevice's own Options/SCORM controls while editing, and the regression reaches the exported SCO assertions. Preserve unsaved-change protection for actions that actually leave the editor; do not merely suppress the modal or skip the test.

Issue #2392, opened 2026-09-13
test: collaborative editor-preservation E2E intermittently reads empty TinyMCE content before remote insertion


erseco:
## Observed failure

During the full Chromium suite for #2390, this unchanged test failed:

`test/e2e/playwright/specs/collaborative/editor-preservation.spec.ts`

Test: **User A editor must remain open when User B creates a new iDevice on the same page**

At line 200, after typing unsaved content, `getTinyMCEContent(pageA)` returned an empty string:

```text
Expected substring: "UNSAVED_EDIT_<timestamp>"
Received string:    ""
```

The failure occurs **before Client B creates the new iDevice**, so the intended remote-insertion assertion is never reached.

## Reproduction and evidence

Base commit: `37922ad85`; macOS ARM64; Chromium; default four Playwright workers.

```sh
make test-e2e
```

Observed result: 549 passed, 11 existing skips, this one failure (21.7 minutes).

The unchanged test then passed in isolation in 13 seconds:

```sh
bun x playwright test --project=chromium test/e2e/playwright/specs/collaborative/editor-preservation.spec.ts
```

This is intermittent; the isolated command is not a deterministic reproducer.

## Investigation

The setup types into a specific iDevice iframe, while the content reader uses `tinymce.activeEditor`. Verify editor identity, initialization and focus before deciding whether this is a test synchronization problem or an application bug. The root cause has not been established.

## Expected result

The test reliably prepares and verifies unsaved content in the intended editor, then exercises remote insertion without losing the regression assertions from #1532. Fix the cause rather than skipping the test or increasing arbitrary delays.

Issue #2391, opened 2026-09-13
test: make test-unit fails coverage gate despite all backend tests passing


erseco:
## Problem

While validating #2390 from base commit `37922ad85`, all 8,790 backend unit tests passed, but `make test-unit` exited with status 2 because two unchanged E2E helpers failed the per-file 90% coverage check:

| File | Line coverage |
| --- | ---: |
| `test/e2e/playwright/helpers/moodle-serving-model.ts` | 33.40% |
| `test/e2e/playwright/helpers/prebuilt-package.ts` | 83.33% |

This blocks unrelated changes even when their own patch coverage meets the required threshold.

## Reproduction

```sh
make test-unit
```

The run reports `8790 pass`, `0 fail`, then `scripts/check-coverage.ts` rejects the two files above.

Observed on macOS ARM64. The #2390 change only modifies ResourceFetcher, its tests, an E2E spec and an ADR; neither helper nor the backend tests were changed.

## Expected result

The unit-test coverage gate should have an intentional, documented scope and pass on the baseline. If these helpers belong in that scope, add the missing coverage; otherwise correct the scope without weakening production-code coverage.

There is also a related command mismatch: AGENTS.md requires `make test-coverage`, but this checkout has no such Make target. Align the documented coverage workflow with an executable command.

## Validation

- `make test-unit` passes without hiding test failures or lowering the required production-code coverage.
- Coverage instructions name commands that exist.
- Preserve meaningful checks for the two helpers in the appropriate suite.

PR #2394, opened 2026-09-13
fix: version resource requests to prevent stale export libraries

### LINKED ISSUE #2390: Stale `libs.zip` served from browser HTTP cache after upgrade can break navigation in exported/previewed sites
Resource bundles (`/api/resources/bundle/libs`, `/api/resources/bundle/theme/:name`) are served with `Cache-Control: public, max-age=31536000, immutable` but their URLs carry no version or hash. After a server upgrade, the browser keeps returning the **old** bundle from its HTTP cache, and `ResourceFetcher` stores that stale content in IndexedDB under the **new** version key. This can produce exported/previewed sites that combine a **new theme** with an **old `exe_export.js`**, which is a breaking combination.

## Is this a real concern?

**Yes.** This is not a theoretical risk: the cache headers, the unversioned bundle URLs and the IndexedDB write under the new key are all in the current code (references below), and the failure sequence only requires ordinary browser behaviour (a `fetch()` honouring `max-age`). It affects the online DB deployment and, with a shorter window, the static/nginx deployment. The Electron desktop app is not affected.

The consequence is severe when it happens: navigation stops working in the in-editor preview and in every exported site (HTML5, SCORM, IMS, EPUB) produced from that browser, and the stale file stays pinned in IndexedDB for the whole lifetime of the new version unless the user clears site data.

## Why this matters for v4.0.4

v4.0.4 changes both the themes and `public/app/common/exe_export.js`. The updated themes call shared helpers that were added to `exe_export.js` (e.g. `$exeExport.setUrlParam`). Two mismatch cases:

- **Old theme + new `exe_export.js`** → harmless (same behaviour as before).
- **New theme + old `exe_export.js`** → `$exeExport.setUrlParam` is undefined and **site navigation breaks** (exports and the in-editor preview).

The second case is reachable through the cache path described below.

## Where it happens

**`public/app/yjs/ResourceFetcher.js`**

- `fetchBaseLibraries()` fetches `${apiBase}/bundle/libs` (no version/hash in the URL).
- `fetchTheme()` fetches `${apiBase}/bundle/theme/${name}` (only site themes get `?v=updatedAt`).
- Static mode: `assembleBundleFromLoose()` fetches `${basePath}/${s}` with no cache-buster (the `?v=BUILD_VERSION` suffix is only applied to URLs inside `index.html`).

**`src/routes/resources.ts`**

- `/api/resources/bundle/libs`, `/bundle/theme/:themeName`, `/bundle/idevices`, `/bundle/common`, `/bundle/content-css` all set `public, max-age=31536000, immutable`. The comment near the bundle endpoints says the version acts as a "virtual cache buster in URLs", but the bundle URLs don't include it.

## Failure sequence (server / DB deployment)

1. User runs v4.0.3; browser HTTP cache holds `libs.zip` (old `exe_export.js`) and the theme bundles they used.
2. Server is upgraded to v4.0.4. `window.eXeLearning.version` changes → IndexedDB cache miss (key `v4.0.4-<libsHash>`).
3. `fetch('/api/resources/bundle/libs')` is answered **from the HTTP cache** (still fresh for a year; it's a `fetch()` call after page load, so a normal reload does not revalidate it).
4. The stale ZIP is written to IndexedDB under the **v4.0.4** key → the old `exe_export.js` is now pinned for the whole lifetime of v4.0.4 on that browser, even after the HTTP cache expires.
5. The user selects a theme they never loaded before (not in HTTP cache), or a re-uploaded site theme (`?v=` busts the cache), or the HTTP cache evicted only the theme entry → **new theme + old `exe_export.js`** → navigation fails in preview and in every export.

## Static deployment (nginx)

`Dockerfile.static` serves files with nginx defaults (no `Cache-Control`, `Last-Modified` = build time). Browsers apply heuristic freshness (~10% of the file's age), so a user who last used the previous version weeks after its release keeps a "fresh" `app/common/exe_export.js` for several days. Same mechanism, shorter window, and the stale file is likewise persisted in IndexedDB under the new version key.

## Electron desktop

Not affected: `app://` responses (`app/main.js`, `protocol.handle`) carry no `Cache-Control`/`Last-Modified`, so Chromium does not cache them, and the IndexedDB key changes with the build version.

## Possible fix

Make the bundle requests cache-busted with data we already have. The bundle manifest (`/api/resources/bundle/manifest`) is fetched without caching and contains a per-bundle `hash`:

- `fetchBaseLibraries()`: `${apiBase}/bundle/libs?v=${manifest.libs.hash}`
- `fetchTheme()`: `${apiBase}/bundle/theme/${name}?v=${manifest.themes[name].hash}` (keep the `updatedAt` param for site themes)
- Same for `idevices`, `common`, `content-css` bundles.
- Static mode: append `?v=${this.version}` (or the manifest hash) in `assembleBundleFromLoose()`.

Since the server ignores query params on these endpoints, no backend change is required. Alternatively (or additionally), drop `immutable` / the one-year `max-age` on the unversioned `/api/resources/bundle/*` endpoints so browsers revalidate them.

## Request

Please evaluate whether the fix above (or a different one) should be applied **before releasing v4.0.4**, and more generally whether the resource-loading path should guarantee that all files bundled into an export/preview come from the same application version, so that this class of theme/common-library mismatch cannot happen on future upgrades either.

## Affected versions

Any upgrade where `exe_export.js` and themes change together; first critical occurrence: upgrading to v4.0.4.

erseco:
Browser upgrades could reuse an old immutable resource bundle and save it under the new IndexedDB version, mixing old export runtimes with new themes and breaking navigation.

Version bundle requests with their manifest hash (application-version fallback), preserve site-theme update tokens, and version static manifests and loose application resources. Theme downloads and separately loaded SCORM/i18n runtimes follow the same policy. The server manifest is revalidated before choosing hashes.

Fixes #2390.

Adds unit coverage and a Playwright regression using real HTTP caching and IndexedDB in server and static modes. Records the cache policy and deployment limitations in ADR-2390-01.

Validation:
- `make fix` and `make architecture-check`: passed.
- Frontend: 15,995 tests passed; all 211 ResourceFetcher tests passed.
- Integration: 770 tests passed.
- Changed executable lines: 19/19 covered (100%).
- Backend: 8,790 tests passed, but `make test-unit` fails its existing per-file coverage gate for `moodle-serving-model.ts` (33.40%) and `prebuilt-package.ts` (83.33%). Neither file is changed by this PR. Tracked in #2391.
- The documented `make test-coverage` target does not exist; patch coverage was checked using the Vitest LCOV report (see #2391).
- Upgrade regression: both modes fail with the original implementation (old runtime returned) and pass with this fix.
- Full Chromium E2E: 549 passed, 11 existing skips, one failure in `collaborative/editor-preservation.spec.ts:200` while reading the initial TinyMCE content, before the remote insertion. The unchanged test passed when rerun in isolation (13 seconds). Tracked in #2392.
- Full static E2E: 440 passed, 134 existing skips, one failure in `scorm12-sco-runtime.spec.ts:683`: an unsaved-changes modal blocks the Options link before export. This fails both in isolation and with the original ResourceFetcher restored, confirming it predates this fix. Tracked in #2393.

No tests were disabled or modified to hide them. Previously poisoned IndexedDB entries are not migrated; deploying this fix with the next application version invalidates those old keys.


ignaciogros (approved):
Thank you for this fix! It's essential for the v4.0.4 release and landed in record time. 

Just one suggestion: change ADR-2390-01 to status: Accepted before merging.

PR #2399, opened 2026-09-13
fix: stop forcing consent on OpenID login

### LINKED ISSUE #2377: Hardcoded prompt=consent makes OIDC login unusable in tenants that require admin consent
Version: 4.0.3 (ghcr.io/exelearning/exelearning:v4.0.3), Team Edition / APP_ONLINE_MODE=1 IdP: Microsoft Entra ID (single tenant), Authorization Code + PKCE Deployment: self-hosted behind an AWS ALB, PostgreSQL 18

Summary

src/routes/auth.ts hardcodes prompt: 'consent' in the authorization request. Because prompt=consent instructs the IdP to re-run the consent flow on every single login regardless of any existing grant, non-administrative users can never complete sign-in in a tenant where end users are not permitted to self-consent. No amount of correctly-granted admin consent satisfies it.

This is not a missing configuration option so much as a functional break: the parameter defeats the mechanism that is supposed to make consent a one-time administrative action.

Reproduction
Register an Entra application (single tenant), redirect URI https://<host>/login/openid/callback.
Set APP_AUTH_METHODS=openid, OIDC_ISSUER=https://login.microsoftonline.com/<tenant-guid>/v2.0, OIDC_CLIENT_ID, OIDC_CLIENT_SECRET, OIDC_SCOPE="openid email profile".
In the enterprise application, set Assignment required = Yes and assign a group.
Grant tenant-wide admin consent for all requested Microsoft Graph delegated permissions.
Sign in as a non-administrative assigned user.

Expected: the user signs in. Consent was granted tenant-wide by an administrator.

Actual: the user is shown "Approval required — this app requires your admin's approval to:" with a justification field, and cannot proceed. Entra sign-in logs report:

Sign-in error code: 90095
Failure reason: Admin consent is required for the permissions requested by this
                application. An admin consent request may be sent to the admin.

A Global Administrator signs in successfully, because administrators can self-consent on demand — which masks the problem during initial setup.

Why existing consent does not help

The tenant-wide grant was verified correct via Microsoft Graph:

GET /v1.0/servicePrincipals/<sp-id>/oauth2PermissionGrants

{
  "consentType": "AllPrincipals",
  "principalId": null,
  "scope": "email offline_access openid profile User.Read"
}

consentType: AllPrincipals with a null principalId is a correct tenant-wide grant covering every requested scope. The user-consent tab was empty (no conflicting per-user grant). Conditional Access evaluated 23 policies with zero failures.

We also submitted and approved an admin consent request for the specific user through the Entra admin consent workflow. The user was still prompted on the next login, and the approval did not appear against the application. This is the expected behaviour of prompt=consent and is what confirms the diagnosis: the parameter forces re-consent unconditionally, so no grant can ever satisfy it.

Location

src/routes/auth.ts, in the /login/openid handler (line ~790 in the 4.0.3 tag):

typescript
const scope = await getSettingString(db, 'OIDC_SCOPE', process.env.OIDC_SCOPE || 'openid email');
const clientId = await getSettingString(db, 'OIDC_CLIENT_ID', process.env.OIDC_CLIENT_ID || '');

const params = new URLSearchParams({
    client_id: clientId,
    redirect_uri: redirectUri,
    response_type: 'code',
    scope,
    state,
    nonce,
    code_challenge: codeChallenge,
    code_challenge_method: 'S256',
    prompt: 'consent',        // <-- hardcoded
});

Note the contrast with the two lines immediately above: scope and clientId are both resolved through getSettingString(db, KEY, process.env.KEY || default). prompt is the only parameter in this object that is not configurable.

Suggested fix

The plumbing already exists. The codebase exposes nine OIDC_* settings through the same helper — OIDC_ISSUER, OIDC_AUTHORIZATION_ENDPOINT, OIDC_TOKEN_ENDPOINT, OIDC_USERINFO_ENDPOINT, OIDC_END_SESSION_ENDPOINT, OIDC_JWKS_URI, OIDC_CLIENT_ID, OIDC_CLIENT_SECRET, OIDC_SCOPE. Adding OIDC_PROMPT follows the established pattern:

typescript
const prompt = await getSettingString(db, 'OIDC_PROMPT', process.env.OIDC_PROMPT || '');

const params = new URLSearchParams({
    client_id: clientId,
    redirect_uri: redirectUri,
    response_type: 'code',
    scope,
    state,
    nonce,
    code_challenge: codeChallenge,
    code_challenge_method: 'S256',
});

// prompt is optional per OIDC Core 3.1.2.1; omit it entirely when unset
if (prompt) {
    params.set('prompt', prompt);
}

On the default value. We would suggest defaulting to empty (parameter omitted) rather than preserving consent:

prompt is optional in the OIDC specification, and omitting it is the conventional behaviour for a relying party that does not need to force re-authentication.
The current default is broken for any tenant with restrictive consent policies, which is the common enterprise configuration.
Forcing consent on every login provides no security benefit to eXeLearning; it does not re-verify anything the application acts on, since identity comes from the ID token claims.

If backwards compatibility is a concern, defaulting to consent with OIDC_PROMPT="" supported as an explicit opt-out would still resolve the problem, though it leaves the broken behaviour as the out-of-box default.

Valid values worth documenting: none, login, consent, select_account (OIDC Core 3.1.2.1).

Impact and current workaround

This blocks all non-administrative users in affected tenants, which for a Team Edition deployment means every author. It is also difficult to diagnose: the IdP error names consent, the administrator's own login succeeds, and the consent grants all appear correct in the portal — so the investigation naturally focuses on IdP configuration rather than on the client request.

We are currently patching the built bundle in the container image:

dockerfile
FROM ghcr.io/exelearning/exelearning:v4.0.3
USER 0
RUN sed -i 's/prompt: "consent"/prompt: "select_account"/g' /app/dist/index.js \
 && grep -q 'select_account' /app/dist/index.js
USER bun

This works but rebases on every upgrade and breaks silently if the line is ever reformatted.

Related

Version 4.0.2 added automatic OIDC endpoint discovery from OIDC_ISSUER, which is a welcome improvement and works correctly against Entra. This request is in the same vein: making the remaining hardcoded piece of the authorization request configurable.

Happy to open a pull request if the maintainers agree with the approach and the proposed default.

erseco:
Fixes #2377.

Remove the forced OIDC consent prompt so providers can honor existing admin consent. Add redirect regression coverage, including PKCE, state and nonce. Includes the two coverage configuration fixes from #2397.

Reviewer checklist (Entra app with tenant-wide admin consent):
- [x] Sign in as an assigned non-admin user; no new approval request appears.
- [x] Sign out and sign in again; access still works.
- [x] Check the authorization URL: `prompt` is absent.

Validation: `make fix`, 109 auth tests and the OIDC Chromium regression pass; all 15,987 frontend tests pass. The regression fails before the fix. Full local validation encountered unrelated translation/integration timeouts under heavy load; remaining runs were stopped and CI will complete validation, as authorized. Coverage configuration is reused from #2397.


ignaciogros (approved):
Avoiding repeated consent prompts and allowing non-administrative users to sign in is essential.

I just updated the CHANGELOG and the documentation. Please confirm if those changes are OK.

Thanks, @erseco!

PR #2401, opened 2026-09-13
fix: preserve current resource cache entries on startup

### LINKED ISSUE #2398: `ResourceCache.clearOldVersions()` deletes current-version `libs` and site-theme entries on every startup
`ResourceCache.clearOldVersions(currentVersion)` (`public/app/yjs/ResourceCache.js:320-346`) keeps only entries whose `version` field is **strictly equal** to the current application version (`cursor.value.version !== currentVersion`). However, `ResourceFetcher` stores some entries under a **composite** version key:

- Base libraries: `${version}-${libsHash.substring(0, 8)}` (`public/app/yjs/ResourceFetcher.js`, `fetchBaseLibraries()`, `cacheVersion`).
- Site themes: `${version}-${updatedAt}` (`fetchTheme()`, `themeCacheVersion`).

Those keys never equal `currentVersion`, so they are treated as "old" and deleted on every page load.

## Where it is called

`public/app/yjs/YjsProjectBridge.js:449-451` runs `clearOldVersions(window.eXeLearning.version)` right after `ResourceCache.init()`, before `ResourceFetcher` is initialised.

## Impact

- `libs.zip` (and any re-uploaded site theme) is never served from IndexedDB across sessions: each startup deletes the entry and the next export/preview fetches it again from the HTTP cache or the network. The IndexedDB persistence layer is effectively disabled for these resource types.
- No functional breakage and no disk-space leak (the cleanup is more aggressive than intended, not less). Themes, iDevices and content CSS use the plain `version` key and are unaffected.
- Unrelated to #2390 / #2394: the stale-runtime problem there came from the HTTP cache, not from IndexedDB.

## Steps to reproduce

1. Open the workarea and trigger an export/preview so `libs` is written to IndexedDB (`exelearning-resources-v1`, store entry key `libs:base:<version>-<hash8>`).
2. Reload the page.
3. Inspect IndexedDB: the `libs:base:*` entry is gone; with `Logger` output enabled the console shows `[ResourceCache] Cleared N old version entries` with N ≥ 1 even though the version did not change.

Unit-level check against the current code (mocked store, as in `ResourceCache.test.js`): after `set('libs','base','v4.0.4-f253c079')`, `set('theme','site-1','v4.0.4-1712345678')` and `set('theme','base','v4.0.4')`, `clearOldVersions('v4.0.4')` returns `2` and only `theme:base:v4.0.4` survives.

Within a single page load the bug is masked by the in-memory `ResourceFetcher.cache` map; it only shows up across reloads.

## Suggested fix

Treat an entry as current when its `version` equals `currentVersion` **or** starts with `${currentVersion}-`, e.g. in `clearOldVersions()`:

```js
const isCurrent = (v) => v === currentVersion || v.startsWith(`${currentVersion}-`);
if (!isCurrent(cursor.value.version)) { ... }
```

Alternatively, store the base application version in a separate field (`appVersion`) and filter on it, keeping the composite key only in `key`/`version`.

Add a unit test in `public/app/yjs/ResourceCache.test.js` covering composite keys (`v4.0.4-abcdef12`, `v4.0.4-1712345678`) being preserved and plain old versions (`v4.0.3`) being removed.

## Affected versions

Present since the composite `libs` cache key was introduced in #907 (January 2026); site themes have been affected since the `updatedAt` cache-buster from #1775 (May 2026). Still present on `main` after #2394.

erseco:
Closes #2398.

Preserve current-version library and site-theme cache entries on startup, including hash/timestamp suffixes, while removing old app versions. Reuses only the coverage changes from #2397 (360b1e6).

Compatible with #2394: cache keys are unchanged and its patch applies cleanly. Follow its deployment guidance to invalidate previously stale entries with a new app version.

Validation:
- Lint, 8,790 backend tests, 15,992 frontend tests and 770 integration tests passed; 100% patch coverage.
- Cache reload regression passed in Chromium and static mode. Full suites: 547 Chromium and 439 static tests passed. Relate fails in both due to #2393 (fixed by #2397); CodeMagic passed 3 isolated reruns after one transient failure. No tests disabled.

Review checklist:

- [x] Open a project and preview it once.
- [x] Reload and preview again; check that `libs:base:*` survives in IndexedDB (`exelearning-resources-v1`).
- [x] Run `bun x playwright test --project=chromium resource-cache-persistence.spec.ts` to verify reload persistence and old-version cleanup.


ignaciogros (approved):
I think it should go in before v4.0.4 so the persistent IndexedDB cache actually works for libs and site themes with #2394.

One low-severity note, not blocking: `isCurrentVersion` retains every composite entry for the current version (e.g. N copies of a re-uploaded site theme, one `libs:base:*` entry per rebuild on v0.0.0-alpha builds). Disk growth, no functional impact. **I suggest tracking it in a separate issue and fix it later.**

erseco:
Thanks for the review. On the `isCurrentVersion` note: confirmed, and it is worth acting on, because this PR is what opens that window. Before it, `clearOldVersions()` deleted composite entries on every startup (bug #2398), so they could never accumulate. Keeping them means each site-theme re-upload or libs rebuild now leaves a superseded copy behind, with nothing to reclaim it until the app version changes. `package.json` is `0.0.0-alpha`, so development and CI builds never bump it — that is the case you flagged, and it is unbounded there.

Tracked in #2405 and fixed in #2406, which keeps only the most recently written entry per `type`/`name` group using the `cachedAt` timestamp already stored on every entry. It is stacked on this branch since both change `clearOldVersions()`, so the retention policy lands as its own reviewable change and this hotfix stays minimal for v4.0.4.

Leaving this PR as-is with your approval.

ignaciogros:
I think it's ready to be merged once the tests have passed.

Thanks, @erseco!

PR #2409, opened 2026-09-14
fix(form): grade against the configured pass rate

erseco:
Supersedes #2375. Credit to @ayushdwivedi-stack for spotting that `gameOver()` grades against a hardcoded 50.

## The bug

An exported Form activity told every learner over 50% *"Congratulations! You passed the test"*, whatever threshold the activity carried:

```js
// public/files/perm/idevices/base/form/export/form.js
$form.showScore(50, data);
```

## Why the one-line swap is not enough

`data.passRate` held a hardcoded **5**, so `showScore(data.passRate, data)` on its own would have passed anyone over 5% — a 1-out-of-10 answer sheet included. Three fields claimed to hold the threshold and none of them reached the score:

| Field | What it did |
|---|---|
| `dropdownPassRate` | The authored value: the edition dropdown (10…100), and where `ScormTestHandler` puts the `passRate` of an imported legacy SCORM Test. `save()` collected it, pushed its id onto `dataIds` — which this iDevice never iterates — and `getDataJson()` left it out. It never left the editor. |
| `passRate` | What the export runtime reads. `getDataJson()` wrote `data.passRate = 5` on every save, and `updateConfig()` defaulted to the same 5. |
| the literal `50` | What `gameOver()` actually graded against, ignoring both. |

## The fix

**Edition** — `public/files/perm/idevices/base/form/edition/form.js`

- `save()` carries the authored value instead of `this.passRate = 50`.
- `getDataJson()` persists it under both names: `passRate` (what the runtime reads) and `dropdownPassRate` (what `loadPreviousValues()` restores the control from, and what the legacy importer writes).
- `loadPreviousValues()` seeds the control with 50 when the activity carries no threshold, so saving an older activity again keeps the grading it has always had instead of persisting the blank option.

**Export** — `public/files/perm/idevices/base/form/export/form.js`

- New pure `toPassRate(value)` / `resolvePassRate(data)`.
- `updateConfig()` resolves the threshold; `gameOver()` passes it to `showScore()`.
- The check button's `data-pass-rate` attribute carried `$form.passRate`, which is never assigned — it now carries the resolved value.

Resolution order:

| Input | Result |
|---|---|
| `dropdownPassRate` in 1…100 | that value |
| `passRate` in 1…100 and not 5 | that value |
| `passRate` === 5 | **50** — the old hardcode, not a 5% threshold. The dropdown only offers 10…100, so 5 is unreachable as an authored value. |
| anything else, or nothing | **50** |

**No stored content is regraded.** 50 is the threshold every existing activity was really graded against, and it is what they all keep.

## Deliberately not in this PR

The control's container is still hidden:

```html
<div class="question-button inline mb-3" style="display:none;">
```

so the values reaching `dropdownPassRate` today come from legacy imports, not from authors. Revealing the control is a product decision and is left out.

No ADR: this restores intended behaviour inside one iDevice rather than taking a durable architecture decision. `make architecture-check` is green.

## Tests

- `export/form.test.js` — `toPassRate`; `resolvePassRate` across authored / stored / legacy 5 / out of range / absent; `updateConfig`; `showScore` passing and failing; `gameOver` forwarding `data.passRate`.
- `edition/form.test.js` — the value travels into the saved data under both names, comes back into the control on reopen, and defaults to 50 when the activity carries none.
- `test/e2e/playwright/specs/idevices/form.spec.ts` — an activity asking for 100% must fail a 1-out-of-2 sheet. Against the old hardcode that same sheet reported *"Congratulations! You passed the test"*.

Every new test was mutation-checked: reverting each production change makes it fail.

### Verification

| Gate | Result |
|---|---|
| `make fix` / `make lint` | exit 0, no warnings in the changed files |
| `make test-unit` | 8790 pass, 0 fail |
| `make test-frontend` | 296 files, 16046 pass |
| `make test-integration` | 770 pass, 0 fail |
| `make test-e2e` | `idevices/form.spec.ts` 9/9 green; the full suite was green through 157 tests locally before being handed over to CI |

Rebased onto `main` (#2401, #2408) and re-verified after the rebase.


ignaciogros (approved):
My concern was that it might affect the average grades, but I see that it doesn't. 

I think it should be included in v4.0.4. 

Thanks, @erseco.

mnarvaezm:
Thanks for the analysis. The proposed behavior contradicts the decision not to apply the individual pass thresholds from legacy iDevices.

The `form` iDevice did not exist as such in eXeLearning 2.9. Since its first implementation, it has used a fixed 50% threshold and has not allowed it to be configured. When legacy activities are converted to `form`, the adopted policy is to apply that criterion and ignore their individual pass thresholds. This was decided for compatibility with the rest of the iDevices, which do not support their own thresholds, and to avoid confusion on pages containing multiple SCORM iDevices.

`ScormTestHandler` preserves the old threshold in `dropdownPassRate`, but preserving that value does not mean it remains an active configuration. This PR would change that policy by starting to use it:

| Content | Correct answers | `main` | This PR |
|---|---|---|---|
| `passRate: 5` saved by previous versions of `form` | 1/4 | Fail | Fail |
| imported `dropdownPassRate: "70"` | 2/4 | Pass | Fail |
| imported `dropdownPassRate: "100"` | 3/4 | Pass | Fail |

The main problem is that a threshold would be applied even though the author cannot view or modify it in the editing interface. The dropdown remains hidden with `display:none`, as it has been since the first implementation of `form`, and this PR does not make it visible either. However, the imported value would be restored when opening the activity and preserved when saving it again.

Currently, `getDataJson()` omits `dropdownPassRate`, so that value is discarded when the activity is saved from the editor. This PR would instead preserve it and use it as the pass criterion, without providing any visible control over it.

The form also does not tell the learner what minimum score is required. An imported activity with a 100% threshold could show a score of 7.5/10 together with a failure message without explaining why. This could easily be perceived as an error, and the author would have no option in the interface to inspect or change that criterion.

It is also worth clarifying the scope of the change:

- Already published packages will keep their current behavior unless they are replaced. The change would affect content run with the updated version, including previews and newly exported packages.
- For the same answers, the numeric score calculated by `getScore()` does not change, nor does this by itself affect the weighted average or the result sent to the LMS. What changes is the form's local pass/fail verdict and its visual presentation.

The objection is therefore about behavior and user experience: an inherited configuration that we have deliberately chosen to ignore would become active again, without being visible to the author or explained to the learner.

If we want to support thresholds other than 50% in the future, this should be addressed as a general change affecting all iDevices, not as behavior specific to `form`: with a consistent configuration model, visible controls for authors, clear information for learners, a defined policy for imported content, and coherent integration with SCORM evaluation across all iDevices.
Captura de pantalla 2026-09-14 154110
@erseco @ignaciogros @cristinavaldera

erseco:
Thanks @mnarvaezm, this distinction changes my reading of the issue.

I agree that the current PR is doing more than fixing the inconsistency between `passRate: 5` and the hardcoded `showScore(50, data)`. For legacy content carrying `dropdownPassRate`, it makes a previously inactive value affect the local pass/fail verdict, so this is an observable behavior change rather than just a bug fix.

I also agree that activating that value while the control remains hidden is problematic. An imported activity could, for example, require 100%, show a learner a score of 7.5/10 together with a failure message, and give the author no visible way to inspect or change that threshold.

There is also a consistency issue with the SCORM policy: the page-level success threshold comes from the LMS `mastery_score`, falling back to 50%. Making an inherited per-Form threshold active could therefore make the Form display "failed" while the SCORM page itself is considered "passed".

The one point I would like to clarify is the historical intent. I can see that the existing implementation behaves as you describe: Form grades locally against 50%, the control is hidden, and imported `dropdownPassRate` is currently discarded on save. What I have not found is where the decision to deliberately ignore legacy individual thresholds was documented. `ScormTestHandler` still explicitly preserves the legacy `passRate` as `dropdownPassRate`, so if there is an issue, commit or previous discussion recording that policy, please link it here.

Given the current behavior and UI, I think the safest fix for v4.0.4 is to keep the Form threshold at 50% and clean up the misleading/dead `passRate` plumbing rather than activating an invisible inherited setting.

If configurable per-activity thresholds are wanted, I agree that this should be handled separately as a product decision, with a visible author control, clear learner feedback, an explicit legacy-import policy, and consistent interaction with the SCORM success threshold.

mnarvaezm:
At the time this iDevice was created, during the eXe 3.0 beta, there was nowhere to document this kind of decision. The first ADR under doc/architecture/adr/ was added on 2026-07-09, more than a year after that beta. As a general rule, during that period, neither decisions about iDevice behavior nor the rules and behavior governing the rest of the application were documented anywhere. This is not an exception; it was simply the usual situation at that stage of development.

@erseco @ignaciogros @cristinavaldera

ignaciogros (commented):
After reading @mnarvaezm's comments:

If I'm not mistaken, the only scenario in which `main` displays a pass/fail verdict that differs from the configured one (even though the grade is correct) is content imported from eXe 2.x (SCORM Quiz / Test) with a pass mark other than 50.

If that's the case, since the Form iDevice has been this way since its creation, I suggest releasing v4.0.4 without this change and keep working on it until all requirements are covered.

I think it's better not to rush its inclusion in the main branch. Please let me know if you agree.

erseco:
> I think it's better not to rush its inclusion in the main branch. Please let me know if you agree.

Totally agree!

ignaciogros:
Thanks. Postponed then.

PR #2412, opened 2026-09-14
Sort theme icons alphabetically in the block icon picker.

### LINKED ISSUE #2411: Icon selection dialog displays the Universal style icons in the wrong order
In the static version, the box icon selection dialog displays the Universal style icons in the wrong order (not sorted by colour/alphabetically). In the Electron and Cloud versions (with a database), they are displayed correctly.

ignaciogros:
In the static build, the block icon picker showed the Universal style icons in filesystem order instead of alphabetically (#2411). The static bundle is built on Linux with Bun, whose `fs.readdirSync` does not sort entries, so the manifest carried raw ext4 hash order. NTFS returns sorted names, which is why local static builds looked correct.

## Solution

- Add `sortThemeIcons()` / `compareThemeIconNames()` in `src/shared/parsers/theme-parser.ts` (single source of truth) and apply it in the three icon scanners (shared parser, `/api/themes` route, static build script).
- Sort icons in the picker (`blockNode.js`) as well, which also covers user themes loaded from ZIP and previously cached data.
- Tests: unit specs for the parser, route, build script and picker; ordering assertion added to the `block-icons` E2E spec.


erseco (approved):
Reviewed at e331d51.

Approving: the fix is correct and tested at every layer, and none of the points below change user-visible behaviour, so I would rather not block on them. I would still like 1, 2 and 3 addressed before merge — they are cheap and two of them are actively misleading comments — and 4 is explicitly a follow-up, not this PR.

Worth noting that the client-side sort in `blockNode.js` is the one that actually fixes all four sources of theme icons (base themes, site themes, user themes imported from an ELP, and the static bundle); the three server-side sorts make the API and bundle payloads deterministic, which is valuable on its own but is no longer something the picker depends on.

Verified locally on macOS:

- `bun test src/shared/parsers/theme-parser.spec.ts src/routes/themes.spec.ts scripts/build-static-bundle-internals.spec.ts` → 131 pass, 0 fail
- `npx vitest run public/app/workarea/project/idevices/content/blockNode.test.js` → 317 pass
- Playwright: opened the picker with the Universal style and the theme icons deliberately delivered in non-alphabetical order (what Bun's `readdirSync` does on the ext4 runner). The colour grouping of the UDL families makes the difference obvious:



---

### 1. `sortThemeIcons()` in `theme-parser.ts` encodes display order in object key order, which JS does not honour for integer-like keys

`src/shared/parsers/theme-parser.ts:144-150` builds a fresh record and relies on insertion order. Integer-like keys are always enumerated first, in ascending numeric order, whatever the insertion order was:

```
input keys    : _extra, 1, 10, beta, 2
collator order: [ '_extra', '1', '2', '10', 'beta' ]
record order  : [ '1', '2', '10', '_extra', 'beta' ]
```

So a theme shipping `1.svg` silently breaks both the doc comment on line 142 ("a copy of an icons record whose keys are in display order") and the contract the three new specs assert. Returning a `ThemeIcon[]` — which the JS twin already does — removes the dependency on a guarantee the language does not give. If keeping the `Record` shape is deliberate (it is the wire format, so it probably is), then the doc comment should say the order is best-effort and integer-like ids are an exception.

### 2. Two exported functions named `sortThemeIcons` with different signatures, documented as twins

- TS (`theme-parser.ts:144`): `Record<string, ThemeIcon> -> Record<string, ThemeIcon>`, sorts by record key, keeps entries whose `value` is empty.
- JS (`blockNode.js:105`): `Record -> Array`, drops entries without `value`, sorts by `id || value`.

Yet `blockNode.js:94` says "JS twin of THEME_ICON_COLLATOR … keep both in sync" and `theme-parser.ts:129-130` says the same in reverse. What is actually shared is the collator, not the function. Someone taking those comments literally and porting one implementation onto the other breaks either `makeModalChangeIconBody()` (it destructures an array) or the API response shape. Suggestion: rename the JS one (`themeIconsInDisplayOrder`, say) and scope both comments to the collator.

Related: the repo already has a TS/JS twin convention *with* a parity test — `src/shared/block-icon.spec.ts:87` reads `public/app/common/blockIconRuntime.js` and asserts the table equals the TS constant. Nothing guards the new collator, so `numeric: true` or the locale can drift on one side with no failing test. `blockIconRuntime.js` is also where the other twin lives, and `YjsProjectBridge.js` / `YjsStructureBinding.js` already consume `window.eXeBlockIconRuntime` — putting it there would make it reusable instead of guaranteeing a third copy the next time a theme-icon list gets rendered.

### 3. Both new server-side comments justify the sort with something this PR makes false

`src/routes/themes.ts:181` says "the picker relies on this order (#2411)". After this PR `makeModalChangeIconBody()` sorts client-side (`blockNode.js:1851`), so the picker does not rely on it — remove the server sort and neither the picker nor any test would notice. Symmetrically, `blockNode.js:98-100` justifies the client sort with "static bundles built on Linux keep Bun's raw readdir order", which `build-static-bundle.ts:523` now fixes. The still-true reason for all of them is "keep the API / bundle payload deterministic across filesystems" — worth saying that instead, so the next reader does not go looking for a consumer that is not there.

### 4. The sort is applied in four places instead of consolidating the three duplicate scanners (follow-up, not this PR)

Nothing imported `src/shared/parsers/theme-parser.ts` before this PR — `src/routes/themes.ts:18` is the first import of it anywhere in the repo. Its `parseThemeConfig` / `scanThemeIcons` have no production caller, so the sort at `theme-parser.ts:191` and the new spec at `theme-parser.spec.ts:236` exercise code that never runs. Meanwhile `routes/themes.ts` keeps its own `scanThemeIcons` (line 142) and `build-static-bundle.ts` keeps a third (line 498), and the three have already diverged:

- `routes/themes.ts` resolves duplicate base names by extension priority (svg > png > gif > jpg > jpeg > webp) — asserted by the new spec at `themes.spec.ts:918`
- `build-static-bundle.ts` and `theme-parser.ts` let the last `readdir` entry win, and neither supports `.webp`

So for a theme shipping both `reto.svg` and `reto.png`, the server and the static bundle can hand the same theme different icon URLs, and the bundle's pick can flip between builds. Key order is now deterministic; the value choice still is not. Per AGENTS.md ("single source of truth"), the deeper change is to have the two real scanners call the shared one and delete their copies.

### Nits

- `compareThemeIconNames` is newly exported from the `src/shared/parsers` barrel but has no consumer outside `theme-parser.ts:146` and its own spec, and that spec (`theme-parser.spec.ts:395-401`) only re-asserts `Intl.Collator` behaviour. Consider leaving it unexported rather than committing a public API nobody calls.
- `block-icons.spec.ts:119-124`: `expect(themeIconIds).toEqual([...themeIconIds].sort(collator.compare))` also passes when `themeIconIds` is `[]`, so renaming `.theme-block-icon` would silently turn the only E2E guard for #2411 into a no-op. One extra `expect(themeIconIds.length).toBeGreaterThan(0)` fixes it. (The assertion also sits inside a `skipInStaticMode` test, so `make test-e2e-static` never runs it — which is a shame given the comments point at the static bundle as the failure mode.)
- `blockNode.test.js:919` reassigns `eXeLearning.app.themes.getThemeIcons` and never restores it; `beforeEach` only calls `vi.clearAllMocks()`, which does not undo a reassigned property. The next test ("adds section titles separating theme and general icons") now runs against the `udl_*` fixture instead of the file-level `icon1`/`icon2` one. It passes today, but it has become order-dependent. The same pattern already exists a few tests below, so one `afterEach` restoring the default fixture would fix both.
- `blockNode.js:108`: `THEME_ICON_COLLATOR.compare(a.id || a.value, …)` — in the id-less fallback the sort key is a full URL, and for user themes `Theme.resolveIconBlobUrls()` (`public/app/workarea/themes/theme.js:427`) rewrites `icon.value` to `blob:<random-uuid>`, so such an icon would land in a different position every session. Sorting by the basename of `value` would be more predictable. Edge case, and the new test at `blockNode.test.js:4431` currently pins the present behaviour.
- The collator locale is hard-coded to `'en'`. That is the right call for the server/bundle payload (reproducible everywhere), less obviously so at the display site, where the UI locale would order non-ASCII icon names the way the user expects. Worth one line of comment saying the fixed locale is intentional.

### Same root cause, still open elsewhere (worth a follow-up issue)

- `scanThemes()` (`routes/themes.ts:265-288`) and `buildThemesList()` (`build-static-bundle.ts:540`) push themes in raw `readdir` order, so the style picker itself still shows the #2411 symptom.
- `scanThemeFiles()` (`routes/themes.ts:117`) returns a theme's `.css` files in `readdir` order, and that order is the cascade order — a theme with more than one stylesheet can render differently depending on the filesystem that served it.

Both are one `.sort()` away and would be a natural second half of this work.

PR #2414, opened 2026-09-14
fix(progress-report): restore the report after the page, block and id model changes

mnarvaezm:
## Scope — this PR touches the progress-report iDevice and nothing else

```
public/files/perm/idevices/base/progress-report/export/progress-report.js       +196 −240
public/files/perm/idevices/base/progress-report/export/progress-report.test.js  +634  −19
```

Two files, both inside `public/files/perm/idevices/base/progress-report/export/`. No exporter, no
backend, no Yjs, no shared frontend module, no `translations/**`. The iDevice's own
`edition/progress-report.js` is untouched too — it had already been adapted; only the export runtime
was left behind. Everything below is a consumer-side fix: the report is catching up with a data
model that moved under it.

## Why: the report was written against the previous page model

The progress-report builds a course map by reading the project's page tree, its blocks and its
iDevice ids. Each of those three things changed, and the iDevice kept reading them the old way.
Seven distinct symptoms, one cause each:

**1. Component ids are regenerated on clone; the copies embedded in the content are not.**
`YjsStructureBinding.cloneComponentMap()` (`public/app/yjs/YjsStructureBinding.js:1389`) assigns a
fresh `id`/`ideviceId` to a cloned component but copies `htmlContent`, `htmlView` and
`jsonProperties` verbatim — so the `data-id` inside them still names the *original*. The exported
wrapper is `<div id="${component.id}">` (`src/shared/export/renderers/IdeviceRenderer.ts:244`), which
is also the id a learner's result is reported under. The report preferred the embedded copy
(`dataIDs.ideviceID || row.ode_idevice_id`), so on a duplicated page every row pointed at the
original activity: scores never reached their row and "see activity" jumped to the wrong page.
→ `b47e20caa`

**2. The same stale id duplicated activities in exported packages.**
`jsonProperties` is written only inside `<odeComponent>`
(`src/shared/export/generators/OdeXmlGenerator.ts:308-313`), so `querySelector('jsonProperties')` on
the parent `<odePagStructure>` descended into the first component and listed it a second time. The
old de-duplication keyed by id collapsed the pair only while both ids matched — after a clone they
don't. → `056d40347`

**3. Order became scoped to the block.**
A component's order counts inside its own block, and two blocks of one page routinely declare the
same order or none. `buildNestedPages` sorted a page's components by `ode_components_sync_order`
alone, interleaving blocks. The row source already carried `blockOrder` and `ode_block_id` — the sort
simply ignored them — and `parseOdeXmlToJson` never read
`<odePagStructureOrder>`/`<odeComponentsOrder>` at all. → `37d317dcd`, `848f1e2fb`

**4. The tree sort skipped whole levels.**
`sortTree` recursed only when `children.length > 1`, so a page with exactly one child left that
child's entire subtree unsorted. → `8ecc3d007`

**5. Exported packages read the search index instead of the page tree.**
`_hasPagesMetadata()` routed them to `loadFromDom`, which reads `window.exeSearchData` /
`#exe-client-search[data-pages]`. That payload
(`src/shared/export/renderers/PageRenderer.ts:1184`) carries `name`, `fileName`, `prePageId`,
`nextPageId` and blocks — but **no `parentId` and no page-level `order`**. The report came out flat
and arbitrarily ordered. `content.xml` is the only shipped source with the tree, so it is now the
source; the metadata path stays as `loadFromContentXml`'s fallback for packages exported without the
editable source. → `f0806c6cf`

**6. Links broke on the cover page.**
`pathname.replace(/\/html(\/.*)?$/i, '')` assumed the report always lives under `html/`. The first
page is written to `index.html` at the package root
(`src/shared/export/exporters/Html5Exporter.ts:580-587`); with no `/html` segment the base kept the
file name and links came out as `/index.html/html/page.html` → 404. → `2fcabd216`

**7. Page file names diverged from the exporter's.**
The iDevice carried its own legacy transliteration table, predating
`BaseExporter.sanitizePageFilename()` (`src/shared/export/exporters/BaseExporter.ts:486`, introduced
by the rewrite in #869 and extended in #1896). The legacy table collapsed runs of separators: a page
titled `Adivina - Acceso` was linked as `adivina-acceso.html` while the exporter had written
`adivina---acceso.html`. Every such link 404'd. The iDevice now applies the exporter's algorithm
verbatim, `'page'` fallback and 50-char truncation included. → `064907f56`

## What changed

- `parseOdeXmlToJson` reads `jsonProperties` from the component that owns it, and uses the payload
  only to fill gaps the `htmlView` left — the payload is the older of the two sources and must not
  override it.
- New `sortComponentsByBlock(components, keys)`: block order first, position in the source as
  tie-break, component order last. Shared by the content.xml path and the row path, which name the
  block id differently (`keys.blockId` lists the candidates) — one ordering rule, not two.
- New `readOrder(node, selector, fallback)`: a missing, blank or non-numeric order resolves to the
  element's position, not to `0`, which would hoist it ahead of everything that declares one.
- New `parseComponentProperties(rawJson)`: a malformed payload yields `{}` and the component is
  still listed.
- New `loadCourseMap(mOption, index, init)`: the source-selection branch was duplicated at both call
  sites; it now lives in one place.
- De-duplication by id switched from an object map to a `Set` filter, so it no longer loses the
  computed order (integer-like legacy ids get hoisted to the front of an object).
- `normalizeFileName` replaced by the exporter's algorithm; ~180 lines of legacy transliteration
  table removed.

## Testing

- `npx vitest run public/files/perm/idevices/base/progress-report/export/progress-report.test.js` →
  **105 passing**, up from 60 cases. No `.skip` / `.todo`.
- New suites cover each regression directly: `ordering regressions`,
  `parseOdeXmlToJson component listing`, `parseComponentProperties`, plus `loadCourseMap`,
  `getURLPage` and `normalizeFileName`.
- `npx biome lint` clean on the touched folder.
- `make test-coverage` does not instrument `public/files/perm/idevices/**`, so patch coverage was
  checked by inspection: every new function and every changed branch has a direct test.


ignaciogros (changes requested):
Thanks, @mnarvaezm.

I think there's a blocking issue. Could you please check it?

`content.xml` is only shipped when `exportSource !== false` (`Html5Exporter.ts:617-626`; the SCORM exporters inherit this). With "export source" disabled and the package served over HTTP (e.g. Moodle), `fetch('../content.xml')` returns a 404 without throwing; `loadFromContentXml` (untouched by this PR) only falls back to `loadFromDom` inside `.catch`, so it parses the 404 HTML body → empty report. Before this PR that package still worked (flat, but with activities) because `_hasPagesMetadata()` took precedence. `loadFromContentXml` needs an `if (!response.ok) throw` (or equivalent) so the metadata fallback actually fires, plus a test for that case — none of the new tests cover the 404 path.

**Minor (non-blocking):**

- No E2E spec although the change affects a user-visible export flow (`AGENTS.md §5.3`). Arguable for an iDevice, but worth checking it.
- De-duplication now keeps the first entry instead of the one carrying `evaluationID`. Acceptable, since genuine duplicates disappear with the `jsonProperties` fix.

ignaciogros (approved):
Thanks, @mnarvaezm!

PR #2420, opened 2026-09-15
fix(icon-picker): inline Material glyphs instead of one external sprite <use> per option

### LINKED ISSUE #2419: Block icon picker freezes the Electron app (Windows/macOS): one sprite fetch per Material icon under app://
## Summary

Opening the block icon picker ("Select an icon" on an iDevice box) in the desktop app freezes the renderer, drives CPU to 70-90 %, grows memory past 2.5 GB, and never finishes rendering: Style (theme) icons never appear, Material icons appear one by one, and the styles menu stops responding. Reported on the Windows installer; reproduced on macOS with `make run-app`, so it is an Electron issue, not a Windows one. The web version is not affected.

## Root cause

Since #1497 the picker renders each of the 3 798 Material options as an external sprite reference:

```html
<svg class="exe-material-icon-sprite"><use href="app://localhost/libs/material-icons/material-icons.svg#alarm"></use></svg>
```

Under the Electron custom `app://` scheme Chromium does not share the external SVG resource document between `<use>` elements. Every `<use>` triggers its own fetch and parse of the 1.6 MB sprite. Measured with Playwright driving the dev Electron build: opening the picker issued **3 798 requests** to the sprite and the renderer process stayed at ~87 % CPU / 2.7 GB RSS until the app was killed. A standalone Electron benchmark confirms it: 300 `<use>` elements pointing at an `app://` sprite produce exactly 300 requests, with or without a `Cache-Control` header.

Over `http://` (web server) Chromium caches and shares the sprite, which is why the E2E suite and the web build never showed the problem.

## Fix

Render Material picker options from the sprite that is already parsed in memory at startup (`blockIconRuntime.loadMaterialSprite`) as self-contained inline `<svg>` elements, with no per-option external reference. The ADR-1247-04 note about the picker's `<use>` variant is amended accordingly.

erseco:
Fixes #2419

## Problem

Opening the block icon picker in the desktop app (Windows installer, also reproduced on macOS with `make run-app`) froze the renderer: CPU at 70-90 %, memory above 2.5 GB, Style icons never shown, Material icons trickling in one by one, styles menu unresponsive.

## Root cause

Since #1497 each of the 3 798 Material options in the picker was an external sprite reference:

```html
<svg class="exe-material-icon-sprite"><use href="app://localhost/libs/material-icons/material-icons.svg#alarm"></use></svg>
```

Under Electron's custom `app://` scheme Chromium does not share the external SVG document between `<use>` elements, so every option fetched and parsed the 1.6 MB sprite again. Measured with Playwright driving the dev Electron build: **3 798 sprite requests** on opening the picker, renderer at 87 % CPU / 2.7 GB RSS until killed. A standalone Electron benchmark confirms 300 `<use>` → 300 requests, with or without `Cache-Control`. Over `http://` Chromium shares the document, which is why the web build and the E2E suite never hit it.

## Fix

- `blockIconRuntime.js`: new `renderMaterialInlineIcon()` (and `getMaterialIconSymbol()`) that emits a self-contained inline `<svg class="exe-material-icon-sprite" viewBox="…" fill="currentColor"><path …/></svg>` from the sprite already parsed at startup by `loadMaterialSprite()`. Before the sprite is loaded it falls back to the existing mask placeholder, hydrated later.
- `blockNode.js`: the picker uses `renderMaterialInlineIcon()`; the external `<use>` renderer and `getMaterialSpritePath()` are removed. No per-option external reference remains.
- ADR-1247-04 amended (picker rendering evidence, consequence, references).

After the fix (same Playwright script, Electron on macOS): picker visible in ~3 s, 0 sprite requests, all 50 Style icons loaded, CPU back to 0 %.

## Tests

- Unit (Vitest): `renderMaterialInlineIcon` before/after sprite load, catalog and `help` fallback; picker options are inline SVG / hydration placeholders and never contain `<use>` or the sprite URL; blockNode delegate and degraded paths.
- E2E (`block-icons.spec.ts`): Material options carry an inline `path` and the picker contains no `use` element.

## Checks

- `make fix` clean (pre-existing warnings only)
- `make architecture-check` OK
- `bun x playwright test --project=chromium test/e2e/playwright/specs/block-icons.spec.ts`: 3 passed
- `make test-frontend`: see PR checks


ignaciogros (approved):
Thanks a lot for fixing this so fast.

ignaciogros:
The issue has also been confirmed not to occur on Linux or Mac.

PR #2430, opened 2026-09-16
fix(preview): self-heal the preview Service Worker registration (#2429)

### LINKED ISSUE #2429: Preview fails after upgrade: "Timeout waiting for SW content ready"
After upgrading a server instance (observed on a Cloud instance moving to a v4.0.5 build), clicking
**Preview** shows *Preview Error* and the browser console logs:

```
[PreviewPanel] Error generating preview: Error: Timeout waiting for SW content ready
```

- Reproduces on every page load, with any project (including an empty one).
- Does **not** reproduce in a private window on the same browser.
- Observed in Firefox. The same failure mode is possible in Chromium-based browsers under the
  race described below, although Chromium usually recovers on reload.

## Diagnosis

The preview system relies on a Service Worker (`public/preview-sw.js`, scope `<basePath>viewer/`).
`app.js` sends the generated preview files to the worker with `postMessage({ type: 'SET_CONTENT' })`
and waits for a `CONTENT_READY` reply on a `MessageChannel` port. If no reply arrives within 10 s
the promise rejects with the error above (`public/app/app.js:534-538`).

On the failing profile:

1. `navigator.serviceWorker.getRegistrations()` returned one registration, scope `/viewer/`,
   `active.state === 'activated'` — i.e. the browser considered the worker healthy.
2. A manual `GET_STATUS` ping over a `MessageChannel` got **no reply at all**.
3. `about:debugging` listed **more than one** `preview-sw.js` entry for the domain.
4. Unregistering the preview worker(s) in `about:debugging` and reloading fixed it immediately and
   permanently.

Conclusion: the profile held a **stale ("zombie") Service Worker registration** — reported as
active by the browser but no longer executing its `message` handler. The application adopted it
without verification and had no recovery path.

### Why stale registrations exist

- `preview-sw.js` changed between the previously deployed build and v4.0.5 (#2254 renamed
  `pdf.min.mjs` → `pdf.min.js` inside the worker). The browser detects the byte change on
  `registration.update()` and starts the install/activate cycle of a new worker. In Firefox this
  cycle can leave the registration pointing at a worker that never answers messages.
- Before v4.0.0-beta1 (#1103) the preview worker was registered with `scope: basePath` (`/`).
  Since then it uses `basePath + 'viewer/'`. Old root-scope registrations are never removed, and
  `getRegistration(basePath)` (`app.js:225`) returns the root-scope one when present, so the app
  may adopt a worker from a pre-4.0.0 build.
- Registrations are keyed by origin + scope and persist in the browser profile indefinitely.
  A private window starts with none, which is why it works there.

## Root cause in the application code

All in `public/app/app.js`:

| # | Problem | Location |
|---|---------|----------|
| 1 | An existing registration is adopted as soon as `registration.active` exists and `scriptURL` ends with `preview-sw.js`. There is **no health check** (no ping/reply) before trusting it. | `app.js:225-239` |
| 2 | `await registration.update()` resolves when the script download finishes, **not** when the new worker is activated. The `updatefound` listener that posts `SKIP_WAITING` is only attached in the fresh-registration branch, so the existing-registration branch never learns the worker changed. | `app.js:235`, `app.js:253-263` |
| 3 | On the 10 s timeout the code simply rejects. There is **no recovery** (no `unregister()` + `register()` + resend). The error is surfaced to the user as *Preview Error*. | `app.js:534-538` |
| 4 | Orphaned registrations (`preview-sw.js` under other scopes, e.g. `/` from pre-4.0.0 or an old `BASE_PATH`) are **never cleaned up**. | `registerPreviewServiceWorker()` |
| 5 | `SW_VERSION` is hard-coded to `'1.0.0'` and never bumped, so a version mismatch between app and worker cannot be detected via `GET_STATUS`. | `public/preview-sw.js:7` |

In short, the flow is built for the happy path (first install) and for an update that Chromium
resolves cleanly. It does not handle "the worker exists but is dead", which is exactly what Firefox
leaves behind after `preview-sw.js` changes between builds.

## Proposed fix

All changes in `public/app/app.js`, plus a one-line change in `public/preview-sw.js`. No server
changes, no data migration.

### 1. Clean up orphaned registrations at startup

Before registering, enumerate all registrations and unregister any `preview-sw.js` whose scope is
not the expected one.

```js
async _unregisterStalePreviewWorkers(expectedScope) {
    const registrations = await navigator.serviceWorker.getRegistrations();
    const expected = new URL(expectedScope, window.location.origin).href;
    await Promise.all(
        registrations
            .filter((r) => {
                const url = (r.active || r.waiting || r.installing)?.scriptURL || '';
                return url.endsWith('preview-sw.js') && r.scope !== expected;
            })
            .map((r) => r.unregister().catch(() => false))
    );
}
```

### 2. Health-check the worker before trusting it

Extract a pure helper that pings the worker with `GET_STATUS` over a `MessageChannel` and resolves
`true`/`false` within a short timeout (2–3 s). Use it in both branches of
`registerPreviewServiceWorker()`.

```js
pingPreviewServiceWorker(sw, timeoutMs = 3000) {
    return new Promise((resolve) => {
        const channel = new MessageChannel();
        const timer = setTimeout(() => { channel.port1.close(); resolve(false); }, timeoutMs);
        channel.port1.onmessage = (event) => {
            clearTimeout(timer);
            channel.port1.close();
            resolve(event.data?.type === 'STATUS');
        };
        try {
            sw.postMessage({ type: 'GET_STATUS' }, [channel.port2]);
        } catch {
            clearTimeout(timer);
            resolve(false);
        }
    });
}
```

### 3. Wait for activation after `update()` and attach `updatefound` in both branches

After `await registration.update()`, if `registration.installing` or `registration.waiting` is set,
wait until that worker reaches `activated` (reusing `_waitForActivation`) and refresh
`this._previewSwRegistration`. Move the `updatefound` listener out of the fresh-registration branch
so it is attached in both cases.

### 4. Recover instead of failing

Introduce a single recovery routine and use it in two places:

```js
async _recoverPreviewServiceWorker(swPath, previewScope) {
    const current = this._previewSwRegistration
        || (await navigator.serviceWorker.getRegistration(previewScope));
    if (current) {
        await current.unregister().catch(() => false);
    }
    const registration = await navigator.serviceWorker.register(swPath, { scope: previewScope });
    await this._waitForActivation(registration);
    await this._tryClaimClients(registration);
    this._previewSwRegistration = registration;
    return registration;
}
```

- In `registerPreviewServiceWorker()`: after adopting/registering, `ping()`; if it fails, run
  `_recoverPreviewServiceWorker()` once and ping again. Only then resolve
  `_previewSwRegistrationPromise`.
- In `sendContentToPreviewSW()`: on timeout, run `_recoverPreviewServiceWorker()` once and resend
  the same payload before rejecting. Guard with a flag so recovery happens at most once per call.

Note: the `ArrayBuffer`s are transferred on the first `postMessage`, so the payload must be cloned
(or regenerated by the caller) before a resend. Simplest option: have `sendContentToPreviewSW`
accept a factory `() => files` from `PreviewPanel`, or clone buffers up-front only when a retry is
needed.

### 5. Make the worker version diagnosable

In `public/preview-sw.js` derive `SW_VERSION` from the app version (e.g. inject it as a query
parameter on the registration URL `preview-sw.js?v=<appVersion>` and read `self.location.search`),
or at minimum bump it on every change to the file. This lets `GET_STATUS` reveal an app/worker
mismatch and gives support a concrete value to ask for.

### Expected user-visible behaviour after the fix

- First preview after an upgrade may take 2–3 s longer (ping timeout + re-register).
- No error is shown; no manual `about:debugging` intervention.
- Orphaned pre-4.0.0 registrations disappear on first load.

## Tests

- `public/app/app.test.js` (Vitest): unit tests for `pingPreviewServiceWorker` (reply / no reply /
  `postMessage` throws), `_unregisterStalePreviewWorkers` (keeps expected scope, removes others,
  ignores non-preview workers), and the recovery path in `sendContentToPreviewSW` (first send times
  out → unregister + register + resend succeeds; second timeout → rejects with the original error).
  Mock `navigator.serviceWorker` with fake registrations.
- `test/e2e/playwright/specs/preview-sw-recovery.spec.ts` (Playwright, Chromium + Firefox):
  1. Load workarea, register a **stub** worker at the preview scope that ignores messages
     (`page.evaluate` registering an inline blob script is not allowed cross-origin; instead
     intercept `preview-sw.js` with `page.route()` and serve a script with no `message` handler).
  2. Reload without the route, open Preview, assert the iframe renders `article` and no
     `[PreviewPanel] Error` is logged.
- Patch coverage ≥ 90% on the changed lines (`make test-coverage`).

## Rollout / support notes

- Until the fix ships, affected users can recover by unregistering `preview-sw.js` for the domain
  (Firefox: `about:debugging#/runtime/this-firefox` → Service Workers → *Unregister*; Chrome:
  DevTools → Application → Service Workers → *Unregister*) and reloading. No data is lost: the
  preview worker keeps only in-memory content regenerated on every preview.
- Keep `APP_VERSION` set to the real release (`v4.0.5-em`) rather than the default `0.0.0-alpha`
  so versioned static assets (`/v<version>/…`, served `immutable`) are not reused from cache across
  builds. This is unrelated to the worker issue but avoids a second class of stale-frontend bugs.

erseco:
Fixes #2429

## Summary

The preview Service Worker registration flow now self-heals instead of trusting whatever registration the browser profile kept:

- **Health check at load.** After registering, the app pings the worker (`GET_STATUS`, 3 s). A worker that never answers is unregistered and registered again once; if that fails too, the registration promise resolves `null` and the preview panel falls back to the blob URL renderer instead of showing *Preview Error*.
- **Deterministic upgrades.** The worker is registered as `preview-sw.js?v=<app version>` and the app runs `registration.update()` when `register()` did not already start an install, then waits for the new worker to activate. A changed `preview-sw.js` is replaced at load, before any content is sent, instead of by a browser-side update in the middle of the first preview.
- **Orphan cleanup.** `preview-sw.js` registrations whose scope *contains* the preview scope (the root scope used before #1103) are unregistered at startup. Sibling scopes of other deployments on the same origin and the PWA `service-worker.js` are never touched.
- **Recovery on send.** `sendContentToPreviewSW()` re-registers the worker once when `SET_CONTENT` is never confirmed and, because the `ArrayBuffer`s were transferred, resends files produced by a `regenerateFiles` callback supplied by the preview panel.
- **No 5 s stall.** `CLAIM_CLIENTS` is only awaited when the page is inside the registration scope. The editor page is not, so the registration promise used to wait the full 5 s controller timeout on every page load.
- **Diagnosable worker.** `GET_STATUS` now reports the script revision (`version: 1.1.0`) and the app version the worker was registered with (`appVersion`); the app warns on a mismatch.

## Reproduction and evidence

All measurements from a Playwright-library harness (headless Chromium 1208 and Firefox 1538) against a local server, empty project, and from a real Chrome profile through the Chrome MCP.

On `main` (8fa25e7b0):

| Scenario | Chromium | Firefox |
|---|---|---|
| Fresh profile | preview in 0.3 s | preview in 0.4 s |
| Registered worker never answers | *Preview Error: Timeout waiting for SW content ready* after 10.3 s | same after 10.5 s |
| Root-scope orphan plus `/viewer/` registration | app adopts `/`, iframe loops `about:blank` / `/viewer/index.html`, never renders | same |
| `preview-sw.js` bytes changed, reload | preview in 3.7 s (browser update mid-preview) | 3.7 s |
| Registration promise settle time | 5.0 s (`CLAIM_CLIENTS` wait) | 5.0 s |

On this branch every scenario renders in 0.3 to 0.4 s on both browsers, the dead worker is re-registered at load, the orphan is removed at load, and a script change is applied at load. Real Chrome (profile with a root-scope orphan and a leftover `/web/exelearning/viewer/` registration from another deployment): orphan removed, sibling kept, preview in 310 ms.

Two findings that bound the scope of the reported failure:

- A clean profile upgrading from v4.0.3 to v4.0.5 does **not** fail (manual Firefox test by @ignaciogros, and the "bytes changed" row above). The error needs a registration that is already dead or duplicated. Long-lived profiles that used pre-4.0.0 builds (root-scope registration) and profiles that saw several deployments on one origin are the population at risk. This is therefore a robustness fix rather than a failure every user hits; the 5 s stall, on the other hand, affects every page load today.
- Playwright cannot route the worker script request in Firefox, and in Chromium a worker installed from a route-fulfilled stub makes the browser reinstall the worker on every in-scope navigation afterwards (pure-browser experiment, no app code; a real script change on disk installs exactly once). The E2E spec therefore simulates the dead worker from the page, which behaves identically in both browsers.

## Scope versus the minimal hotfix proposed in the issue

The issue comment suggests shipping only the registration-time ping for v4.0.5 and deferring the rest. Each deferred item is included here because the reproduction showed it matters, and each concern is addressed:

- *Orphan removal could unregister the good worker.* The rule only removes registrations whose scope is a **prefix** of `<basePath>viewer/`; a wrong base path can never match the good `/viewer/` registration or a sibling deployment. The orphan case reproduces as a blank preview on both browsers.
- *Waiting for activation after `update()` changes Chrome behaviour.* Without it, Chromium keeps the old worker after `register()` (verified: a stale worker survived a reload with the new script) and replaces it mid-preview, which is the race the issue describes. `update()` is skipped when `register()` already started an install.
- *Version in the URL forces a reinstall per release.* That reinstall now happens at load and is covered by the same health check, and the query defeats proxies that cache `preview-sw.js`. `SW_VERSION` is bumped to `1.1.0` as suggested; the app version is reported separately.
- *Recovery inside `sendContentToPreviewSW()` needs buffer cloning.* Not needed: the preview panel passes a `regenerateFiles` callback, so nothing is cloned on the happy path.

## Tests

- `public/app/app.test.js`: registration flow (version query, `update()` handling, health check, recovery, failure to `null`), `pingPreviewServiceWorker`, `_unregisterStalePreviewWorkers`, `_recoverPreviewServiceWorker`, `_tryClaimClients` scope guard, `sendContentToPreviewSW` retry paths, `waitForPreviewServiceWorker` precedence.
- `public/app/workarea/interface/elements/previewPanel.test.js`: `regenerateFiles` callback, blob URL fallback after a failed send.
- `public/preview-sw.test.js`: script revision and app version resolution.
- `test/e2e/playwright/specs/preview-sw-recovery.spec.ts` (Chromium and Firefox): a worker that never answers is re-registered and the preview renders without `[PreviewPanel] Error`; a root-scope orphan is removed and the preview renders.

## Documentation

- New `doc/architecture/adr/ADR-2429-01-self-heal-preview-service-worker-registration.md` (status Proposed).
- `doc/architecture.md` §8: new 8.6 "Registration Lifecycle and Self-Healing", `GET_STATUS` in the protocol table, corrected worker path.

## Verification

Run locally on this branch (macOS, Bun 1.4.0):

- `make fix`: clean (only pre-existing warnings).
- `make test-unit` (backend): 8807 tests, 0 failures.
- `make test-integration`: 770 tests, 0 failures.
- `make test-frontend`: 296 files, 16136 tests, 0 failures. Patch coverage of the changed executable lines in `app.js` and `previewPanel.js`: 96.3 % measured, then five more tests were added for the remaining error branches.
- `make architecture-check`: OK.
- `test/e2e/playwright/specs/preview-sw-recovery.spec.ts`: 4 passed on `--project=chromium` and `--project=firefox`.
- Reproduction harness (5 scenarios, Chromium and Firefox): every scenario renders in 0.3 to 0.4 s; real Chrome profile via the browser extension: orphan removed, preview in 310 ms.
- Full E2E suite: left to CI for this PR (a local full run was started but stopped to keep the machine free for the Firefox subset).

## Notes for release

- `v4.0.5` is already tagged. Given that a clean profile upgrade works, this can ship in the next release rather than by moving the tag; the workaround for affected profiles remains unregistering `preview-sw.js` in `about:debugging` / DevTools.
- Operators should run containers with `APP_VERSION` set to the real release (see the issue); unrelated to the worker but it also makes the `?v=` query meaningful.


ignaciogros:
The solution is complete, apart from a few details that could be fixed later (I will provide a full review at another time). However, given the scope of the change, I think it would be better to integrate it into the next version (v4.0.6) and try to include a minimal fix for now.

I will open a PR with a proposal as soon as possible. I suggest considering its inclusion and leaving this complete solution for later.

Thank you.

ignaciogros (dismissed):
Thanks!

**Verdict after two independent automatic reviews: OK to merge after the details below.**

## To fix / improve

1. **`SW_VERSION` not bumped.** `public/preview-sw.js` still reports `1.1.0`, which is what v4.0.5 already shipped with #2432. This PR changes the script: it adds `SW_APP_VERSION`, and `GET_STATUS` now returns `appVersion`. Please bump it to `1.2.0` and update `preview-sw.test.js` and the ADR to match.

2. **Send-time recovery failure does not degrade.** When `_recoverPreviewServiceWorker()` fails inside `sendContentToPreviewSW()`, `_previewSwUnavailable` is never set. The load-time path does set it. As a result, `getPreviewServiceWorker()` can still return a worker, `isServiceWorkerPreviewAvailable()` stays `true`, and the user sees *Preview Error* instead of the blob URL fallback. Please set the flag on that failure and add a unit test for it.

3. **Rebase drops #2432 safeguards.** `_toPathname()` has been removed. Scopes and script URLs are now compared with `new URL(..., window.location.origin)`, and in `_unregisterStalePreviewWorkers()` that call sits outside any `try`. The `Boolean(scopePath)` guard is also gone, so an empty scope now counts as stale. Please keep the origin-independent pathname helper and the guard from #2432. Pathname comparison already ignores the new `?v=` query.

4. **Unit tests weaker than in #2432.** `app.test.js` brings back `getRegistration` mocks, which the registration flow no longer calls. The "PWA SW not preview SW" test cannot fail. The healthy-adoption test lost its `GET_STATUS` / `toBe(viewerReg)` assertions, and the re-register test lost its exact warn-message check. Please restore the #2432 assertions and drop the dead mocks. Also add tests for `_getPreviewServiceWorkerPaths()` without a version and for the `_isPageWithinScope()` catch branch.

5. **ADR-2429-01 predates #2432.** Context and Options present Option 1 as a suggestion that was not chosen, but it shipped in v4.0.5 as #2432 (and it also removes orphans). Please say that this ADR completes #2432 and list what each PR contributes. Add `2432` to `related.prs`. The "before" baseline (`main @ 8fa25e7b0`) is also no longer the state this PR changes. The Validation section should not list `preview-sw-recovery.spec.ts` as new (it came from #2432). The manual test attribution differs: the ADR credits @erseco and the PR body credits @ignaciogros.

6. **PR description is out of date.** It lists the health check and the orphan cleanup as changes, but both are already in `main`. The evidence table also still shows those scenarios failing on `main`. Please rewrite it as the delta over #2432.

7. **CHANGELOG.** There is no entry. Add a line for the next release covering what users will notice: the first preview no longer waits about 5 s, and the preview recovers when the worker stops answering during a session.

8. **E2E coverage of the new paths.** The existing spec only covers recovery at load and orphan cleanup. Please add a case where the worker dies after the first preview and the second preview still renders (re-register plus `regenerateFiles`). The `ZombieState` helper in the existing spec can be reused for this.

9. **ADR status.** Change `Proposed` to `Accepted` when merging.

erseco:
@ignaciogros thanks for the review. I rebased the branch on `main` (it now includes #2431) and addressed all nine points in 93e497157:

1. **`SW_VERSION`**: bumped to `1.2.0`. `preview-sw.test.js` and the ADR are updated to match.
2. **Send-time recovery failure**: `sendContentToPreviewSW()` now sets `_previewSwUnavailable` when `_recoverPreviewServiceWorker()` fails, so `getPreviewServiceWorker()` returns `null` and the panel falls back to blob URLs. A new unit test covers this with a stale preview controller still present, and it fails without the fix.
3. **#2432 safeguards**: `_toPathname()` is restored and used again by `_isPreviewServiceWorkerScript()` and `_unregisterStalePreviewWorkers()`, together with the `Boolean(scopePath)` guard. There is a test for an empty scope.
4. **Unit tests**: I rebuilt the preview SW block of `app.test.js` from the #2432 version, so the `GET_STATUS` / `toBe(viewerReg)` checks, the exact re-register warning and the secure-context spies are back. The dead `getRegistration` mocks and the "PWA SW not preview SW" test are gone. New tests cover `_getPreviewServiceWorkerPaths()` with and without a version (and that it is memoised), the `_isPageWithinScope()` catch branch and `_checkPreviewServiceWorkerVersion()` with unknown versions. Patch coverage against `main`: `app.js` 113/113 and `previewPanel.js` 13/13 changed executable lines.
5. **ADR-2429-01**: rewritten as completing #2432. The context starts from the post-#2432 state, Option 1 is now "shipped in v4.0.5 as #2432", `2432` is in `related.prs`, the `8fa25e7b0` evidence is labelled as pre-#2432 with a note on which rows each PR fixes, and Validation lists only the new E2E case as added here. For the manual clean-upgrade test, the ADR and the PR body disagreed on who ran it, so both now describe it without naming anyone.
6. **PR description**: rewritten as the delta over #2432.
7. **CHANGELOG**: added a `v4.0.6 – Unreleased` / Fixed entry: the first preview no longer waits about 5 s, and the preview recovers when its worker stops responding during a session.
8. **E2E**: `preview-sw-recovery.spec.ts` has a new case. The worker dies after the first preview (using the existing `ZombieState` simulation), and the next refresh still renders. The test checks that the new document came from the re-registered worker (`/viewer/` iframe, not marked unavailable) and not from the blob fallback. The spec passes 3/3 on Chromium and 3/3 on Firefox.
9. **ADR status**: set to `Accepted` in this PR, since it lands with the merge.

Local checks: `make fix` clean; backend unit 8807/0; frontend 16151/0; integration 770/0; `make architecture-check` OK. The full E2E suite is left to CI.

ignaciogros (approved):
Thanks for fixing those details so fast!

PR #2431, opened 2026-09-16
ci: disable the metadata-action latest flavor on tag builds

erseco:
## Problem

#2425 moved `latest` / `latest-static` out of the build workflows and into `promote-latest.yml`, on the `released` event. It did not work: the v4.0.5 tag push wrote `latest` again on Docker Hub and GHCR while the release is still a pre-release.

`latest` and `v4.0.5` share the same digest, written by the tag-push CI run, not by a promotion:

| tag | digest | last_updated |
|---|---|---|
| `latest` | `sha256:e380b75…` | 2026-09-16 06:49 |
| `v4.0.5` | `sha256:e380b75…` | 2026-09-16 06:49 |
| `latest-static` | `sha256:d8e4e2d…` (= `v4.0.3-static`) | 2026-09-15 15:34 |

`promote-latest.yml` has exactly one run, the manual dispatch from 2026-09-15. The `released` event never fired.

## Cause

#2425 removed this line from `ci.yml`:

```yaml
type=raw,value=latest,enable=${{ github.ref_type == 'tag' && !contains(github.ref_name, '-') }}
```

That line was redundant, not the source. `docker/metadata-action` defaults to `flavor: latest=auto`, and with `latest=auto` a `type=ref,event=tag` entry always produces `latest` — unlike `type=semver`, the ref type has no prerelease check at all, and the GitHub "pre-release" checkbox is not visible to the action either way.

From the [v4.0.5 tag run](https://github.com/exelearning/exelearning/actions/runs/35064811296):

```
Processing tags input
latest=auto
...
Docker tags
  exelearning/exelearning:v4.0.5
  exelearning/exelearning:latest
  ghcr.io/exelearning/exelearning:v4.0.5
  ghcr.io/exelearning/exelearning:latest
"tag-names":["v4.0.5","latest"]
```

`static-release.yml` was unaffected because `type=raw` only yields `latest` under `latest=true`, never under `auto` — which is why `latest-static` stayed on v4.0.3.

## Changes

- **`ci.yml`**: `flavor: latest=false` on the metadata step. A tag push now builds only `vX.Y.Z`.
- **`static-release.yml`**: the same line. It is a no-op for `type=raw`, but it states the intent so both build workflows read the same and a future `type=semver`/`type=ref` entry cannot reintroduce this silently.

No change to `promote-latest.yml` or to the release flow documented in `doc/development/version-control.md`; this only makes the build workflows behave the way that document already describes.

## Registry state

`latest` is at v4.0.5 on both Docker Hub and GHCR and has to be rolled back to v4.0.3 with a manual run of "Promote release to latest" (`version: v4.0.3`). `latest-static` is already correct on both registries. Both `v4.0.3` and `v4.0.3-static` exist in both registries, so the promotion has a valid source.

## Verification

CI-config-only change, no code under `src/` or `public/`, so there is no unit/e2e coverage to add. `make fix` is clean and both workflow files parse. The change is confirmed by the run log above: `latest=auto` is what emitted the tag, and `latest=false` is the documented way to suppress it. The real check is the next tag push producing only `vX.Y.Z`.


erseco:
Rollback done — ["Promote release to latest" run 35093283063](https://github.com/exelearning/exelearning/actions/runs/35093283063) (`version: v4.0.3`) finished successfully. Verified against both registries:

| tag | Docker Hub | GHCR |
|---|---|---|
| `latest` | `sha256:5f89abf5…` | `sha256:5f89abf5…` |
| `v4.0.3` | `sha256:5f89abf5…` | `sha256:5f89abf5…` |
| `latest-static` | `sha256:d8e4e2d6…` | `sha256:d8e4e2d6…` |
| `v4.0.3-static` | `sha256:d8e4e2d6…` | `sha256:d8e4e2d6…` |

`latest` is back on v4.0.3 and `v4.0.5` (`sha256:e380b751…`) is untouched, still reachable by its own tag while the release is a pre-release.

mnunezcedec (approved):

ignaciogros (approved):
Thanks!

One note: the "Registry state" section is outdated. v4.0.5 has since been published as a full release and `promote-latest` ran on `released`, so `latest` / `latest-static` already point to v4.0.5. No rollback to v4.0.3 is needed.

PR #2432, opened 2026-09-16
Fix preview after upgrade

ignaciogros:
This PR fixes the preview re-registering a preview Service Worker that does not answer. It's just a provisional solution for `v4.0.5`.

Refs #2429

This is a deliberately small subset of #2430, which contains the complete solution and is scheduled for the next release once it is approved. This PR only makes sure affected users get a working preview now; it does not change how healthy installations behave.

## What this PR does

- When the editor loads, it now asks the preview worker whether it is alive. If it does not answer, the worker is removed and registered again. If that also fails, the preview quietly switches to the fallback renderer instead of showing an error.
- Preview workers from versions before 4.0.0 (registered at the site root) are removed so they can no longer get in the way.
- `SW_VERSION` in `preview-sw.js` is raised to `1.1.0`. This value is only used for diagnostics: it lets support tell, from the worker's own status reply, whether a browser is running the worker shipped with this release or an older one. It is a manual constant on purpose; tying it to the application version would force the browser to reinstall the worker on every release, which is one of the behaviour changes left for #2430.


erseco (approved):
Approved as a scoped hotfix for v4.0.5. The startup health check, targeted cleanup of ancestor-scope preview registrations, and blob fallback address the immediate failure without introducing the broader lifecycle changes from #2430.

Non-blocking follow-ups for #2430:

- Avoid waiting for controller acquisition when the editor is outside the preview worker's scope. The existing five-second wait is repeated during recovery.
- Keep recovery during content delivery and deterministic worker updates in scope; a successful startup ping does not guarantee a later `SET_CONTENT` will succeed.
- Strengthen browser coverage for worker replacement. The current E2E simulation resumes message delivery after `unregister()` is called, so it verifies the recovery flow but does not establish that a genuinely unresponsive worker was replaced.

Validation: reviewed f6636f2 and ran four isolated checks using the actual registration methods with simulated registrations: healthy worker, successful recovery, fallback after failed recovery, and ancestor cleanup preserving sibling scopes. The full Vitest and browser E2E suites were not run in this review environment.

Please keep #2429 tracked until the remaining lifecycle work is addressed.

PR #2433, opened 2026-09-16
Keep the local iDevice editor alive and its content intact during remote structure changes

### LINKED ISSUE #2427: Collaborative editing: adding an iDevice in the same block closes another user’s active edit and causes data loss
There is a problem in collaborative mode when two users are working in the same block.

If User A is editing an iDevice and User B adds another iDevice to the same block, the edit session of User A is automatically closed and any unsaved changes are lost.

Steps to reproduce

- User A starts editing an iDevice.
- User B adds a new iDevice to the same block, or moves an existing idevice into the block.
- User A’s iDevice editing mode closes automatically.
- Any unsaved changes made by User A are lost.

erseco:
## Summary

Fixes #2427.

Two independent paths lost User A's unsaved work when User B changed the same page in collaborative mode:

1. **Stale save after a remote insertion.** Rendering any iDevice in export mode called `restartExeIdeviceValue()`, which cleared the global `$exeDevice`. That global is the edition object of the iDevice User A is editing. Remote iDevices are rendered in export mode when User B adds one (same block or a new block), so User A's next save found no `$exeDevice`, kept the previously saved data and silently discarded everything typed in the editor.
2. **Forced page reload after a remote move.** Remote transactions containing deletions (moving an iDevice into another block, reordering, moving blocks) scheduled a full reload of the page content, which tears down every iDevice, including the open editor.

## Fix

- `IdeviceNode.restartExeIdeviceValue()` leaves `$exeDevice` alone while another iDevice is in edition on this client.
- `YjsProjectBridge.schedulePageReloadIfCurrent()` defers the reload while an iDevice is being edited; `IdevicesEngine.updateMode()` flushes it as soon as no iDevice is in edition (save, cancel or delete), and the reload is skipped if the user has meanwhile moved to another page. The late-asset refresh and the `StructureEngine` remote handler go through the same guarded method, so there is a single place that decides when the page content may be rebuilt (the `StructureEngine` reload is also debounced with the bridge's own one now, instead of running twice).

Behaviour change to be aware of: while an iDevice is being edited, **structural** changes made by other users on the current page (moves, reorders, deletions) are applied when the edition ends instead of immediately. Additions and content updates of other iDevices keep appearing in real time as before.

## Tests

- **Unit (Vitest)** — `ideviceNode.test.js` (`$exeDevice` kept while another iDevice is edited, cleared when it is this one), `YjsProjectBridge.test.js` (deferral, flush, flush after leaving the page, late-asset deferral), `idevicesEngine.test.js` (`updateMode` flushes only when nothing is in edition), `structureEngine.test.js` (remote reload routed through the bridge).
- **E2E (Playwright, two clients)** — `test/e2e/playwright/specs/collaborative/editor-preservation.spec.ts`, two new tests that fail on `main`:
  - User A saves the editor content, not the stale version, after User B creates an iDevice.
  - User A's editor survives User B moving another iDevice into the block being edited; after saving, the page shows the moved iDevice in that block and User B sees User A's content.

```bash
make fix && make test-unit
bun x playwright test --project=chromium test/e2e/playwright/specs/collaborative/ test/e2e/playwright/specs/collaboration.spec.ts
```

## How to verify in a deployment

1. Log in as two different users in two browsers. User A opens a project, shares it as public and User B opens the share link. Both select the same page, which has at least two iDevices (each in its own block).
2. User A opens the first iDevice for editing and types some text **without saving**.
3. User B adds a new iDevice to the page (drop it inside User A's block or click it in the sidebar). User A's editor must stay open; User A saves and the saved content must be the text just typed. Before this fix the previous content was saved and the typed text was lost.
4. User A edits the first iDevice again and types without saving. User B drags the second iDevice into User A's block (or moves a block with the block arrows). User A's editor must stay open with the text; when User A saves, the page refreshes to the new structure and User B sees the saved text.

## Screenshots (User A, while editing with unsaved text)

| Before | After |
|---|---|
| User B moves an iDevice into the block: the editor is closed by the forced reload. Whether the typed text survives depends on `$exeDevice` still being set (path 1), so in practice the changes vanish. | The editor stays open with the typed text; the moved iDevice is applied when the edition ends. |
|  |  |

After User A saves (fixed build): the content typed in the editor is what gets persisted and the page reflects the remote move.

Evidence for path 1 on `main` (new E2E test, User A saves after User B created an iDevice):

```
Expected substring: UNSAVED_EDIT_1789566723031
Received string:    Seed content 1789566718994
```


cristinavaldera (approved):
Working right! Thanks!

PR #2434, opened 2026-09-16
Run export post-render hooks for remote iDevice updates in collaborative mode

### LINKED ISSUE #2428: Collaborative mode: newly added iDevices are not fully rendered for other users until page refresh
There is a refresh/rendering issue in Cloud collaborative mode when two users are working on the same page.

When User A creates and saves an iDevice, User B sees the new iDevice appear immediately, but its content may be empty or only partially rendered.

This happens with iDevices such as A-Z, Guess and GeoGebra, but also with TinyMCE features. For example, with music notation, User B sees the source code instead of the rendered score.

If User B navigates to another page and then returns, the content is displayed correctly.

Steps to reproduce

- Two users open the same page in collaborative mode.
- User A creates an iDevice (e.g. A-Z, Guess, or GeoGebra) and saves it.
- The new iDevice appears for User B, but its content is empty or incomplete.
- Alternatively, add music notation using TinyMCE: User B sees the code instead of the rendered score.
- User B navigates to another page and returns.
- The content is now rendered correctly.

erseco:
## Summary

Fixes #2428.

In collaborative mode, when User A saves an iDevice, User B receives it through Yjs and renders it incrementally (`renderRemoteIdevice` / `updateRemoteIdeviceContent`) instead of reloading the page. That incremental path only ran `loadInitScriptIdevice('export')` and skipped the page-level steps that a page load or a local save run afterwards. User B therefore saw:

- **Music notation** (TinyMCE ABC plugin) as its raw `<pre class="abc-music">` source, because `$exeABCmusic.init()` never ran on the new HTML.
- **A-Z quiz / Guess / GeoGebra activity** as an inert body with only the instructions text. These are HTML-type iDevices whose export script initialises once, on document ready (`$(function () { $azquizgame.init() })`). On User B the script had been loaded while the iDevice was still an empty locked placeholder, so when the real content arrived nobody called `init()` again.

Navigating away and back "fixed" it because a page load performs those steps.

## Fix

`IdevicesEngine.reloadExportRuntime()` groups what a local save already did inside `resetCurrentIdevicesExportView()`:

1. remove and re-insert the export scripts, so their document-ready bootstraps run over the new HTML;
2. run the legacy functionalities (`$exeFX`, `$exeGames`, `$exeHighlighter`, `$exeABCmusic`, `$exe`);
3. wire `exe-node:` internal links.

The local save path and both remote paths now call it, so there is a single place for "what has to happen after export HTML lands in the DOM".

## Tests

- **Unit (Vitest)** — `idevicesEngine.test.js`: call order of `reloadExportRuntime`; `renderRemoteIdevice` and `updateRemoteIdeviceContent` run it right after `loadInitScriptIdevice('export')`; lock-only updates and iDevices being edited locally do not trigger it.
- **E2E (Playwright, two clients)** — `test/e2e/playwright/specs/collaborative/remote-idevice-render.spec.ts`: music notation saved by User A renders as a score for User B; an A-Z quiz saved by User A shows its game board for User B. Both tests fail on `main` and pass with this change.

```bash
make fix && make test-unit
bun x playwright test --project=chromium test/e2e/playwright/specs/collaborative/
```

## How to verify in a deployment

1. Log in as two different users in two browsers. User A opens a project, shares it as public and User B opens the share link. Both select the same page.
2. User A adds a **Text** iDevice, inserts music notation (TinyMCE → Insert → *ABC music*, or paste `<pre class="abc-music">X:1\nK:C\nC D E F|</pre>` in source mode) and saves.
3. User B, **without navigating**, must see the rendered score (with the ABC / MIDI toggles), not the notation source.
4. User A adds an **A-Z quiz** (or *Guess* / *GeoGebra activity*), fills one word and its definition and saves.
5. User B must see the game board immediately. Before this fix only the instructions paragraph appeared until the page was re-opened.

## Screenshots (User B, same page, no navigation)

| | Before | After |
|---|---|---|
| Music notation |  |  |
| A-Z quiz |  |  |


cristinavaldera (dismissed):
Not possible to Save an idevice. After clicking in the Save button and waiting for long, this warning appears:

Image

erseco:
@cristinavaldera thanks for catching this — you were right, and it was this PR. Fixed in the last commit.

**What was happening**

The modal you saw comes from a single place: the export-load retry loop for `component-type=json` iDevices (Text is one of them). It gives up after 50 attempts, and each attempt that gets as far as rendering also fires a save request — which is why it hangs for a long time before the alert instead of failing immediately.

The cause was the helper this PR introduced. `reloadExportRuntime()` calls `clearNeedlessScripts()`, which removes **every** `head > script:not(.exe)` — edition scripts included — and then puts back only the export ones. That is fine for a local save, where `resetCurrentIdevicesExportView()` rebuilds the whole page anyway, but not for the incremental remote path: a remote update can land at any moment, and one remote save produces three or four of them. Measured on the branch: after a single remote update the receiving client had **zero** edition script tags left in `<head>`, and every export script of the page had been re-downloaded and re-executed. That breaks the "tag in `<head>` means the global is loaded" invariant that `loadScriptDynamically()` relies on for the shared `$exeDevice` global, so a later local save could render an empty view and fall into the retry loop.

**The fix**

The remote paths now use a scoped `reloadExportRuntimeForIdevice(ideviceNode)` that re-executes only the export scripts of the iDevice type that just arrived, and only when it is HTML-type — JSON-type ones are initialised explicitly through their export object, so they never needed it. It no longer touches any other script on the page. `reloadExportRuntime()` is unchanged and still used by the local save path.

**Tests**

Five unit tests for the new helper, plus an E2E in this PR's own spec asserting that a remote render leaves the receiving client's loaded scripts in place and that the user can still edit and save afterwards. I checked the E2E has teeth: reverting only the source change makes it fail on that assertion.

One unrelated note from testing: if User B has an editor open when User A adds an iDevice, B's draft is lost. That is #2427, fixed by #2433, so I kept it out of this test to avoid conflating the two.

Could you take another look when you have a moment? Thanks again for the careful review.

